PR #8, opened 2015-01-23
Export open type family IsElem' to allow custom API types in Links.

christian-marie:
Also:
* Export unLink function to be able to use the contents of a Link
* Add Raw to the isLink'' family.

--------------------------

Furthermore. You'll note that the documentation is updated in this commit to what typechecks. This is not consistent with the rest of the library's use of Proxy. Perhaps we should use Proxy instead of undefined?

It's a breaking change, but I'm not so sure anyone was using this module as you can't get any String out of the Link.

### DIFF
--- a/src/Servant/Utils/Links.hs
+++ b/src/Servant/Utils/Links.hs
@@ -6,29 +6,49 @@
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE UndecidableInstances #-}
+
 -- | Type safe internal links.
 --
 -- Provides the function 'mkLink':
 --
 -- @
---   type API = Proxy ("hello" :> Get Int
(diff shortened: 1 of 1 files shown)


christian-marie:
I'll address one more thing:

This changes error messages on invalid links. Instead of ghc telling us that we can't unify True with False (which is incredibly unhelpful), it will now tell us how it tried to unify the types and failed, like this:

```
Test.hs:27:15:
    Couldn't match type ‘Servant.Utils.Links.Or
                           (IsElem' (Get Int) (Thing :> Get Int))
                           (IsElem'
                              ("hello" :> Get Int)
                              ("bye" :> (QueryParam "name" String :> Post Bool)))’
                  with ‘'True’
    In the expression: mkLink link1 api
```

christian-marie:
I'll fix the tests up on Monday.

christian-marie:
Closed in favor of #10

PR #9, opened 2015-01-27
Jkarni/content types

jkarni:
Do not merge yet - QQ needs to fixed.


jkarni:
I'm happy to merge this without QQ so all the downstream packages will start building against master, and then have the QQ as a separate PR. @alpmestan @tvh @christian-marie review?

tvh:
Where should the instances for `MimeRender` and `MimeUnreder` live? I think they should live in a package that is used both by `servant-server` and `servant-client`. This could either mean a different package `servant-instances`, or just moving them in here. I prefer the latter. Related to this: I think there should be only types in here that are actually supported. Having something that sais `XML` is only confusing if there is no actual implementation backing it.

I also don't like upper bounds where they just guess if it will still work or not. With Stackage we have a powerful tool to catch those broken builds.

Apart from this I like how it looks.

christian-marie:
I agree to all of these points.

I think servant-instances would be not so nice, I'd prefer just sticking them in servant.

As for XML, if it's not got instances it should be added to a servant-xml package with instances. Vice versa probably for anything that isn't JSON.

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+    -- Content-Type and response (serialization of @a@ into the appropriate
+    -- mimetype).
+    handleAcceptH :: Proxy list -> AcceptHeader -> a -> Maybe (ByteString, ByteString)
+
+instance ( AllMimeRender ctyps a, IsNonEmpty ctyps
+         ) => AllCTRender ctyps a where
+    handleAcceptH _ (AcceptHeader accept) val = M.mapAcceptMedia lkup accept
+      where pctyps = Proxy :: Proxy ctyps
+            amrs = amr pctyps val
+            lkup = zip (map fst amrs) $ map (\(a,b) -> (cs $ show a, b)) amrs
```
christian-marie: This is bending my mind a little, so this may not be 100% correct, but isn't this equivalent to:

```haskell
fmap (\(k,(a,b)) -> (k, (cs $ show a, b)) amrs
```

Which is probably equivalent to:
```haskell
(fmap . fmap) (\(a,b) -> (cs $ show a, b))
```

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+    amr :: Proxy ls
+        -> a                              -- value to serialize
+        -> [(M.MediaType, ByteString)]    -- content-types/response pairs
+
+instance ( MimeRender ctyp a ) => AllMimeRender '[ctyp] a where
+    amr _ a = [(contentType pctyp, toByteString pctyp a)]
+        where pctyp = Proxy :: Proxy ctyp
+
+instance ( MimeRender ctyp a
+         , MimeRender ctyp' a        -- at least two elems to avoid overlap
```
tvh: Is this really necessary? Why not 
```.haskell
instanceAllMimeRender '[] a where
    amr _ _ = []
```
If the non-emptyness of ls is a concern, then maybe 
```.haskell
class IsNonEmpty ls => AllMimeRender ls a where
instance ( MimeRender ctyp a ) => AllMimeRender '[ctyp] a where
instance ( MimeRender ctyp a
         , AllMimeRender (ctyp' ': ctyps) a
         ) => AllMimeRender (ctyp ': ctyp' ': ctyps) a where
    amr _ a = (contentType pctyp, toByteString pctyp a)
            : (amr pctyps a)
        where pctyp = Proxy :: Proxy ctyp
              pctyps = Proxy :: Proxy (ctyp' ': ctyps)
```
That way the recursion is clean and there is less code-duplication.

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
```
tvh: Is there a way to go without this?

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+         , AllMimeUnrender ctyps a
+         ) => AllMimeUnrender (ctyp ': ctyp' ': ctyps) a where
+    amu _ val = (contentType pctyp, fromByteString pctyp val)
+               :(contentType pctyp', fromByteString pctyp' val)
+               :(amu pctyps val)
+        where pctyp = Proxy :: Proxy ctyp
+              pctyps = Proxy :: Proxy ctyps
+              pctyp' = Proxy :: Proxy ctyp'
+
+type family IsNonEmpty (ls::[*]) :: Constraint where
```
tvh: Maybe make this into a typeclass instead? I know this looks nice, but it introduces UndecidableInstances as far as I can tell.

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+
+--------------------------------------------------------------------------
+-- * Utils (Internal)
+
+
+--------------------------------------------------------------------------
+-- Check that all elements of list are instances of MimeRender
+--------------------------------------------------------------------------
+class AllMimeRender ls a where
+    amr :: Proxy ls
```
christian-marie: I'd personally prefer if we avoided acronyms such as amr, as it forces people to guess what that means.

src/Servant/API/Post.hs
```diff
@@ -12,6 +14,6 @@ import Data.Typeable ( Typeable )
 -- >            -- POST /books
 -- >            -- with a JSON encoded Book as the request body
 -- >            -- returning the just-created Book
--- > type MyApi = "books" :> ReqBody Book :> Post Book
-data Post a
+-- > type MyApi = "books" :> ReqBody Book :> Post '[JSON] Book
+data Post (cts::[*]) a
```
christian-marie: content_types would be clearer here, to me.

src/Servant/API/ReqBody.hs
```diff
@@ -6,5 +7,5 @@ module Servant.API.ReqBody where
 -- Example:
 --
 -- >            -- POST /books
--- > type MyApi = "books" :> ReqBody Book :> Post Book
-data ReqBody a
+-- > type MyApi = "books" :> ReqBody '[JSON] Book :> Post Book
+data ReqBody (ls::[*]) a
```
christian-marie: Inconsistent naming here? ls

src/Servant/Utils/Links.hs
```diff
@@ -147,24 +160,27 @@ type family IsElem endpoint api :: Constraint where
...
 
--- | A safe link datatype.
--- The only way of constructing a 'Link' is using 'safeLink', which means any
--- 'Link' is guaranteed to be part of the mentioned API.
-data Link = Link
-  { _segments :: [String] -- ^ Segments of "foo/bar" would be ["foo", "bar"]
-  , _queryParams :: [Param Query]
-  } deriving Show
 
+type family IsSubList a b :: Constraint where
```
christian-marie: I'm glad this worked out with a minimum of fuss. I wonder if 'True ~ 'False is needed, though?

Wouldn't you get a better type error if you simply removed that rule and had the unification fail?

src/Servant/API/ContentTypes.hs
christian-marie: We could probably cull some out, but seeing as all of these are used at some point within other modules, it shouldn't really preclude it.

Which leads me to a question: Should we add lens as a dep everywhere? It'd clean up some stuff.

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+         ) => AllMimeUnrender (ctyp ': ctyp' ': ctyps) a where
+    amu _ val = (contentType pctyp, fromByteString pctyp val)
+               :(contentType pctyp', fromByteString pctyp' val)
+               :(amu pctyps val)
+        where pctyp = Proxy :: Proxy ctyp
+              pctyps = Proxy :: Proxy ctyps
+              pctyp' = Proxy :: Proxy ctyp'
+
+type family IsNonEmpty (ls::[*]) :: Constraint where
+    IsNonEmpty '[] = 'False ~ 'True
```
christian-marie: I'd prefer if this simply didn't type, also. 'False ~ 'True will probably make for useless errors. Not confirmed, though.

src/Servant/API/ContentTypes.hs
```diff
@@ -0,0 +1,181 @@
...
+            amrs = amr pctyps val
+            lkup = zip (map fst amrs) $ map (\(a,b) -> (cs $ show a, b)) amrs
+
+
+
+
+--------------------------------------------------------------------------
+-- * Unrender
+class Accept ctype => MimeUnrender ctype a where
+    fromByteString :: Proxy ctype -> ByteString -> Maybe a
```
tvh: Can we please make this be
```.haskell
fromByteString :: Proxy ctype -> ByteString -> Either String a
```

thsutton:
I've opened a pull request on servant-docs which updates it in line with the changes here.

https://github.com/haskell-servant/servant-docs/pull/12

tvh:
There is one for servant-jquery as well.

https://github.com/haskell-servant/servant-jquery/pull/9

src/Servant/API/ContentTypes.hs
jkarni: I agree with using `Either`, but I'm not sure I understand your `UnrenderError` point. If the mime type is not supported, we'll get a `Nothing :: Maybe (Either String ByteString)` in AllCTUnrender, as opposed to a `Just $ Left str` - so we'll know which one it was.

src/Servant/API/ContentTypes.hs
jkarni: Nice catch. No, non-emptiness was a concern at some point, before I decided to separate a proof that it isn't into a separate type family, but now you're right that we could provide an instance for the empty list.

src/Servant/API/ContentTypes.hs
jkarni: UndecidableInstances is also needed for lines 101 and 122 (the constraint isn't smaller than the instance head)

src/Servant/API/ContentTypes.hs
jkarni: I strongly oppose lens in `servant`. `servant` should be lightweight, since it's a dependency for all sorts of things.

If `servant-server` would benefit from it, then I'm not that strongly opposed.

src/Servant/API/ContentTypes.hs
tvh: Yes, you are right. That solution looks good.

jkarni:
ugh, switching to `IsNonEmptyList` constraints has servant-server acting up. I'm investigating further, but hopefully this won't change servant-docs and servant-jquery again...

alpmestan:
Great work guys, this looks good to merge! Let the content-type-driven goodness spread =)

jkarni:
Thanks a lot everyone for the reviews and ideas!

PR #10, opened 2015-01-28
New safe link API

christian-marie:
Rewrite Utils.Links to export a nicer API.

* Add support for parameters.
* Clean up type level magic, invalid links no longer have a type.
* Proxies are now used to be consistent with the rest of the servant ecosystem.


src/Servant/Utils/Links.hs
```diff
@@ -6,77 +7,135 @@
...
+-- >>> let api = Proxy :: Proxy API
 --
--- @
---   type API = Proxy ("hello" :> Get Int
---                :<|> "bye" :> QueryParam "name" String :> Post Bool)
+-- It is possible to generate links that are guaranteed to be within 'API' with
+-- 'safeLink'. The first argument to 'safeLink' is a type representing the
+-- endpoint you would like to point to. This will need to end in a verb like
+-- Get, or Post. The second argument is the API in which you would like to
+-- ensure the endpoint is within. Further arguments may or may not be required
```
jkarni: I think the `or may not` is redundant.

src/Servant/Utils/Links.hs
```diff
@@ -6,77 +7,135 @@
...
 --
---   link1 :: Proxy ("hello" :> Get Int)
---   link1 = proxy
+-- >>> let hello = Proxy :: Proxy ("hello" :> Get Int)
+-- >>> print (safeLink hello api :: URI)
+-- hello
 --
---   link2 :: Proxy ("hello" :> Delete)
---   link2 = proxy
+-- If the API has an endpoint with parameters then we can generate links with
```
jkarni: `optional` parameters?

src/Servant/Utils/Links.hs
```diff
@@ -6,77 +7,135 @@
...
+-- or without those:
 --
---   mkLink link1 API  --  typechecks, returns 'Link "/hello"'
+-- >>> let with = Proxy :: Proxy ("bye" :> QueryParam "name" String :> Delete)
+-- >>> print $ safeLink with api "Hubert"
```
jkarni: How does this work if there are multiple query params, and I want to provide an argument for, say, the second but not the first?

src/Servant/Utils/Links.hs
```diff
@@ -6,77 +7,135 @@
...
-    Or 'False 'False = 'False
-    Or 'True b       = 'True
-    Or a 'True       = 'True
-
-type family And a b where
-    And 'True 'True = 'True
-    And a 'False    = 'False
-    And 'False b    = 'False
+-- | You may use this type family to tell the type checker that your custom type
+-- may be skipped as part of a link. This is useful for things like
```
jkarni: You can do more than just tell it to skip, no? You can describe the logic for determining that something with that combinator is an element of an api quite generally

src/Servant/Utils/Links.hs
```diff
@@ -84,48 +143,190 @@ type family IsElem a s where
...
+    f xs = init xs <> [g (last xs)]
+    -- Modify the segment at the "top" of the stack
+    g :: String -> String
+    g seg =
+        case param of
+            SingleParam k v    -> seg <> ";" <> k <> "=" <> escape (unpack v)
+            ArrayElemParam k v -> seg <> ";" <> k <> "[]=" <> escape (unpack v)
+            FlagParam k        -> seg <> ";" <> k
+
+instance Monoid Link where
```
jkarni: Hmm, not sure about this. Doesn't it provide a way of producing non-safe links?

src/Servant/Utils/Links.hs
```diff
@@ -84,48 +143,190 @@ type family IsElem a s where
...
+--
+-- This function will only typecheck if `endpoint` is part of the API `api`
+safeLink
+    :: forall endpoint api. (IsElem endpoint api, HasLink endpoint)
+    => Proxy endpoint -- ^ The API endpoint you would like to point to
+    -> Proxy api -- ^ The whole API that you this endpoint is a part of
+    -> MkLink endpoint
+safeLink endpoint _ = link endpoint mempty
+
+-- | Construct a link for an endpoint.
```
jkarni: Very cool that this ends up taking the same form as all our other classes!

jkarni:
Wow!

src/Servant/Utils/Links.hs
```diff
@@ -6,77 +7,135 @@
...
--- returns a 'Link', but only typechecks if the link proxy is a valid link,
--- and part of the sitemap.
+-- Attempting to construct a link to an endpoint that does not exist in api
+-- will result in a type error like this:
 --
--- __N.B.:__ 'mkLink' assumes a capture matches any string (without slashes).
+-- >>> let bad_link = Proxy :: Proxy ("hello" :> Delete)
+-- >>> safeLink bad_link  api
+-- <BLANKLINE>
+-- <interactive>:56:1:
```
jkarni: I'll see if I can add some sort of wildcard to doctest to make this less fragile

src/Servant/Utils/Links.hs
christian-marie: Will remove this.

src/Servant/Utils/Links.hs
christian-marie: Simply leave it out of the link you are trying to generate. If it is not needed as per IsElem then it will still typecheck.

src/Servant/Utils/Links.hs
christian-marie: See below.

src/Servant/Utils/Links.hs
christian-marie: Yes, I'm sure you could do all kinds of crazy things. I can't think of a useful thing to do other than skip, though.

src/Servant/Utils/Links.hs
christian-marie: I can remove it quite easily. And will.

src/Servant/Utils/Links.hs
jkarni: Ah, perfect!

src/Servant/Utils/Links.hs
```diff
@@ -84,48 +143,185 @@ type family IsElem a s where
...
+escape :: String -> String
+escape = escapeURIString isUnreserved
+
+-- | Create a valid (by construction) relative URI with query params.
+--
+-- This function will only typecheck if `endpoint` is part of the API `api`
+safeLink
+    :: forall endpoint api. (IsElem endpoint api, HasLink endpoint)
+    => Proxy endpoint -- ^ The API endpoint you would like to point to
+    -> Proxy api -- ^ The whole API that you this endpoint is a part of
```
soenkehahn: Typo: remove the 'you'.

soenkehahn:
This is indeed pretty awesome!

Just some bikeshedding: wouldn't it be more intuitive, if the arguments to `safeLink` were reversed?

One nice thing that you can already do is declaring your own custom link creation function `l`:
``` haskell
type GlobalApi =
       QueryParam "foo" Int -> Get String
  :<|> BarRoute
globalApi :: Proxy GlobalApi = Proxy

type BarRoute = QueryParam "bar" Int -> Get String
barRoute :: Proxy BarRoute = Proxy

l routeProxy = safeLink routeProxy globalApi
```

This then would allow you to use `l` for creating links for your custom api: `l barRoute`. Without having to refer to your global api all the time.

This also would be slightly easier if the arguments to `safeLink` were reversed.

Enough bikeshedding...

jkarni:
@soenkehahn that makes sense to me. @christian-marie ?

christian-marie:
Done. The type signature that needed to go with requires an understanding of what's going on, but there's an example in the docs now.

I also renamed "link" from the typeclass to toLink (to avoid shadowing). This cropped up as we need to expose some more things to make such a partial application possible (to write the type signature).

soenkehahn:
Many thanks!

PR #13, opened 2015-02-20
Add support for application/x-www-form-urlencoded

tvh:
This depends on https://github.com/haskell-servant/servant/pull/9. It support for url encoded request bodys. To use a custom type, There is these typeclasses:

```.haskell
class ToFormUrlEncoded a where
  toFormUrlEncoded :: a -> [(TextL.Text, TextL.Text)]

class FromFormUrlEncoded a where
  fromFormUrlEncoded :: [(TextL.Text, TextL.Text)] -> Either String a
```

### DIFF
--- a/servant.cabal
+++ b/servant.cabal
@@ -94,3 +94,4 @@ test-suite spec
     , servant
     , string-conversions
     , text
+    , url
--- a/src/Servant/API/ContentTypes.hs
+++ b/src/Servant/API/ContentTypes.hs
@@ -12,11 +12,14 @@
 module Servant.API.ContentTypes where
 
 import           Control.Arrow           (left)
+import           Control.Monad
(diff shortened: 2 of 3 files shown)


christian-marie:
I'm going to wait till the dependent PR is merged to review

alpmestan:
Looks good, once #9 is merged in. We'll just have to rebase this PR to make sure travis gives us the go-ahead. Cheers!

jkarni:
Good stuff. Mind adding a few more tests? Since the encoding/decoding is in-house, just checking inverse laws doesn't quite seem like enough.

tvh:
I just added 2 more roundtrip-tests using a different library..

The reason I don't use that library in the first place is
 1. Avoid lots of string-conversions
 2. Avoid extra import for very simple functionality

jkarni:
Thanks!

Issue #14, opened 2015-02-20
Tests for Servant.Common.Text


jkarni:
In particular, check left/right inverse relationships for the instances.

jkarni:
Closed by #16

PR #16, opened 2015-02-20
Fix Integer FromText instance.

jkarni:
And add FromText/ToText tests.


jkarni:
closes #14

jkarni:
@alpmestan can you confirm there wasn't a non-obvious reason for making `Integer` unsigned?

alpmestan:
I confirm, I somehow messed up that one but not the others =)

PR #23, opened 2015-03-09
Canonical types

alpmestan:
Hello folks,

While thinking about the more formal aspects of servant with @jkarni, I mentionned that I was uncomfortable with the fact that we couldn't state some kinds of "laws" for *interpretations* of an API type, like:

``` haskell
Server (a :<|> b) = Server a :<|> Server b -- this one holds
Server (a :> (b :<|> c)) = Server (a :> b) :<|> Server (a :> c) -- this one doesn't!
```

Julian seemed just as comfortable with this as I am and after giving it some thought, we found a solution:

``` haskell
-- | Turn an API type into its canonical form.
--
-- The canonical form is defined and will basically turn:
--
-- > "hello" :> (Get Hello :<|> ReqBody Hello :> Put Hello)
--
-- into
--
-- > ("hello" :> Get Hello) :<|> ("hello" :> ReqBody Hello :> Put Hello)
--
-- i.e distributing all ':>'-separated bits into the subsequent ':<|>'s.
type family Canonicalize api :: * where
  -- requires UndecidableInstances
  Canonicalize (a :> (b :<|> c)) = ((a :> Canonicalize b)  :<|>  (a :> Canonicalize c))
  Canonicalize (a :> b)          = Redex b (Canonicalize b) a
  Canonicalize (a :<|> b)        = Canonicalize a :<|> Canonicalize b
  Canonicalize a                 = a

type family Redex a b c :: * where
  Redex a a first = Canonicalize first :> a
  Redex a b first = Canonicalize (first :> b)

canonicalize :: Canonicalize layout ~ t => Proxy layout -> Proxy t
canonicalize Proxy = Proxy
```

We can now just rely on `Caninocalize` to turn:

``` haskell
Capture "name" String :> (Get User :<|> ReqBody User :> Put User)
```

into

``` haskell
(Capture "name" String :> Get User) :<|> (Capture "name" String :> ReqBody User :> Put User)
```

The `Server` handlers type for the former would be:

``` haskell
String -> (EitherT (Int, String) IO User :<|> (User -> EitherT (Int, String) IO User))
```

whereas for the latter it would be:

``` haskell
(String -> EitherT (Int, String) IO User) :<|> (String -> User -> EitherT (Int, String) IO User)
```

See accompanying PRs for the 4 other libraries in order to automatically guarantee our 2 laws from above by always running `Canonicalize` before "interpreting" the API type with some class/recursion:

- haskell-servant/servant-server#32
- haskell-servant/servant-client#15
- haskell-servant/servant-jquery#10
- haskell-servant/servant-docs#16

### DIFF
--- a/src/Servant/API.hs
+++ b/src/Servant/API.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 module Servant.API (
 
   -- * Combinators
@@ -43,6 +46,9 @@ module Servant.API (
   module Servant.Common.Text,
   -- | Classes and instances for types that can be converted to and from @Text@
 
+  -- * Canonicalizing (flattening) API types
+  Canonicalize,
+
   -- * Utilities
   module Servant.Utils.Links,
   -- | Type-safe internal URIs
@@ -69,3 +75,25 @@ import           Servant.API.ReqBody      (ReqBody)
 import           Servant.API.Sub          ((:>))
 import           Servant.Utils.Links      (HasLink (..), IsElem, IsElem',
                                            URI (..), safeLink)
+
+-- | Turn an API type into its canonical form.
+--
+-- The canonical form is defined and will basically turn:
+--
+-- > "hello" :> (Get Hello :<|> ReqBody Hello :> Put Hello)
+--
+-- into
+--
+-- > ("hello" :> Get Hello) :<|> ("hello" :> ReqBody Hello :> Put Hello)
+--
+-- i.e distributing all ':>'-separated bits into the subsequent ':<|>'s.
+type family Canonicalize api :: * where
+  -- requires UndecidableInstances
+  Canonicalize (a :> (b :<|> c)) = ((a :> Canonicalize b)  :<|>  (a :> Canonicalize c))
+  Canonicalize (a :> b)          = Redex b (Canonicalize b) a
+  Canonicalize (a :<|> b)        = Canonicalize a :<|> Canonicalize b
+  Canonicalize a                 = a
+
+type family Redex a b c :: * where
+  Redex a a first = Canonicalize first :> a
+  Redex a b first = Canonicalize (first :> b)
--- a/src/Servant/API.hs
+++ b/src/Servant/API.hs
@@ -80,7 +80,12 @@ import           Servant.Utils.Links      (HasLink (..), IsElem, IsElem',
 
 -- | Turn an API type into its canonical form.
 --
--- The canonical form is defined and will basically turn:
+-- The canonical form of an API type is basically the all-flattened form
+-- of the original type. More formally, it takes a type as input and hands you
+-- back an /equivalent/ type formed of toplevel `:<|>`-separated chains of `:>`s,
+-- i.e with all `:>`s distributed inside the `:<|>`s.
+--
+-- It basically turns:
 --
 -- > "hello" :> (Get Hello :<|> ReqBody Hello :> Put Hello)
 --
@@ -101,5 +106,5 @@ type family Redex a b c :: * where
   Redex a a first = Canonicalize first :> a
   Redex a b first = Canonicalize (first :> b)
 
-canonicalize :: Canonicalize layout ~ t => Proxy layout -> Proxy t
+canonicalize :: Proxy layout -> Proxy (Canonicalize layout)
 canonicalize Proxy = Proxy
--- a/src/Servant/API.hs
+++ b/src/Servant/API.hs
@@ -48,12 +48,14 @@ module Servant.API (
 
   -- * Canonicalizing (flattening) API types
   Canonicalize,
+  canonicalize,
 
   -- * Utilities
   module Servant.Utils.Links,
   -- | Type-safe internal URIs
   ) where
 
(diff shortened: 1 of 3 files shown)


src/Servant/API.hs
```diff
@@ -69,3 +77,29 @@ import           Servant.API.ReqBody      (ReqBody)
...
+  Canonicalize ((a :<|> b) :> c) = a :> Canonicalize c :<|> b :> Canonicalize c
+  Canonicalize (a :> b)          = Redex b (Canonicalize b) a
+  Canonicalize (a :<|> b)        = Canonicalize a :<|> Canonicalize b
+  Canonicalize a                 = a
+
+type family Redex a b c :: * where
+  Redex a a first = Canonicalize first :> a
+  Redex a b first = Canonicalize (first :> b)
+
+canonicalize :: Canonicalize layout ~ t => Proxy layout -> Proxy t
```
jkarni: Does `Proxy layout -> Proxy (Canonicalize layout)` work? Seems cleaner, and perhaps better for type inference.

jkarni:
LGTM, with or without the change mentioned above.

src/Servant/API.hs
alpmestan: Oh of course, forgot to put that back to `Proxy layout -> Proxy (Canonicalize layout)`. This was just the result of an intermediary implementation of the canonicalization process. I'm about to push this along with some more docs and a changelog entry.

alpmestan:
Just pushed the fix you suggested, a changelog entry + a tentative definition for "canonical type" that you might want to review/reword.

soenkehahn:
Doesn't this increase backtracking?
``` haskell
type Api = ("name" :> foo) :<|> ("name" :> bar)
```
has to do the check for `"name"` twice when `foo` fails, whereas
``` haskell
type Api = "name" :> (foo :<|> bar)
```
only has to do it once. Or am I missing something?

jkarni:
Hmm, right. This could be bad if the LHS happens to be something like `ReqBody`. @alpmestan what do you think? What we *might* be able to do is have an extra instance to handle sharing:

```
instance ... => HasServer (a :> b :<|> a :> c) where
     route ... = <who knows what, but probably a "let">
```
But if not, it might make sense to revert?

thsutton:
I played around with the idea of replacing backtracking search with translation to predicates which check a whole root-to-leaf path in one go. The implementation of such a scheme could be structured to do cheap tests (http method, presence of params, headers, etc.) before more expensive ones (parsing the request body), reducing the likelihood of repeated work.

If this does change get reverted please leave the canonicalisation machinery because I'd quite like to have another go at this alternative approach.

alpmestan:
The canonicalization machinery will stay for sure. I think @soenkehahn is just bothered by server side of things.

Either we use it at the expense of a couple of milliseconds on each request or we end up with `Server MyAPI` possibly being `Text -> (EitherT (Int, String) IO Foo) :<|> (Int -> EitherT (Int, String) IO Blah))`.

I suggest we go for the former *for now* until one of us takes the canonicalized version of API types to produce a nicely optimized routing machinery, I wouldnt really feel comfortable having to explain to someone why they end up having to write handlers with the type above... =)

However as usual all voices count so please let your voice be heard if you disagree.

jkarni:
I don't very much care (though a couple of milliseconds sounds pretty bad to me - I wish we actually had benchmarks to guide use here). 

I do think there must be some better design in the vicinity, though. For instance, adding extra instances of the form `a :> (b :<|> c)` which share the decoding/path-matching/whatever wouldn't suffer the performance penalty (and indeed would offer a performance advantage if we went the *other* way with `canonicalize`). I just don't yet know how to work out `a :> b :> (c :<|> d)`.

mf59816:
hm, is it possible to canonicalize into the other direction, and thus even optimize the code where it is repeating prefices?

    Canonicalize (a :> b :<|> a :> c)) = a :> Canonicalize (b :<|> c)

probably not.  i haven't really followed this issue, just giving it a shot.

alpmestan:
We could, but what we want is not to end up with `Server SomeAPI` generating a type like:
`Text -> (EitherT (Int, String) IO b :<|> (Double -> EitherT (Int, String) IO c))` which would mean you'd have to do:

``` haskell
server :: Server SomeAPI
server txt = handler1 :<|> handler2

  where handler1 = ...
             handler2 dbl = ...
```

However, with a completely canonicalized API type, you can factor everything you want and figure out an optimized routing table for that API. My initial intention with the canonicalization is 1/ to make the obvious "laws" hold and 2/ to avoid the weird handler type issues I've mentionned. It would also probably be used for generating an optimized routing machinery, but for now it's just for 1/ and 2/.

jkarni:
I say let's stay with the canonicalization, and start making tickets for benchmarking, and then optimization. To make sure we actually do it, I also propose not merging anything major after 0.3 before we get a benchmark infrastructure.
@soenkehahn ?

mf59816:
I don't feel i should have a say in this, but just in case: +1  (-:

Possibly relevant:

    git clone https://github.com/fhartwig/pronk -b barely-working-state

this is an extension to pronk that (a) has been observed at least once to work with current hackage, and (b) features state machines for generation of context-dependent request sequences.  It's highly unstable, but we could try moving it upstream if you are interested.

soenkehahn:
Yes, I'm completely fine with this. :+1:

PR #26, opened 2015-03-17
Fix a HasLink instance, also fix float text handling/tests.

### LINKED ISSUE #17: Sporadic test failures
Looks like some of the fromByteString/toByteString and Text tests sometimes fail (empty input, and very very small floats).

christian-marie:
I got distracted by broken tests and fixed #17 whilst correcting a breakage when moving to the new definition of ReqBody with content types. This fixes both.

Need review on:

* Importing GHC.Float, is this OK?
* Conversion is now going via Double for float, as the rational function does some weird round offs, is this OK?
* Are 15 significant bits for Double and 7 for Float a reasonable amount of precision to expect?


jkarni:
Yeah, all of these sound good! Thanks!

Issue #25, opened 2015-03-16
Responce Headers


tvh:
Currently it is impossible to specify in any way which headers are returned. It would be nice if there was a way to specify this. Related issues: https://github.com/haskell-servant/servant-server/issues/25 https://github.com/haskell-servant/servant-client/issues/16.

As far as I can tell a `ResponseHeader X a` similar to `Header` should be implementable by collecting them in a type-level list on the way. I am however unsure about what the response type should be.

christian-marie:
I agree that this seems implementable with a ResponseHeader '["MAGIC1", "MAGIC2"] kind of thing.

alpmestan:
Do we only lack a way to set response headers or are there other things we might want to set in the response? (we already have the response body -- obviously -- and the content-type)

The thing is, it's a bit hard to compose response-related combinators because we conceptually only have one "slot" to describe the response: the last element in a `:>`-separated chain of combinators, which necessarily has to be one of `Get`, `Put`, `Post`, `Delete` or `Patch`. So how would you see it?

``` haskell
-- * first approach
type Api1 = "u" :> Header "X-Preferred-Haskell-Library" Text :> ResponseHeader "X-Preferred-Haskell-Library" Text (Get [JSON] User)
-- we stack headers by "composing" the combinator with itself
type Api1' = ResponseHeader "a" Text (ResponseHeader "b" Text (Get [JSON] User))

-- * second approach
type Api2 = "u" :> Header "X-Preferred-Haskell-Library" Text :> ResponseHeaders [("X-Preferred-Haskell-Library", Text)] (Get [JSON] User)
-- all response headers are specified in the same place

-- * any other approaches in mind?
```

And how would the combinator affect the handler type? `EitherT (Int, String) IO a` becoming `EitherT (Int, String) IO (a, Text)`?

jkarni:
A modified content-type approach should work:

```
type Api a = SetsHeader '[ETag, Location] :> Get '[JSON] a
```
Would check that there is a class, `C`, s.t. `C ETag a` and `C Location a`, which defines a method `mkHeader :: a -> (ByteString, ByteString)`. That way, you don't have to repeatedly tuple anything, and can still use whatever type you want.

alpmestan:
We might want to add another argument somewhere, a "response thats being built up", and have the instances modify it? So that we can tweak the response as we go and keep the combinatorial approach for response-generation.

christian-marie:
[...]

PR #27, opened 2015-03-18
Add Support for response headers

tvh:
This is addressing #25.


alpmestan:
We finally went with #28.

PR #53, opened 2015-04-24
Add a servant-examples dir with 3 "classical" examples (and more to come?)

alpmestan:
- `hackage/hackage.hs` shows how to quickly derive client functions for a couple of JSON endpoints in hackage's API;
- `wai-middleware/wai-middleware.hs` shows how to apply a WAI middleware, in particular the one that logs every request to stdout.
- `auth-combinator/auth-combinator.hs` is a small adaptation of the code presented at the Haskell SG meetup by Soenke & Julian.

I also added `servant-examples` to the sources & the nix file.


jkarni:
Looks good. Probably in the future we should convert to markdown lhs so that this can easily be viewed on github or on the website.

Issue #43, opened 2015-04-20
Use `HTTP.Status.Types.Status` instead of `(Int, String)` for errors


jkarni:
_From @davidsd on April 4, 2015 18:4_

Might it be simpler/safer to return a `Status` instead of `(Int, String)` in the case of a handler error? That way, we could re-use already-defined statuses like `ok200` or `forbidden403`. If we want custom error messages, we could just use `mkStatus` on our own.

_Copied from original issue: haskell-servant/servant-server#37_

jkarni:
I think from discussions on IRC it seems we all think `Left` of the `EitherT` needs a more general overhaul (currently you can return bodies or headers). One suggestion was just having it be a WAI `Response`, but I agree that if we keep status code and reason string, we should combine them into `Status`.

jkarni:
_From @alpmestan on April 5, 2015 10:48_

Yeah but sometimes people also want to give back a response body with the response. Does that mean we should put a `Status` and some response body with its content type in a data type and have this be the Left part of our Either? The more I see it, the more I want to nuke EitherT altogether...

Issue #54, opened 2015-04-24
Make Canonicalize optional


jkarni:
As Andres noted, `Canonicalize` breaks modularity. It shouldn't be the default behaviour. Moreover, we should add something like `Unit` to prevent it from traversing into specific types.

`Canonicalize` should also be named something else (maybe `Distribute`) since now it's no longer the canonical form (there isn't one), and it really just applies one transformation.

PR #57, opened 2015-05-02
Last tasks for 0.3

### LINKED ISSUE #40: Instrument other monads
_From @jkarni on April 19, 2015 10:45_

https://gist.github.com/rschatz/3c0ebd3cbf319c8061a6 shows how we could do this, though I think we could come up with a variation of `HandlerT` that takes a variable number of arguments (not just one) and thus make e.g. `enter`-ing from `Either (Int,String) a` not require a dummy `()` argument, and things that require multiple arguments not have to be curried.

This also would include the appropriate instances for common `mtl`/`tranformers` monads. 

_Copied from original issue: haskell-servant/servant-server#43_

### LINKED ISSUE #46: Handlers can return an error response body
_From @thsutton on February 20, 2015 0:14_

Currently the `(Int, String)` values represent an HTTP status only (i.e. the status code and reason phrase used in the HTTP response status line). It would be nice if servant handlers could supply a body to send to the client when they use `throwError`.

Alternatively, we could allow handlers to specify an HTTP status code and reason when they succeed, as well as fail.

This would make it possible to implement APIs like, for example, OAuth2 which issues 400 Bad Request responses which include a JSON document describing the specific error.

_Copied from original issue: haskell-servant/servant-server#27_

### LINKED ISSUE #52: servant-blaze and servant-lucid
Add packages for HTML. They should only be a handful of lines each.

jkarni:
There's still a little work to do on this PR, but it should be stable enough to start warranting a look. 
  - Fixes #52
  - Fixes #46
  - Addresses #43 
  - Fixes #40


jkarni:
I know I should have broken this up into separate PRs, but the interdependencies always get to be a pain...

jkarni:
And now addresses #54 also (removes `canonicalize` - it'd be nicer to have it be optional, thought).

alpmestan:
Can't we just have the `Canonicalize` type family and:

``` haskell
canonicalize :: Proxy api -> Proxy (Canonicalize api)
canonicalize Proxy = Proxy
```

sit somewhere? So that people can *manually* use it? As in:

``` haskell
type API = ...

api :: Proxy API
api = Proxy

server :: Server API
server = ...

myApp :: Application
myApp = serve (canonicalize api) server
```

?

This prevents the need for wrappers around `ServerT` that use `canonicalize` while still being very simple to "enable".

alpmestan:
Everything else looks good to me. Incredible amount of work, thank you!

jkarni:
I think we'd need some `DontDistribute` datatype, and then instances for it. Now that we don't pretend that there is this "canonical form" though, I think we maybe should step back and ask what transformations are actually *useful* for end-users. I don't have a good sense for the use-case of `canonicalize` yet. 

One problem I'm beginning to see is that there's an overlap developing between (a) the new `Nat` stuff; (b) the Canonicalize, type family approach; and (c) the overlapping instances approach of `() == NoContent` and `Headers ... = <someheaders>`. The `Nat` stuff might require it's own form of `DontDistribute`, and may even be sufficient to encode the headers/`()`. The `Nat` logic only works on the leafs of the tree, whereas arguably canonicalize doesn't; but that, I think, is just because our use-case for `Nat` happened to be leaf modification.

The overlapping instances approach ((c)) is going to land us in trouble in the future. The number of cases is exponential in behaviour.

All of this is to say, my vote is for leaving this until we understand the lay of the land better. Adding it is a backwards-compatible change anyhow.

jkarni:
To add to that: why `distribute`? There are generic-like functions, such as mapping a maybe over every `User`, and then functions that just switch the order of, say, headers and request bodies, and tons of other things; which ones to we want to support out of the box? Why?

jkarni:
Only somewhat relatedly, our tests aren't great. We can add or remove canonicalize without changing test code, *and everything still compiles*

PR #59, opened 2015-05-06
modify delete to allow for response body

### LINKED ISSUE #33: DELETE doesn't allow for response body
_From @Fuuzetsu on March 25, 2015 0:16_

I have an API here that I DELETE to and which returns 200 (so we get `Left "HTTP DELETE request failed with status: 200 - OK"`, another issue) and some body which we can not process because unlike `Put` or `Post`, we can't parametrise the `Delete`.

_Copied from original issue: haskell-servant/servant-client#19_

codedmart:
@alpmestan @jkarni This could use a good check over just to be sure I got this right. Please let me know if I missed anything or did something wrong. All tests are passing in servant, servant-client, servant-docs, servant-jquery, and servant-server. Fixes #33.

### DIFF
--- a/servant-docs/example/greet.hs
+++ b/servant-docs/example/greet.hs
@@ -53,13 +53,8 @@ instance ToParam (MatrixParam "lang" String) where
                   "Get the greeting message selected language. Default is en."
                   Normal
 
-instance ToSample () Greet where
-  toSample _ = Just $ Greet "Hello, haskeller!"
-
-  toSamples _ =
-    [ ("If you use ?capital=true", Greet "HELLO, HASKELLER")
-    , ("If you use ?capital=false", Greet "Hello, haskeller")
-    ]
+instance ToSample () () where
+  toSample _ = Just ()
 
 instance ToSample Greet Greet where
   toSample _ = Just $ Greet "Hello, haskeller!"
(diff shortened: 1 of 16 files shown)


servant-docs/example/greet.hs
```diff
@@ -53,6 +53,14 @@ instance ToParam (MatrixParam "lang" String) where
                   "Get the greeting message selected language. Default is en."
                   Normal
 
+instance ToSample () Greet where
```
jkarni: I don't think this is doing the right thing. `ToSample` should pretty much always have it's first and second arguments the same (there is only one exception, `Headers ls a`, which is kind of sticking out like a sore thumb right now and caused us to add the second param in the first place). In essence, the first argument is what this is a sample for, and the second is what the type of the sample is (this might need to be better documented -- sorry). So in this case, you probably want `ToSample () ()`, or *maybe* another special case, `ToSample () String` with toSample ~~ "NO CONTENT".

jkarni:
This is great! I was feeling pretty bad about releasing with fixing #33... Could you add entries to the changelogs of all the repos this touches?

servant-docs/example/greet.hs
codedmart: Yeah this part seemed wrong to me. I will try your suggestion for now.

codedmart:
I will update the change logs and update this pr. Thanks!

codedmart:
All updated and tests are passing. Let me know if I can do anything else.

jkarni:
Thanks again.

PR #60, opened 2015-05-08
make BaseUrl an argument to 'client' instead of each function produced by 'client'

codedmart:
Let me know if you need anything else on this.


jkarni:
There's a `Setup` binary involved, but otherwise looks good. (I can fix the binary file.)

jkarni:
Thanks again - you've helped a lot getting this out the door!

PR #61, opened 2015-05-08
Fix base url

jkarni:
#60 without Setup

### DIFF
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -36,8 +36,6 @@ import           Network.HTTP.Media
 import qualified Network.HTTP.Types         as H
 import qualified Network.HTTP.Types.Header  as HTTP
 import           Servant.API
-import           Servant.API.ResponseHeaders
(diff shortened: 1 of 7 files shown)


servant-client/src/Servant/Client.hs
```diff
@@ -52,16 +52,17 @@ import           Servant.Common.Req
...
 -- | This class lets us define how each API combinator
 -- influences the creation of an HTTP request. It's mostly
 -- an internal class, you can just use 'client'.
 class HasClient layout where
-  type Client layout :: *
-  clientWithRoute :: Proxy layout -> Req -> Client layout
+  type Client' layout :: *
+  clientWithRoute :: Proxy layout -> Req -> BaseUrl -> Client' layout
 
+type Client layout = Client' layout
```
jkarni: Why this? Why not just `Client` to start with?

alpmestan: Oh, oops. This is a leftover from the short stay of `Canonicalize`.

PR #64, opened 2015-05-09
Tutorial examples

alpmestan:
This PR adds many examples that are studied in the tutorial, along with an executable that lets us run the appropriate server for a given exemple. Hopefully this makes things easier for user to follow along, although as @jkarni suggested to me we could add a script that sets up sandboxes and builds all the libraries and examples so that the user just has to launch this, get a coffee, read the news for a while and then can play with the various parts.

Do not merge yet: we may want to rename the getting-started executable "tutorial". We can discuss potential changes, perform them and merge when we're happy.


servant-examples/getting-started/getting-started.hs
```diff
@@ -0,0 +1,39 @@
...
+app :: String -> (Application -> IO ()) -> IO ()
+app n f = case n of
+  "1" -> f GS1.app
+  "2" -> f GS2.app
+  "3" -> f GS3.app
+  "4" -> f GS4.app
+  "5" -> f GS5.app
+  "6" -> f GS6.app
+  "7" -> f GS7.app
+  "8" -> f GS3.app
```
jkarni: probably should be`GS8`

servant-examples/getting-started/gs8-main.hs
```diff
@@ -0,0 +1,4 @@
+import GS8
+
+main :: IO ()
```
jkarni: why a separate file?

servant-examples/getting-started/getting-started.hs
alpmestan: no that's on purpose, because "8" is about *servant-client*, running against the webservice from the 3rd example.

servant-examples/getting-started/gs8-main.hs
alpmestan: To have separate executables for the client and the server.

jkarni:
LGTM. I guess we can upload everything to hackage after this merge

alpmestan:
Yup =)

PR #71, opened 2015-05-11
Fix missing test modules

### LINKED ISSUE #68: Test suite missing modules
```
Resolving dependencies...
Configuring servant-client-0.4.0...
Building servant-client-0.4.0...
Preprocessing library servant-client-0.4.0...
[1 of 3] Compiling Servant.Common.BaseUrl ( src/Servant/Common/BaseUrl.hs, dist/build/Servant/Common/BaseUrl.o )
[2 of 3] Compiling Servant.Common.Req ( src/Servant/Common/Req.hs, dist/build/Servant/Common/Req.o )
[3 of 3] Compiling Servant.Client   ( src/Servant/Client.hs, dist/build/Servant/Client.o )
In-place registering servant-client-0.4.0...
Preprocessing test suite 'spec' for servant-client-0.4.0...

test/Spec.hs:1:8:
    Could not find module ‘Servant.ClientSpec’
    Perhaps you meant
      Servant.Client (needs flag -package-key servant-client-0.4.0@serva_JSTqdDtJ1hWEvU0Dj4ZDIR)
      Servant.Client (from servant-client-0.4.0@serva_6VXrsrfyUXEJrNxZS1Cd6A)
```

jkarni:
Fixes #68. 
Also adds a little bit of tooling.


jkarni:
The option was either removing or reinstalling in a different sandbox with sdist. Since that was going to be a big blowup in the length of travis checks, I removed `--only-dependencies`. I don't myself see the issue (after all, most, if not all, packages are going to be installed anyhow), but I might be missing something.

soenkehahn:
`bump-versions.sh` does not append trailing zeros when bumping positional version 0, 1 or 2. So for example it will bump to `0.5` not `0.5.0.0`. I don't know if we care about that, but it's inconsistent with the current version.

jkarni:
> bump-versions.sh does not append trailing zeros when bumping positional version 0, 1 or 2. So for example it will bump to 0.5 not 0.5.0.0. I don't know if we care about that, but it's inconsistent with the current version.

I don't particularly care. It'll become consistent before release anyhow, though.

jkarni:
Closing in favour of individual PRs for each change (e.g. #96 and #97)

PR #74, opened 2015-05-13
servant: clean up some white spaces, add link explaining why Or works.

christian-marie:
Should be non-functional

### DIFF
--- a/servant/src/Servant/Utils/Links.hs
+++ b/servant/src/Servant/Utils/Links.hs
@@ -137,10 +137,9 @@ data Link = Link
   } deriving Show
 
 -- | If either a or b produce an empty constraint, produce an empty constraint.
---
--- This works because of:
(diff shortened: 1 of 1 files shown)


servant/src/Servant/Utils/Links.hs
```diff
@@ -137,16 +137,19 @@ data Link = Link
   } deriving Show
 
 -- | If either a or b produce an empty constraint, produce an empty constraint.
+--
+-- This works because of:
+-- https://ghc.haskell.org/trac/ghc/wiki/NewAxioms/CoincidentOverlap
```
jkarni: Won't this show up in the haddocks? I feel like it's a good comment to have in the source code, but not relevant to someone just browsing the haddocks.

christian-marie:
Comments have been actioned.

PR #79, opened 2015-05-14
fix doctests for cabal-install-1.18

soenkehahn:
Not the most elegant code, but I think this is the best fix.


christian-marie:
Thanks! I'm okay with it.

PR #82, opened 2015-05-15
Add support for wai-app-static-3.1.0

### LINKED ISSUE #81: servant-server-0.4.0 broken by API change in system-filepath
Citing from http://hydra.cryp.to/build/841883/nixlog/1/raw:

~~~~~
src/Servant/Utils/StaticFiles.hs:36:41:
    Couldn't match type ‘system-filepath-0.4.13.4:Filesystem.Path.Internal.FilePath’
                   with ‘[Char]’
    Expected type: FilePath
      Actual type: system-filepath-0.4.13.4:Filesystem.Path.Internal.FilePath
    In the first argument of ‘defaultFileServerSettings’, namely
      ‘(decodeString (documentRoot ++ "/"))’
    In the first argument of ‘staticApp’, namely
      ‘(defaultFileServerSettings (decodeString (documentRoot ++ "/")))’
~~~~~

triplepointfive:
system-filepath was just deprecated http://www.yesodweb.com/blog/2015/05/deprecating-system-filepath
This also fixes #81

### DIFF
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -53,10 +53,11 @@ library
       , split              >= 0.2  && < 0.3
       , string-conversions >= 0.3  && < 0.4
       , system-filepath    >= 0.4  && < 0.5
+      , filepath           >= 1
       , text               >= 1.2  && < 1.3
       , transformers       >= 0.3  && < 0.5
       , wai                >= 3.0  && < 3.1
-      , wai-app-static     >= 3.0  && < 3.1
+      , wai-app-static     >= 3.0  && < 3.2
       , warp               >= 3.0  && < 3.1
   hs-source-dirs: src
   default-language: Haskell2010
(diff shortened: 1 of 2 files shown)


jkarni:
Thanks a lot! 

I was thinking of using CPP to continue supporting wai-app-static < 3.1 (people might still be using libraries with that upper bound, and then there'd be woe). I did some of that work [here](https://github.com/haskell-servant/servant/pull/71/files#diff-d39f28a5b4d74389b94f41e137fae2eeR40), but that PR is getting too big and diffuse, and I might not have enough time to fix all the issues with it in the next couple of days, so if you want to add that, I'd be happy to merge!

triplepointfive:
Indeed. It could support both versions, thank you pointed it out!

alpmestan:
@jkarni satisfied with this PR? Can we merge it?

jkarni:
@triplepointfive thanks a lot!

Issue #83, opened 2015-05-15
Cookie Combinator


seanhess:
I'm using JWTs for auth. I'm implementing the logic by hand, which is fine for now, but it is slightly more convenient to have a `Cookie` combinator so I don't have to parse the header by hand. 

Here's an example. It works exactly like `Header`, doesn't assume anything about authentication, it just provides a cookie value to your code. 

https://gist.github.com/seanhess/f4cf71ea58e4483d8ee4

Is there a place for a low-level Cookie combinator like this? I think it could live side-by-side with higher-level authentication code. If so I'd be happy to clean this up and issue a PR.

alpmestan:
Great!

Such a "dummy" Cookie combinator (dummy as in it doesn't do anything fancy) would definitely be something I would like to add support for in all servant packages. The definition would go in a `Servant.API.Cookie` module and we would have instances of all the appropriate classes in the respective packages.

Anyone could then provide a higher level auth machinery on top of `Cookie` or `Header`. This would in fact be useful for the ongoing work on auth.

seanhess:
Sounds good! I'll close this and open a PR soon.

seanhess:
Closed in favor of PR.

jkarni:
Closing again since we have `servant-auth` & `servant-auth-cookie`

Issue #70, opened 2015-05-11
Authentication


aaronlevin:
Hi everyone! 

I thought I'd open up an issue to act as a container for authentication-related discussion. The issue will be considered *closed* when the first iteration of authentication has been implemented, at which point further improvements will have their own issues. 

## Authentication

We want to support various types of authentication for http servers and clients:

- Basic authentication
- Digest authentication
- Cookie-based authentication
- Token-based authentication (for [OAuth 1.0](http://oauth.net/core/1.0/), [OAuth 2.0](http://tools.ietf.org/html/rfc6749), [Amazon HMAC-style](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html), and any others)

Further, we want to account for the following (not a total list):

- maintain type-safety as per servant's design goals, but also allow for flexibility of various authentication methods and protocols (both standardized and not standardized).
- per-[`realm`](https://tools.ietf.org/html/rfc2617) authentication
- per-endpoint or site-wide authentication

Out of scope for this iteration (but worth thinking about):

- role-based authentication

References:

- example [cookie-based authenatication](https://github.com/haskell-servant/HaskellSGMeetup2015/blob/master/examples/authentication-combinator/AuthenticationCombinator.hs) in servant.

I welcome any thoughts, ideas, or suggestions! I will be creating an `auth` branch to feature work and will let everyone know when that's done.

jkarni:
Thanks a lot for opening the issue!

Another thing to think about is whether to have basic auth at all, or just go with digest auth. My inclination is to go with both, in case someone wants to provide authentication without the challenge + nonce dance, since we do provide HTTPS by default.

One design @aaronlevin and I discussed for Basic/Digest auth is as follows. A combinator, `BasicAuth (realm :: Symbol) lookup`, and a class

```
class BasicAuthLookup lkup a | lkup -> a where
   basicAuthLookup :: Proxy lkup -> String -> IO (Maybe a)
```
With
```
instance (HasServer a, BasicAuthLookup lkup val) => BasicAuth realm lookup :> a where
    type (Server (BasicAuth realm lookup :> a)) = val -> Server a
    route = ...
```

[...]

PR #88, opened 2015-05-16
Added Cookie, a low-level combinator similar to Header

seanhess:
See #83. Works just like Header but parses the Cookie field.

### DIFF
--- a/servant/src/Servant/API/Cookie.hs
+++ b/servant/src/Servant/API/Cookie.hs
@@ -1,0 +1,23 @@
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveFunctor      #-}
+{-# LANGUAGE PolyKinds          #-}
+{-# OPTIONS_HADDOCK not-home    #-}
+module Servant.API.Cookie where
+
+import Data.ByteString    (ByteString)
+import Data.Typeable      (Typeable)
+import GHC.TypeLits       (Symbol)
+
+-- | Extract the given cookie's value as a value of type @a@.
+--
+-- Example:
+--
+-- >>>            -- GET /test
+-- >>> type MyApi = "test" :> Cookie "token" Text :> Get '[JSON] Text
+
+data Cookie (sym :: Symbol) a = Cookie a
+                              | MissingCookie
+                              | UndecodableCookie ByteString
+                              deriving (Typeable, Eq, Show, Functor)
+
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -59,6 +59,7 @@ library
       , wai                >= 3.0  && < 3.1
       , wai-app-static     >= 3.0  && < 3.2
       , warp               >= 3.0  && < 3.1
+      , cookie             >= 0.3  && < 0.5
   hs-source-dirs: src
   default-language: Haskell2010
   ghc-options: -Wall
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -38,7 +38,7 @@ import           Network.Wai                 (Application, Request, Response,
                                               requestMethod, responseLBS,
                                               strictRequestBody)
 import           Servant.API                 ((:<|>) (..), (:>), Capture,
-                                               Delete, Get, Header,
+                                               Cookie, Delete, Get, Header,
                                               MatrixFlag, MatrixParam, MatrixParams,
                                               Patch, Post, Put, QueryFlag,
                                               QueryParam, QueryParams, Raw,
@@ -51,6 +51,7 @@ import           Servant.API.ResponseHeaders (Headers, getResponse, GetHeaders,
 import           Servant.Common.Text         (FromText, fromText)
 
 import           Servant.Server.Internal.ServantErr
+import           Web.Cookie                  (parseCookiesText)
 
 data ReqBodyState = Uncalled
                   | Called !B.ByteString
@@ -437,6 +438,35 @@ instance (KnownSymbol sym, FromText a, HasServer sublayout)
     let mheader = fromText . decodeUtf8 =<< lookup str (requestHeaders request)
     route (Proxy :: Proxy sublayout) (subserver mheader) request respond
 
+      where str = fromString $ symbolVal (Proxy :: Proxy sym)
+
+-- | If you use 'Cookie' in one of the endpoints for your API,
+-- this automatically requires your server-side handler to be a function
+-- that takes an argument of the type specified by 'Cookie'.
+-- This lets servant worry about extracting it from the request and turning
+-- it into a value of the type you specify.
+--
+-- All it asks is for a 'FromText' instance.
+--
+-- Example:
+--
+-- >            -- GET /test
+-- > type MyApi = "test" :> Cookie "token" Text :> Get '[JSON] Text
+-- >
+-- > server :: Server MyApi
+-- > server = test
+-- >   where test :: Maybe Text -> EitherT ServantErr IO Text
+-- >         test token = return token
+instance (KnownSymbol sym, FromText a, HasServer sublayout)
+      => HasServer (Cookie sym a :> sublayout) where
+
+  type ServerT (Cookie sym a :> sublayout) m =
+    Maybe a -> ServerT sublayout m
+
+  route Proxy subserver request respond = do
+    let mheader = lookup "cookie" (requestHeaders request)
+        mc = fromText =<< lookup str =<< fmap parseCookiesText mheader
+    route (Proxy :: Proxy sublayout) (subserver mc) request respond
       where str = fromString $ symbolVal (Proxy :: Proxy sym)
 
 -- | When implementing the handler for a 'Post' endpoint,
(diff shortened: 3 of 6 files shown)


seanhess:
I'm having trouble building this locally. I created a sandbox in both the `servant` and `servant-server` folders, built `servant` successfully, then I ran `cabal sandbox add-source ../servant` from `servant-server`, but when I try to build it, I get: 

```
Failed to load interface for ‘Servant.API.Cookie’
There are files missing in the ‘servant-0.4.0’ package,
try running 'ghc-pkg check'.
cabal: Error: some packages failed to install:
servant-server-0.4.0 failed during the building phase. The exception was:
ExitFailure 1
```

alpmestan:
Here's my suggestion (and how I usually do when I work on servant):

First, wipe out all the `.cabal-sandbox` directories and `cabal.config` you have around. Then go to the root of the servant repo and:

``` bash
cabal sandbox init && cabal sandbox add-source servant/ servant-server/ servant-client/ servant-docs/ servant-jquery/ # and -blaze/-lucid if you want to use them
for d in servant*/; do
  cd $d && cabal sandbox init --sandbox ../.cabal-sandbox && cd ..
done
# you can now enter the directory you're interested in and
# issue 'cabal install --enable-tests --dep',
# do some changes, then cabal build / cabal repl
```

Let me know if you encounter any problem with this.

seanhess:
@alpmestan Thank you! In addition to that, I had forgotten to add `Servant.API.Cookie` to the `exposed-modules` of `servant.cabal`. 

Ok, the tests are passing (well, the ones I didn't break. Master doesn't build right now. I switched back to master to make sure I hadn't broken anything). 

What else do I need to add or change to get this merged?

seanhess:
@alpmestan can you review this?

alpmestan:
Hi!

Great work so far! To make `Cookie` an actual first-class citizen of the DSL, we just need to add instances for it for the other interpretations. This would be what's left to do to get this PR merged.

- `HasClient`: I think clients shouldn't have to worry about the value set for a `Cookie`, which would mean we could basically have an `HasClient (Cookie sym a :> api)` instance that just ignores the `Cookie` combinator. Clients probably shouldn't let users control what values to put in cookies. Do you agree?
- `HasDocs`: maybe we could just add a "list of cookie names" to the information we already have about each endpoint, and every time we meet a `Cookie` combinator, we add it there? So that the docs can say "this endpoint uses the following cookie values: *foo*, *bar* and *baz*.
- `HasJQ`: just ignore the cookie combinator, I guess?

jkarni:
Thanks for the PR. A few comments:

What about `SetCookie`? I'd like to have an integrated story about this. 

This also differs from how aaronlevin's `BasicAuth` is attacking this (#70). There we just ask for the library user to provide a lookup function (indexed by dummy datatypes) so that they don't have to figure out which user a cookie is associated with manually on every request. Personally I like that approach, but I think even if one doesn't, we have to be consistent.

jkarni:
One interesting possibility (I'm not saying it's necessarily a good one, or that it'll even work) is trying to make a combinator for middleware (that way you could easily handle both the request and the response with a single combinator, and do both the getting and the setting of the cookie in a consistent and unified way). Then something like [wai-session](https://hackage.haskell.org/package/wai-session-0.3.1/docs/Network-Wai-Session.html) could be tweaked to our needs.

(Exploring the middleware combinator space would also be very interesting from a broader perspective. It'd unify what so far have been distinct things, and allow for middleware to be applied selectively to only some parts of the server easily.)

seanhess:
@alpmestan I can try doing those if you like. I'll be slower at it than you guys will. Of the libraries I've only used `servant` and `servant-server`. 

@jkarni Sure, I could make a `SetCookie`. 

I think you should have both higher level auth plugins as well as a low level cookie plugin, because you can use cookies for things other than auth, or to implement auth strategies that you guys haven't anticipated yet. See #83 for where I brought this up with @alpmestan. Do you disagree? I don't want to write all this up if you guys won't merge it :)

jkarni:
Ok, I see your point, and I agree that this is useful already. What worries me a little is not having a clear story about how this will progress, and so not knowing whether we'll have to break backwards compatibility when we add a combinator that does something like `wai-session`. Or whether we'll end up with a `GetCookie` and `SetsCookie` combinator that don't really relate to (and much less compose to form) as `WithCookie` combinator. But maybe I'm making the mistake of trying to figure out everything before doing anything at all - I defer decision to @alpmestan et al.

I would prefer a different name - probably `GetCookie`?

servant/src/Servant/API/Cookie.hs
```diff
@@ -0,0 +1,23 @@
...
+
+-- | Extract the given cookie's value as a value of type @a@.
+--
+-- Example:
+--
+-- >>>            -- GET /test
+-- >>> type MyApi = "test" :> Cookie "token" Text :> Get '[JSON] Text
+
+data Cookie (sym :: Symbol) a = Cookie a
+                              | MissingCookie
```
jkarni: Are the constructors actually used for anything?

seanhess: No, I just copied `Header` and didn't know if I would need them at the time. I'll see if they come up when I implement `HasClient` etc and if not remove them.

seanhess:
@jkarni Sounds good. Why `GetCookie`? `Header` isn't called `GetHeader`. I was planning on following the API for `Header` almost exactly. 

I'm not confident that I'm skilled enough to implement the setting side, but I can try. I was thinking a syntax similar to `Header`, so `addCookie` and `Get (Cookies '[Cookie "token" Text] User)`. Although, having used that with Header, I'm not sure I love that syntax. It'll look a little crazy if you try to add cookies and headers, right? Any ideas?

jkarni:
@seanhess the setting side is hard! To be honest, I think we (well, my fault, really) screwed up a little with response headers, and I don't know if we want to continue going down that route (the [release announcement](http://haskell-servant.github.io/posts/2015-05-10-servant-0.4-released.html#conclusions) describes some of the problems with the approach). I'd open an issue for us to discuss design options for that.

>Although, having used that with Header, I'm not sure I love that syntax. It'll look a little crazy if you try to add cookies and headers, right? 

Agreed.

> Why GetCookie? Header isn't called GetHeader

That's because we use `Header` in both getting and setting the header. Again, that might not have been the best idea...

servant/src/Servant/API/Cookie.hs
jkarni: Yeah, in `Header` it's used for setting headers more easily, but since I don't think we should worry about setting cookies as part of that PR, I'd leave it out for now.

alpmestan:
I think this can go in but we should think about how we're going to let people set cookies too. Being able to read but not to set them is... kind of weird, isn't it? =)

Can you guys think of any solution that could be used both by the `Header` and `Cookie` machinery, that would allow getting and setting? There must be a nice solution that abstracts over both and would require less work from us.

seanhess:
So I think of the difference between request headers and response headers as being comparable to the difference between a request body and response body. I like the syntax for the response body currently. 

What about this?

    -- send the "something" header with the request
    "request-test" :> ReqHeaders '[Header "something" Text] :> Post ()

    -- set the "something" header in our response
    "response-test" :> ResHeaders '[Header "something" Text] :> Get User

And do the same thing for cookies. 

Is it possible to throw them all in the same array?

        "request-test" :> Request '[Header "something" Text, Cookie "asdf" Text] :> Post ()

I don't really know. I'm just brainstorming.

alpmestan:
We've not forgotten about this. We really would like to have a good story for the whole business of getting/setting headers/cookies (and other things maybe?).

Your code so far definitely constitutes a good beginning. We however ought to find a nicer solution than what we currently have with response headers. One that we can generalize and reuse for `Cookie`s, ideally.

seanhess:
Cool, I agree. Let me know when you come up with something.

BartAdv:
I'm doing auth related things right now too, and would find it quite useful. @seanhess do you have local build with this combinator, or are you not using it yet?

seanhess:
@BartAdv great! Here is the file: https://github.com/seanhess/serials/blob/master/server/Serials/Lib/ServantCookie.hs

And here it is in use: https://github.com/seanhess/serials/blob/master/server/Serials/Api.hs

jkarni:
Closing this since I think we'd prefer a more integrated approach (with auth more generally, and with sessions).

jml:
FWIW, I think Servant _should_ provide an API at this level.

While proper session and auth support would be great, there will always be some circumstances when you need something more specialized.

jekor:
A use case for raw cookies: reading a uid cookie set by nginx.

rvion:
@jekor +1 (or by fb in case of pure client flow login, or by keycloak, etc)

saurabhnanda:
> Closing this since I think we'd prefer a more integrated approach (with auth more generally, and with sessions).

@jkarni possible to elaborate on this? On what level is a more integrated approach expected? Is it mostly in terms of underlying code, which should be common between auth, sessions and raw cookies? Or is in terms of the API surface exposed to library users?

There is definitely a use-case for accessing raw cookies. For example, setting the preferred currency for a unauthenticated user, without using sessions.

adnelson:
Any more movement on this? I've started writing a server using servant and declarative support for cookies would definitely be nice.

jkarni:
@adnelson There's [servant-auth-cookie](https://hackage.haskell.org/package/servant-auth-cookie-0.3.0.2) (and a [tutorial](https://www.stackbuilders.com/tutorials/haskell/servant-auth/) for it). 


There's also my own [servant-auth](https://github.com/plow-technologies/servant-auth), though it's a little early in it's life. 

Does one of those cover your use cases?

Issue #80, opened 2015-05-14
ServantError instance of Exception


PierreR:
Wouldn't make sense to make [ServantError](https://github.com/haskell-servant/servant/blob/master/servant-client/src/Servant/Common/Req.hs#L34) an instance of Control.Exception ?

alpmestan:
I guess we could. Since (for now) we force `EitherT` down people's throats there wouldn't be any use for an `Exception` instance I think. Do you have any? I do agree that we could make an instance just because it makes to though =)

PierreR:
I do. With the current situation I eventually need to map ServantError to a custom applicative one. Ideally I would prefer to have a function that accepts any `Control.Exception` and keep `ServantError` exposed until it reaches that function.

Generally speaking it seems that an exception would add more flexibility.

I have to say I am not convinced the `EitherT` approach is so great either ;-).

alpmestan:
Well with 0.4 you can write handlers in oter monads, which is the first step in improving the situation.

Feel like putting a little PR together?

PierreR:
Sure I can PR the instance for `exception` if you are ok with it. That should be quite easy ;-)

PierreR:
I actually need another change and will PR accordingly if I have your ok.

Could `parseBaseUrl` type goes from  `String -> Either String BaseUrl` to ` String -> Either ServantError BaseUrl` ? I actually would like it to be an `Exception` to improve composability.

alpmestan:
You could even put it in its own module. Regarding `parseBaseUrl` I think it would be ok since it's internal. So if you need to tweak the internals a bit, why not, I'm sure @jkarni won't mind either.

PierreR:
>  So if you need to tweak the internals a bit, why not, I'm sure @jkarni won't mind either.

Is it really internal ? The whole module is [re-exported](https://github.com/haskell-servant/servant/blob/master/servant-client/src/Servant/Client.hs#L20).

jkarni:
@PierreR can you explain a little more the reason for changing the signature of `parseBaseUrl`?

PierreR:
My motivation is to re-conciliate exception handlings from different libraries when I build the command line of my application. Ideally I would like to use one function with this signature `checkError :: Exception e => Doc -> Either e a -> IO a`. 
Of course I could just have  `checkError :: Show  e => Doc -> Either e a -> IO a` and it would work just fine. It is just less precise.

[...]

Issue #51, opened 2015-04-20
Support GHCJS


jkarni:
[this](https://github.com/arianvp/ghcjs-servant-client) is a good starting point.

tvh:
With the change in the error type this got harder in 0.4. The `ConnectionError` leaks the underlying Implementation. Not sure what the best thing would be here, but this is definitely a step back.

alpmestan:
Could you expand on why this got harder?

tvh:
`http-client` won't work with GHCJS because of its dependency on `network`. Previously it was possible to just remove the dependency on `http-client`, as there was nothing in the interface referencing that package. The new `ServantError` however catches the `HttpException`.

alpmestan:
Do you think we can make that bit be conditionally compiled when the compiler is not ghcjs?

tvh:
I would like to avoid that. The issue here is, that any client-side pattern matches would need the same annotations. I think it would be better to replace the exception with a simplified datatype of our own. That could then be populated with appropriate errors from either `XMLHttpRequest` or `http-client`.

alpmestan:
Do you feel like putting a PR together?

tvh:
Can do, but I'd like some input on the granularity. `HttpException` (http://hackage.haskell.org/package/http-client-0.4.11.2/docs/Network-HTTP-Client.html#t:HttpException) exposes a lot of details. I would prefer something simpler.

On actually making GHCJS work: I would probably user `ghcjs-dom` to abstract the foreign calls away. This is a bit of a moving target though. `ghcjs-base` is currently being rewritten and `ghcjs-dom` depends on those changes already.

jkarni:
Some development related to this is happening at https://github.com/plow-technologies/ghcjs-servant-client

soenkehahn:
I just pushed a new branch here: https://github.com/haskell-servant/servant/tree/client-ghcjs

And some explanation about it here: https://github.com/haskell-servant/servant/blob/client-ghcjs/GHCJS.md

I'm not planning to put that much more work into this, but I hope that this branch may serve as a basis for other people to improve the support for `ghcjs`.

tdietert:
Thanks for this. In the next few weeks I may try to dig deeper and fix some of the problems you've outlined in the branch readme.md, but most likely only with respect to the failing test case. I don't feel like playing with javascript enough to tackle the json parsing issue.

[...]

PR #89, opened 2015-05-20
Use MonadThrow instead of Either in the signature of parseBaseUrl

PierreR:
See #80

### DIFF
--- a/servant-client/src/Servant/Common/BaseUrl.hs
+++ b/servant-client/src/Servant/Common/BaseUrl.hs
@@ -2,8 +2,10 @@
 {-# LANGUAGE ViewPatterns #-}
 module Servant.Common.BaseUrl where
 
+import Control.Monad.Catch (MonadThrow, throwM)
 import Data.List
 import GHC.Generics
+import Network.HTTP.Client (HttpException(InvalidUrlException))
 import Network.URI
 import Safe
 import Text.Read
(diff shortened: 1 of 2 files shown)


PierreR:
I have overlooked the tests. So I will need to fix that.

jkarni:
I like the change, but as it stands it would be backwards incompatible. Can we maybe DEPRECATE the old `parseBaseUrl` and have the new one have a different name? (Unfortunately, I can't think of anything that's as good as `parseBaseUrl`...)

soenkehahn:
I would consider keeping the name and force people to manually migrate. They will get a type error that is more or less easy to fix, right?

jkarni:
But that means another major release (for something that seems quite minor).

soenkehahn:
I personally don't see that as a big downside. At least pre `1.*` releases.

soenkehahn:
Or more generally I don't see bumping the second digit in the version for something minor as a problem.

PierreR:
> Unfortunately, I can't think of anything that's as good as parseBaseUrl...
Me neither ... Here are some ideas:
* parseUrl (if users are not expected to use both `http-client` and `servant-client)
* toBaseUrl
* fromString

Personally  I would force people to manually migrate, bump a major release but include this change in a global set of changes that allow users to move away from EitherT completely ...

christian-marie:
I've already switched to Network.URI in the safe link code, I'd like to move this across to Network.URI here too to remove duplication if everyone's OK with that.

For a release tomorrow, though, let's just make a decision on this. I don't personally want to break peoples builds. So I'd prefer to deprecate or do major releases quickly. I don't see a problem with going to 0.5 for this release. People are free to bump their majors and be pleasantly surprised that nothing has actually broken.

As for the function name if we wish to ship two functions that do the same thing that are very similar (which I'm dubious about):

parseUrl may or may not be better, I'd like to deprecate the BaseUrl type and use Network.URI directly. Hence it's going to have its type change but not its name. This makes me vote no.

toBaseUrl seems good, fromString isn't partial though. That said, I think we can combine the best of these two with:

```haskell
{-# LANGUAGE RankNTypes #-}

import Network.URI
import Data.String
import Control.Monad.Catch

newtype ToBaseUrl = ToBaseUrl { toBaseUrl :: forall m. MonadThrow m => m URI }

instance IsString ToBaseUrl where
    fromString _ = 
        ToBaseUrl (throwM $ userError "catch!")

main :: IO ()
main = do
    print (toBaseUrl "http://ponies.io" :: Maybe URI)
    toBaseUrl "http://ponies.io" >>= print
```

Or just

christian-marie:
Wrong button.

PierreR:
I have just add another tiny (and orthogonal) change (the exception instance for ServantError) that I wish to be considered if there is a new `0.5` release.

Cheers

christian-marie:
If there is a 0.5 release it is likely that it will not be this week. Is there anything in your PR that you require on hackage in the near future? If not, I see no reason we can't do this PR properly and break all of the things via a major release.

That is, if you agree that breaking changes are going to clean up the code base/remove duplicate code.

PierreR:
No I am not in a hurry.

Exceptions/error handling is an important topic so I would proceed with care. There is #91 (if it is the way to go) that I believe should be part of `0.5`as well.

tvh:
I don't really like the idea of reusing an exception from `http-client`. This would just make #51 harder.

PierreR:
@tvh I see your point. I have used it because `http-client` is already a dependencies but you're right we don't need such a "complicated" data type.

alpmestan:
To be honest, I don't feel strong about this PR since I *very* rarely use *servant-client*. It's however great that we're little by little polishing many dirty corners of servant, whatever the final solution ends up being. :+1:

jkarni:
@christian-marie agreed that we should try to expose as little as possible of `http-client`.
Has no one yet written an `http-client-compat` package for GHCJS?

PierreR:
I have made a couple changes as suggested. There is a new type "InvalidBaseUrlException" which is isomorphic to String. I only expose the type. Is that OK ?

Travis CI is saying that the build is on Error but it passes on ghc 7.10 and fails for unknown reason on ghc 7.8. I don't know why Travis is so unstable ...

alpmestan:
I've just restarted the build, maybe it's just Travis timing out.

PierreR:
A litlle question: I use `newtype` to define `InvalidBaseUrlException" but maybe I should use `data` instead.?

PierreR:
Tests now pass for both ghc version.

jkarni:
Thanks!

arianvp:
#245 actually shows the tests fail for this change because servant-client's test suite wasn't running properly.

PR #90, opened 2015-05-21
Shahn/cabal fixes second try

soenkehahn:
I've switched to installing everything into the user package db. This seems to work.

We might run into problems at some point because we don't use cabal's constraint solver to figure out the dependencies all at once. I would consider switching to stackage then. But for now this works.

@jkarni: What do you think?


jkarni:
This means we can't really use the script for local development, but I guess we were hoping for too many things at once.
(The StaticFiles changes in my original branch are no longer needed.)
LGTM

PR #94, opened 2015-05-25
Don't export HttpException

tvh:
This simplifies the error type. This allows for `http-client` to be swapped out (#51).

### DIFF
--- a/servant-client/src/Servant/Common/Req.hs
+++ b/servant-client/src/Servant/Common/Req.hs
@@ -46,12 +46,12 @@ data ServantError
     { responseContentType       :: MediaType
     , responseBody              :: ByteString
     }
-  | ConnectionError
-    { connectionError           :: HttpException
-    }
   | InvalidContentTypeHeader
     { responseContentTypeHeader :: ByteString
     , responseBody              :: ByteString
+    }
+  | ConnectionError
+    { connectionError           :: String
     }
   deriving (Show)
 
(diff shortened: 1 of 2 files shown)


alpmestan:
LGTM. @jkarni ?

servant-client/src/Servant/Common/Req.hs
```diff
@@ -148,11 +148,11 @@ performRequest reqMethod req isWantedStatus reqHost = do
...
 
   eResponse <- liftIO $ __withGlobalManager $ \ manager ->
-    catchHttpException $
-    Client.httpLbs request manager
+    catchConnectionError $
+      Client.httpLbs request manager
   case eResponse of
     Left err ->
-      left $ ConnectionError err
+      left $ ConnectionError (show err)
```
christian-marie: Do we have to lose the underlying exception here? Can we just pass this thing along as a SomeException existential? Then the user can recover the exception, or just throw/show it.

jkarni:
Hm, :+1: on moving away from exposing `http-client` internals, but I think `String` is not informative enough. Should we have our own `HttpException` datatype that's basically the same?

christian-marie:
That might fix the things. I'm going to make a PR for this PR so we can merge whilst we merge.

christian-marie:
This is my meta-PR: https://github.com/haskell-servant/servant/pull/118

christian-marie:
Any thoughts on using SomeException?

christian-marie:
Back to you @tvh

tvh:
I only had a preliminary look at the errors returned by `XMLHttpRequest`. It seems to me though that there is a lot less information to be had there. I am all for `SomeException` here. If you want to handle the error, then you are free to do so, without losing information.

christian-marie:
I think this is ready for merge.

alpmestan:
LGTM. @jkarni ?

jkarni:
Sure. I still fear we have less information than before, but if everyone thinks this is fine I'm good with it.

Issue #92, opened 2015-05-21
make safeLink safer


jkarni:
Andres pointed out that:
```
type MyAPI = Capture "id" Int :> Get '[JSON] Int
GHCI> safeLink (Proxy :: Proxy MyAPI) (Proxy :: Proxy ("foo" :> Get '[JSON] Int)
foo
```
Not good. To be truly safe, only `Capture x y` should match `Capture z y`. The `Int` should be provided as an argument.

christian-marie:
Not sure how to regression test this without doctests, which is really nasty. Do we have any better ideas there?

tvh:
What about `hint`?

```.haskell
runInterpreter $ do
    setImports ["Servant.API","Data.Proxy"]
    typeChecks "safeLink (Proxy :: Proxy (Capture \"id\" Int :> Get '[JSON] Int)) (Proxy :: Proxy (\"foo\" :> Get '[JSON] Int)"
````

christian-marie:
That's certainly cleaner than the current approach.

mpickering:
One way to write tests to make sure certain expressions don't type check is to use `-fdefer-type-errors`.

jkarni:
:+1: on `defer-type-errors`.

christian-marie:
Sounds good to me!

jkarni:
Done is #95

PR #95, opened 2015-05-26
More strict matching for Capture (safeLink)

tvh:
This is a fix for #92.

The change itself is straight forward. To test for regression, it is checked (with `-fdefer-type-errors`) that `IsElem` doesn't yield an empty `Constraint`.

### DIFF
--- a/servant/src/Servant/Utils/Links.hs
+++ b/servant/src/Servant/Utils/Links.hs
@@ -72,8 +72,7 @@
 --
 -- >>> let bad_link = Proxy :: Proxy ("hello" :> Delete '[JSON] ())
 -- >>> safeLink api bad_link
--- <BLANKLINE>
--- <interactive>:64:1:
+-- ...
 --     Could not deduce (Or
 --                         (IsElem' (Delete '[JSON] ()) (Get '[JSON] Int))
 --                         (IsElem'
@@ -168,7 +167,8 @@ type family IsElem endpoint api :: Constraint where
     IsElem (e :> sa) (e :> sb)              = IsElem sa sb
     IsElem sa (Header x :> sb)              = IsElem sa sb
     IsElem sa (ReqBody y x :> sb)           = IsElem sa sb
-    IsElem (e :> sa) (Capture x y :> sb)    = IsElem sa sb
+    IsElem (Capture z y :> sa) (Capture x y :> sb)
+                                            = IsElem sa sb
     IsElem sa (QueryParam x y :> sb)        = IsElem sa sb
     IsElem sa (QueryParams x y :> sb)       = IsElem sa sb
     IsElem sa (QueryFlag x :> sb)           = IsElem sa sb


tvh:
Travis fails because of `-WError`. I have no idea how to silence that warning though.

christian-marie:
This seems to make type error messages un-informative and adds noise to the type of safeLink. I think this is complicated enough without extra noise and artificial errors. I'm glad we got something to work, but this isn't quite there I think.

Can we please do the fix for IsElem and think about this further? Maybe you can submit that by itself as a PR. I'm pretty confident that it's the correct fix.

servant/test/Servant/Utils/LinksSpec.hs
```diff
@@ -42,7 +48,7 @@ type BadTestLink'2 = "greet" :> Get '[OctetStream] Bool
 type NotALink = "hello" :> Capture "x" Bool :> Get '[JSON] Bool
 type NotALink2 = "hello" :> ReqBody '[JSON] 'True :> Get '[JSON] Bool
 
-apiLink :: (IsElem endpoint TestApi, HasLink endpoint)
+apiLink :: (IsElem endpoint TestApi ~ (() :: Constraint), HasLink endpoint)
```
jkarni: Is this necessary?

jkarni:
Relatedly, kosmikus came up with some nice refactoring of this in section 5.2 of the [paper](http://www.andres-loeh.de/Servant/servant-wgp.pdf)

christian-marie:
Seems similar to how it was was at some point?

We ended up with the type error that True couldn't unify with False.

I don't particularly want to go back to that. I like not unifying, taking type error information surely is not a good thing.

tvh:
Okay, this now just the change itself. I noticed there are basically no tests for the `safe` in `safeLink`, so we definitely need a story for that.

christian-marie:
Agreed. I think your ... is going to break in the doctest tests, though?

tvh:
I put the `...` in so that it is not dependent on line numbers. It is a wildcard.

christian-marie:
Didn't know that worked, awesome.

christian-marie:
:+1:, would merge. @jkarni, @alpmestan, this could go in an upcoming point release if that wasn't done yet.

kosmikus:
I'm happy with this as an incremental change.

@christian-marie There are two aspects of the paper version that are different from the current code, but independent. (1) The more systematic use of an open top-level type family where all Servant constructs are treated uniformly. It may require some additional cases in total, but in my opinion has the advantage of not giving any unfair (and unnecessary) advantages in treatment to built-in constructs over user-defined ones. (2) The use of Bool over Constraint.

Issue (1) is more important to me. I do like (2), because I think it makes everything clearer and less subtle. I agree that if we apply (2) without any further changes, the error messages get worse. However, in my opinion, there are good ways to recover good error messages on top of (2) by calling wrapper type functions and producing type errors explicitly.

jkarni:
LGTM. We can figure out how to deal with `Werror` and add test cases later.
(Incidentally, Simon and I added `...` to doctest precisely because we noticed how hard to maintain this stuff was going to get.)

PR #101, opened 2015-05-27
Check sdist

jkarni:


alpmestan:
This has been superseeded by another PR, right? Just asking so that we can kill this PR and the associated branch.

jkarni:
No, it hasn't. I'll try fixing it, though.

jkarni:
This is bit-rotting - I'll reopen when I actually get back to it.

PR #105, opened 2015-05-29
Fix outdated docs

jkarni:


alpmestan:
The cabal descriptions don't need to be updated though do they?

jkarni:
What do you mean?

alpmestan:
Just wondering about the examples we have in cabal descriptions and READMEs etc.

jkarni:
No, those are fine.

Issue #73, opened 2015-05-13
Allow failures in arbitrary order


jkarni:
This is something we've discussed quite a bit before, but I just realized there's no ticket for it. We're stuck in this position where failures (from `failWith`) happen without trying matches further down the `:>` chain, resulting in some bad error messages (e.g. `Bad Request` instead of `Method Not Found`), hard to implement authentication ( #70 ), and inefficiency (it's better not to try decoding a body if there's no matching route anyhow).

This relates to #48. It also likely relates to improved routing. 

# Design issues

Do we want only the leaves of the tree (the methods) to be able to adjudicate error messages? Or do we want combinator instances to be *allowed* (but not required) to adjudicate between failure and continuation as well? I think the former would be sufficient to conform to the behaviour of most frameworks, but the second may also allow for room for interesting possibilities (what comes to mind immediately is auth-protecting a set of endpoints without even allowing non-authenticated clients to figure out what the endpoints being protected are - not a crucial feature, but also almost certainly not the only one).

# Implementation ideas

?

aaronlevin:
I think the latter (*combinator instances to be* allowed *(but not required) to adjudicate between failure and continuation as well*) is preferable from an authentication standpoint. It makes sense to wrap sets of endpoints under a specific authentication scheme. 

Something to consider: authentication (and other combinators) may introduce `IO` and other effects *during* the routing process. Therefore, whatever new process is introduced for routing it's important that authors of a web-service are able to reason about the effects that happen during matching. This is why I think the latter choice is a good one.

thsutton:
I have the start of a thing which constructs a traditional check-the-predicate-then-call-the-handler router from an API type in https://github.com/anchor/servant-route.

It has an overlapping instances problem I haven't managed to resolve yet, but I think that an approach like this is more likely to be a "correct" approach than mixing all these success-but-failure responses in with backtracking search.

jkarni:
@thsutton come to the servant channel on freenode, we're discussing just this.

jkarni:
Fixed in #111

Issue #48, opened 2015-04-20
Generalize RouteMismatch


jkarni:
_From @jkarni on January 30, 2015 4:52_

https://github.com/haskell-servant/servant-server/pull/22 adds the ability to have some overriding custom errors in `RouteMismatch`. It'd be nice to generalize this to a more flexible priority system, as discussed in the PR.

_Copied from original issue: haskell-servant/servant-server#23_

jkarni:
_From @christian-marie on January 30, 2015 23:26_

The most obvious and simple way appears (to me) to be just a typeclass, such that you can failWith these things if they can produce an appropriate HTTP response.

jkarni:
_From @alpmestan on January 30, 2015 23:32_

The thing is, we have to be able to "compare" them. Some mismatches mean "stop right here, i don't care about whether other endpoints could match", whereas some aren't as strong and the routing can actually lead to a successful match later on with another endpoint, despite the fact than an earlier endpoint returned a mismatch.

jkarni:
_From @christian-marie on January 31, 2015 0:0_

Right, surely this information can be encoded in the typeclass?

jkarni:
_From @alpmestan on January 31, 2015 10:25_

Have you read the discussion from [this PR](https://github.com/haskell-servant/servant-server/pull/22) ? I'm not sure how we could go with a typeclass, especially considering we would have to look up the "priority" of other mismatches by going to the source code, so not very handy. If you have any idea/suggestion, we're all ears! :)

jkarni:
_From @thsutton on February 1, 2015 8:11_

I wonder if a better solution might be a priority from a small set (as in, e.g., syslog) together with an HTTP status: the priority controls routing behavior (stop now vs backtrack, etc.) and the HTTP status is what gets sent to the client.

This allows HasServer instances to distinguish between "nothing to see here, move along" and "if this was meant for me, the client made mistake X" and "mayday, mayday, mayday: abort now" without having to overload the HTTP status codes with additional meaning or try to tell a nice story about they way you assign Rational or `Ord a` values consistently in your handlers.

jkarni:
"Stop now" vs. "continue" is one axis (`succeedWith` vs `failWith`) ; `RouteMismatch` is another, and I don't think we should or need to intermingle them.

jkarni:
_From @christian-marie on February 1, 2015 19:48_

My intention with the typeclass was that it would have a priority field (ugly, but we've all seen it before).

```haskell

[...]

Issue #42, opened 2015-04-20
Refactoring - abstract away common instance functionality


jkarni:
_From @jkarni on April 14, 2015 14:24_

A lot of the instances share code which could be abstracted into separate functions now that the number of instances is getting pretty large.

_Copied from original issue: haskell-servant/servant-server#40_

jkarni:
Largely done in #111

PR #111, opened 2015-06-04
Improved routing

kosmikus:
I'd like to open this up for discussion and (ideally) merging.

This is to a certain extent ongoing work and should probably be refined further before the next major release, to keep breaking changes for end users to a minimum.

The two main aspects this PR addresses are:

   * Complexity of routing is much better. We don't perform linear testing where possible, but instead dispatch via lookup tables. @jkarni has performed some benchmarking and it looks quite promising.

   * We delay matching of "non-path" components (such as request headers and the request body) and of captures until we are sure that a route otherwise matches. This saves work, and allows us to produce better error codes in many cases. This addresses issue #73, and will hopefully allow progress on #70. (I have tested that the ongoing work on authentication can be ported to this approach.)

We do not yet tackle #48, but that's going to be on top of this PR next.

We also perform some refactoring, which at least partially addresses issue #42.

/cc @jkarni @alpmestan

### DIFF
--- a/servant-examples/auth-combinator/auth-combinator.hs
+++ b/servant-examples/auth-combinator/auth-combinator.hs
@@ -28,14 +28,15 @@ data AuthProtected
 instance HasServer rest => HasServer (AuthProtected :> rest) where
   type ServerT (AuthProtected :> rest) m = ServerT rest m
 
-  route Proxy a request respond =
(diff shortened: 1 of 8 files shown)


alpmestan:
This is one amazing PR. Thanks a lot!

aaronlevin:
:+1:

PR #119, opened 2015-06-12
servant-client: Hide HttpException by wrapping it in SomeException

christian-marie:
This is what I meant by using SomeException

### DIFF
--- a/servant-client/src/Servant/Common/Req.hs
+++ b/servant-client/src/Servant/Common/Req.hs
@@ -53,7 +53,7 @@ data ServantError
     , responseBody              :: ByteString
     }
   | ConnectionError
-    { connectionError           :: String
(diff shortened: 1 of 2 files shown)


alpmestan:
I'm fine with this change personally, although you can't do much with `SomeException` except show it. @jkarni ?

christian-marie:
Well you can point ```fromException :: SomeException -> Maybe e``` at it, right?

I agree that if you could only show it there's no benefit.

christian-marie:
I should write a test that shows this, I'll find some time today.

christian-marie:
Didn't get around to this today, nor does it look like I will soon.

christian-marie:
@jkarni?

jkarni:
LGTM

PR #124, opened 2015-06-16
servant-docs: Fix docsWith.

kantp:
Hi,

thanks for this nice library :)

I noticed that the enpoints vanished from the documentation when adding extra info via docsWith. This pull-request fixes that.

Best,
Philipp

### DIFF
--- a/servant-docs/servant-docs.cabal
+++ b/servant-docs/servant-docs.cabal
@@ -70,6 +70,7 @@ test-suite spec
       base
     , aeson
     , hspec
+    , lens
     , servant
     , servant-docs
     , string-conversions
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -338,7 +338,7 @@ extraInfo p action =
 docsWith :: HasDocs layout => [DocIntro] -> ExtraInfo layout -> Proxy layout -> API
 docsWith intros (ExtraInfo endpoints) p =
     docs p & apiIntros <>~ intros
-           & apiEndpoints %~ HM.unionWith combineAction endpoints
+           & apiEndpoints %~ HM.unionWith (flip combineAction) endpoints
 
 
 -- | Generate the docs for a given API that implements 'HasDocs' with with any
--- a/servant-docs/test/Servant/DocsSpec.hs
+++ b/servant-docs/test/Servant/DocsSpec.hs
@@ -7,7 +7,9 @@
 {-# OPTIONS_GHC -fno-warn-orphans  #-}
 module Servant.DocsSpec where
 
+import           Control.Lens
 import           Data.Aeson
+import           Data.Monoid
 import           Data.Proxy
 import           Data.String.Conversions (cs)
 import           GHC.Generics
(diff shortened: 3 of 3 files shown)


alpmestan:
Hello Philipp and thanks for the PR!

Upon seeing your PR, @jkarni realized that much of that "information merging logic" seems a bit fragile. We're wondering whether we could actually revamp some of that logic and find a smarter, sound & principled solution to this.

So while your PR fixes the problem you were having, we're kind of afraid that someone else might encounter a similar dark corner tomorrow because of a similar problem.

I would be much more comfortable having a lossless information merging logic, where we never drop any bit of information.

tl;dr: thanks for fixing the issue you found, we might merge it as is or use this opportunity to try and find a more robust merging strategy.

kantp:
Hi Alp,

thanks for the quick reply.

I'm actually having some very similar 'lopsided merging' code in one of my own projects, so I would be very interested if you can come up with something more robust, or to participate in a brainstorming phase.  Of course, if I have a sudden flash of wit and find something, I'll let you know as well.

Best,
Philipp

alpmestan:
Would you mind just adding a little test for a minimal-version of the problem you encountered and which gets fixed by this PR? Just to make sure a future rewrite gets it right too.

kantp:
Sure, I'll do that.

kantp:
The CI failure seems to be unrelated to the commit (apparently, the random package failed to install).

kantp:
I also gave the problem of making the merging more robust some thought.

One thing that I think should be possible is to use indexed types, as in REPA (http://www.cse.unsw.edu.au/~benl/papers/guiding/guiding-Haskell2012-sub.pdf). You'd have two different Actions, let's call them Action A and Action B, where Action A is a complete Action containing all information derived from the API datatype. An ExtraInfo would contain Action B, which cannot have information about Responses. 

combining functions would have signatures Action A -> Action B -> Action A and Action B -> Action B -> Action B, but not Action A -> Action A -> Action A, so the type system would prevent you from accidentally discarding information by providing arguments in the wrong order, or by trying to combine two Actions with Reponses.

It might be overkill, but servant already does a lot of magic on the type level, so maybe it fits in. What do you think?

alpmestan:
Re: the build failure, I've just restarted it because it seems to a random timeout or something rather than an actual problem.

alpmestan:
There you go. The build succeeds. So we're going to merge this and make a bugfix release in the upcoming days and then we're probably going to end up with a rewrite in 0.5.

One of most promising directions here is to just rely on something like [pandoc-type's `Block` ADT](https://hackage.haskell.org/package/pandoc-types-1.12.4.4/docs/Text-Pandoc-Definition.html). This is strictly more principled, tested and richer than our little ADTs and very simple to understand. We even have [servant-pandoc](https://hackage.haskell.org/package/servant-pandoc) to guide us in the rewrite, @jkarni =)

Julian: good to merge?

jkarni:
@alpmestan yup on my end.

@kantp hi! and thanks!

alpmestan:
Thanks @kantp!

kantp:
You're very welcome, thanks for handling this so quickly, and for writing the library in the first place!

@jkarni Hi there, hope to see you at the next meetup!

christian-marie:
Cherry picking this.

PR #132, opened 2015-06-18
Tweak the HasLink instance for QueryParam, fixes #129

### LINKED ISSUE #129: Link for 'QueryParam sym v' asks for a value of type 'v' instead of 'Maybe v'
From `Servant.Utils.Link`:

``` haskell
instance (KnownSymbol sym, ToText v, HasLink sub)
    => HasLink (QueryParam sym v :> sub) where
    type MkLink (QueryParam sym v :> sub) = v -> MkLink sub
```

Shouldn't that be

``` haskell
type MkLink (QueryParam sym v :> sub) = Maybe v -> MkLink sub
```

We may very well want to link to the endpoint/page without specifying any value for a query param, which is not possible with the current instance.

alpmestan:
This can only go in 0.5 though, can't be cherry-picked in 0.4.

### DIFF
--- a/servant/src/Servant/Utils/Links.hs
+++ b/servant/src/Servant/Utils/Links.hs
@@ -273,10 +273,10 @@ instance (KnownSymbol sym, HasLink sub) => HasLink (sym :> sub) where
 -- QueryParam instances
 instance (KnownSymbol sym, ToText v, HasLink sub)
     => HasLink (QueryParam sym v :> sub) where
-    type MkLink (QueryParam sym v :> sub) = v -> MkLink sub
-    toLink _ l v =
-        toLink (Proxy :: Proxy sub)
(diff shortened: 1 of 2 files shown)


alpmestan:
@jkarni care to review / merge? I would do it given how trivial the change is, but I prefer sticking to the policy whatever the size of the change is as long as it affects the code.

jkarni:
Do MatrixParams require similar changes?

jkarni:
And I guess this still needs a changelog entry

alpmestan:
Oops, there you go =) Yes, Travis, please get back to work.

alpmestan:
Good to merge? I'll update the changelogs right before the minor bugfix release, and update the ones for the 0.5 branch as well.

jkarni:
Yup!

PR #140, opened 2015-06-23
Fix ordering of CPP directive for aeson test

oswynb:
If the version of aeson used is below 0.9 the current ordering
results in an empty do block.


alpmestan:
Hello,

Nice catch! I've restarted the failing travis build, it errored out because of a timeout.

christian-marie:
+1 nice catch, would merge.

Issue #134, opened 2015-06-19
Host combinator


alpmestan:
This one has been discussed in the paper we submitted to WGP if I remember correctly, and I still have the file around so I thought I might as well propose the inclusion of a `Host` combinator to get the host/ip of the client.

``` haskell
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeOperators #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE ScopedTypeVariables #-}
import Network.Socket (SockAddr)
import Network.Wai
import Servant
import Servant.Client

data Host

instance HasServer api => HasServer (Host :> api) where
  type ServerT (Host :> api) m = SockAddr -> ServerT api m

  route Proxy serv req respond =
    route (Proxy :: Proxy api) (serv h) req respond

    where h = remoteHost req

instance HasClient api => HasClient (Host :> api) where
  type Client (Host :> api) = Client api

  clientWithRoute Proxy req baseurl =
    clientWithRoute (Proxy :: Proxy api) req baseurl
```

Nothing fancy really. Request handlers for endpoints which use `Host` get an argument of type `SockAddr` (from `Network.Socket`).

jkarni:
:+1: Can we also add `isSecure`, `vault`, and maybe `httpVersion`? These are all almost exactly the same as above.

jkarni:
oh, and can we name it `RemoteHost` instead of `Host`?

alpmestan:
Yeah, will do. And yes for the renaming too, it's much clearer. Expect a PR very soon =)

alpmestan:
It's called `RemoteHost` and has been merged as part of #141.

PR #141, opened 2015-06-23
Add HttpVersion, IsSecure, RemoteHost and Vault combinators with inst…

alpmestan:
…ances for all interpretations (all but the server ones being 'ignore that combinator').

Do not merge yet. I need to add a couple of tests that actually make use of those combinators. This is obviously going in 0.5.

### DIFF
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -52,6 +52,7 @@ library
       , network-uri        >= 2.6  && < 2.7
       , mtl                >= 2    && < 3
       , mmorph             >= 1
+      , network            >= 2.6  && < 2.7
       , safe               >= 0.3  && < 0.4
       , servant            == 0.4.*
       , split              >= 0.2  && < 0.3
@@ -63,6 +64,7 @@ library
       , wai                >= 3.0  && < 3.1
       , wai-app-static     >= 3.0  && < 3.2
       , warp               >= 3.0  && < 3.1
+
   hs-source-dirs: src
   default-language: Haskell2010
   ghc-options: -Wall
--- a/servant-server/test/Servant/ServerSpec.hs
+++ b/servant-server/test/Servant/ServerSpec.hs
@@ -37,9 +37,9 @@ import           Test.Hspec.Wai             (get, liftIO, matchHeaders,
 import           Servant.API                ((:<|>) (..), (:>),
                                              addHeader, Capture,
                                              Delete, Get, Header (..), Headers,
-                                             JSON, MatrixFlag, MatrixParam,
-                                             MatrixParams, Patch, PlainText,
-                                             Post, Put, QueryFlag, QueryParam,
+                                             HttpVersion, IsSecure(..), JSON, MatrixFlag,
+                                             MatrixParam, MatrixParams, Patch, PlainText,
+                                             Post, Put, RemoteHost, QueryFlag, QueryParam,
                                              QueryParams, Raw, ReqBody)
 import           Servant.Server             (Server, serve, ServantErr(..), err404)
 import           Servant.Server.Internal.RoutingApplication
@@ -93,6 +93,7 @@ spec = do
   prioErrorsSpec
   errorsSpec
   responseHeadersSpec
+  miscReqCombinatorsSpec
 
 
 type CaptureApi = Capture "legs" Integer :> Get '[JSON] Animal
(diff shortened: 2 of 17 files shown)


servant-server/servant-server.cabal
```diff
@@ -52,6 +52,7 @@ library
       , network-uri        >= 2.6  && < 2.7
       , mtl                >= 2    && < 3
       , mmorph             >= 1
+      , network            >= 2.6
```
jkarni: I'm in favour of upper bounds too - we get reports via stackage as soon as a new version comes out, and between being certainly briefly not up on the latest version and possibly being broken and not knowing it, I'd take the former

alpmestan: Yeah we can add `< 2.7`, no problem, I just wrote that as a placeholder.

servant/src/Servant/API/IsSecure.hs
```diff
@@ -0,0 +1,7 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+module Servant.API.IsSecure where
+
+import Data.Typeable
+
+data IsSecure = Secure | NotSecure
```
jkarni: Can we add the note about what `Secure` means from `Network.Wai`?

alpmestan: Sure. I'm not done on the PR yet, I still have to push some tests for these new combinators and write decent haddocks, `IsSecure` will be documented and link to wai's `Bool`-returning field accessor.

jkarni:
@alpmestan if you don't have time, do you want me to take over this PR and make the remaining fixes?

jkarni:
Closes #134

alpmestan:
I actually wanted to take a stab at tests & docs for this PR today. So unless you really want to do it, I can take care of that =)

alpmestan:
Alright, this PR is probably much closer to being mergeable now. The tests are quite dumb but I just wanted some runnable code that makes use of all the new combinators, there's really nothing fancy we might want to do there. Thoughts?

servant/servant.cabal
```diff
@@ -57,6 +61,7 @@ library
     , text >= 1 && < 2
     , string-conversions >= 0.3 && < 0.5
     , network-uri >= 2.6
+    , vault >= 0.3
```
jkarni: Can we put an upper bound?

servant/test/Servant/API/ContentTypesSpec.hs
```diff
@@ -163,7 +163,6 @@ spec = describe "Servant.API.ContentTypes" $ do
 #if MIN_VERSION_aeson(0,9,0)
     -- aeson >= 0.9 decodes top-level strings
     describe "eitherDecodeLenient" $ do
-
```
jkarni: Any reason to remove this? Seems more consistent with.

servant/servant.cabal
alpmestan: Done.

servant/test/Servant/API/ContentTypesSpec.hs
alpmestan: I've put it back.

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -667,3 +668,41 @@ errorsSpec = do
...
+miscServ = versionHandler
+      :<|> secureHandler
+      :<|> hostHandler
+      :<|> vaultHandler
+
+  where versionHandler = return . show
+        secureHandler Secure = return "secure"
+        secureHandler NotSecure = return "not secure"
+        hostHandler = return . show
+        vaultHandler _ = return ()
```
jkarni: Can we at least be strict on the vault? Otherwise the test really isn't testing anything.

alpmestan: Well, the vault is always there, it's not like this could fail. It's just a field in the `Request`, and not a `Maybe`. But if you insist I can make it be `vaultHandler v = v `seq` return ()`.

alpmestan: Final word on this, @jkarni ?

jkarni: I'd prefer either not having the vault test at all, or having it test *something* ;) If you thinking just seq-ing the vault doesn't actually make a difference, then we maybe shouldn't have the test?

alpmestan: Fine by me. I'll remove it right away.

jkarni:
LGTM (though maybe some commits can be squashed?)

alpmestan:
This commit addresses should fix the travis build and contains the whole PR. I'll merge when travis says it's OK (if it is).

PR #146, opened 2015-07-06
Bugfix release

christian-marie:
WIP, review only, no merge yet.

I'm thinking of naming this: 0.4.3

I think it's a nice name.

If we are all OK with the name and the commits included I'll bump and add some words to the changelog, and make the tests all happy, etc.

PR #148, opened 2015-07-08
servant-jquery: set content type to application/json on POST

### LINKED ISSUE #122: $.ajax should set content-type on post
if we're posting JSON, the content-type should be set explicitly to application/json.

christian-marie:
**I HAVE NO IDEA WHAT I'M DOING**

Can someone please verify that this is the correct thing to do? I don't javascript very often and have never used this library.

Closes #122

Will be included in bugfix release #146

### DIFF
--- a/servant-jquery/src/Servant/JQuery.hs
+++ b/servant-jquery/src/Servant/JQuery.hs
@@ -37,6 +37,7 @@ generateJS req = "\n" <>
  <> "  $.ajax(\n"
  <> "    { url: " <> url <> "\n"
  <> "    , success: onSuccess\n"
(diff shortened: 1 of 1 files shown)


thsutton:
Does it maybe want an `accepts` in there too? http://api.jquery.com/jQuery.ajax/

christian-marie:
Apparently jquery is smart about that (probably looks at the response headers). I didn't want to introduce something that might break things that currently work. That was my rationale for not including accepts, happy to swing either way.

christian-marie:
Also I'm not sure how dataType interacts with accepts, it didn't seem worth messing with.

PR #149, opened 2015-07-08
alternative cotnent-type

bwo:
@alpmestan @christian-marie I would think something like this, but I am also not an expert here, sadly. Who was the original author?

### DIFF
--- a/servant-jquery/src/Servant/JQuery.hs
+++ b/servant-jquery/src/Servant/JQuery.hs
@@ -37,6 +37,7 @@ generateJS req = "\n" <>
  <> "  $.ajax(\n"
  <> "    { url: " <> url <> "\n"
  <> "    , success: onSuccess\n"
+ <> contentType
  <> dataBody
  <> reqheaders
  <> "    , error: onError\n"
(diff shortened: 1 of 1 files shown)


christian-marie:
I'm not sure who the original author was, @jkarni seems to know things about it.

This seems fine too, I don't thing content-type affects anything but posts so it seems cleaner not to make it conditional, is all? Much of a muchness, I won't have anything to say if either of these gets merged.

Could you please fix up your commit message, more along the line of mine in #148?

@alpmestan, @jkarni: Could one of you please merge this or the other (#148) depending on your tastes? I'll then include that in our bugfix release (#146)

christian-marie:
Actually, there is a difference.

```
 If you explicitly pass in a content-type to $.ajax(), then it is always sent to the server (even if no data is sent).
```

So, perhaps this one is more "correct"? I don't think it would break anything either way.

alpmestan:
Well, this PR would do. Could you just make [this warning](https://travis-ci.org/haskell-servant/servant/jobs/70007753#L1127) disappear?

christian-marie:
So that's:
- [ ] Fix the commit message
- [ ] Fix the warning

And we're good to go.

christian-marie:
This PR has been improved upon in #150.

PR #156, opened 2015-07-14
Make HasDocs instance for :<|> overlappable.

jkarni:
That way it's easier to document that the endpoints on the LHS "fall
        through" to the RHS.


jkarni:
@alpmestan good to merge?

PR #159, opened 2015-07-17
Adding a generic way to generate js variants

freezeboy:
Support is added for Vanilla XMLHttpRequest, JQuery and Angular.
Angular generators contain Top Level Functions or with a simple Service.


servant-jquery/src/Servant/JQuery/Angular.hs
```diff
@@ -0,0 +1,121 @@
...
+        prologue :: String -> String -> String,
+        epilogue :: String
+    }
+
+defAngularOptions :: AngularOptions
+defAngularOptions = AngularOptions {
+        serviceName = "",
+        prologue = \svc m -> m <> "service('" <> svc <> "', function($http) {\n"
+                               <> "  return ({",
+                               epilogue = "});\n});\n"
```
jkarni: Weird indentation here.

servant-jquery/src/Servant/JQuery/Internal.hs
```diff
@@ -25,8 +25,29 @@ import GHC.Exts (Constraint)
 import GHC.TypeLits
 import Servant.API
 
+data CommonGeneratorOptions = CommonGeneratorOptions {
+        -- Function transforming function names
+        functionName :: String -> String,
```
jkarni: `functionName` doesn't seem like the right name for this field. `functionNameModifier`, `functionRenamer` or some such?

servant-jquery/src/Servant/JQuery/Internal.hs
```diff
@@ -25,8 +25,29 @@ import GHC.Exts (Constraint)
 import GHC.TypeLits
 import Servant.API
 
+data CommonGeneratorOptions = CommonGeneratorOptions {
```
jkarni: Could you write Haddock documentation for this, and for the individual fields?

servant-jquery/test/Servant/JQuerySpec.hs
```diff
@@ -46,51 +49,107 @@ customHeaderProxy = Proxy
...
+    where
+        specLabel = "generateJS(" ++ (show test) ++ ")"
+        --output = putStrLn
+        output _ = return ()
+        testName = "MyService"
+        ngOpts = NG.defAngularOptions { NG.serviceName = testName }
+        genJS req = NG.wrapInService ngOpts req
+    
+generateJSSpec :: TestNames -> (AjaxReq -> String) -> Spec
+generateJSSpec n gen = describe specLabel $ do
```
jkarni: Can we tweak the tests so that if one fails with a specific `TestName` constructor, we get to see what the constructor is?

(Possibly this could be done by wrapping all the `it`s, in a `context (show name)` do block.

freezeboy: In fact if you run the test, the describe hierarchy seems logical to me, but I don't mind adding the context to be more precise

jkarni: My bad, I didn't see `specLabel`.

jkarni:
This is amazing! Thanks a lot!
We'll need to rename some files, as well as the package itself. 
This isn't necessary, but one thing that might make a lot of the string splicing easier to understand is using `interpolate` rather than just `<>`

freezeboy:
Ok I add this suggestion on my toto-list

alpmestan:
Are we good to go on this? I'll probably clean up and shuffle things around a bit, but I'll probably do so in a subsequent PR, it seems to me we can merge this. I'll make sure it's documented and mentionned in the changelogs in the next PR (I can't push to @freezeboy's fork, so it'd be harder to do it all here).

freezeboy:
As you prefer between pushing to my repo or do two PR.
I'm glad It can be merged

alpmestan:
OK so in the spirit of trying to keep master as clean and "releasable" as possible, I've borrowed your commits and am working further on this in a new branch in this repo. So we'll merge after I'm done with the clean-up phase. I don't think there will be drastic changes, I'll just try to make this release ready.

@jkarni @freezeboy shall we proceed and rename the package on the way? If yes, `servant-js` or `servant-javascript`? Similarly, `Servant.JS` or `Servant.Javascript` for the (main) module name?

freezeboy:
I'd vote for changing the name at the same time, but I let you guys decide which solution is nicer

alpmestan:
@freezeboy Alright, we'll probably go with `servant-js` and `Servant.JS`.

By the way, I'm wondering if we shouldn't make `JavaScriptGenerator` be:

``` haskell
type JavaScriptGenerator = [AjaxReq] -> String
-- instead of:    AjaxReq -> String
```

to give the liberty to the generators to map over the reqs the way they want, adding stuffs before and after if needed etc. I feel like this would greatly simplify the code for Angular, for instance. What do you think?

freezeboy:
@alpmestan  That's a good idea, it removes the need for the specific "wrapInService" function.
I used the AjaxReq -> String type because I was just 'patching' the original version, let's even mainstream the service this way.

alpmestan:
This is now part of #161 which extends upon this work.

PR #160, opened 2015-07-21
Fix isElem for Header.

jkarni:
The intention was obviously to ignore them, but we weren't doing so
    correctly.


alpmestan:
There's an actual error here, but we can't see it entirely on travis -- "Last 10 lines".

jkarni:
@alpmestan fixed.

PR #163, opened 2015-07-23
newpackage: servant-mock, to generate mock servers from API types

alpmestan:
As discussed in [this WIP post](https://github.com/haskell-servant/haskell-servant.github.io/blob/hakyll/posts/2015-07-22-pulling-mock-servers-out-of-thin-air.md), Julian and I came up with this idea so I just went ahead and made a new package for this, with some decent docs and an example. I'd say that if the build succeeds, we can probably already merge this in master, unless anyone has some alternative solution to suggest or tweaks to do on the code?


ggreif:
It is fun to see this :-) Incidentally the last few days I've been implementing a small web-frontend for a maths trainer based on QuickCheck I wrote a while back. `servant` has been a painless experience for me. Here is the code: https://github.com/ggreif/leleka/blob/master/Main.hs

jkarni:
Looks great to me. Simple, and very useful!

servant-mock/servant-mock.cabal
```diff
@@ -0,0 +1,46 @@
+name:                servant-mock
+version:             0.4
```
jkarni: Can we have the same version number as everything else? (It'll get bumped soon enough, to be fair...)

alpmestan: 0.4.3.1?

jkarni: 0.4.1, but point well taken. Doesn't really matter :)

alpmestan:
Good to merge?

jkarni:
Yup! Looks great

PR #167, opened 2015-07-25
add stack.yaml and use it in travis file

soenkehahn:
Please, don't merge yet, opening this PR for discussion.

`travis` is throwing an error because definitions from `cabal_macros.h` are not available. This should be fixed in `stack-0.1.2.1` (on master). Here's the build:

https://travis-ci.org/haskell-servant/servant/builds/72568559

So we can either wait for the next stack release or try to get rid of the cabal macros. I think at least some of them could be avoided through using `base-compat`.

Also, I haven't figured out caching yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ cabal.config
 *.aux
 *.hp
 Setup
+.stack-work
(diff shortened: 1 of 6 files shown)


soenkehahn:
I will split this up into two separate PRs. Closing this.

PR #168, opened 2015-07-25
add stack file

soenkehahn:


soenkehahn:
Please review. Shouldn't interfere with normal cabal uses.

alpmestan:
Could you just add the very young [servant-mock](https://github.com/haskell-servant/servant/tree/master/servant-mock) package to stack.yml? Looks good otherwise.

soenkehahn:
Done and force-pushed.

Issue #143, opened 2015-06-23
Generated JavaScript function naming collision


parsonsmatt:
Given an API type:
```haskell
type ExampleAPI = "resources" :> Get '[Text] [String]
             :<|> "resources" :> Capture "id" Int :> Get '[Text] String
```
The generated jQuery looks like:
```js
function getresources(onSuccess, onError) { /* ... */ }
function getresources(id, onSuccess, onError) { /*...*/ }
```
Since JavaScript doesn't have function overloading, the second function overrides the first one.

This is related to #39, and I think a solution to that issue would fix this as well. The code that I'd *like* to generate would look something like:
```js
function getResource(id, onSuccess, onError) {}
function getResources(onSuccess, onError) {}
```
where the name generator can infer (or is told about) a singular vs collection resource and pluralize the function name appropriately. In addition to a camelCase formatter, you could pass a function to resolve name collisions. We'd also want this to be generalizable for languages with other semantics around function overloading. Java would have no problem with the above example, for instance.

I'd like to contribute here, and I have some ideas for how it might be handled. Before getting started, I'd like to know what other people's takes are on this, and what the best solution might be.

A current work-around is to give the resources different names, but this breaks RESTful resource naming in URLs: `resources` vs `resource/:id`

alpmestan:
Hello @parsonsmatt,

Yes, the way we generate function names is hacky at best. But we needed something, to get the ball rolling and actually publish this package. The problem being that with the huge focus of everyone on *servant* and *servant-server*, we haven't taken much time to re-engineer this bit of *servant-jquery*, one of the reasons, I think, being that it's not used all that much.

If you are willing to give it a shot, I can help you understand anything unclear in the code and would stay around for any question. I unfortunately don't have time to tackle this myself these days but would more than glad to help you do it. The first thing you have to do really is "simply" to find an algorithm to generate  the function name just using whatever combinators you see on the way.

The way it works now is that in many (but not all) `HasJQ` instances, we add something to the front of the "current name", [e.g](https://github.com/haskell-servant/servant/blob/master/servant-jquery/src/Servant/JQuery/Internal.hs#L218):

``` haskell
instance Elem JSON list => HasJQ (Delete list a) where
  type JQ (Delete list a) = AjaxReq

  jqueryFor Proxy req =
    req & funcName  %~ ("delete" <>)  -- HERE: we prepend "delete" to the current function name
        & reqMethod .~ "DELETE"
```

[...]

PR #171, opened 2015-07-27
Replace functionName with an array, and use functionRenamer to build …

freezeboy:
…the real function name

### DIFF
--- a/servant-js/examples/www/angular/index.html
+++ b/servant-js/examples/www/angular/index.html
@@ -21,11 +21,11 @@ counterApp.controller('CounterCtrl', ['$scope', '$http', '$interval', function (
   var publishCounter = function (response) { $scope.counter = response.value; };
   
   $scope.getCounter = function() {
-    getcounter($http)
+    getCounter($http)
         .success(publishCounter);
   }
   $scope.incCounter = function() {
-    postcounter($http)
+    postCounter($http)
         .success(publishCounter);
   }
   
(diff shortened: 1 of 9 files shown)


freezeboy:
In order to resolve the conflicts I would also suggest to add in the instance for Capture, to add some "By{Capture}"

For instance:
```haskell
data MyAPI = "users" :> GET ...
                :<|> "users" :> Cap "id" Int :> GET
````
would generate (with camelcase)
```javascript
  var getUsers;
  var getUsersById;
```

And even maybe:
```haskell
data MyAPI = "groups" :> Cap "id" Int :> "users" :> Cap "name" Int :> GET
````
This ?
```javascript
  var getGroupsByIdUsersByName;
```

freezeboy:
Could someone with the permission restart the travis test because I don't see the reason why it failed

alpmestan:
Wouldn't we want to make the `camelCaseRenamer` the default one? That's, I'd say, the most commonly used convention. I'm also wondering we could find a better word than "renamer" -- since it doesn't actually "rename" but barely *puts the pieces of the name together*.

Aside from these two points, looks good to me!

freezeboy:
I vote for CamelCase by default, I let you find a better name for the Renamer part

codedmart:
@alpmestan @freezeboy Why not `concatCase` `snakeCase` `camelCase` just drop renamer? camelCase for sure.

alpmestan:
@codedmart Sounds good to me. @freezeboy ?

freezeboy:
very smart solution :)

freezeboy:
Fixes #143

alpmestan:
Great, thanks (again)! =)

PR #178, opened 2015-07-31
Change Raw now works with Enter.

### LINKED ISSUE #157: Enter and Raw don't interact well
As per the error [illustrated here](https://gist.github.com/codedmart/c965ca9cbd158802bfc8#file-error), or [this one](https://stackoverflow.com/questions/31020207/serving-static-files-with-servant-wai) where you can see _the same (cryptic) error message_, `Enter` and `Raw` don't play well together. That's because Enter tries to "peef off a monad stack" of `Application`, which reduces to trying to peel a monad stack off of `Request -> (Response -> IO ResponseReceived) -> IO ResponseReceived`, which reduces to trying to peel a monad stack off of `IO ResponseReceived` per the "distribute the arguments" instances of `Enter`, which usually won't work.

We need an instance for `Enter` that says that whatever the source and destination monads for `Enter`, if we encounter `Raw`/`Application` please leave it alone, as-is.

Something like:

``` haskell
instance Enter Application (m :~> n) Application where
  enter _ app = app
```

codedmart:
Reviews welcome, this should resolve #157.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/Enter.hs
+++ b/servant-server/src/Servant/Server/Internal/Enter.hs
@@ -8,7 +8,9 @@
 {-# LANGUAGE TypeFamilies           #-}
 {-# LANGUAGE TypeOperators          #-}
 {-# LANGUAGE UndecidableInstances   #-}
+#if !MIN_VERSION_base(4,8,0)
 {-# LANGUAGE OverlappingInstances   #-}
+#endif
 module Servant.Server.Internal.Enter where
 
 #if !MIN_VERSION_base(4,8,0)
@@ -51,12 +53,19 @@ instance C.Category (:~>) where
     id = Nat id
     Nat f . Nat g = Nat (f . g)
 
+instance
+#if MIN_VERSION_base(4,8,0)
+    {-# OVERLAPPING #-}
+#endif
+    Enter (m a) (m :~> n) (n a)  where
+        enter (Nat f) = f
 
-instance Enter (m a) (m :~> n) (n a)  where
-    enter (Nat f) = f
-
-instance (Raw m' ~ m, Raw n' ~ n) => Enter (m a) (m' :~> n') (n a)  where
-    enter _ (Raw a) = Raw a
+instance
+#if MIN_VERSION_base(4,8,0)
+    {-# OVERLAPPABLE #-}
+#endif
+    (Raw m' ~ m, Raw n' ~ n) => Enter (m a) (m' :~> n') (n a)  where
+        enter _ (Raw a) = Raw a
 
 -- | Like `lift`.
 liftNat :: (Control.Monad.Morph.MonadTrans t, Monad m) => m :~> t m
--- a/servant-server/src/Servant/Server/Internal/Enter.hs
+++ b/servant-server/src/Servant/Server/Internal/Enter.hs
@@ -8,9 +8,7 @@
 {-# LANGUAGE TypeFamilies           #-}
 {-# LANGUAGE TypeOperators          #-}
 {-# LANGUAGE UndecidableInstances   #-}
-#if !MIN_VERSION_base(4,8,0)
 {-# LANGUAGE OverlappingInstances   #-}
-#endif
 module Servant.Server.Internal.Enter where
 
 #if !MIN_VERSION_base(4,8,0)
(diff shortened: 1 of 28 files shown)


alpmestan:
LGTM! I definitely think that despite the code breakage that this is going to introduce, it will overall make "the servant experience" less surprising. This `Enter` instance is probably worth it.

@jkarni Feel free to merge if/when you're happy with this PR.

codedmart:
@jkarni I can rebase/fix conflicts Monday morning. Any feedback or suggestions on this one?

servant-server/src/Servant/Utils/StaticFiles.hs
```diff
@@ -35,10 +36,14 @@ import Filesystem.Path.CurrentOS (decodeString)
...
-serveDirectory :: FilePath -> Server Raw
-serveDirectory =
+serveDirectoryWith :: StaticSettings -> ServerT (Raw Application m) n
+serveDirectoryWith settings = Raw (staticApp settings)
+
+serveDirectory :: FilePath -> ServerT (Raw Application m) n
+serveDirectory = serveDirectoryWith . defaultFileServerSettings .
 #if MIN_VERSION_wai_app_static(3,1,0)
-    staticApp . defaultFileServerSettings . addTrailingPathSeparator
+  id .
```
jkarni: Not that is matters much, but couldn't you use `#if !MIN... decodeString .` and not have an `else` clause (and no `id .`)?

codedmart: Yes good point. Thanks!

servant/src/Servant/API/Raw.hs
```diff
@@ -11,4 +12,5 @@ import Data.Typeable (Typeable)
 -- In addition to just letting you plug in your existing WAI 'Application's,
 -- this can also be used with 'Servant.Utils.StaticFiles.serveDirectory' to serve
 -- static files stored in a particular directory on your filesystem
-data Raw deriving Typeable
+newtype Raw a (m :: * -> *) = Raw a
```
jkarni: I'd write out all the instances `Identity` (from [Data.Functor.Identity](https://hackage.haskell.org/package/base-4.8.0.0/docs/Data-Functor-Identity.html)) here. And maybe also have an `unRaw` record.

I think the documentation should say something like 

> the `m` type parameter is a phantom type used to allow <http://hackage.haskell.org/package/servant-server/docs/Servant-Server.html#g:4 `enter`> to treat endpoints uniformly. 

We're going to need a better explanation for that, but I haven't really figured out how best to phrase it...

servant/test/Servant/Utils/LinksSpec.hs
```diff
@@ -26,7 +26,7 @@ type TestApi =
   :<|> "put" :> Put '[JSON] ()
   :<|> "post" :> ReqBody '[JSON] 'True :> Post '[JSON] ()
   :<|> "delete" :> Header "ponies" String :> Delete '[JSON] ()
-  :<|> "raw" :> Raw
+  :<|> "raw" :> Raw () IO
```
jkarni: Can all these `IO`s be just `m`? Semantically, I would expect it to be `EitherT ServantErr IO`, but I guess when `enter` isn't used it doesn't matter?

servant/src/Servant/API/Raw.hs
jkarni: [Tagged](http://hackage.haskell.org/package/tagged-0.7.2/docs/Data-Tagged.html) may have a few more instances that we should also add.

jkarni:
I mentioned this on IRC, but having `Raw` take an explicit param gives us the ability to have documentation that's more tailored to the specific details (e.g. that it servers a specific directory, or that it's a websocket endpoint). I think maybe that should be a different PR, but if so, we should open an issue so we don't forget.

codedmart:
@jkarni @alpmestan I added `unRaw` and removed the redundant id in `StaticFiles.hs`. If you do come up with a solution to swap `Raw` I am happy to finish this up, but I will be out for the remainder of this week and can't get to it till next week. Feel free to make the changes if you need it done sooner. Thanks!

codedmart:
I rebased master into here as well so it is up to date.

codedmart:
@jkarni I thought I had left this here before, but http://lpaste.net/138722#line27. I am not sure what is going on, but it is only a problem with ghc 7.8. Any thoughts/ideas?

servant-server/src/Servant/Server/Internal/Enter.hs
```diff
@@ -33,13 +36,15 @@ import           Servant.API
 class Enter typ arg ret | typ arg -> ret, typ ret -> arg where
     enter :: arg -> typ -> ret
 
+data IsRaw
```
jkarni: This is no longer necessary, I think

jkarni:
@alpmestan @kosmikus Any thoughts on the error codedmart posted above? It seems 7.8 is picking the *more* general instance, which is baffling to me.

servant-server/src/Servant/Server/Internal/Enter.hs
codedmart: Yes I will remove it. Thanks!

alpmestan:
No clue to what's going on here. What if we switch the `OVERLAPPABLE`/`OVERLAPPING`? I.e `OVERLAPPABLE` for the "general case" and `OVERLAPPING` for the `Raw`s?

codedmart:
@alpmestan That won't matter because `OVERLAPPABLE/OVERLAPPING` only apply to 7.10.

alpmestan:
Oh right the problem is with 7.8, sorry.

codedmart:
@jkarni @alpmestan Don't merge yet if the tests pass. As per my convo with @jkarni I am going to add a few more complex api test to check Enter changes.

jkarni:
Closing in favour of #248

PR #185, opened 2015-08-09
Authentication

aaronlevin:
This implements the initial ideas of our Authentication framework, including a simple example of how it's used for `Basic` authentication.

The basic design is outlined here: https://github.com/haskell-servant/servant/issues/70#issuecomment-122256550 

Most of the changes are additive. However, we've added a new constructor to `RouteMismatch` that holds a generic `Response` object. This required the removal of both the `Eq` and `Ord` instances from `RouteMismatch`. This wasn't too destructive, but is probably the only controversial change (aside from design choices for authentication itself).


aaronlevin:
You can see usage of the authentication framework in one of the tests here: https://github.com/aaronlevin/servant/blob/auth/servant-server/test/Servant/ServerSpec.hs#L758

aaronlevin:
I need to update the `auth-combinator` example.

aaronlevin:
@alpmestan so, I think everything is done. I haven't implemented the auth stuff in `servant-client` or some of the other things in the `servant` ecosystems, but the server stuff is complete. 

Would you like to review the PR?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -51,8 +51,21 @@ data RouteMismatch =
   | WrongMethod        -- ^ a more informative "you just got the HTTP method wrong" error
   | UnsupportedMediaType -- ^ request body has unsupported media type
   | InvalidBody String -- ^ an even more informative "your json request body wasn't valid" error
-  | HttpError Status (Maybe BL.ByteString)  -- ^ an even even more informative arbitrary HTTP response code error.
-  deriving (Eq, Ord, Show)
+  | HttpError Status [Header] (Maybe BL.ByteString)  -- ^ an even even more informative arbitrary HTTP response code error.
+  | RouteMismatch Response -- ^ an arbitrary mismatch with custom Response.
+
+instance Show RouteMismatch where
+    show = const "hello"
```
alpmestan: Can you remove this instance?

aaronlevin: ah, yes, sorry. left it in there by accident!

aaronlevin:
build failed with this error:

```
abal: Error: some packages failed to install:
hspec-2.1.10 depends on hspec-discover-2.1.10 which failed to install.
hspec-discover-2.1.10 failed during the configure step. The exception was:
/home/travis/.cabal/setup-exe-cache/setup-Simple-Cabal-1.22.2.0-x86_64-linux-ghc-7.10.1:
runProcess: runInteractiveProcess: exec: resource busy (Text file busy)
```

Perhaps it just needs to be run again.

alpmestan:
Triggered a rebuild.

alpmestan:
Looks good to me. I would have preferred to have `HasClient`, `HasDocs` etc instances as well, but I'm happy to merge this to get it in and then work on those other instances separately. Folks have been working on this for a while so it'd be great to see this merged now.

Given that I participated to the design for the code in this PR, I'll let @jkarni make the final review/call.

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -7,59 +7,78 @@
...
 
-type PublicAPI = Get '[JSON] [PublicData]
+    notAuthenticated :: CookieAuth -> IO Response
+    notAuthenticated (CookieAuth cookie) = return $
+        responseBuilder status401 [] ("Invalid cookie: " <> byteStringCopy cookie)
 
-type API = "private" :> AuthProtected :> PrivateAPI
-      :<|> PublicAPI
+-- | 'AuthData' is a typeclass that provides a method to extract authentication
+-- data from a 'Reqest'
```
jkarni: `Request`

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -7,59 +7,78 @@
...
+instance AuthData CookieAuth where
+    authData req = fmap CookieAuth (lookup "Cookie" (requestHeaders req))
 
+-- | some data we will return from our API that is protected
```
jkarni: `Some`

aaronlevin:
@alpmestan I'd like `HasClient` and `HasDoc`, too, but I think we should merge this and I'll create two issues to implement those and can try to commit some time myself or potentially someone else can. I know there is a deadline by end of august to release `0.5`.

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -7,59 +7,78 @@
...
 import Network.Wai
 import Network.Wai.Handler.Warp
 import Servant
+import Servant.API.Authentication
 import Servant.Server.Internal
+import Servant.Server.Internal.Authentication (strictProtect, AuthHandlers(AuthHandlers))
 
--- Pretty much stolen/adapted from
--- https://github.com/haskell-servant/HaskellSGMeetup2015/blob/master/examples/authentication-combinator/AuthenticationCombinator.hs
+-- | An example of a custom authentication framework that checks a Cookie for a
```
jkarni: I'd move this to the very top.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -232,6 +259,75 @@ instance
...
+
+    route _ subserver = WithRequest $ \req ->
+        route (Proxy :: Proxy sublayout) $ do
+            -- Note: this may perform IO for each attempt at matching.
+            rr <- routeResult <$> subserver
+
+            case rr of
+                -- Successful route match, so we extract the author-provided
+                -- auth data.
+                Right authProtectionStrict ->
```
jkarni: Would there be less drift to the right margin if we used `MaybeT IO` as the monad here, and then just checked for `Nothing` once at the end?

servant-server/src/Servant/Server/Internal/Authentication.hs
```diff
@@ -0,0 +1,107 @@
...
+                        , subServerStrict :: subserver
+                        }
+
+-- | concrete type to provide when in 'Lax' mode.
+data instance AuthProtected authData usr subserver 'Lax =
+    AuthProtectedLax { checkAuthLax :: authData -> IO (Maybe usr)
+                     , subServerLax :: subserver
+                     }
+
+-- | handy function to build an auth-protected bit of API with a Lax policy
```
jkarni: `'Lax'` or `@Lax@`

servant-server/src/Servant/Server/Internal/Authentication.hs
```diff
@@ -0,0 +1,107 @@
...
+laxProtect :: (authData -> IO (Maybe usr)) -- ^ check auth
+           -> subserver                    -- ^ the handlers for the auth-aware bits of the API
+           -> AuthProtected authData usr subserver 'Lax
+laxProtect = AuthProtectedLax
+
+-- | handy function to build an auth-protected bit of API with a Strict policy
```
jkarni: `Strict` or `@Strict@`

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -51,8 +51,18 @@ data RouteMismatch =
   | WrongMethod        -- ^ a more informative "you just got the HTTP method wrong" error
   | UnsupportedMediaType -- ^ request body has unsupported media type
   | InvalidBody String -- ^ an even more informative "your json request body wasn't valid" error
-  | HttpError Status (Maybe BL.ByteString)  -- ^ an even even more informative arbitrary HTTP response code error.
-  deriving (Eq, Ord, Show)
+  | HttpError Status [Header] (Maybe BL.ByteString)  -- ^ an even even more informative arbitrary HTTP response code error.
+  | RouteMismatch Response -- ^ an arbitrary mismatch with custom Response.
+
+-- | specialized 'Less Than' for use with Monoid RouteMismatch
```
jkarni: Is there a reason for not declaring this `Ord`?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -51,8 +51,18 @@ data RouteMismatch =
   | WrongMethod        -- ^ a more informative "you just got the HTTP method wrong" error
   | UnsupportedMediaType -- ^ request body has unsupported media type
   | InvalidBody String -- ^ an even more informative "your json request body wasn't valid" error
-  | HttpError Status (Maybe BL.ByteString)  -- ^ an even even more informative arbitrary HTTP response code error.
-  deriving (Eq, Ord, Show)
+  | HttpError Status [Header] (Maybe BL.ByteString)  -- ^ an even even more informative arbitrary HTTP response code error.
+  | RouteMismatch Response -- ^ an arbitrary mismatch with custom Response.
```
jkarni: Does `Eq` not make sense on `Response`s?

servant-server/src/Servant/Server/Internal.hs
aaronlevin: @jkarni there are two failures here represented by the Nothing:

1. request is missing authentication (e.g.. doesn't have a header)
2. request has authentication, but authentication fails (e.g. wrong password)

in each case there is a different `Response`. So, even though the failure handling is the same (`failWith (RouteMismatch resp)`) the `resp` is different in each case.

I could wrap it in a `EitherT e IO` with a custom `e` or we can rejig the `AuthProtected` data family instance to use `EitherT e IO` as opposed to returning values like `IO (Maybe a)`.

servant-server/src/Servant/Server/Internal/Authentication.hs
aaronlevin: My haddocks knowledge is really weak. Is `@Lax@` preferred?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: `Response` doesn't (and cannot) have an `Ord` instance.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: The `ResponseRaw` constructor makes `Eq` impossible for `Response`. See: https://hackage.haskell.org/package/wai-3.0.3.0/docs/src/Network-Wai-Internal.html#Response

servant-server/src/Servant/Server/Internal/Authentication.hs
alpmestan: No, `'Lax'` is. It will link to `Lax`'s haddocks.

alpmestan:
@aaronlevin Could you add CHANGELOG entries, while you're at it?

aaronlevin:
@alpmestan done.
@jkarni updated the haddocks changes you suggested.

aaronlevin:
@alpmestan @jkarni any remaining blockers?

alpmestan:
LGTM. We can merge if @jkarni doesn't have any other feedback.

jkarni:
LGTM, but as @alpmestan mentioned, I'd prefer `HasClient` etc. as well. As I see it, we should try to keep master always releasable. @aaronlevin if you don't have time for that, let us know and I think I'll have time for that today or tomorrow.

aaronlevin:
@jkarni I move across the ocean in two weeks. there is a small chance I might get a spare moment, but I'm entering that "starting to feel overwhelmed" territory. So, I'd say I won't have time :(

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -728,3 +758,70 @@ miscReqCombinatorsSpec = with (return $ serve miscApi miscServ) $
...
+                liftIO $ do
+                    decode' (simpleBody response) `shouldBe` Just jerry
+
+            it "rejects requests with the incorrect username and password" $ do
+                authGet "/foo" base64UserColonPassword `shouldRespondWith` 401
+                authGet "/bar" base64UserColonPassword `shouldRespondWith` 401
+
+            it "does not respond to non-authenticated requests" $ do
+                get "/foo" `shouldRespondWith` 401
+                get "/bar" `shouldRespondWith` 401
```
jkarni: We should also have 403 changes (I can do this when working on `HasClient` etc - just wanted to make a note here before I forget).

aaronlevin: So, when I was reading the `Basic` spec it did not mention any `403` errors. The `Authentication` framework is flexible enough to allow generic `Response` codes, but for the case of `Basic` there are no `403` errors (as far as I could see - open to being wrong here).

aaronlevin:
@jkarni I am in the waiting room at a hospital (long story) and have some time today. Working on `servant-client` implementation.

aaronlevin:
@jkarni do you want to review the `servant-client` implementation?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -94,6 +100,24 @@ type Api =
...
+base64ServantColonServer = "c2VydmFudDpzZXJ2ZXI="
+
+type AuthUser = T.Text
+
+basicAuthCheck :: BasicAuth "realm" -> IO (Maybe Person)
+basicAuthCheck (BasicAuth user pass) = if user == "servant" && pass == "server"
+                                       then return (Just $ Person "servant" 17)
+                                       else return Nothing
+
+instance AuthenticateRequest (BasicAuth realm) where
```
alpmestan: Don't we want this instance exposed in the `Servant.Client` module?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -248,6 +275,15 @@ spec = withServer $ \ baseUrl -> do
...
         Right val -> getHeaders val `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
 
+    it "Handles Authentication appropriatley" $ withServer $ \ _ -> do
+      (Arrow.left show <$> runEitherT (getPrivatePerson (BasicAuth "servant" "server"))) `shouldReturn` Right alice
+
+    it "returns 401 when not properly authenticated" $ do
+      Left res <- runEitherT (getPrivatePerson (BasicAuth "xxx" "yyy"))
+      case res of
+        FailureResponse (Status 401 _) _ _ -> return ()
+        _ -> fail $ "expcted 401 response, but got " <> show res
```
jkarni: "expected"

servant-client/test/Servant/ClientSpec.hs
aaronlevin: Good call. I'll try to find some time to add that.

aaronlevin:
@alpmestan now exporting `BasicAuth` instance.

alpmestan:
Your travis build failed (but not because of you). Just restarted it.

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -1,65 +1,85 @@
...
+-- | Our full API as a type with authentication
+type API = AuthProtect CookieAuth User 'Strict :> "private" :>  PrivateAPI
+      :<|> PublicAPI
+
 api :: Proxy API
 api = Proxy
 
 server :: Server API
-server = return prvdata :<|> return pubdata
+server = strictProtect isGoodCookie cookieAuthHandlers (const (return prvdata))
```
teh: Trying to understand how one would access `usr` in the subserver. Maybe it's worth extending example with that? Fantastic work btw!

aaronlevin: Where the 'const' is you can have a function that takes 'usr', for example: '\usr -> return prvdata'

But when I get time I will extend it as you suggest to make it more explicit.

alpmestan: @teh FWIW, there's an example [here](https://github.com/haskell-servant/servant/issues/70#issuecomment-122256550), at the top of the code box in my comment, for `Lax`-protection (i.e no erroring out when the user is not authenticated, just passing a `Maybe usr` argument equal to `Nothing` to the handlers. `Strict` is similar but passes a `usr` directly and errors out when the user is not authenticated.

luigy:
any thoughts on having a way to "sign" a `Request` in `servant-client`?

maybe having another field in `Req`
```hs
-- for `BasicAuth` it can default to `return` in `defReq`
signRequest :: MonadIO io => Request -> io Request
```


then in [`reqToRequest`](https://github.com/haskell-servant/servant/blob/4238a58f926a4b5886e5e245a477795ce3118a44/servant-client/src/Servant/Common/Req.hs#L100-L102)
```hs
reqToRequest :: (Functor m, MonadThrow m, MonadIO m) => Req -> BaseUrl -> m Request
reqToRequest req (BaseUrl reqScheme reqHost reqPort path) =
    setheaders . setAccept . setrqb . setQS <$> sign =<< parseUrl url
    where 
      sign = signRequest req
```

aaronlevin:
@jkarni @alpmestan hey, so I'm in Berlin now. my moving hasn't totally settled but I wanted to know if I can merge this or not. I am working on including documentation, but I'd like to get this merged and I can submit another PR shortly with documentation if you'd like. 

let me know!

aaronlevin:
PS - by "documentation" i mean `servant-docs`

jkarni:
I'd prefer to have everything in one PR.

aaronlevin:
@alpmestan @jkarni ok, so I pushed the `servant-docs` changes. I didn't really know what the best approach to this was as authentication is kind of global but applies to only specific endpoints. So, the main idea is:

- a new class `ToAuthInfo` that needs to be implemented for any `AuthProtected` type (usually you'd only need to do this once, or twice depending on how many policies you're using)
- when generating docs, each endpoint gets authentication information prepended to the docs. So, it currrently says: "this endpoint is protected" and then there is some user-supplies information as per `ToAuthInfo` where the api author can state some details about what is expected for authentication.

aaronlevin:
@jkarni @alpmestan It'd be nice if the author only needed to implement `ToAuthInfo` for their own `AuthProtected` once, instead of for each `policy` as a policy thing is only pertinent to api authors. However, I don't know how to do this. Essentially, I want something like:

```haskell
instance ( HasDocs sublayout
         , ToSample auth auth
         , ToSample usr usr
         , Or (policy ~ 'Strict) (policy ~ 'Lax)    -- ^ how to do something like this?
         , ToAuthInfo (AuthProtect auth usr policy)
         )
```

arianvp:
@aaronlevin @jkarni  maybe its useful if i can push my changes onto Aarons
fork so we have digest author in the same PR?

alpmestan:
@arianvp That would be very handy indeed, if @aaronlevin doesn't mind giving you push access to his fork.

@aaronlevin 

Final step guys... adding an auth section to the tutorial =)

alpmestan:
Oh wait, we'll probably want -js/-mock support. Pinging @freezeboy for the -js bits =)

@aaronlevin Actually, don't have any feedback except that it seems you don't export `AuthenticationInfo` from `Servant.Docs`. Aside from that, LGTM.

aaronlevin:
@arianvp I added you as a collaborator on my fork. feel free to merge your changes!

arianvp:
@aaronlevin Awesome thanks. I'll start rebasing my changes

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -728,3 +758,70 @@ miscReqCombinatorsSpec = with (return $ serve miscApi miscServ) $
...
+base64UserColonPassword = "dXNlcjpwYXNzd29yZA=="
+
+authGet :: ByteString -> ByteString -> WaiSession SResponse
+authGet path base64EncodedAuth = Test.Hspec.Wai.request methodGet path [("Authorization", "Basic " <> base64EncodedAuth)] ""
+
+authRequiredSpec :: Spec
+authRequiredSpec = do
+    describe "Servant.API.Authentication" $ do
+        with (return $ serve authRequiredApi authRequiredServer) $ do
+            it "allows access with the correct username and password" $ do
```
arianvp: This doesn't really test anything as it tests the behaviour of the example API used in the test (which is defined in line 771-779) , not the actual authentication API itself.   A  400 or a 401 or a 500 could still have a body which theoretically could decode to `Just alice` 
Something like:
`authGet "/foo" base64ServantColonServer `shouldRespondWith` 200` seems a lot more rigid

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -728,3 +758,70 @@ miscReqCombinatorsSpec = with (return $ serve miscApi miscServ) $
...
+                response <- authGet "/foo" base64ServantColonServer
+                liftIO $ do
+                    decode' (simpleBody response) `shouldBe` Just alice
```
arianvp: This doesn't really test anything as it tests the behaviour of the example API used in the test (which is defined in line 771-779) , not the actual authentication API itself.   A  400 or a 401 or a 500 could still have a body which theoretically could decode to `Just alice` 
Something like:
```
authGet "/foo" base64ServantColonServer `shouldRespondWith` 200
```
seems a lot more rigid

aaronlevin: You have a good point. However, one subtlety is that there are two authentication realms on this api (`foo-realm` and `bar-realm`) and if we just return a `200` are we sure we are authenticating for the appropriate realm? One realm protects an endpoint that returns only `alice :: Person` while the other returns `jerry :: Animal`.

aaronlevin: I agree that perhaps `alice` and `jerry` are not good proxies for ensuring we are accessing the right endpoints in the right realms, so perhaps a specific `PlainText` response would be more appropriate?

aaronlevin:
Just checking in on everyone!

@arianvp : how is your merge coming? Do you need some support from me?
@freezeboy : were you able to look at possible `-js` and `-mock` integration? Can I help at all?

freezeboy:
@aaronlevin Sorry, couldn't find some time to play with it, but I do see a challenge.

There are some things to be careful about:
Each auth type will produce a "value pair": e.g. 
  - "WWW-Authenticate" and "Basic base64(login:pwd)"
  - "WWW-Authenticate" and "Digest md5(...)"
  - the Token stuff (header + value)

The problem is that depending on the js library, the way used to define a header is not always the same.
I don't see a clean way to handle cookie based auth (without breaking other cookies).

arianvp:
@aaronlevin I've been running in some trouble with digest auth.  For digest auth to work (if I've read the spec correctly), the server needs to remember what Nonce it sent to the client and make sure the client hasn't modified this in the next request. This means we need to persist some sessions state across a request which is extremely annoying.

The flow simplified flow is:

* Client sends normal request
* Server responds with `401` and a `nonce`
* Client sends a `username:realm:password+nonce+cnonce`  to the server
* Server checks if nonce
* Server responds with 200

You could short-circuit this flow by the following:

* The client generates both nonce and cnonce
* Client sends a `username:ream:password+none+cnonce` hash to server
* Server responds with 200

see how the client doesn't request a nonce from the server but still succesfully authenticates? I'm not sure if this is a valid flow. and I can't really deduce it from the RFC. Maybe I'm misreading.

It seems kinda weird that a client can arbitiraly choose its own nonce. But otoh, it seems weird the server needs to persist state across requests.




For the JWT part, I have that locally and all seems to work, and I'm planning on merging that to your branch soon.

arianvp:
Discussed it with @jkarni on IRC. apparently my second scenario seems to be the desired behaviour (as it's currently implemented). 

`HasClient` seems to be a bit more tricky though . As the client needs to make two requests . One to get the nonce, one to actually authenticate. And I don't really see how to do that with the current `servant-client`  .

Anybody got some thoughts?

aaronlevin:
@arianvp hmm, I'm a bit confused. Do you have to performer two requests (as per your initial flow) or one request (per your second flow that you discussed with @jkarni)?

arianvp:
@aaronlevin A well-behaving client should ask for a nonce from the server.

Some servers (read: not servant)  generate the nonce by  hashing its content with a secret key, similar to JWT. Currently in servant-server, a nonce is randomly generated and  isn't checked at all. meaning a client could choose any random string as the nonce value and it would work..    We could in the future hash an expiry date with a secret key to generate the nonce, for example.  This means a client cannot know what a well-formed nonce is without asking it to the server for a well-formed nonce.

So if we want servant-client to work with other servers than servant-server, then we need to have a second roundtrip for requesting a nonce.

Also it's probably better if servant-server generates "non-guessable" nonces as well. to be more spec coherent.

arianvp:
@aaronlevin I have rebased master onto auth. Hence I'll require to force-push to auth. Please let me know if this okay. I don't want to conflict with any local changes that you might have made.

arianvp:
Either `jwt` needs to be added to `extra-deps` or we can use http://haddock.stackage.org/lts-3.8/jose-jwt-0.6.2/Jose-Jwt.html instead. Might require some rewriting.

jkarni:
I'm okay with the extra dependency. Recursively it seems like we're taking in very little new.

3noch:
I'm interested in seeing this merged into master. Is there an ETA?

arianvp:
@3noch "soon".   Once I'm finished with exams I'll rebase onto master, rewrite the code to work with the latest master and update documentation.   I can't make any promises but I hope to finish it somewhere next week.

aaronlevin:
@arianvp how far have we deviated from `master`?

3noch:
@arianvp Great. Hope exams go well.

KaneTW:
Just dropping in to say that I'm interested in this, too.

aaronlevin:
to keep people updated: I'm currently rebasing onto `master`.

aaronlevin:
@alpmestan  @jkarni  @arianvp 

I'm wondering if it's too late to make a refactoring. I'm running into some issues with incorporating this into the `Delayed` framework and it's hitting a consistent issue I've had with the authentication API as it stands.

### Background

Right now we have the `AuthProtected` data family that looks like:

```haskell
data family AuthProtected authData usr subserver :: AuthPolicy -> *
```

And we have two instances:

```haskell
data instance AuthProtected authdata usr subserver 'Strict = 
  AuthProtectedStrict { checkAuthStrict :: authData -> IO (Maybe usr)
                      , authHandlers :: AuthHandlers authData -- methods that just return IO Response
                      , subServerStrict :: subserver
             }
```

note that we don't do much with the `subserver` other than old it in memory.

### The Problem

When we write instances of `HasServer` we end up putting the entire api into the `subserver` param, which makes it difficult to extract the authentication handlers without doing `IO`. So our `HasServer` instance looks like:

```haskell
    type ServerT (AuthProtect authdata usr 'Strict :> sublayout) m = AuthProtected authdata usr (usr -> ServerT sublayout m) 'Strict
```

Here instead of having a function that returns `ServerT sublayout m` we have the entire result type wrapped in `AuthProtected`. This has made it an issue to actually retrieve the `AuthProtectedStrict`.

It always bugged me, but I figured there was no other way until I realized that we never actually use the `subserver` in `AuthProtected`.

### Solution

Note that we never really use the `subserver` from within the `AuthProtected` instance. So, instead, I'd like to do this how we do all our `Capture`s:

```haskell

-- | remove the subserver part form the AuthProtected data family
data family AuthProtected authData usr :: AuthPolicy -> *

-- | an example instance:
data instance AuthProtected authdata usr subserver 'Strict = 
  AuthProtectedStrict { checkAuthStrict :: authData -> IO (Maybe usr)
                      , authHandlers :: AuthHandlers authData -- methods that just return IO Response
                      }

-- | the new ServerT type.
    type ServerT (AuthProtect authdata usr 'Strict :> sublayout) m = AuthProtected authdata usr  'Strict -> ServerT sublayout m
```

I wouldn't need to change very much and it would make dealing with values of the `AuthProtected` type much, much easier and idiomatic.

aaronlevin:
With my suggestion above, we can also integrate Authentication better with the new `Delayed` type.

kosmikus:
@aaronlevin I'm somewhat confused here. In the original version, it looks like whoever writes the handler is responsible for creating a value of type `AuthProtected`, so they have to *provide* a function of type `authData -> IO (Maybe usr)`. In the new version, it looks like the handler *receives* a value of type `AuthProtected` and in particular a function of type `authData -> IO (Maybe usr)`. Where would this function be coming from?

aaronlevin:
@kosmikus you are correct, the person implementing the server needs to provide a value of `AuthProtected` when implementing the handler. However, this is a static value, so it can just be threaded through (or perhaps provided via `Reader` monad using `Enter`?)

Otherwise, the only way I can see to grab the current `AuthProtected` data as it stands is to pattern match on `Delayed` and run the actions in an arbitrary order. Ala:

```haskell
route proxy (Delayed captures method body server) = do
    a <- captures
    b <- body
    case server a b of
        Fail _ -> ...
        FailFatal -> ...
        RouteResult authProtectedData -> ... -- finally we have our authprotected value!
``` 

This is not great for a few reasons:

1. `Delayed` is a nice framework to handle the ordering problem of status codes, but we subvert that here (I think? perhaps propagating failure is what we want, but it seems tat `401` should have higher priority according to the `Delayed` docs)
2. it's not clear how to integrate authentication into `Delayed`.

thoughts?

aaronlevin:
(note I purposefully left out running the `method` check to illustrate ow this fails).

aaronlevin:
I did some more work tonight and I think I can get this working with `Delayed`, although it requires me to run `runDelayed` to extract the `AuthProtected` data.

arianvp:
@aaronlevin What still needs to happen to make this work? I might have time to take alook.

aaronlevin:
@arianvp I just pushed changes to use the new `Delayed` type. I need to do the same for the `Lax` police then integrate this into `HasServer` and then update the examples. Also it looks like we'll need to rebase on `master` again. 

Maybe take a look at this most recent commit and see if it changes anything for you. I had to change one of our arbitrary return types into `ServantErr` to make it more compatible with the upstream changes.

aaronlevin:
slog: just rebased onto `master`

aaronlevin:
@arianvp the updated `jwt` tests are failing for some reason. it may or may not be related to my changes. can you take a look?

arianvp:
@aaronlevin  the 7.8.4 fail seems to be an AMP issue (`pure` is not in scope).  and   the 7.10 is failing on identation error in `ClientSpec` it seems. But I'll have a look

aaronlevin:
@arianvp I can fix those errors (but feel free if you want! :smile:), but if you run `cabal test` in `servant-server` you'll see the `jwt` errors I'm talking about.

arianvp:
@aaronlevin Tests should pass now =)

servant-foreign/src/Servant/Foreign.hs
```diff
@@ -37,3 +37,274 @@ module Servant.Foreign
 
 import Servant.API
 import Servant.Foreign.Internal
+import           Control.Lens (makeLenses, (%~), (&), (.~), (<>~))
```
jkarni: Hmm, looks like a git issue?

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -1,32 +1,46 @@
...
 
-data AuthProtected
-
-instance HasServer rest => HasServer (AuthProtected :> rest) where
-  type ServerT (AuthProtected :> rest) m = ServerT rest m
+-- | Response handlers: what do we do when authentication doesn't work.
+cookieAuthHandlers :: AuthHandlers CookieAuth
+cookieAuthHandlers = AuthHandlers missingAuth notAuthenticated
+  where
+    missingAuth :: IO Response
```
jkarni: Shouldn't these be `ServantErr` instead of `Response`?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -217,6 +273,15 @@ sucessSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
         Left e -> assertFailure $ show e
         Right val -> getHeaders val `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
 
+    it "Handles Authentication appropriatley" $ withServer $ \ _ -> do
```
jkarni: nitpicks: `appropriately`, and lowercase `handles`.

servant-client/src/Servant/Client/Authentication.hs
```diff
@@ -0,0 +1,31 @@
...
+    authReq :: a -> Req -> Req
+     
+
+instance AuthenticateRequest (BasicAuth realm) where
+    authReq (BasicAuth user pass) req =
+        let authText = decodeUtf8 ("Basic " <> encode (user <> ":" <> pass)) in
+            addHeader "Authorization" authText req
+
+instance AuthenticateRequest JWTAuth where
+  authReq (JWTAuth token) req =
```
jkarni: Consistent indentation

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -245,7 +252,8 @@ defResponse = Response
 -- You can tweak an 'Action' (like the default 'defAction') with these lenses
 -- to transform an action and add some information to it.
 data Action = Action
-  { _captures :: [DocCapture]                -- type collected + user supplied info
+  { _authInfo :: Maybe AuthenticationInfo    -- type collected + user supplied info
```
jkarni: What about when there are nested authentication combinators?

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -262,8 +270,8 @@ data Action = Action
 -- 'combineAction' to mush two together taking the response, body and content
 -- types from the very left.
 combineAction :: Action -> Action -> Action
-Action c h p n m ts body resp `combineAction` Action c' h' p' n' m' _ _ _ =
-        Action (c <> c') (h <> h') (p <> p') (n <> n') (m <> m') ts body resp
+Action a c h p n m ts body resp `combineAction` Action _ c' h' p' n' m' _ _ _ =
```
jkarni: If we do left-biased merge of `authInfo`, the comment above should be updated. (But note that, unlike body and content-type, this would probably not be the correct behaviour.)

servant-mock/src/Servant/Mock.hs
```diff
@@ -139,6 +143,22 @@ instance (KnownSymbol s, HasMock rest) => HasMock (QueryFlag s :> rest) where
 instance (KnownSymbol h, FromHttpApiData a, HasMock rest) => HasMock (Header h a :> rest) where
   mock _ = \_ -> mock (Proxy :: Proxy rest)
 
+instance (HasMock rest, AuthData authdata, Arbitrary usr)
```
jkarni: Can the behaviour for `mock` for auth be described in a comment?

servant-server/servant-server.cabal
```diff
@@ -65,7 +67,8 @@ library
       , wai                >= 3.0  && < 3.1
       , wai-app-static     >= 3.0  && < 3.2
       , warp               >= 3.0  && < 3.2
-
+      , word8              >= 0.1.0
```
jkarni: We've tried to stick to having both upper and lower bounds, and updating as new versions get released.

servant/src/Servant/API/Authentication.hs
```diff
@@ -0,0 +1,41 @@
...
+-- Lax: all handlers under 'AuthProtect' take a 'Maybe usr' argument.
+--      when auth fails, we call the handlers with 'Nothing'.
+data AuthPolicy = Strict | Lax
+
+-- | the combinator to be used in API types
+data AuthProtect authdata usr (policy :: AuthPolicy)
+
+-- | what we'll ask user to provide at the server-level when we see a
+-- 'AuthProtect' combinator in an API type
+data family AuthProtected authdata usr subserver :: AuthPolicy -> *
```
jkarni: I've been on the sidelines on this one, so apologies in advance for having to ask, but can someone explain exactly why it is that we are using a data family here, rather than a class with an associated type synonym as we usually do?

jkarni: Thinking about this more, I think my question is: why not have a GADT? As far as I can tell, we don't really want more instances than the two we provide.

jkarni:
We're distinguishing between lax and strict, but there's unauthorized, unauthenticated, and authenticated. It would seem to me like we want to allow distinctions between all these (also, the handler only gets a `Maybe`, so doesn't know the exact type of failure).

servant-server/src/Servant/Server/Internal/Authentication.hs
```diff
@@ -0,0 +1,141 @@
...
+-- | concrete type to provide when in 'Strict' mode.
+data instance AuthProtected authData usr subserver 'Strict =
+    AuthProtectedStrict { checkAuthStrict :: authData -> IO (Maybe usr)
+                        , authHandlers :: AuthHandlers authData
+                        , subServerStrict :: subserver
+                        }
+
+-- | concrete type to provide when in 'Lax' mode.
+data instance AuthProtected authData usr subserver 'Lax =
+    AuthProtectedLax { checkAuthLax :: authData -> IO (Maybe usr)
```
jkarni: What if something else besides `IO` would be convenient (e.g. a `Reader` for a connection pool)?

servant-examples/auth-combinator/auth-combinator.hs
aaronlevin: yep. I switched over after switching to the `Delayed` stuff and still have a few things to update.

servant-foreign/src/Servant/Foreign.hs
aaronlevin: Will update.

servant-client/test/Servant/ClientSpec.hs
aaronlevin: thanks, will update.

servant-client/src/Servant/Client/Authentication.hs
aaronlevin: thanks, will update!

servant-docs/src/Servant/Docs/Internal.hs
aaronlevin: I'll think about this once I update the examples and address some other issues.

servant-mock/src/Servant/Mock.hs
aaronlevin: /cc @arianvp

servant-server/servant-server.cabal
aaronlevin: thanks. I'll add an upper bound.

aaronlevin:
@jkarni on the `authenticated` vs `authorized` front: we are concerning ourselves only with `authentication` at this point as `authorization` is a lot more general / tougher to solve. I.e. we are only concerned with extracting some kind of primitive and authenticating them/it. In the case of Basic, this primitive is the username/password, in the case of JWT it's a valid JWT token.

servant-server/src/Servant/Server/Internal/Authentication.hs
aaronlevin: these eventually get used in `Delayed` (https://github.com/aaronlevin/servant/blob/auth/servant-server/src/Servant/Server/Internal/RoutingApplication.hs#L183), so we'd need to figure out how to thread a different context/monad through there.

It should be easy enough to close over a database pool or something like that.

aaronlevin:
@arianvp how did you deal with the fact that `HeaderArg` can't have an `Eq` instance in `servant-foreign`? The code compiles but the tests rely on an `Eq` instance for `Req`.

jkarni:
@aaronlevin oops, I should have said invalid vs non-existent auth data

servant/src/Servant/API/Authentication.hs
jkarni: E.g.:
```
data AuthStrictness = StrictAuth | LaxAuth
data OnMissing  a where
    LaxMissing :: OnMissing Lax
    StrictMissing :: IO ServantErr -> OnMissing Strict

data OnUnauthenticated authData stricness where
   LaxUnauthenticated :: OnUnathenticated Lax
   StrictUnauthenticated :: (authData -> IO ServantErr) -> OnUnauthenticated Strict


data AuthProtected authData missingStrictness unauthStrictness subserver = AuthProtected
  { onMissing :: OnMissing missingStrictness
  , onUnathenticated :: OnUnathenticated authData unauthStrictness
  , subserver :: subserver -- but note commment below
  }

```

Re: `subserver`. The analogue of `->` is `(,)`, so I wonder if we're doing the right thing by having the subserver insider the `AuthProtected` datatype. If we want to keep the server regular, I think `(,)` makes more sense. This comes up right now already with e.g. `enter`. @alpmestan what do you think?

servant-server/src/Servant/Server/Internal/Authentication.hs
jkarni: Yeah, might not be crucial - I guess you can always put stuff into MVars.

servant/src/Servant/API/Authentication.hs
jkarni: (Continuing on with `->` ~ `(,)`:)
Some of the interesting possibilities that come up when the server is regular are e.g. transforming things into a `Reader` monad over e.g. a header (in the case of `->`); for `(,)`, we have the `State` monad, which can be similarly mechanically transformed. Off the top of my head, this seems promising: we can imagine writing subapis with nested but different auth requirements in a state monad. Pretty fancy - not sure if practically relevant in this specific case, but assuming `(,)` cases multiply, something is bound to come up.

This brings up the question of order wrt `,`. From the `State` perspective, it seems like we'd want the subserver on the LHS. OTOH from the perspective of the tuple functor instance, we'd want it on the RHS...

servant-server/src/Servant/Server/Internal/Authentication.hs
alpmestan: No need. You can define `checkAuthXXX` only when your connection pool is in scope.

servant/src/Servant/API/Authentication.hs
alpmestan: Your GADT based solution would suit me as well -- feel free to go ahead and write it if @aaronlevin and @arianvp approve it.

Note: In your code sample, we could kind-restrict the type param of `OnMissing` and friends to `AuthStrictness`. And `OnUnauthenticated` is missing a type param on the two lines that define its constructors.

aaronlevin:
@jkarni ok, for the lax case then I can return an `Either` indexed by an error type.

servant/src/Servant/API/Authentication.hs
aaronlevin: @jkarni @alpmestan  unrelated to overall API design: this code lives in `servant` core, whereas `ServantErr` lives in `servant-server`. I think this was the original motivation for the data family.

Now, I could make a single instance of the data family in `servant-server` rather than the two instances.

jkarni: Maybe just turn `IO ServantErr` into another polymorphic type, which then `servant-server` can require be `IO ServantErr`?

jkarni: Alternatively, maybe we should move `ServantErr`. We've been talking about unifying `ServantErr` and `ServantError` a bit more, and for that we'd need to move it to `servant` anyhow.

aaronlevin: ok, for now I'll paramaterize over the error type and when `ServantErr` and `ServantError` get merged we can update.

servant/src/Servant/API/Authentication.hs
```diff
@@ -0,0 +1,86 @@
...
+-- authData: the type of authData present in a request (e.g. JWT token)
+-- errorIndex: an index of error to give a user-provided, semantic meaning to the authentication failure.
+-- usr: a data type extracted from the authenticated data. This data is likely fetched from a database.
+-- subserver: the rest of the servant API.
+data AuthProtected m e (missingPolicy :: AuthPolicy) (unauthPolicy :: AuthPolicy) errorIndex authData usr subserver =
+    AuthProtected { onMissing :: OnMissing m e missingPolicy
+                  , onUnathenticated :: OnUnauthenticated m e unauthPolicy errorIndex authData
+                  , checkAuth :: authData -> m (Either errorIndex usr)
+                  , subserver :: subserver
+                  }
```
aaronlevin: @jkarni @alpmestan here is the GADT-based design. There is quite a lot of type-level specification:
- the monad it's run in (`IO` for now)
- the server error (`ServantErr` for now)
- the missing and unauth policy (lets users have strict misisng but lax unauth, for example)
- an `errorIndex`: this was for @jkarni who requested that server writers can provide their own index of errors when trying to authenticate a user (such an index may be: `data ErrorIndex = InvalidUserName | InvalidEmail | InvalidPassword | RateLimited`.
- the `authData` extracted from the request
- the `usr` type fetched from the database
- `subserver`

This proves to be quite flexible. For example, I was able to take the older `Basic` and `JWT`-based auth code and just use a unital index (`()`).

The only annoying part is specifying `IO` and `ServantErr`, but I've exposed combinators in `servant-server` so that clients of the library shouldn't have to worry about that too much.

servant-mock/src/Servant/Mock.hs
arianvp: I'm actually not all that happy with the current mock behaviour.   I'm gonna have a new look at it. But i'll document it in the process =)

aaronlevin:
@jkarni @alpmestan ok, i've incorporated the new GADT-based authentication type. The type is fully indexed by:
- monad (right now `IO`)
- server return type (right now `ServantErr`)
- OnMissing policy
- OnMissing error type (so you can capture semantic failures during extraction)
- OnUnauthenticated policy
- OnUnauthenticated error type (so you can capture semantic failures when checking)

The side-effect of this is `AuthProtect` and `AuthProtected` have considerably more parameters. However, I think with some good documentation (which I will write soon) we can explain all of the type params and it won't be so bad.

Overall I'm much happier with this new design. Now I need to update all the downstream projects :frowning: 

/cc @arianvp

alpmestan:
Looks good, thanks for the heroic effort guys. I'll take a closer look when it's closer to ready.

Merry christmas everyone.

aaronlevin:
I've got everything building on `7.10`. Just need to fix this `7.8` stuff.

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -543,6 +557,55 @@ routerSpec = do
...
+prioErrorsApi :: Proxy PrioErrorsApi
+prioErrorsApi = Proxy
+
+-- | Test the relative priority of error responses from the server.
+--
+-- In particular, we check whether matching continues even if a 'ReqBody'
+-- or similar construct is encountered early in a path. We don't want to
+-- see a complaint about the request body unless the path actually matches.
+--
+prioErrorsSpec :: Spec
```
jkarni: Didn't this get (re)moved?

servant-server/src/Servant/Server/Internal/Authentication.hs
```diff
@@ -0,0 +1,131 @@
...
+import            Web.JWT                    (decodeAndVerifySignature, JWT, VerifiedJWT, Secret)
+import qualified  Web.JWT as JWT             (decode)
+
+-- | Class to represent the ability to extract authentication-related
+-- data from a 'Request' object.
+class AuthData a e | a -> e where
+    authData :: Request -> Either e a
+
+-- | combinator to create authentication protected servers.
+authProtect :: OnMissing IO ServantErr missingPolicy missingError
```
jkarni: Is there any point to having these functions?

servant/src/Servant/API/Authentication.hs
```diff
@@ -0,0 +1,96 @@
...
+-- responseError: an error response type. For now just ServantErr.
+-- authData: the authentication data extracted from the request
+-- errorIndex: an index of error to give a user-provided, semantic meaning to the authentication failure.
+-- policy: the policy to handle OnUnauthenticated actions.
+data OnUnauthenticated m responseError (policy :: AuthPolicy) errorIndex authData where
+    LaxUnauthenticated :: OnUnauthenticated m e 'Lax errorIndex authData
+    StrictUnauthenticated :: (errorIndex -> authData -> m responseError)
+                          -> OnUnauthenticated m responseError 'Strict errorIndex authData
+
+-- | A GADT representing the data and functions required to protect a reasource for authentication.
```
jkarni: Not really a GADT - just an ADT.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -155,32 +159,111 @@ data Delayed :: * -> * where
...
+               -> IO (RouteResult (Either mError auth))
+               -> Delayed a
+addAuthCheckLL = genAuthCheck (\(LaxMissing) cont e -> (return . Route . cont) (Left (Left e)))
+                              (\(LaxUnauthenticated) cont e _ -> (return . Route . cont) (Left (Right e)))
+                              Right
+
+-- | Add an auth check by supplying OnMissing policies and OnUnauthenticated policies.
+addAuthCheck :: SAuthPolicy mPolicy
+             -> SAuthPolicy uPolicy
+             -> Delayed (AuthProtected IO ServantErr mPolicy mError uPolicy uError auth usr (AuthDelayedReturn mPolicy mError uPolicy uError usr -> a))
```
jkarni: Why do we need the two singleton arguments? Judging by the signature, there is no information they provide besides what's already within `AuthProtected`.

servant/src/Servant/API/Authentication.hs
```diff
@@ -0,0 +1,96 @@
...
+
+-- | we can be either Strict or Lax.
+-- Strict: all handlers under 'AuthProtect' take a 'usr' argument.
+--         when auth fails, we call user-supplied handlers to respond.
+-- Lax: all handlers under 'AuthProtect' take a 'Maybe usr' argument.
+--      when auth fails, we call the handlers with 'Nothing'.
+data AuthPolicy = Strict | Lax
+
+-- | Singleton types for AuthPolicy
+data SAuthPolicy (p :: AuthPolicy) where
```
jkarni: I'm not sure we need the singletons. I *think* we really only need the type-level tags.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -155,32 +159,111 @@ data Delayed :: * -> * where
...
+type family AuthDelayedReturn (mP :: AuthPolicy) mE (uP :: AuthPolicy) uE usr :: * where
+    AuthDelayedReturn 'Strict mE 'Strict uE usr = usr
+    AuthDelayedReturn 'Strict mE 'Lax    uE usr = Either uE usr
+    AuthDelayedReturn 'Lax    mE 'Strict uE usr = Either mE usr
+    AuthDelayedReturn 'Lax    mE 'Lax    uE usr = Either (Either mE uE) usr
+
+-- | Internal method to generate auth checkers for various policies. Scary type signature
+-- but it does help with understanding the logic of how each policy works. See
+-- examples below.
+genAuthCheck :: (OnMissing IO ServantErr mPolicy mError -> (AuthDelayedReturn mPolicy mError uPolicy uError usr -> a) -> mError -> IO (RouteResult a))
```
jkarni: This signature is pretty hard to make sense of!

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -155,32 +159,111 @@ data Delayed :: * -> * where
...
 addMethodCheck :: Delayed a
                -> IO (RouteResult ())
                -> Delayed a
-addMethodCheck (Delayed captures method body server) new =
-  Delayed captures (combineRouteResults const method new) body server
+addMethodCheck (Delayed captures method auth body server) new =
+  Delayed captures (combineRouteResults const method new) auth body server
+
+-- | helper type family to capture server handled values for various policies
+type family AuthDelayedReturn (mP :: AuthPolicy) mE (uP :: AuthPolicy) uE usr :: * where
```
jkarni: Do we need the type family `AuthDelayedReturn`? It seems like it could be a GADT and be cleaner that way.

servant-server/test/Servant/ServerSpec.hs
aaronlevin: not sure, it could have been a casualty of the rebase. I'm unsure what to do here.

servant-server/src/Servant/Server/Internal/Authentication.hs
aaronlevin: yeah, I use it. I find it more idiomatic when defining the api to a server to use a function rather than a constructor, but ymmv.

servant/src/Servant/API/Authentication.hs
aaronlevin: fixed. thanks.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: this allows someone to use `addAuthCheck` like:

```
addAuthCheck SStrict SLax
```

instead of relying on the convention of ending each combinator with `S` or `L`.

servant/src/Servant/API/Authentication.hs
aaronlevin: I used them in the `Delayed` code.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: Agreed. I was reluctant to use it. However, in doing this I actually exposed a few subtlties in the error handling logic! In my original implementation, I was doing a lot of boilerplate around the `bindRouteResult` and inspecting errors and the like. The combinators downstream are much clearer and really benefited from this.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: It probably looks scarier than it is. It just captures the desired return values of the four different scenarios for auth policy:

- `'Strict x 'Strict`: handler gets a `usr`
- `'Strict x 'Lax`: handler gets a `Either e usr` where `e` is errors encountered when authenticating the user.
- `'Lax x 'Strict`: handler gets a `Either e usr` where `e` is errors encountered when extracting data from a request.
- `'Lax x 'Lax`: handler gets `Either (Either e1 e2) usr` where `e1` and `e2` are errors encountered extracting data from a request or authenticating the user.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: I can remove them if need be. I personally like them but I'm not attached.

servant/src/Servant/API/Authentication.hs
aaronlevin: correction: I use them in `servant-server`.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: see my comments above.

servant-server/src/Servant/Server/Internal/Authentication.hs
alpmestan: I tend to agree with Aaron here.

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -149,83 +166,93 @@ failServer = serve failApi (
...
+            (left show <$> runExceptT (getBody p)) `shouldReturn` Right p
+
+        it "Servant.API.QueryParam" $ \(_, baseUrl) -> do
+            let getQueryParam = getNth (Proxy :: Proxy 4) $ client api baseUrl manager
+            left show <$> runExceptT (getQueryParam (Just "alice")) `shouldReturn` Right alice
+            Left FailureResponse{..} <- runExceptT (getQueryParam (Just "bob"))
+            responseStatus `shouldBe` Status 400 "bob not found"
+
+        it "Servant.API.QueryParam.QueryParams" $ \(_, baseUrl) -> do
+            let getQueryParams = getNth (Proxy :: Proxy 5) $ client api baseUrl manager
```
jkarni: Any reason to change the indentation?

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -1,66 +1,110 @@
...
+
+-- | We will look up authentication data in the database and extract a User.
+type DBLookup = CookieAuth -> IO (Either UserFromCookieError User)
+
+-- | Method that tests for authentication and extracts a User type.
+userFromCookie :: DBLookup
+userFromCookie (CookieAuth cookie) = if cookie == "good cookie" then return (Right "one user") else return (Left NoUserAssociatedWithCookie)
+
+-- | a handler that takes a cookie error and returns a ServantErr in IO
+missingCookieHandler :: OnMissing IO ServantErr 'Strict CookieFromRequestError
```
jkarni: Cool examples.

servant-foreign/test/Servant/ForeignSpec.hs
```diff
@@ -57,58 +57,87 @@ testApi = listFromAPI (Proxy :: Proxy LangX) (Proxy :: Proxy TestApi)
...
 listFromAPISpec :: Spec
 listFromAPISpec = describe "listFromAPI" $ do
-    it "generates 4 endpoints for TestApi" $ do
+    it "generates 4 endpoints for TestApi" $
         length testApi `shouldBe` 4
 
     let [getReq, postReq, putReq, deleteReq] = testApi
 
+    -- test if two requests are equal for testing purposes
+    let reqEq req1 req2 = do
```
jkarni: What's the difference between this and equality?

servant-foreign/test/Servant/ForeignSpec.hs
```diff
@@ -57,58 +57,87 @@ testApi = listFromAPI (Proxy :: Proxy LangX) (Proxy :: Proxy TestApi)
...
+            _reqMethod req1 `shouldBe` _reqMethod req2
+            _reqBody req1 `shouldBe` _reqBody req2
+            _reqReturnType req1 `shouldBe` _reqReturnType req2
+            _funcName req1 `shouldBe` _funcName req2
+
+            let h = case (_reqHeaders req1, _reqHeaders req2) of
+                    ([], []) -> True
+                    ([HeaderArg a], [HeaderArg b]) -> a == b
+                    ([ReplaceHeaderArg a1 b1], [ReplaceHeaderArg a2 b2]) -> a1 == a2 && b1 == b2
+                    ([HeaderArgGen a1 f1], [HeaderArgGen a2 f2]) -> a1 == a2 && f1 "" == f2 ""
```
jkarni: Ah, I guess these function params don't allow for equality?

servant-foreign/test/Servant/ForeignSpec.hs
```diff
@@ -57,58 +57,87 @@ testApi = listFromAPI (Proxy :: Proxy LangX) (Proxy :: Proxy TestApi)
...
+                    _ -> False
```
jkarni: If the lists contain multiple elements, why do they always compare unequal?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -155,32 +159,111 @@ data Delayed :: * -> * where
...
+-- examples below.
+genAuthCheck :: (OnMissing IO ServantErr mPolicy mError -> (AuthDelayedReturn mPolicy mError uPolicy uError usr -> a) -> mError -> IO (RouteResult a))
+             -> (OnUnauthenticated IO ServantErr uPolicy uError auth -> (AuthDelayedReturn mPolicy mError uPolicy uError usr -> a) -> uError -> auth -> IO (RouteResult a))
+             -> (usr -> (AuthDelayedReturn mPolicy mError uPolicy uError usr))
+             -> Delayed (AuthProtected IO ServantErr mPolicy mError uPolicy uError auth usr (AuthDelayedReturn mPolicy mError uPolicy uError usr -> a))
+             -> IO (RouteResult (Either mError auth))
+             -> Delayed a
+genAuthCheck missingHandler unauthHandler returnHandler d@(Delayed captures method _ body _) new =
+    let newAuth =
+            runDelayed d `bindRouteResults` \ authProtection ->
```
jkarni: Is this the right thing to do? Aren't we destroying the order set up?

servant-foreign/test/Servant/ForeignSpec.hs
aaronlevin: ah, good catch. let me update. it should be a fold.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
aaronlevin: I've made note of this a few times and it's been one of my concerns about holding the API in the data type. There is no way for us to get a hold of the `authProtection` data type without running the `delayed`. We had this same problem before `Delayed`, where we had to run the server before we could extract the `AuthProtected` datatype.

However, recall that `Delayed` is holding this all as a continuation, so if you look below, we should only be running this entire handler if authentication is present; and it will run in order.

servant-foreign/test/Servant/ForeignSpec.hs
aaronlevin: correct. however, for running tests I think this is enough.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
jkarni: > I've made note of this a few times and it's been one of my concerns about holding the API in the data type.

Agreed. It's one of the reasons why I always had doubts this approach could work.

> However, recall that Delayed is holding this all as a continuation, so if you look below, we should only be running this entire handler if authentication is present; and it will run in order. 

What do you mean? We are going to run e.g. the body check before auth (if `ReqBody` appear sooner), no?

jkarni: I'll try adding a config param and see how it goes. I think we'd benefit from it for other reasons too (e.g. having combinators that require certain things to be initialized).

aaronlevin: So, to be honest, my understanding of how `Delayed` is used is still a bit shaky. For example, it seems like anyone using `runDelayed` must provide a function that accepts authentication params (which could be `()` but I don't see that being used anywhere).

To your question: this can change, but as it stands I've placed the order as: `captures`, `method`, `auth`, `body`. So, `captures` and `method` will run first, then `auth`, then `body`.

As for destroying the order setup, my comment was this: if we add an `authCheck` what we are returning is a `Delayed` with an authentication param in it. So, when we eventually `runDelayed` the order is restored, because we run first the `captures`, then the `method`, then the `auth`, and finally the `body` checks.

Now, if I understand things correctly, what I think happens because we need to extract the `AuthProtected` data is that the `captures` and `body` checks happen twice, which is unfortunate.

aaronlevin: I don't really see any way around having to call `runDelayed` since we need to extract `AuthProtected` from the request.

jkarni:
I'm a little worried about the `jwt` library. From the haddocks:

```
    Only HMAC SHA-256 algorithm is currently a supported signature algorithm
    There is currently no verification of time related information (exp, nbf, iat).
    Registered claims are not validated
    This implementation uses the term JWTHeader instead of JOSEHeader (changed in version 23 of the JWT draft). Future versions will move to the offical term once that has stabilised.
```

arianvp:
https://hackage.haskell.org/package/jose could be an alternative. It's more complete and supports all the algorithms in the current spec.

The current `jwt` implementation in servant is very _simple_. The only thing that we check is the HMAC actually.

jkarni:
> The current jwt implementation in servant is very simple. The only thing that we check is the HMAC actually

What do you mean?

arianvp:
We only verify that the token was signed with the correct key. We don't do
any further verification currently.

tylerholien:
I'm currently evaluating the different JWT libraries, and the [jose-jwt](https://hackage.haskell.org/package/jose-jwt) package seems the most complete. The [jose](http://hackage.haskell.org/package/jose) package hasn't yet implemented encrypted tokens (JWE).

arianvp:
@tylerholien I've seen the `jose-jwt` package and considered using it. But I wasn't sure what to do with the `MonadRandom m` constraint in all the functions as I was not sure how to thread a random generator into servant handlers...   That's why I opted for using `jwt` which only supports `HS256` which requires no `MonadRandom` constraint.

aaronlevin:
@arianvp with the new `Config` setup, we can much more easily thread that kind of dependency to the handler. I'll keep this use case in mind when I meet with @jkarni and we chat more.

jkarni:
Closing in favour of #311

Issue #189, opened 2015-08-12
Need a way to return 201 Created as the result of a PUT


Porges:
Per [RFC7321](https://tools.ietf.org/html/rfc7231#section-4.3.4):

> If the target resource does not have a current representation and the
   PUT successfully creates one, then the origin server MUST inform the
   user agent by sending a 201 (Created) response.  If the target
   resource does have a current representation and that representation
   is successfully modified in accordance with the state of the enclosed
   representation, then the origin server MUST send either a 200 (OK) or
   a 204 (No Content) response to indicate successful completion of the
   request.

At the moment 200 `OK` (or 204 `No Content`) is always returned. It would be nice to have a way to indicate that a resource was created.

jkarni:
Closing this as a duplicate of #190 . For people who might come across is, currently the suggested way about it is defining your own combinator, with a `HasServer` instance that differs from the current one only in having 201 rather than 200 as the status code.

i-am-the-slime:
I tried to backtrack to the actual solution but couldn't really find anything. Can you go into more detail about this "own combinator" approach? Thanks!

alpmestan:
@i-am-the-slime Hi! [This PR](https://github.com/haskell-servant/servant/pull/276) changed all existing combinators to be simple type synonyms of [`Verb`](https://github.com/haskell-servant/servant/pull/276/files#diff-37b524dc8396a175938c788d35d9d2a6R24) with different parameters. You can use this to define your own `Post` that returns something else than *201* by defining a new type synonym similar to [servant's default `Post`](https://github.com/haskell-servant/servant/blob/master/servant/src/Servant/API/Verbs.hs#L39) but specifying another status code than 201. You need servant from git for now though to be able to use that as this will only be part of the next release.

i-am-the-slime:
Ah, I'm using stack. I'll probably wait for the next LTS including this then and just return a 200 in any case.

alpmestan:
You can use packages from git repos with stack as well, FWIW: http://docs.haskellstack.org/en/stable/yaml_configuration/ (scroll down a bit and look for "git")

rvion:
eg:

```yaml
resolver: lts-5.5

# Local packages, usually specified by relative directory name
packages:
- .
- location:
    git: git@github.com:haskell-servant/servant.git
    commit: 5188e842a9288ac3986b30b93cb1a5d58cb2933d
  extra-dep: true
  subdirs:
  - servant
  - servant-blaze
[...]

PR #190, opened 2015-08-12
Make it easier to define more HTTP methods

Porges:
Sending this for a bit of discussion.

This change factors out common functionality from the HasServer instances for
the various existing HTTP methods, so that there is only one real
datatype (`HttpMethod`) and the existing implementations are all just type
aliases for this.

This should hopefully make it near-trivial to define new HTTP methods.

I think this should be mostly source-compatible with the previous implementation?

Probably need to do something about the 'default' status codes as well (per #189).

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -25,6 +25,7 @@ import           Control.Applicative         ((<$>))
 #endif
 import           Control.Monad.Trans.Either  (EitherT)
 import qualified Data.ByteString             as B
+import qualified Data.ByteString.Char8       as B8
 import qualified Data.ByteString.Lazy        as BL
 import qualified Data.Map                    as M
 import           Data.Maybe                  (mapMaybe, fromMaybe)
@@ -34,7 +35,7 @@ import           Data.Text                   (Text)
 import qualified Data.Text                   as T
 import           Data.Text.Encoding          (decodeUtf8, encodeUtf8)
 import           Data.Typeable
-import           GHC.TypeLits                (KnownSymbol, symbolVal)
+import           GHC.TypeLits                (KnownSymbol, symbolVal, KnownNat, natVal)
 import           Network.HTTP.Types          hiding (Header, ResponseHeaders)
 import           Network.Socket              (SockAddr)
 import           Network.Wai                 (Application, lazyRequestBody,
@@ -43,9 +44,8 @@ import           Network.Wai                 (Application, lazyRequestBody,
                                               isSecure, vault, httpVersion, Response,
                                               Request)
 import           Servant.API                 ((:<|>) (..), (:>), Capture,
-                                              Delete, Get, Header,
-                                              IsSecure(..), MatrixFlag, MatrixParam,
-                                              MatrixParams, Patch, Post, Put,
+                                              Header, HttpMethod, IsSecure(..),
+                                              MatrixFlag, MatrixParam, MatrixParams,
                                               QueryFlag, QueryParam, QueryParams,
                                               Raw, RemoteHost, ReqBody, Vault)
 import           Servant.API.ContentTypes    (AcceptHeader (..),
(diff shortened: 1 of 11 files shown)


Porges:
Caution: I haven't actually tried to use this yet (stack build had some issues in the other projects about disabled components) so it's possible the entire idea is unworkable.

servant/src/Servant/API/Delete.hs
```diff
@@ -12,9 +12,7 @@ import Data.Typeable ( Typeable )
 --
 -- >>>            -- DELETE /books/:isbn
 -- >>> type MyApi = "books" :> Capture "isbn" Text :> Delete
-data Delete (contentTypes :: [*]) a
-  deriving Typeable
-
+type Delete (contentTypes :: [*]) a = HttpMethod "DELETE" 200 contentTypes a
```
jkarni: The doctest should say `Delete <content-type> ()` (our mistake - only now are the doctests failing here because previously it'd still compile)

jkarni:
Great idea, by the way. Thanks!

servant/src/Servant/API/Delete.hs
Porges: Yes, I got quite confused starting out because the documentation says you can use `Delete` without arguments in a few places.

Porges:
Had a think about how to achieve #190:

1) Drop the default status code from the `HttpMethod` definition, so it's just the method name.
2) Create a type family to map e.g. `HttpMethod "GET" = 200`, so we can preserve existing behaviour.
3) Create a new result type `ServantSucc(ess)` (to go with `ServantErr`) defined as something like `data ServantSuccess r = ServantSuccess Status r`. If this is returned from a handler then the corresponding status code is used (this will be another overlap for `HasServer` - need to see how this plays with headers etc).
4) Add some helpers like `ok r = ServantSuccess ok200 r` and `created r = ServantSuccess created201 r`. Then a handler looks something like `handler _ = do { _; return $ created _ }`

This should continue to have backwards compatibility while allowing more flexibility in status codes.

I'll write this up after work today and see how it goes.

codedmart:
I haven't been following along but any change I feel should allow the user to provide a status code. Regardless of what the specs say in case someone wants to provide a custome status code they should be able to. Is that where this is going?

Porges:
Just realized I referenced #190 (this PR) instead of #189 in that comment - fixed. But yes, that's the idea.

servant/src/Servant/API/Methods.hs
```diff
@@ -0,0 +1,12 @@
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE KindSignatures     #-}
+{-# OPTIONS_HADDOCK not-home    #-}
+
+module Servant.API.Methods where
+
+import           Data.Typeable (Typeable)
+import           GHC.TypeLits (Nat, Symbol)
+
```
jkarni: Maybe some documentation saying you can use this to define your own status code response/method?

jkarni:
Looks great. 

This changes the behaviour of methods to not return 204 when the type is `()`, as was previously the case. This is a trade-off: on the one hand, the correct thing in those cases is almost always 204; on the other, there are many other cases (sum types, more particularly) where there should be a 204 but there isn't, and thus the behaviour is incomplete and a little surprising. Also, our previous overlapping instances approach made certain things ugly (exponential blow-up in the number of instances). I'm on the fence about that change, but slightly in favour of it (not being too opinionated about things seems good). @alpmestan what do you think?

Otherwise (and aside from the small comment), this looks good to me.

alpmestan:
I'm really not sure about this. How wrong would it be to not give 204 when appropriate, but e.g 200 instead? Would it mess things up with external tools? Would it annoy people?

Porges:
@jkarni the intention is to *not* remove special-casing 204 for `()` (and indeed, not to affect the behaviour of any existing code). The overlapping instances will still be there (but won't have `n*m` instances).

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -307,139 +238,82 @@ instance (KnownSymbol sym, FromText a, HasServer sublayout)
...
-         {-# OVERLAPPABLE #-}
-#endif
-         ( AllCTRender ctypes a) => HasServer (Put ctypes a) where
+#endif 
+        ( AllCTRender ctypes a
+        , KnownSymbol method
+        ) => HasServer (HttpMethod method ctypes (ServantSuccess a)) where
 
-  type ServerT (Put ctypes a) m = m a
+  type ServerT (HttpMethod method ctypes (ServantSuccess a)) m = m (ServantSuccess a)
```
jkarni: I like this idea. Another option maybe worth discussing is just putting in a `StateT` in the midst of the stack with the status code (initialized to 200 for GET, 201 for POST, 204 for `()`, etc.). In theory I guess we could then even remove the `EitherT` then (and have users themselves declare sum types when they may return very different datatypes. That would be a pretty fundamental change, though.

jkarni: Hm, thinking more on it, I think we want to go in the direction of making HTTP status codes more rather than less knowable statically. (And the handlers less rather than more HTTP-aware.) Maybe the status code can be a (Nat-kinded) param to `HttpMethod`?

jkarni:
@Porges I like the type synonyms, but I'm not so sure about `ServantSuccess`. Do you want to separate out the changes, so we can discuss them independently?

Porges:
Sorry about lack of movement on this - my personal laptop deaded itself and I haven't replaced it yet...

jkarni:
Closing in favour of #276 (and thanks!)

PR #196, opened 2015-08-17
Contributing

jkarni:


jkarni:
Most significant are the README changes, and the addition of `.stylish-haskell.yaml` and `HLint.hs`. Everything else is the result of applying the changes suggested by both those tools.

PR #204, opened 2015-08-25
Fix #203; add path to BaseUrl

### LINKED ISSUE #203: Allow for paths in BaseUrl
I like being able to generate clients with `Servant.Client.client`. However I wish that I could add a path to my `Servant.Common.BaseUrl`. For example, let's say I have this API:

``` hs
type API = "widgets" :> Get '[JSON] [Widget]
```

I would like to be able to create clients that could consume that API at `example.com` and `example.org/api/v1`. 

- `http://example.com/widgets`
- `http://example.org/api/v1/widgets`

I can create `BaseUrl` values that have paths, but the client ignores them.

``` hs
host = BaseUrl Http "example.org/api/v1" 80
                                ^^^^^^^ - This part is ignored.
```

tfausak:
This pull request fixes #203. It adds another field, `path`, to the `BaseUrl` value.

### DIFF
--- a/servant-client/src/Servant/Common/BaseUrl.hs
+++ b/servant-client/src/Servant/Common/BaseUrl.hs
@@ -15,7 +15,7 @@ import           Control.Monad.Catch (Exception, MonadThrow, throwM)
 import           Data.List
 import           Data.Typeable
 import           GHC.Generics
-import           Network.URI
+import           Network.URI hiding (path)
 import           Safe
 import           Text.Read
 
@@ -31,11 +31,12 @@ data BaseUrl = BaseUrl
   { baseUrlScheme :: Scheme -- ^ URI scheme to use
   , baseUrlHost   :: String   -- ^ host (eg "haskell.org")
   , baseUrlPort   :: Int      -- ^ port (eg 80)
+  , baseUrlPath   :: String   -- ^ path (eg "/a/b/c")
   } deriving (Show, Eq, Ord, Generic)
 
 showBaseUrl :: BaseUrl -> String
-showBaseUrl (BaseUrl urlscheme host port) =
-  schemeString ++ "//" ++ host ++ portString
+showBaseUrl (BaseUrl urlscheme host port path) =
+  schemeString ++ "//" ++ host ++ portString ++ path
     where
       schemeString = case urlscheme of
         Http  -> "http:"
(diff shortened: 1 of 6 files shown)


servant-client/src/Servant/Common/Req.hs
```diff
@@ -109,7 +109,7 @@ reqToRequest req (BaseUrl reqScheme reqHost reqPort) =
                                          , uriRegName = reqHost
                                          , uriPort = ":" ++ show reqPort
                                          }
-                             , uriPath = reqPath req
+                             , uriPath = path ++ reqPath req
                              }
```
tfausak: Is there a reason why this doesn't re-use `showBaseUrl`? That would reduce some duplication.

servant-client/src/Servant/Common/BaseUrl.hs
```diff
@@ -31,11 +31,12 @@ data BaseUrl = BaseUrl
...
   , baseUrlHost   :: String   -- ^ host (eg "haskell.org")
   , baseUrlPort   :: Int      -- ^ port (eg 80)
+  , baseUrlPath   :: String   -- ^ path (eg "/a/b/c")
   } deriving (Show, Eq, Ord, Generic)
 
 showBaseUrl :: BaseUrl -> String
-showBaseUrl (BaseUrl urlscheme host port) =
-  schemeString ++ "//" ++ host ++ portString
+showBaseUrl (BaseUrl urlscheme host port path) =
+  schemeString ++ "//" ++ host ++ portString ++ path
```
tfausak: This should probably handle paths both with and without leading slashes.

alpmestan:
Could you update accordingly the `CHANGELOG` for *servant-client*? LGTM. @jkarni ?

alpmestan:
Well, actually, there are some errors with Travis, have you seen them?

[here](https://travis-ci.org/haskell-servant/servant/jobs/77092923#L1024)

tfausak:
Ok, I updated the change log and (hopefully) fixed the build.

servant-client/src/Servant/Common/BaseUrl.hs
jkarni: `</>` from `System.FilePath`? Or just roll our own?

tfausak: I don't think `</>` does what I want here.

``` hs
"a/" </> "/b"
-- actual: "/b"
-- desired: "a/b"
```

servant-client/src/Servant/Common/Req.hs
tfausak: In particular, I though this could be:

``` hs
url = showBaseUrl baseUrl ++ reqPath req
```

tfausak:
The build failed, but I don't think it's my fault. It does the same thing on master. Check out [this build](https://travis-ci.org/haskell-servant/servant/jobs/77156625#L393), for example.

codedmart:
@tfausak I restarted the failed build.

tfausak:
:+1: That did the trick.

servant-client/src/Servant/Common/BaseUrl.hs
jkarni: Ugh, of course.

jkarni:
Thanks

tfausak:
Thank you!

PR #205, opened 2015-08-27
Remove broken links

jkarni:
Fixes #202.


jkarni:
@alpmestan I think just the tutorial is a better place to go, no? We can link everything from there.

alpmestan:
Yes, I agree.

PR #206, opened 2015-08-29
Add servant-mock to 0.4

jkarni:
So we can release it already.


jkarni:
Please don't merge yet.

PR #207, opened 2015-08-29
Add servant-cassava.

jkarni:


jkarni:
This allows for using non-default encode and decode options, but doesn't allow requests or responses without headers. If we're to allow that, we probably need another type parameter (e.g. `(CSV', MyEncodeOpts, NoHeader)`. Thoughts?

Issue #209, opened 2015-08-31
Converting to raw should not alter behaviour


jkarni:
When writing a `Middleware` combinator, I noticed it was not possible to do the obvious thing, which is to convert the sub-api to an `Application` and then apply the middleware. This is because the http errors that come out of `Raw` are not treated in the same way as framework errors. Relatedly, user-sent errors (with `Left errXXX`) are also not treated in the same way as framework errors. They all should. 

This solves more than the middleware issue; it also fixes the currently quite crazy issues with `Raw` (e.g. having `Raw` be the first, top-level combinator makes everything else in the API be ignored.

alpmestan:
Fixing this would be great!

fizruk:
What's the status of this? Has this been fixed already? If not, can you elaborate on the problem?

jkarni:
As mentioned [here](https://github.com/haskell-servant/servant/issues/108#issuecomment-173187131), I don't think this will be possible.

Issue #198, opened 2015-08-18
Default content-type


jkarni:
Right now we return a 415 if the `Content-Type` header is not set in the request. We should probably just assume the first content-type in the list as the default.

EDIT: Actually we assume octet-stream.

alpmestan:
Yep, +1.

jkarni:
Actually, the current behaviour is the correct one. Added a link to the appropriate RFC in #226.

Issue #197, opened 2015-08-18
415 falls through


jkarni:
type Api = "hi" :> ReqBody '[JSON] Int :> X
                :<|> Raw

In the above API, a request to "hi" with a body content-type of `text/plain` (or none at all) will fall through to the `Raw` endpoint. This is internally consistent with the general servant behaviour, but likely to be very surprising. I think it makes more sense to just always return 415.

alpmestan:
So, we change the semantics of `:<|>`?

jkarni:
I think we can just change the semantics of content-type mismatch.

kosmikus:
I disagree with this proposed change. The current behaviour is the only behaviour that makes sense to me. How else would I modularly (i.e., separately) specify two handlers for the same endpoint for different body content types?

alpmestan:
Just posted a comment in #214 that explains many things among which this issue. Hoping to get all the eyes and brains we can there to find a nice and solid solution to this whole thing with routing and combining "routing results" and the problems it causes.

jkarni:
> How else would I modularly (i.e., separately) specify two handlers for the same endpoint for different body content types?

Fair point, but it also seems to favour underlying elegance (at the cost of practicality) a little too much. Being able to have separate handlers for different content-types is neat, but I don't see an obvious use, especially since in my opinion *content-type should not affect behaviour*. If an endpoint can handle JSON and XML request bodies, it *should not* return an `Int` and a `String` in each case, but rather always the same datatype. And in practice I'd imagine being bitten by routing continuing is a much more common issue than wanting that form of modularity.

jkarni:
Closed in #218 .

PR #214, opened 2015-09-05
Routing and errors (breaks tests).

fisx:
Routing needs improvement.  I added a few test cases to model what I think we may want (comments welcome), and I get these

```
Failures:

  src/Test/Hspec/Wai.hs:101:
  1) Servant.Server, Handlers can trigger re-routing by responding with error status 404, re-route from /un handler, /un
       status mismatch:
         expected: 200
         but got:  404
       body mismatch:
         expected: "2"
         but got:  ""

  src/Test/Hspec/Wai.hs:101:
  2) Servant.Server, Broken request body triggers error response (not re-routing), happy handlers, ("/sum","[1, 2]")
       status mismatch:
         expected: 200
         but got:  415
       body mismatch:
         expected: 3
         but got:  unsupported media type

  src/Test/Hspec/Wai.hs:101:
  3) Servant.Server, Broken request body triggers error response (not re-routing), parse error, ("/sum","@@@")
       status mismatch:
         expected: 400
         but got:  415

  src/Test/Hspec/Wai.hs:101:
  4) Servant.Server, Different methods on same end-point, ("get","/")
       status mismatch:
         expected: 200
         but got:  404
       body mismatch:
         expected: 3
         but got:  not found

  src/Test/Hspec/Wai.hs:101:
  5) Servant.Server, Different methods on same end-point, ("put","/")
       status mismatch:
         expected: 405
         but got:  404
```

I suspect 2) is my fault, but I wanted to push this anyway.  1) is currently seriously biting us, and it would be great to get a quick fix at least in this branch.  @jkarni suspects it may be a matter of replacing a single `succeedWith` with a `failWith`.

I put this out here FYI, will probably keep working on it.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
+++ b/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
@@ -1,7 +1,9 @@
 {-# LANGUAGE CPP                        #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE ViewPatterns               #-}
 module Servant.Server.Internal.RoutingApplication where
 
 #if !MIN_VERSION_base(4,8,0)
@@ -10,6 +12,7 @@ import           Data.Monoid                        (Monoid, mappend, mempty)
 #endif
 import           Control.Monad.Trans.Either         (EitherT, runEitherT)
 import qualified Data.ByteString                    as B
+import qualified Data.ByteString.Char8              as BS
 import qualified Data.ByteString.Lazy               as BL
 import           Data.IORef                         (newIORef, readIORef,
                                                      writeIORef)
@@ -41,6 +44,7 @@ newtype RouteResult a =
 instance Monoid (RouteResult a) where
   mempty = RR $ Left mempty
 
+  unrecoverable@(RR (Left (HttpError (statusCode -> s) _))) `mappend` RR (Right _) | s /= 404 = unrecoverable
   RR (Left x)  `mappend` RR (Left y)  = RR $ Left (x <> y)
   RR (Left _)  `mappend` RR (Right y) = RR $ Right y
   r            `mappend` _            = r
@@ -119,7 +123,7 @@ runAction action respond k = do
       e <- runEitherT a
       respond $ case e of
         Right x  -> k x
-        Left err -> succeedWith $ responseServantErr err
+        Left err -> failWithServantError err
     go (RR (Left err)) = respond $ failWith err
 
 feedTo :: IO (RouteResult (a -> b)) -> a -> IO (RouteResult b)
@@ -136,10 +140,15 @@ extractR (RR (Left err))         = RR (Left err)
 failWith :: RouteMismatch -> RouteResult a
 failWith = RR . Left
 
+-- FIXME: add headers field to HttpError constructor.
+failWithServantError :: ServantErr -> RouteResult a
+failWithServantError ServantErr{..} = failWith $ HttpError status {- errHeaders -} (Just errBody)
+  where
+    status = mkStatus errHTTPCode (BS.pack errReasonPhrase)
+
 succeedWith :: a -> RouteResult a
 succeedWith = RR . Right
 
 isMismatch :: RouteResult a -> Bool
 isMismatch (RR (Left _)) = True
 isMismatch _             = False
-
--- a/servant-server/test/Servant/ServerSpec.hs
+++ b/servant-server/test/Servant/ServerSpec.hs
@@ -30,8 +30,8 @@ import           Network.Wai                (Application, Request, pathInfo,
 import           Network.Wai.Test           (defaultRequest, request,
                                              runSession, simpleBody)
 import           Test.Hspec                 (Spec, describe, it, shouldBe)
-import           Test.Hspec.Wai             (get, liftIO, matchHeaders,
-                                             matchStatus, post, request,
+import           Test.Hspec.Wai             (ResponseMatcher, get, liftIO, matchHeaders,
+                                             matchStatus, post, put, request,
                                              shouldRespondWith, with, (<:>))
 import           Servant.API                ((:<|>) (..), (:>),
                                              addHeader, Capture,
@@ -94,6 +94,9 @@ spec = do
   errorsSpec
   responseHeadersSpec
   miscReqCombinatorsSpec
+  errorRoutingUser404Spec
+  errorRoutingBodyParseErrorSpec
+  errorRouting405
 
 
 type CaptureApi = Capture "legs" Integer :> Get '[JSON] Animal
@@ -728,3 +731,78 @@ miscReqCombinatorsSpec = with (return $ serve miscApi miscServ) $
       go "/host" "\"0.0.0.0:0\""
 
   where go path res = Test.Hspec.Wai.get path `shouldRespondWith` res
+
+
+type ErrorRoutingUser404
+  =    "un"   :> Get '[JSON] Int
+  :<|> "deux" :> Get '[JSON] Int
+  :<|>           Get '[JSON] Int
+
+errorRoutingUser404Spec :: Spec
+errorRoutingUser404Spec =
+  describe "Handlers can trigger re-routing by responding with error status 404" $ do
+    describe "happy handlers" $ do
+      let hs = return 1 :<|> return 2 :<|> return 3
+      go hs "/un"   "1"
+      go hs "/deux" "2"
+      go hs "/"     "3"
+      go hs "/n/a"  404
+    describe "re-route from /un handler" $ do
+      let hs = left err404 :<|> return 2 :<|> return 3
+      go hs "/un"   "2"
+      go hs "/deux" "2"
+      go hs "/"     "3"
+      go hs "/n/a"  404
+  where
+    go :: Server ErrorRoutingUser404 -> String -> ResponseMatcher -> Spec
+    go hs path resp = with (return $ serve (Proxy :: Proxy ErrorRoutingUser404) hs) $
+      it path $ Test.Hspec.Wai.get (cs path) `shouldRespondWith` resp
+
+
+type ErrorRoutingBodyParseError
+  =    "sum" :> ReqBody '[JSON] [Int] :> Post '[JSON] Int
+  :<|> "const"                        :> Post '[JSON] Int
+
+errorRoutingBodyParseErrorSpec :: Spec
+errorRoutingBodyParseErrorSpec =
+  describe "Broken request body triggers error response (not re-routing)" $ do
+    let hs = (return . sum) :<|> return 2
+    describe "happy handlers" $ do
+      go hs "/sum"   "[1, 2]" "3"
+      go hs "/const" ""       ("2" { matchStatus = 201 })
+    describe "parse error" $ do
+      go hs "/sum"   "@@@"    400
+      go hs "/const" ""       ("2" { matchStatus = 201 })
+  where
+    go :: Server ErrorRoutingBodyParseError -> String -> String -> ResponseMatcher -> Spec
+    go hs path body resp = with (return $ serve (Proxy :: Proxy ErrorRoutingBodyParseError) hs) $
+      it (show (path, body)) $
+        Test.Hspec.Wai.post (cs path) (cs body) `shouldRespondWith` resp
+
+
+type ErrorRouting405
+  =    "content" :> Get '[JSON] Int
+  :<|> "content" :> Post '[JSON] Int
+  :<|> ""        :> Get '[JSON] Int
+
+errorRouting405 :: Spec
+errorRouting405 =
+  describe "Different methods on same end-point" $ do
+    let hs = return 1 :<|> return 2 :<|> return 3
+    go hs "get"  "/content" "1"
+    go hs "post" "/content" ("2" { matchStatus = 201 })
+    go hs "put"  "/content" 405
+    go hs "get"  "/"        "3"
+    go hs "put"  "/"        405
+    go hs "get"  "/n/a"     404
+    go hs "put"  "/n/a"     404
+  where
+    go :: Server ErrorRouting405 -> String -> String -> ResponseMatcher -> Spec
+    go hs method path resp = with (return $ serve (Proxy :: Proxy ErrorRouting405) hs) $ do
+      let runReq = case method of
+            "get"  -> Test.Hspec.Wai.get  (cs path)
+            "post" -> Test.Hspec.Wai.post (cs path) ""
+            "put"  -> Test.Hspec.Wai.put  (cs path) ""
+            _      -> error $ "errorRouting405: unknown method: " ++ show method
+          msg = show (method, path)
+      it msg $ runReq `shouldRespondWith` resp
--- a/servant-server/test/Servant/ServerSpec.hs
+++ b/servant-server/test/Servant/ServerSpec.hs
@@ -762,14 +762,16 @@ errorRoutingUser404Spec =
 type ErrorRoutingBodyParseError
   =    "sum" :> ReqBody '[JSON] [Int] :> Post '[JSON] Int
   :<|> "const"                        :> Post '[JSON] Int
+  :<|> Raw
 
 errorRoutingBodyParseErrorSpec :: Spec
 errorRoutingBodyParseErrorSpec =
   describe "Broken request body triggers error response (not re-routing)" $ do
-    let hs = (return . sum) :<|> return 2
+    let hs = (return . sum) :<|> return 2 :<|> (\_ cont -> cont (responseLBS ok200 [] "Raw"))
     describe "happy handlers" $ do
       go hs "/sum"   "[1, 2]" "3"
       go hs "/const" ""       ("2" { matchStatus = 201 })
+      go hs "/n/a"   ""       "Raw"
     describe "parse error" $ do
       go hs "/sum"   "@@@"    400
       go hs "/const" ""       ("2" { matchStatus = 201 })


jkarni:
Thanks, this is very useful! 

I thought (3) was the one we had issues with? (1) ( #209 ) is certainly not a single-line change.

(Also relevant is #197).

fisx:
You may be right: let's fix 3) first.

fisx:
Just realized that this was still not exactly what goes wrong over at liqd/thentos#285.  The last patch changes errors 2), 3):

```
  src/Test/Hspec/Wai.hs:101:
  2) Servant.Server, Broken request body triggers error response (not re-routing), happy handlers, ("/sum","[1, 2]")
       body mismatch:
         expected: 3
         but got:  Raw

  src/Test/Hspec/Wai.hs:101:
  3) Servant.Server, Broken request body triggers error response (not re-routing), parse error, ("/sum","@@@")
       status mismatch:
         expected: 400
         but got:  200
```

Still not sure what's going wrong in 2), but 3) is now exactly our problem: we what the parse error in the body to be reported to the user, but instead, the `Raw` catch-all handler gets called.

fisx:
(Don't really understand what I'm doing in the last commit.  Hope it helps anyway.)

alpmestan:
Very helpful! Also forces us to take a look at this sooner rather than later. Thanks

ChristianSi:
I don't understand the difference between (2) and (3). "Broken request body triggers error response (not re-routing)" is indeed the thing we urgently need for Thentos, but I don't see how there could be "happy handlers" if parsing the request body has already failed.

About (4, 5), I agree that different methods on the same endpoint should still fall through. E.g. if there is a POST handler registered for "my_custom_endpoint", OPTION requests to that endpoint should still fall through to a final Raw endpoint, if defined. Indeed, we rely on that behavior for the Thentos proxy.

About (1), I doubt that it's a good idea if "Handlers can trigger re-routing by responding with error status 404". If I throw a 404, I would expect that 404 to show up at the caller! It may indeed be a good idea to allow handlers to  trigger re-routing, but a special kind/instance of ServantErr should be used for that, not a standard status code.

kosmikus:
I'm lacking a bit of context here. Are these tests regressions between 0.4 and 0.5? I'm sure we have all sorts of dark corner cases right now, because we don't have a clear semantics. I'd much prefer if we could actually clarify the intended semantics and then "fix" whatever necessary with respect to that, rather than making a quick fix here and there.

fisx:
@kosmikus: not sure when that behavior was introduced, but i would expect that these tests never worked.  nevertheless, at least 3) is vital for our use case.

makes perfect sense to do this right.  since we are in pain, we will are interested in a quick fix anyway, but we can always work on a branch until this has been resolved properly.

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -728,3 +731,80 @@ miscReqCombinatorsSpec = with (return $ serve miscApi miscServ) $
...
+  =    "sum" :> ReqBody '[JSON] [Int] :> Post '[JSON] Int
+  :<|> "const"                        :> Post '[JSON] Int
+  :<|> Raw
+
+errorRoutingBodyParseErrorSpec :: Spec
+errorRoutingBodyParseErrorSpec =
+  describe "Broken request body triggers error response (not re-routing)" $ do
+    let hs = (return . sum) :<|> return 2 :<|> (\_ cont -> cont (responseLBS ok200 [] "Raw"))
+    describe "happy handlers" $ do
+      go hs "/sum"   "[1, 2]" "3"
```
jkarni: The reason this is failing is because the `Content-Type` header is not set, so we don't know how to decode it. There's an issue for trying to decode as the first content-type in the CT list ( #198 ), but I think we shouldn't try to do all of that in one go.

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -728,3 +731,80 @@ miscReqCombinatorsSpec = with (return $ serve miscApi miscServ) $
...
+      go hs "/const" ""       ("2" { matchStatus = 201 })
+      go hs "/n/a"   ""       "Raw"
+    describe "parse error" $ do
+      go hs "/sum"   "@@@"    400
```
jkarni: Ditto here.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -119,7 +123,7 @@ runAction action respond k = do
       e <- runEitherT a
       respond $ case e of
         Right x  -> k x
-        Left err -> succeedWith $ responseServantErr err
+        Left err -> failWithServantError err
```
jkarni: I think this is generally the right idea, but it's really unclear to me whether we want to just generally fail without further ado *here*.

jkarni:
I'm going to play around with immediate fixes [here](https://github.com/haskell-servant/servant/tree/jkarni/http-error-fixes), but I agree with @kosmikus - we really need to work the semantics for errors before making a release with major changes.

servant-server/test/Servant/ServerSpec.hs
fisx: ah.  should have guessed that from the error message.  i agree, i'll fix the bug and #198 should fixed elsewhere.

alpmestan:
I suggest we do that before we try to fix anything =) I/we (with Julian and other folks from **#servant**)  have actually had discussions about this very topic *many* times. I'll try and write down a concise but hopefully structured summary of those discussions/my thoughts below.

**tl;dr:** It seems to me everything boils down to chosing between backtracking and precise errors. If I'm wrong, please, pretty please let me know and prove me wrong in a comment below!

## Routing

The current routing implementation is really not that complicated. Whenever we see `endpointA :<|> endpointB` in an API type, the corresponding router combines "the router for A" and "the router for B", trying to make it reasonably efficient with a few smart tricks from @kosmikus' brain.

A [router](https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/Router.hs#L11) overall is something that can take a look at a [`Request`](https://hackage.haskell.org/package/wai-3.0.3.0/docs/Network-Wai.html#t:Request) and if it passes some some "checks" (e.g the path the request is made to starts with `"hello"`), we run the handler on it, get the result and encode it to finally respond to the client. Now, when one of those checks fails, we use the fact that our router type has ['RoutingApplication'](https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/RoutingApplication.hs#L28)s at its leaves instead of just "standard" WAI [`Application`](https://hackage.haskell.org/package/wai-3.0.3.0/docs/Network-Wai.html#t:Application)s, meaning that we can also say "hey, couldn't find a handler in our router for this request because *X*".

*X* could be "the request was made to */php* but we don't have any handler for requests made there, we only know about */haskell* and */rust*", as well as "we were well on our way to find a handler, but when trying to decode the JSON request body as a Haskell value it failed" and other reasons. [`RouteMismatch`](https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/RoutingApplication.hs#L50) precisely exists to desribe those reasons. 

This means the `RoutingApplication`s we find at the leaves of the `Router`s are just WAI `Application` that are aware of routing failures at a higher level than just with status codes in responses.

## Combining "routing results" ?

Ok so, when we see `endpointA :<|> endpointB` in an API type, we have a router that "makes a choice". It behaves in a smart way when `endpointA` and `endpointB` share some common prefix string etc.

If we try to route the request to A and it succeeds, we return the response it hands us back. But if it fails (incorrect request body content type or content "shape", route not found, etc), we try the one for B. If it succeeds at routing and gets a result from the handler, it returns that response. If we fail to route the request to B too though, what do we do?

Currently, [this](https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/RoutingApplication.hs#L50-L65). So each route mismatch reason has a "priority" basically, `NotFound` being the one with "minimum priority". When we fail to route a request to both A and B, we then combine the two route mismatches we got according to the `Monoid` instance highlighted in that link. `a <> b = max a b`& `mempty = NotFound`.

This might look fishy. It's most likely because it actually is. One reason why I'm writing all of this down is that with more eyes and brainpower on this I'm hoping we can simply rewrite those bits. Once we have good semantics it should be easy.

## The mighty `Left`

Things are about to get dirtier. So once a request is successfully routed to an handler (i.e the path matches, we're able to capture all the `Capture`s we need, we're able to read the request body from the right format into a haskell value, etc), we give everything to the handler and wait for it to return a result.

What's the handler type for a simple endpoint like the one below?

``` haskell
type Endpoint = Get '[JSON] String
```

The answer is... `Server (Get '[JSON] String)`, which is `EitherT ServantErr IO String`. We all know and love `Either` and `EitherT` for their ability to let us state that we are returning something that's either of one type or of another. In this case, the "normal", "if successful" results are just `String`s. But we can return an "untyped" HTTP response if things go wrong somewhere in the handler by just using [`left`](https://hackage.haskell.org/package/either-4.4.1/docs/Control-Monad-Trans-Either.html#v:left) and types/fields/values from the [`ServantErr`](https://hackage.haskell.org/package/servant-server-0.4.4.2/docs/Servant-Server-Internal-ServantErr.html) module.

What this concretely means is that there are two levels of failure. At the request handler level with this `left` machinery, and at the routing level when we write combinators and the `HasServer` instances for them, where we want to say "routing failed because of this precise reason". Whether the two should be unified... probably. @jkarni and I have discussed this a couple of times and both think this most likely is the right direction. What I'm saying here is that the semantics should go this far and take into account handler-level failures too. It really feels like there's a simpler solution, as in the code for it will be short, simple and reliably do what we want it to do.

## Backtracking

Now, have you ever considered/tried such an example:

``` haskell
type API = ReqBody '[JSON] Int :> Get '[JSON] Int
      :<|> ReqBody '[FormUrlEncoded] Blah :> Get '[JSON] String
```

Here, we're describing an API that "has two root endpoints". Imagine we're doing this because we want to handle `Int`s and `String`s. differently. In this case, the router will try to decode some JSON request body into an `Int` and run the handler for that case. If the request body can't be decoded into an Int, we try to route the request to the second endpoint, which will try to read the request body as a url-encoded form and then convert it to a value of type `Blah`, then run the handler for that case.

This means that when there's no particularly smart way to see if any of two endpoints separated by `:<|>` matches our request (e.g one is under "hello" and the other under "world", in which case we use a map to efficiently send the router to the appropriate handler by looking up the path the request is made to in that map), we'll just try them in order and return the response for the first one that matches; If none of them match, see above, we try and combine the "failures" so that we keep "the most relevant one", which doesn't actually happen in many cases, as evidenced by this PR & several IRC discussions / bug reports.

## Problems

Obviously, the problems arising from these few tests are only a small fraction of the overall problems people have seen with servant. Another common problem (reported by *dmj* just the other day on IRC) is that having an API with `ReqBody '[JSON] SomeType :> Get '[JSON] Blah   :<|>   Raw` will systematically return the error reported by `Raw`'s router instead of the one for the JSON endpoint, when you would have a malformed JSON body. This is *terrible*.

Why does this happen? Well, the JSON is malformed so the request body decoding doesn't work. This yields a `RouteMismatch`. However, routing to a `Raw` handler always succeeds (as in doesn't return a `RouteMismatch`), because it's an arbitrary wai `Application` so there's no constraint on the request to call the handler, there's nothing to lookup for the handler, the handler will just do it on its own in an "untyped and unsafe" way if needed.

Because the first endpoint gave us a `RouteMismatch`, we try the second one (remember? that's how we interpret `endpointA :<|> endpointB` in terms of routers). It's a `Raw` so it suceeds, but if that `Raw` is just `serveDirectory "./www"`, and we really meant to target the first endpoint, the odds are we'll get back a 404 response, or "Only GET or HEAD is supported" for when we're sending a POST, PUT or DELETE request.

Now it becomes interesting. On one hand,trying to route the request to the endpoint we mean to target yields a `RouteMismatch` because of our malformed JSON request body. On the other hand, routing the request to the `Raw` endpoint succeeds (that's easy, there's no single constraint to satisfy for the `Request`) but yields an HTTP error response (404, wrong method etc). We want to get the error about the malformed request body, but because we prioritize successful routing, we'll keep `Raw`'s crappy, unrelated error message. This clearly hints that there shouldn't be two levels of "failure" in my opinion.

## Solution

I'm hoping this comment provides the necessary context & pointers for anyone interested in giving a shot at redesigning these bits of *servant-server*'s internals. I'm sure we can unify the "failure levels" and simplify the whole thing around `RoutingApplication`. I'm unfortunately really, really busy these days so I can't take a stab at it for now. I'd be happy to provide guidance to the brave soul(s) =)

jkarni:
Picking up, here's an attempt to concretize where we want to be rather than
where we are.

# Semantics

This is based off of [webmachine's
 diagram](https://raw.githubusercontent.com/wiki/webmachine/webmachine/images/http-headers-status-v3.png).
 But it has some changes that I think are warranted for our use case, and
 simplifies (ignores) a lot of the things we don't yet care about.

1) Check whether one or more endpoints have the right path. Otherwise return 404.
2) Check whether the one of those have the right method. Otherwise return
405. If so, pick the first. We've now commited to calling at *most* one handler. \*
3) Check whether the Content-Type is known. Otherwise return 415.
4) Check whether that one deserializes the body. Otherwise return 400. If there
was no Content-Type, try the first one of the API content-type list.
5) Check whether the request is authorized. Otherwise return a 401.
6) Check whether the request is forbidden. If so return 403.
7) Check whether the request has a known Accept. Otherwise return 406. 
8) Check whether Accept-Language, Accept-Charset and Accept-Encoding exist and
match. We can follow the webmachine order here.
9) Call the handler. Whatever it returns, we return.

\* We can, I think, statically check that there is only one endpoint per method
here. The exception is Raw.

# Raw

Ideally, if we have an API

```
type Api1 = ... :<|> X :<|> ...
server1 :: Server Api1
server1 = ... :<|> x :<|> ...
```

The behaviour of this should be identical to:

```
type Api2 = ... :<|> Raw :<|> ...
server2 :: Server Api2
server2 = ... :<|> serve (Proxy :: Proxy X) x :<|> ...
```

In addition to making adding middleware to sub-apis easy, this is a litmus test
for composition with Raw being well-behaved. Right now for example, having Raw
at the start of the API is a disaster.

The solution might have been easy (just peek at the first line of the response
to check the status code) but:

# The issue

If a handler returns 404, we just pass that along. If a handler doesn't match,
however, we want to try the next one. That by itself is feasible, I think, but
it comes into conflict with the rule for Raw. Raw has a handler that may return
404 either because a handler did so, or because no handler matched. Similarly
with 405 (though perhaps we can assume no handler will return a 405). 404 and
405 are, by the description above, special, since they are the only checks that
occur before we committed to a single handler.

The only option I see is making two kinds of handler-404s. We could do this by
e.g throwing an exception when and endpoint doesn't match. Or passing a
continuation. This might be ugly, but it has the advantage of making it easy to
have custom 404s.

imalsogreg:
Re: the code @alpmestan posted, showing how `RouteMismatch`s are `mappend`ed together.. it might be nice if instead of reducing to a single `RouteResult`, we had `NodeMismatch` type, and while walking the API, we add nodes to the tree labeled by what happened there (success or failure). If the api is exhausted with no match, we return the whole tree of failures. If only a Raw matched, run the Raw handler and color that Raw node as 404 if the Raw handler is 404.

On success (when any non-raw route matches, or when a Raw route matches and runs without 404) discard the route history tree away return the handler's result. ? Not sure if it makes sense :)

ChristianSi:
I think the resolution algorithm described by @jkarni is a good one. It would certainly solve the problems we have with Thentos, and it would avoid the "terrible" effects described by @alpmestan.

Regarding whether 404 should be returned as failure or merely trigger re-routing, I would suggest using a different special error, maybe called "WrongRoute", to indicate that another handler should be called.

jkarni:
https://github.com/haskell-servant/servant/pull/218 has the tests for the semantics I gave above.

fisx:
comments on @jkarni's semantics:

re. 5), 6) isn't this stuff that needs to be implemented in servant?  these points sound like it's somehow baked into the framework.

re. 9) and `Raw`: so...  make RouteMismatch an exception and make servant wrap catchers around all handlers (or only `Raw` handlers?), and distinguish on the type level between type-level "route not found" and the run-time variant thrown by a, say, `Raw`?

@imalsogreg: how would you construct an HTTP response from a tree of 4 404s, one 405, and one 400?  (not sure if that makes sense either.  :-) is what you are suggesting implementable in terms of changes to the `Monoid` instance of `RouteMismatch`?)

jkarni:
Let's continue on #214 since this discussion is really about general ideas/issues rather than specific code review/PR

alpmestan:
On #224, you mean?

jkarni:
Right!

dredozubov:
In response to https://github.com/haskell-servant/servant/pull/214#issuecomment-138731211 proposal by @jkarni 

As i mentioned on IRC, i find 4) wrong and misleading. I believe there should be an option for overriding it by the user. Currently we can't have well-formed API responses for errors. Let's say we want to implement an API with a JSON response in the lines of: `{ "status": <status>, "response": <response> }`. `<status>` can be `ok`/`error`, and `response` can contain error message or response body itself. It's impossible to do, because there is no way of overriding decoding step to return something more meaningful than http://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/RoutingApplication.hs#L54. This can be resolved in two ways:
1. decoding step should be able to result in the same type as the corresponding request handler(associated type family, probably?)
2. handler must have access to the `RouteMismatch` returned by the previous steps.

I agree with @fisx - 5) and 6) doesn't seem like a servant material, at least at the moment.

jkarni:
@dredozubov I agree it'd be nice to configure the 400 errors more easily. We're already on it though. https://github.com/haskell-servant/servant/pull/218/files#diff-ff8d2e0e7588e514320a616559029328R42 includes work that would allow changing the 400 error response with combinators.

I disagree with your (2) though.  `RouteMismatch` (soon to be folded into `RouteResult`) is precisely for determining whether to call the handler, and which one to call. It doesn't make sense for a handler to have access to a 404 error, because well, what handler would it be that has access to it? And we decided (after quite a lot of discussion) that 400 errors should be allowed to trigger re-routing, so 400 will be analogous to 404. If the *handler* itself really wants to handle decoding errors, we have a perfectly good mechanism for that - just decode into an `Either DecodeError a` instead of an `a`. That decoding always succeeds, and then your handler can inspect the `Left` and throw whatever 400 it wants. 

Note also that we've already moved to `ServantErr` internally too in the `jkarni/http-errors` branch. So `HasServer` instances *can* set the body of the 404 to whatever they want.

dredozubov:
@jkarni I'd vote iin favor of this direction. I actually mentioned `RouteMismatch` in 2) because it currently contains `InvalidBody` constructor. It's not a question of concrete datatype, it's about ability to get to the "parsing context" in user handler.

jkarni:
Fixed in #218

PR #215, opened 2015-09-06
make travis builds use stack

soenkehahn:
This PR switches our travis builds to stack. Build times of this branch are about 4 minutes.

The one caveat (that has been discussed before) of this switch is that we don't test for hackage breakage through travis anymore. I talked about this with @jkarni and we thought it would be best to set up a separate CI (not through travis) for this. He's working on that.

cc: @alpmestan

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,27 @@
-language: haskell
+sudo: false
+
+addons:
+  apt:
+    packages:
+      - libgmp-dev
 
 env:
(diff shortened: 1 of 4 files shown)


jkarni:
LGTM

alpmestan:
Fine, fine, we might want to switch, travis is really almost not helping us anymore these days so we've got to do something. LGTM, merge this when you want guys =)

PR #217, opened 2015-09-08
Fix nix script

jml:
I tried running the nix script as per the README but got errors saying it couldn't change into each of the various directories. This was because it never changed out of the directory it changed into. 

This patch updates `update-nix-files` to use `pushd` and `popd`, and also contains the results of a run of `update-nix-files.sh`.

[Review on Reviewable](https://reviewable.io/reviews/haskell-servant/servant/217)


codedmart:
@alpmestan @jkarni I really think we should remove/ignore the nix files per sub dir and just keep main `update-nix-files.sh` so users can easily generate the files they need. It gets annoying having the sub nix files as part of all the PR's. Any objections?

jkarni:
@jml Thanks a lot!
Does it perhaps make more sense to remove (and add to gitignore) the `default[/shell].nix` files, and just keep `update-nix-files.sh` (and maybe rename that to `create-nix-files.sh`?

codedmart:
@jkarni Haha, yes it does. @jml Do you want to add that to this PR or better state remove those files and gitignore them?

jml:
Absolutely. I was surprised they were in there in the first place. 

I chose 'generate' rather than 'create' because it has fewer connotations of making something that doesn't exist.

codedmart:
@jml Fantastic thanks!

jkarni:
Thanks a lot!

Issue #41, opened 2015-04-20
`:<|>` tries subsequent `Raw`s even for bad requests


jkarni:
_From @soenkehahn on April 17, 2015 10:37_

@phunehehe and I came up with a test case for this: https://github.com/haskell-servant/servant-server/compare/badRequest-bug?expand=1

I'm not sure this is a bug or wanted behaviour though.

_Copied from original issue: haskell-servant/servant-server#42_

jkarni:
Ouch, tricky one. My intuition is that the `Raw` endpoint should be tried (any non-complete match should try other endpoints), but if the `Raw` endpoint fails, `400` should be returned. We'd have to check whether the `Raw` endpoint returned `2XX` though (not impossible, just annoying).

phunehehe:
My intuition is a little different :)
I think the URL on its own should determine the handler. If the handler says "bad request" or throw an error, that should be the final answer. This mapping feels "clean" to me.
But of course, that's just an intuition.

jkarni:
The problem is that it's a *feature* that `Raw` doesn't need to match on endpoints. It means one can mount an entire other web server, with its own routing, there. So we can't assume that the route doesn't in fact match just because from the outside, so to speak, it looks that way. Internally, the `Raw` endpoint may well match against the route too.

phunehehe:
Ah I was talking about other, "normal", endpoints. Do you think Raw should still be tried if the request already matches another endpoint?

alpmestan:
@jkarni With the new routing machinery, is this issue solved?

jkarni:
No, I don't think so. I'm still unsure of what should be considered an appropriate solution. We've so far not had the semantics that body mismatches prevent future endpoints from being tried (so you could have the same endpoint with different `ReqBody`s and it would make sense). @phunehehe's point that this doesn't match up with how most web frameworks (where URL alone determines handlers) behave is well taken. OTOH from within the logic of servant the current behaviour also makes sense: URL match, method match, request-body match, etc. are all just predicates that stand on the same footing, and a handler is picked when all predicates are true. 

So we have to decide: with something like:

         ReqBody '[JSON] Person :> Get '[JSON] ()
         ReqBody '[JSON] Thing   :> Get '[JSON] ()

Would a request with a body that can be deserialized as a `Thing` but not `Person` get a 400 or a 204? Note that this also applies to e.g. differing content-types, and other things besides.

[...]

PR #218, opened 2015-09-09
Fix error priority and retries

jkarni:
Looking pretty bad:
```
servant-server-0.5: test (suite: doctests, args: -m "HTTP Errors")
Examples: 7  Tried: 7  Errors: 0  Failures: 0
servant-server-0.5: test (suite: spec, args: -m "HTTP Errors")

Servant.Server.Error
  HTTP Errors
    HTTP error order
      has 404 as its highest priority error FAILED [1]
      has 405 as its second highest priority error
      has 415 as its third highest priority error
      has 400 as its fourth highest priority error
      has 406 as its fifth highest priority error FAILED [2]
      returns handler errors as its lower priority errors
    Handler search
      should continue when URLs don't match FAILED [3]
      should continue when methods don't match FAILED [4]
      should not continue when Content-Types don't match
      should not continue when body can't be deserialized FAILED [5]
      should not continue when Accepts don't match FAILED [6]

Failures:

  test/Servant/Server/ErrorSpec.hs:65:
  1) Servant.Server.Error, HTTP Errors, HTTP error order, has 404 as its highest priority error
       src/Test/Hspec/Wai.hs:101:
       status mismatch:
         expected: 404
         but got:  405

  test/Servant/Server/ErrorSpec.hs:81:
  2) Servant.Server.Error, HTTP Errors, HTTP error order, has 406 as its fifth highest priority error
       src/Test/Hspec/Wai.hs:101:
       status mismatch:
         expected: 406
         but got:  402

  test/Servant/Server/ErrorSpec.hs:124:
  3) Servant.Server.Error, HTTP Errors, Handler search, should continue when URLs don't match
       src/Test/Hspec/Wai.hs:101:
       status mismatch:
         expected: 201
         but got:  405
       body mismatch:
         expected: 5
         but got:  method not allowed

  test/Servant/Server/ErrorSpec.hs:128:
  4) Servant.Server.Error, HTTP Errors, Handler search, should continue when methods don't match
       src/Test/Hspec/Wai.hs:101:
       status mismatch:
         expected: 200
         but got:  415
       body mismatch:
         expected: 4
         but got:  unsupported media type

  test/Servant/Server/ErrorSpec.hs:136:
  5) Servant.Server.Error, HTTP Errors, Handler search, should not continue when body can't be deserialized
       src/Test/Hspec/Wai.hs:101:
       status mismatch:
         expected: 400
         but got:  201

  test/Servant/Server/ErrorSpec.hs:140:
  6) Servant.Server.Error, HTTP Errors, Handler search, should not continue when Accepts don't match
       src/Test/Hspec/Wai.hs:101:
       status mismatch:
         expected: 406
         but got:  201

Randomized with seed 515273500

Finished in 0.0034 seconds
11 examples, 6 failures
Test suite failure for package servant-server-0.5
    spec:  exited with: ExitFailure 1
Logs printed to console

```

### DIFF
--- a/servant-examples/auth-combinator/auth-combinator.hs
+++ b/servant-examples/auth-combinator/auth-combinator.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 {-# LANGUAGE TypeOperators       #-}
+
 import           Data.Aeson
 import           Data.ByteString          (ByteString)
 import           Data.Text                (Text)
@@ -31,12 +32,12 @@ instance HasServer rest => HasServer (AuthProtected :> rest) where
   route Proxy a = WithRequest $ \ request ->
     route (Proxy :: Proxy rest) $ do
       case lookup "Cookie" (requestHeaders request) of
-        Nothing -> return $ failWith $ HttpError status401 (Just "Missing auth header.")
+        Nothing -> return $! failFatallyWith err401 { errBody = "Missing auth header" }
         Just v  -> do
           authGranted <- isGoodCookie v
           if authGranted
             then a
-            else return $ failWith $ HttpError status403 (Just "Invalid cookie.")
+            else return $! failFatallyWith err403 { errBody = "Invalid cookie" }
 
 type PrivateAPI = Get '[JSON] [PrivateData]
 
--- a/servant-examples/auth-combinator/auth-combinator.hs
+++ b/servant-examples/auth-combinator/auth-combinator.hs
@@ -10,7 +10,6 @@ import           Data.Aeson
 import           Data.ByteString          (ByteString)
 import           Data.Text                (Text)
 import           GHC.Generics
-import           Network.HTTP.Types
 import           Network.Wai
 import           Network.Wai.Handler.Warp
 import           Servant
@@ -29,15 +28,16 @@ data AuthProtected
 instance HasServer rest => HasServer (AuthProtected :> rest) where
   type ServerT (AuthProtected :> rest) m = ServerT rest m
 
-  route Proxy a = WithRequest $ \ request ->
-    route (Proxy :: Proxy rest) $ do
-      case lookup "Cookie" (requestHeaders request) of
-        Nothing -> return $! FailFatal err401 { errBody = "Missing auth header" }
-        Just v  -> do
-          authGranted <- isGoodCookie v
-          if authGranted
-            then a
-            else return $ FailFatal err403 { errBody = "Invalid cookie" }
+  route Proxy subserver = WithRequest $ \ request ->
+    route (Proxy :: Proxy rest) $ addAcceptCheck subserver $ cookieCheck request
+      where
+        cookieCheck req = case lookup "Cookie" (requestHeaders req) of
+            Nothing -> return $ FailFatal err401 { errBody = "Missing auth header" }
+            Just v  -> do
+              authGranted <- isGoodCookie v
+              if authGranted
+                then return $ Route ()
+                else return $ FailFatal err403 { errBody = "Invalid cookie" }
 
 type PrivateAPI = Get '[JSON] [PrivateData]
 
--- a/servant-examples/auth-combinator/auth-combinator.hs
+++ b/servant-examples/auth-combinator/auth-combinator.hs
@@ -32,12 +32,12 @@ instance HasServer rest => HasServer (AuthProtected :> rest) where
   route Proxy a = WithRequest $ \ request ->
     route (Proxy :: Proxy rest) $ do
       case lookup "Cookie" (requestHeaders request) of
-        Nothing -> return $! failFatallyWith err401 { errBody = "Missing auth header" }
+        Nothing -> return $! FailFatal err401 { errBody = "Missing auth header" }
         Just v  -> do
           authGranted <- isGoodCookie v
           if authGranted
             then a
-            else return $! failFatallyWith err403 { errBody = "Invalid cookie" }
+            else return $ FailFatal err403 { errBody = "Invalid cookie" }
 
 type PrivateAPI = Get '[JSON] [PrivateData]
 
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -113,11 +113,10 @@ instance (KnownSymbol capture, FromHttpApiData a, HasServer sublayout)
      a -> ServerT sublayout m
 
   route Proxy subserver =
-    DynamicRouter $ \ first ->
-      route (Proxy :: Proxy sublayout)
-            (case captured captureProxy first of
-               Nothing  -> return $ failWith NotFound
-               Just v   -> feedTo subserver v)
+    DynamicRouter $ \ first -> case captured captureProxy first of
+       Nothing  -> LeafRouter (\_ r -> r $ failWith err404)
+       Just v   -> route (Proxy :: Proxy sublayout) (feedTo subserver v)
+
     where captureProxy = Proxy :: Proxy (Capture capture a)
 
 allowedMethodHead :: Method -> Request -> Bool
@@ -131,7 +130,7 @@ processMethodRouter :: forall a. ConvertibleStrings a B.ByteString
                     -> Maybe [(HeaderName, B.ByteString)]
                     -> Request -> RouteResult Response
 processMethodRouter handleA status method headers request = case handleA of
-  Nothing -> failWith UnsupportedMediaType
+  Nothing -> failFatallyWith err406
   Just (contentT, body) -> succeedWith $ responseLBS status hdrs bdy
     where
       bdy = if allowedMethodHead method request then "" else body
@@ -150,8 +149,8 @@ methodRouter method proxy status action = LeafRouter route'
                 handleA = handleAcceptH proxy (AcceptHeader accH) output
             processMethodRouter handleA status method Nothing request
       | pathIsEmpty request && requestMethod request /= method =
-          respond $ failWith WrongMethod
-      | otherwise = respond $ failWith NotFound
+          respond $ failWith err405
+      | otherwise = respond $ failWith err404
 
 methodRouterHeaders :: (GetHeaders (Headers h v), AllCTRender ctypes v)
                     => Method -> Proxy ctypes -> Status
@@ -167,8 +166,8 @@ methodRouterHeaders method proxy status action = LeafRouter route'
               handleA = handleAcceptH proxy (AcceptHeader accH) (getResponse output)
           processMethodRouter handleA status method (Just headers) request
       | pathIsEmpty request && requestMethod request /= method =
-          respond $ failWith WrongMethod
-      | otherwise = respond $ failWith NotFound
+          respond $ failWith err405
+      | otherwise = respond $ failWith err404
 
 methodRouterEmpty :: Method
                   -> IO (RouteResult (ExceptT ServantErr IO ()))
@@ -180,8 +179,8 @@ methodRouterEmpty method action = LeafRouter route'
           runAction action respond $ \ () ->
             succeedWith $ responseLBS noContent204 [] ""
       | pathIsEmpty request && requestMethod request /= method =
-          respond $ failWith WrongMethod
-      | otherwise = respond $ failWith NotFound
+          respond $ failWith err405
+      | otherwise = respond $ failWith err404
 
 -- | If you have a 'Delete' endpoint in your API,
 -- the handler for this endpoint is meant to delete
@@ -558,8 +557,9 @@ instance HasServer Raw where
   route Proxy rawApplication = LeafRouter $ \ request respond -> do
     r <- rawApplication
     case r of
-      RR (Left err)  -> respond $ failWith err
-      RR (Right app) -> app request (respond . succeedWith)
+      HandlerVal app  -> app request (respond . succeedWith)
+      Retriable  e    -> respond $ failWith e
+      NonRetriable e  -> respond $! failFatallyWith e
 
 -- | If you use 'ReqBody' in one of the endpoints for your API,
 -- this automatically requires your server-side handler to be a function
@@ -599,8 +599,8 @@ instance ( AllCTUnrender list a, HasServer sublayout
       mrqbody <- handleCTypeH (Proxy :: Proxy list) (cs contentTypeH)
              <$> lazyRequestBody request
       case mrqbody of
-        Nothing -> return $ failWith $ UnsupportedMediaType
-        Just (Left e) -> return $ failWith $ InvalidBody e
+        Nothing -> return $! failFatallyWith err415
+        Just (Left e) -> return $! failFatallyWith err400 { errBody = cs e }
         Just (Right v) -> feedTo subserver v
 
 -- | Make sure the incoming request starts with @"/path"@, strip it and
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -1,15 +1,15 @@
-{-# LANGUAGE CPP                  #-}
-{-# LANGUAGE DataKinds            #-}
-{-# LANGUAGE FlexibleContexts     #-}
-{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE OverloadedStrings    #-}
-{-# LANGUAGE PolyKinds            #-}
-{-# LANGUAGE ScopedTypeVariables  #-}
-{-# LANGUAGE TypeFamilies         #-}
-{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE PolyKinds                  #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE TypeOperators              #-}
 #if !MIN_VERSION_base(4,8,0)
-{-# LANGUAGE OverlappingInstances #-}
+{-# LANGUAGE OverlappingInstances       #-}
 #endif
 
 module Servant.Server.Internal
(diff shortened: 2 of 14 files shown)


jkarni:
A couple of the later failures were my mistake, and I opened a separate issue for (4) ( #219 ). Besides that, I fixed all but the first failure, which is baffling me still.

jkarni:
Closes #197, #41, #48

jkarni:
Maybe [`prioErrorSpec`](https://github.com/haskell-servant/servant/blob/master/servant-server/test/Servant/ServerSpec.hs#L617) should be moved to `ErrorSpec.hs`?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -30,39 +26,12 @@ type RoutingApplication =
...
-  --
-  -- "As one judge said to the other, 'Be just and if you can't be just, be
-  -- arbitrary'" -- William Burroughs
-  mappend = max
+data RouteResult a =
+    Retriable ServantErr      -- ^ Keep trying other paths. The @ServantErr@
+                              -- should only be 404 or 405.
+  | NonRetriable ServantErr   -- ^ Stop trying.
+  | HandlerVal a
+  deriving (Eq, Show, Read, Functor)
```
alpmestan: This brilliantly cleans up the internals of servant quite significantly. However, I suggest we rename the constructors of this data type. `Retriable`/`NonRetriable` hold `ServantErr`s, and the errors themselves are not "retriable" or "non retriable". They are either "final" (as in "this is the definitive error we will return") or not. I suggest we directly reflect this in the names.

How about:

``` haskell
data RouteResult a
  = Fail ServantErr
  | Fatal ServantErr
  | Routed a
```

PS: I've also been wondering whether we really want `ServantErr` instead of `Response` so that folks could just use the functions from WAI and not record-update our `errXXX`.

alpmestan:
@jkarni Yeah, definitely.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
jkarni: Sounds reasonable. How about then also removing the `failWith`, `failFatallyWith`, and `succeedWith` functions? Also, mind if instead of `Routed` we go with `Route`? I'd like to be consistent about using the verb form if we go that way.

jkarni: Well, `fatal` is also not a verb. How about `Fail`, `FailFatal`, and `Route`?

alpmestan: `FailFatal` really looks/sounds ugly to me.

jkarni: Other ideas? I'd really like to have a single part-of-speech for all constructors.

alpmestan:
OK, fine, `FailFatal` it is then, I don't have a better suggestion. =)

jkarni:
@kosmikus if you have time I'd love to hear your thoughts as well.

jkarni:
This PR would also make it an option to remove the request body memoization that we've been doing.

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -31,12 +32,12 @@ instance HasServer rest => HasServer (AuthProtected :> rest) where
   route Proxy a = WithRequest $ \ request ->
     route (Proxy :: Proxy rest) $ do
       case lookup "Cookie" (requestHeaders request) of
-        Nothing -> return $ failWith $ HttpError status401 (Just "Missing auth header.")
+        Nothing -> return $! FailFatal err401 { errBody = "Missing auth header" }
```
kosmikus: Why the `$!`?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -116,11 +118,10 @@ instance (KnownSymbol capture, FromText a, HasServer sublayout)
...
   route Proxy subserver =
-    DynamicRouter $ \ first ->
-      route (Proxy :: Proxy sublayout)
-            (case captured captureProxy first of
-               Nothing  -> return $ failWith NotFound
-               Just v   -> feedTo subserver v)
+    DynamicRouter $ \ first -> case captured captureProxy first of
+       Nothing  -> LeafRouter (\_ r -> r $ Fail err404)
+       Just v   -> route (Proxy :: Proxy sublayout) (feedTo subserver v)
+
```
kosmikus: This is in deviation from the desired behaviour we discussed earlier, because it means that captures are checked immediately. In particular, this also means that if we have several routes such as
```
     "foo" :> Capture ... :> "bar"
:<|> "foo" :> Capture ... :> "baz"
```
then the `Capture` will be performed multiple times unless you left-factor by hand.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -181,10 +182,10 @@ methodRouterEmpty method action = LeafRouter route'
     route' request respond
       | pathIsEmpty request && allowedMethod method request = do
           runAction action respond $ \ () ->
-            succeedWith $ responseLBS noContent204 [] ""
+            Route $! responseLBS noContent204 [] ""
```
kosmikus: Again, why the `$!`. Ok, here it has an effect. But if you want to say that `Route` should always have a WHNF-evaluated argument, you should make the constructor strict, not each call.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -134,8 +135,8 @@ processMethodRouter :: forall a. ConvertibleStrings a B.ByteString
                     -> Maybe [(HeaderName, B.ByteString)]
                     -> Request -> RouteResult Response
 processMethodRouter handleA status method headers request = case handleA of
-  Nothing -> failWith UnsupportedMediaType
-  Just (contentT, body) -> succeedWith $ responseLBS status hdrs bdy
+  Nothing -> FailFatal err406
```
kosmikus: I don't see why this should be fatal. In my opinion,
```
Put '[Foo, Bar] T
```
should be equivalent to
```
     Put '[Foo] T
:<|> Put '[Bar] T
```
for reasons of modularity. This doesn't seem to be supported if we make the failure fatal at this point.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -134,8 +135,8 @@ processMethodRouter :: forall a. ConvertibleStrings a B.ByteString
...
+  Just (contentT, body) -> Route $! responseLBS status hdrs bdy
```
kosmikus: One more `$!`.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -718,8 +720,8 @@ instance ( AllCTUnrender list a, HasServer sublayout
       mrqbody <- handleCTypeH (Proxy :: Proxy list) (cs contentTypeH)
              <$> lazyRequestBody request
       case mrqbody of
-        Nothing -> return $ failWith $ UnsupportedMediaType
-        Just (Left e) -> return $ failWith $ InvalidBody e
+        Nothing -> return $ FailFatal err415
```
kosmikus: Again, I don't see why this should be fatal.

servant-server/src/Servant/Server/Internal.hs
jkarni: Why shouldn't captures be checked immediately? Currently by delaying it we let an endpoint with a capture followed by a method return a 405 (bad method) instead of a 404 if both the method and the url don't match. (Sorry, I don't remember this discussion!)

> then the `Capture` will be performed multiple times unless you left-factor by hand.

That does seem undesirable.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -718,8 +720,8 @@ instance ( AllCTUnrender list a, HasServer sublayout
...
+        Just (Left e) -> return $ FailFatal err400 { errBody = cs e }
```
kosmikus: Dito.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -30,39 +26,12 @@ type RoutingApplication =
...
-  -- The following isn't great, since it picks @InvalidBody@ based on
-  -- alphabetical ordering, but any choice would be arbitrary.
-  --
-  -- "As one judge said to the other, 'Be just and if you can't be just, be
-  -- arbitrary'" -- William Burroughs
-  mappend = max
+data RouteResult a =
+    Fail ServantErr           -- ^ Keep trying other paths. The @ServantErr@
+                              -- should only be 404 or 405.
+  | FailFatal ServantErr      -- ^ Don't other paths.
```
kosmikus: "Don't try other paths."

servant-examples/auth-combinator/auth-combinator.hs
jkarni: No real reason - at some point in the writing of this there was a function instead of a constructor on the RHS and the `$!` seemed to make sense in that context, but not anymore. I'll remove

servant-server/test/Servant/Server/ErrorSpec.hs
```diff
@@ -0,0 +1,251 @@
...
+
+import           Servant
+
+
+-- The semantics of routing and handling requests should be as follows:
+--
+--    1) Check whether one or more endpoints have the right path. Otherwise
+--    return 404.
+--    2) Check whether the one of those have the right method. Otherwise return
+--    405. If so, pick the first. We've now commited to calling at most one
```
kosmikus: committed

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -55,18 +59,23 @@ runRouter (StaticRouter table) request respond =
...
-
+  runRouter r1 request $ \ mResponse1 -> case mResponse1 of
+    Fail _ -> runRouter r2 request $ \ mResponse2 ->
+      respond (highestPri mResponse1 mResponse2)
+    _      -> respond mResponse1
+  where
+    highestPri (Fail e1) (Fail e2) =
+      if errHTTPCode e1 == 404 && errHTTPCode e2 /= 404
+        then Fail e2
+        else Fail e1
```
kosmikus: This looks a bit ad-hoc, and incomplete. (But I guess you can afford to have this incomplete because you make more failures fatal than I would.)

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -55,18 +59,23 @@ runRouter (StaticRouter table) request respond =
...
+    highestPri (Fail _) y = y
+    highestPri x _ = x
```
kosmikus: This is the point where `FailFatal` is actually treated as being fatal, right? This should be documented, and possibly be made explicit in pattern matching rather than hidden in a catch-all.

servant-server/src/Servant/Server/Internal/Router.hs
jkarni: The idea is that only path and method are used to identify resources/handlers. So anything that's not a path or method mismatch (404 or 405) is fatal.

servant-server/src/Servant/Server/Internal.hs
kosmikus: That seems strange, but is probably something that should rather be fixed in the method combinator, by delaying the method check until after processing all other delayed checks. But I'm pretty positive that delaying captures is a good thing. Static matching is the fastest dispatch method we have, and first checking all static path components before checking any dynamic ones seems like a huge improvement of the new routing scheme that I'm not willing to easily give up again.

kosmikus:
Done with my pass now. Design looks ok to me, although I'd really like to avoid the use of `FailFatal` as much as possible.

Do we still need some kind of "greedy or" or some "turn `Fail` into `FailFatal`" though, as far as I can see, in order to support the "`Raw` as a last resort" pattern?

servant-server/src/Servant/Server/Internal.hs
jkarni: Good point. I'll try to figure out how to delay captures until just before the method check.

servant-server/src/Servant/Server/Internal.hs
jkarni: Also supported would be 
```
     Put '[Foo] A
:<|> Put '[Bar] B
```
Which as I've mentioned elsewhere I think is a bad idea, since content-type should not be allowed to affect the nature (type) of the resource, only it's presentation.

servant-server/src/Servant/Server/Internal.hs
jkarni: Here there's the added advantage that we can get rid of request body memoization, which to my mind seems like a big win. And this gives a somewhat more principled notion of error messages - there can only be one endpoint that can ever deserialize the body, so you always get the correct error message from deserialization errors.

 On the other hand, there isn't the argument (that I think there is with content-types) about it actually leading to better design.

servant-server/src/Servant/Server/Internal.hs
kosmikus: If you rule it out, though, it still means I cannot do
```
type MyAPI c = ... Put '[c] T ...
```
and then define severs for `MyAPI Foo` and `MyAPI Bar` separately and subsequently `:<|>` them.

So you'd argue that because we cannot rule out that types might be different, we should rule out the whole pattern, and because we cannot rule that out either, we effectively make all but the first branch unreachable?

jkarni: OK, good point. `Fail` it is.

alpmestan: @kosmikus They would be equivalent only if in

``` haskell
     Put '[Foo] T
:<|> Put '[Bar] T
```

both endpoints do the thing that the "one endpoint only" version does. Right?

servant-server/src/Servant/Server/Internal.hs
kosmikus: Ok, getting rid of request body memoization would be nice indeed. But the price seems high.

servant-server/src/Servant/Server/Internal/Router.hs
alpmestan: +1 on this.

servant-server/src/Servant/Server/Internal.hs
jkarni: I'm going to keep arguing just because it helps to clarify things. Ok, so let's say instead of:

> type MyAPI c = ... Put '[c] T ...

You have to write

> type MyAPI c = ... Put c T ...

Was much lost in the process?

(As a side note, we *can* rule out the pattern - we can just run some checks over the API type if we need to.)

servant-server/src/Servant/Server/Internal/Router.hs
kosmikus: Right. I would have liked to add content types and conversion methods into the mix. So this seems to be the source of our disagreement here.

servant-server/src/Servant/Server/Internal.hs
alpmestan: > OK, good point. `Fail` it is.

So we can have it be a `Fail` without getting the wrong response?

jkarni:
I removed the memoReqBody stuff, though that's not to say I'm committed to `FailFatal` on 400.

kosmikus:
@jkarni What actually happens if we do `ReqBody` twice *on the same route*?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -91,55 +62,191 @@ toApplication ra request respond = do
...
+-- it does not provide an input for the handler. Method checks
+-- are comparatively cheap.
+--
+-- 3. Body and accept header checks. The request body check can
+-- cause both 400 and 415. This provides an input to the handler.
+-- The accept header check can be performed as the final
+-- computation in this block. It can cause a 406.
+--
+data Delayed :: * -> * where
+  Delayed :: IO (RouteResult a)
```
jkarni: Would this be better as `State` over the four args?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -60,10 +63,11 @@ import           Servant.Server.Internal.Router
 import           Servant.Server.Internal.RoutingApplication
 import           Servant.Server.Internal.ServantErr
 
+
 class HasServer layout where
   type ServerT layout (m :: * -> *) :: *
 
-  route :: Proxy layout -> IO (RouteResult (Server layout)) -> Router
+  route :: Proxy layout -> Delayed (Server layout) -> Router
```
jkarni: It's not quite as obvious how to do `IO` in the combinator instances (as for example in [auth-combinator.hs](https://github.com/haskell-servant/servant/blob/master/servant-examples/auth-combinator/auth-combinator.hs#L36)). I imagine we'd need a `MonadIO` instance (which to my mind is another reason to consider switching to `StateT`).

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
jkarni: Actually, I'm not sure I know how that'd work...

kosmikus: So you'd want to see the whole thing as a state monad (or perhaps just a reader monad)? Possible, but hardly a win, I'd say.

servant-server/src/Servant/Server/Internal.hs
jkarni: Actually, that won't work either...

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
jkarni: Discussed in the servant IRC channel, but for posterity: this datatype looks a lot like `webmachine`'s `Resource`, and maybe that could be inspiration for how many fields (and which) we should have.

jkarni:
Another idea discussed on IRC is that we could use the same existential mechanism to allow for combinators that modify how e.g. ReqBody parsing errors are displayed.

jkarni:
Sorry, this still has changes that I meant to remove. Not yet ready for review.

jkarni:
Ok, this is ready for review. I'll squash some commits after review.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -91,55 +60,192 @@ toApplication ra request respond = do
...
-  routingRespond (Right response) =
-    respond response
-
-runAction :: IO (RouteResult (ExceptT ServantErr IO a))
+  routingRespond :: RouteResult Response -> IO ResponseReceived
+  routingRespond (Fail err)      = respond $ responseServantErr err
+  routingRespond (FailFatal err) = respond $ responseServantErr err
+  routingRespond (Route v)       = respond v
+
+-- TODO: The above may not be quite right yet.
```
arianvp: Can you elaborate? Or can this be removed now?

jkarni: @kosmikus might know. I don't...

kosmikus: You mean this is my code? I don't remember, but I don't think this looks very suspicious.

jkarni: To me neither. I'll remove it then.

PR #221, opened 2015-09-12
Drop EitherT in favor of ExceptT

### LINKED ISSUE #220: Use ExceptT instead of EitherT
Drop the `either` dependency and use `ExceptT` from `transformers` instead.

int-index:
closes #220


alpmestan:
LGTM. @jkarni ? So do we flip the switch for the next major release?

jkarni:
LGTM

jkarni:
@int-index Thanks!

PR #225, opened 2015-09-15
fix for lens-4.13

soenkehahn:
`List` is a pattern synonym added in `lens-4.13`. (It doesn't show up in the documentation on hackage currently -- I assume haddock doesn't support pattern synonyms?)


jkarni:
LGTM (though the fact that we are breaking on a PVP-conformant change is an argument in favour of explicit or qualified imports...)

soenkehahn:
Yeah, although situations like this won't occur very often, I hope.

Issue #211, opened 2015-09-02
ToSample grievances


jkarni:
The two args for `ToSample` are necessary internally, but there should be a different class that just has one, so it's simple to implement, with an instance for it against the underlying `ToSample`. 

Moreover some basic instances (`()`, `ToSample a a => ToSample [a] [a]`, `ToSample a, ToSample b => ToSample (a, b)`) are missing.

(See also #151)

fizruk:
Why does `ToSample` need two arguments internally?

jkarni:
@fizruk sorry for taking so long to respond. The reason is that in *one* case (`ResponseHeaders`) the datatype we get and the datatype we need a sample for are different. So the first argument is `ResponseHeaders hdrs a` and the second is `a`.

jkarni:
I took a look at having two separate classes, and don't think it'll work due to ambiguity issues. Closing since the rest of this task has been done in #227 .

fizruk:
Okay, but I can't see do we need `ToSample` instance for `Headers ls a` anyway: it is not used anywhere in the code (unless you consider nested `Headers ls (Headers ls' a)`, which should not happen AFAICT).

I have just removed the second argument from `ToSample` together with `ToSample (Headers ls a)` instance, built and run tests and everything looks fine.

So is there any problem I don't see?

jkarni:
>I have just removed the second argument from ToSample together with ToSample (Headers ls a) instance, built and run tests and everything looks fine.
> So is there any problem I don't see?

Missing tests :).

You can have an endpoint such as `EitherT ... (Headers ls a)`, and we want to be able to provide a sample response. But the sample response (body) should be an `a`, not a `Headers ls a`. So the reasoning was we'd have the second param to indicate the sample type in that case would be an `a`. It doesn't need to show up in the code, though.

One option would be to have a single param, and then `ToSample a => ToSample (Headers ls a)`. But of course then we need `ToJSON a => ToJSON (Headers ls a)`. That's not bad, but the idea is that we can have samples in other content-types extensibly, so users declaring a new content-type would have to do something similar (for `ToXML` or whatever). It still might be a better idea.

Does that make sense? Any thoughts?

fizruk:
> You can have an endpoint such as `EitherT ... (Headers ls a)`...

Sorry, I don't understand what you mean by this. Do you mean something like

```haskell
type TestApi1 = Get '[JSON, PlainText] (Headers '[Header "Location" String] Int)
```

[...]

PR #227, opened 2015-09-16
ToSample instances for tuples.

fisx:
On request by @jkarni.

Not sure about the cross product for the individual tuple components.  Are there better ways?  I guess you can always overload the instance if output gets too big.

I couldn't come up with interesting tests.  Can you?

Would you like support for 12-tuples?  What's the right number to stop at?

Lists?  Maybe?  Either?  What else?  I think it's most interesting for tuples, but I am happy to add more.

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -1006,3 +1006,69 @@ instance ( ToSample a a
                  ", " <> te <>
                  ")"
                 , (va, vb, vc, vd, ve))
+
+instance ( ToSample a a
+         , ToSample b b
+         , ToSample c c
+         , ToSample d d
+         , ToSample e e
+         , ToSample f f
+         ) => ToSample (a, b, c, d, e, f) (a, b, c, d, e, f) where
+    toSample _ = (,,,,,)
+        <$> toSample (Proxy :: Proxy a)
+        <*> toSample (Proxy :: Proxy b)
+        <*> toSample (Proxy :: Proxy c)
+        <*> toSample (Proxy :: Proxy d)
+        <*> toSample (Proxy :: Proxy e)
+        <*> toSample (Proxy :: Proxy f)
+    toSamples _ = render
+        <$> toSamples (Proxy :: Proxy a)
+        <*> toSamples (Proxy :: Proxy b)
+        <*> toSamples (Proxy :: Proxy c)
+        <*> toSamples (Proxy :: Proxy d)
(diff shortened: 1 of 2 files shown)


jkarni:
Sweet! 
Re: tests - that the `ToSamples` look like you'd expect?
Re: other instances. More tuples (7? who knows) would be welcome, but list, Maybe, and Either all have issues (which `ToSample` does one pick?).

jkarni:
Sweet, thanks a lot!

fizruk:
I know I'm a bit late for this, but I guess these tuple instances can be made a bit more general:

```haskell
instance (ToSample a a', ToSample b b') => ToSample (a, b) (a', b') where
  ...
```

Also regarding `[]` and `Maybe`, I think something like this would work:

```haskell
instance ToSample a b => ToSample [a] [b] where
  toSamples _ = case toSamples (Proxy :: Proxy a) of
    []      -> [("empty", [])]
    samples -> [("empty", []), ("non-empty", map snd samples)]

instance ToSample a b => ToSample (Maybe a) (Maybe b) where
  toSamples _ = case toSample (Proxy :: Proxy a) of
    Nothing     -> [("empty", Nothing)]
    Just sample -> [("empty", Nothing), ("non-empty", Just sample)]
```

fizruk:
Well, the similar instance can be constructed for `Either`:

```haskell
instance (ToSample a a', ToSample b b') => ToSample (Either a b) (Either a' b') where
  toSamples _ = lefts ++ right
    where
      lefts  = second Left <$> toSamples (Proxy :: Proxy a)
      rights = second Right <$> toSamples (Proxy :: Proxy b)
```

I understand there is no preferred way to define `toSample`, but we can always override these instances with `newtype`. I have `newtype`s and custom `data`s all over my code, so I think I would actually prefer above instances to be _derived_ (with TH or Generics) so that I would be able to derive things for my `newtype`s and custom response/request data types.

I perhaps can provide a corresponding PR sometime if this makes sense to you.

fisx:
Sorry for the lack of generality, i think you are right there.

I don't really have a say in this, but I would like to see these changes.  Only perhaps put rights before lefts in the either instance so that `toSample` gives you the first right, not the first left.

fizruk:
@fisx most of my suggestions have been already merged in #228.

As for the lack of generality — actually with functional dependency on it is not possible to make instances more general as I have suggested. The problem is that GHC is unable to infer `(a, b) -> (a', b')` dependency from `a -> a'` and `b -> b'`.

In practice we only need two parameters internally (see #211), so maybe we could just leave one parameter sometime in the future.

PR #228, opened 2015-09-18
Generic ToSample

### LINKED ISSUE #151: Generic ToSample
We could (and should) provide a generic derivation of `ToSample`. @kosmikus' generic-sop shows how.

fizruk:
Closes #151.

We can add more instances for basic types, like `Int` or `Text`, but I personally think it should be up to library users to choose good samples for those.

Note that this implementation generates infinite lists for recursive datatypes (e.g. `[]`).
Perhaps we don't want that.


fizruk:
Looks like `default toSamples` implementation breaks code, where `toSample` only is defined:

```haskell
instance ToSample T3.ClientInfo T3.ClientInfo where
  toSample _ = Just ci
```

With Generics in action I'd rather remove `toSample` method from `ToSample` class and provide helpers to define `toSamples`:

```haskell
-- | Single sample without description.
singleSample :: a -> [(Text, a)]
singleSample x = [("", x)]

-- | No samples.
noSamples :: [(Text, a)]
noSamples = []

-- | This is now just a function, not a class method.
toSample :: ToSample a b => Proxy a -> [(Text, b)]
toSample proxy = snd <$> listToMaybe (toSamples proxy)
```

Usage example:

```haskell
instance ToSample T3.ClientInfo T3.ClientInfo where
  toSamples _ = singleSample ci
```

fizruk:
It seems that `toSample` is used for request samples, while `toSamples` — for response samples.
Didn't notice that before as documentation does not seem to mention it.

Nonetheless, if we need different request/response samples, I think we should then have to separate `ToSample` classes. So I suggest we move `toSample` method out of `ToSample` class.

fizruk:
I have moved `toSample` method out of the class in 98af812491e0972e3ecb40adb1849199cfe1d197 and am ready to bring changes here.

fizruk:
Just to be complete, an alternative fix is to remove default signature from `toSamples` and export `defaultSamples`. This should not break any of the existing code, but requires extra typing for each default instance:

```haskell
instance ... => ToSample ... where toSamples = defaultSamples
```

This alternative change is available in 747a4726a32aca0ec6b4f9afe3e59ddc51f6929a and is also ready to be here.

alpmestan:
I have to say I quite like the `Generic` version. Kills some more boilerplate, and if someone doesn't want it, all that person has to do is actually provide a definition in the `ToSample` instance. @jkarni ?

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -920,155 +962,22 @@ instance HasDocs sublayout => HasDocs (Vault :> sublayout) where
...
-                 ")"
-                , (va, vb, vc, vd, ve, vf, vg))
+instance (ToSample a a, ToSample b b) => ToSample (a, b) (a, b)
+instance (ToSample a a, ToSample b b, ToSample c c) => ToSample (a, b, c) (a, b, c)
+instance (ToSample a a, ToSample b b, ToSample c c, ToSample d d) => ToSample (a, b, c, d) (a, b, c, d)
+instance (ToSample a a, ToSample b b, ToSample c c, ToSample d d, ToSample e e) => ToSample (a, b, c, d, e) (a, b, c, d, e)
+instance (ToSample a a, ToSample b b, ToSample c c, ToSample d d, ToSample e e, ToSample f f) => ToSample (a, b, c, d, e, f) (a, b, c, d, e, f)
+instance (ToSample a a, ToSample b b, ToSample c c, ToSample d d, ToSample e e, ToSample f f, ToSample g g) => ToSample (a, b, c, d, e, f, g) (a, b, c, d, e, f, g)
+
+instance ToSample a a => ToSample (Maybe a) (Maybe a)
```
jkarni: instances for `Const a b` (given `a`); maybe for all the monoid newtypes (`All`, `Any`, `Sum`, `Product`, `First`, `Last`, `Dual`)?

jkarni: And `ZipList`.

jkarni:
Could you add some tests?
The other question is whether we want to actually add instances for other base types (like `Char` and `Int`). Big downsides, of course (forcing newtypes everywhere), but it would make the generic machinery much more useful.

jkarni:
:+1: on removing `toSample`. We really should unify the treatment of requests and responses.

jkarni:
Perfect! Would you mind adding your change to the changelog?

fizruk:
With this PR some derived instances result in infinite list of samples. While it is okay for requests (since we only display one sample now), it is definitely bad for responses. I suggest we add something like

```haskell
data DocOptions = DocOptions
  { docOptionsMaxSamples :: Int
  , -- probably more options
  }

defaultDocOptions :: DocOptions
defaultDocOptions = DocOptions
  { docOptionsMaxSamples = 5
  , -- other default options
  }
```

We can then pass this to `docsWith` to tell it how many samples we want to get in case of infinitely many of them.

jkarni:
:+1:

jkarni:
Thanks again!

PR #232, opened 2015-09-21
Common code generation types and helpers extracted to servant-foreign

dredozubov:
For the step one of this task i targeted only `servant-js`, so the current refactoring was pretty mechanical. What i wanted to achieve is:
* extract `Servant.JS.Internal` to make unification with other libraries(such as [lackey](https://hackage.haskell.org/package/lackey)) easier
* `HasJS` and `JS` got renamed to `HasForeign` and `Foreign`, but basically stayed the same
* `servant-js` should depend only on `servant-foreign`. There are currently an example which depends on `servant` though.

What do you think? How do you think we should proceed? If it's ok with you, we can have this by `0.5` release and push `servant-foreign` to hackage by that time. I'm not really familiar with servant release cycle, so please correct me if i'm wrong with my assumptions somewhere.

[Review on Reviewable](https://reviewable.io/reviews/haskell-servant/servant/232)

### DIFF
--- a/servant-foreign/CHANGELOG.md
+++ b/servant-foreign/CHANGELOG.md
@@ -1,4 +1,4 @@
-0.1
+0.5
 -----
 
 * Extract javascript-oblivious types and helpers to *servant-foreign*
--- a/servant-foreign/CHANGELOG.md
+++ b/servant-foreign/CHANGELOG.md
@@ -1,0 +1,4 @@
+0.1
+-----
+
+* Extract javascript-oblivious types and helpers to *servant-foreign*
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -1,5 +1,5 @@
 name:                servant-foreign
-version:             0.1.0.0
+version:             0.5.0.0
 synopsis:            Abstract all necessary info for client code-generation in arbitrary programming language.
 description:         Consolidates such projects as servant-js and lackey.
 license:             BSD3
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -20,10 +20,8 @@ source-repository head
 library
   exposed-modules:     Servant.Foreign
   build-depends:       base == 4.*
-                     , charset
                      , lens >= 4
                      , servant >= 0.5
-                     , text
   hs-source-dirs:      src
   default-language:    Haskell2010
   ghc-options:         -Wall
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -19,7 +19,7 @@ source-repository head
 
 library
   exposed-modules:     Servant.Foreign
-  build-depends:       base >=4.8 && <5
+  build-depends:       base == 4.*
                      , charset
                      , lens >= 4
                      , servant >= 0.5
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -1,7 +1,14 @@
 name:                servant-foreign
-version:             0.5.0.0
-synopsis:            Abstract all necessary info for client code-generation in arbitrary programming language.
-description:         Consolidates such projects as servant-js and lackey.
+version:             0.5
+synopsis:            Helper types and functions for generating client functions for servant APIs in any programming language
+description:
+  Helper types and functions for generating client functions for servant APIs in any programming language
+  .
+  This package provides types and functions that collect all the data needed to generate client functions in the programming language of your choice. This effectively means you only have to write the code that "pretty-prints" this data as some code in your target language.
+  .
+  See the servant-js module for an example
+  .
+  <https://github.com/haskell-servant/servant/blob/master/servant-foreign/CHANGELOG.md CHANGELOG>
 license:             BSD3
 license-file:        LICENSE
 author:              Denis Redozubov
@@ -19,8 +26,8 @@ source-repository head
 
 library
   exposed-modules:     Servant.Foreign
-  build-depends:       base == 4.*
-                     , lens == 4.*
+  build-depends:       base    == 4.*
+                     , lens    == 4.*
                      , servant == 0.5.*
   hs-source-dirs:      src
   default-language:    Haskell2010
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -1,0 +1,29 @@
+name:                servant-foreign
+version:             0.1.0.0
+synopsis:            Abstract all necessary info for client code-generation in arbitrary programming language.
+description:         Consolidates such projects as servant-js and lackey.
+license:             BSD3
+license-file:        LICENSE
+author:              Denis Redozubov
+maintainer:          denis.redozubov@gmail.com
+copyright:           2015 Denis Redozubov, Alp Mestanogullari
+category:            Web
+build-type:          Simple
+cabal-version:       >=1.10
+extra-source-files:
+  CHANGELOG.md
+  README.md
+source-repository head
+  type: git
+  location: http://github.com/haskell-servant/servant.git
+
+library
+  exposed-modules:     Servant.Foreign
+  build-depends:       base >=4.8 && <5
+                     , charset
+                     , lens >= 4
+                     , servant >= 0.5
+                     , text
+  hs-source-dirs:      src
+  default-language:    Haskell2010
+  ghc-options:         -Wall
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -20,8 +20,8 @@ source-repository head
 library
   exposed-modules:     Servant.Foreign
   build-depends:       base == 4.*
-                     , lens >= 4
-                     , servant >= 0.5
+                     , lens == 4.*
+                     , servant == 0.5.*
   hs-source-dirs:      src
   default-language:    Haskell2010
   ghc-options:         -Wall
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -1,12 +1,12 @@
 name:                servant-foreign
 version:             0.5
-synopsis:            Helper types and functions for generating client functions for servant APIs in any programming language
+synopsis:            Helpers for generating clients for servant APIs in any programming language
 description:
   Helper types and functions for generating client functions for servant APIs in any programming language
   .
   This package provides types and functions that collect all the data needed to generate client functions in the programming language of your choice. This effectively means you only have to write the code that "pretty-prints" this data as some code in your target language.
   .
-  See the servant-js module for an example
+  See the servant-js package for an example
   .
   <https://github.com/haskell-servant/servant/blob/master/servant-foreign/CHANGELOG.md CHANGELOG>
 license:             BSD3
--- a/servant-foreign/src/Servant/Foreign.hs
+++ b/servant-foreign/src/Servant/Foreign.hs
@@ -1,13 +1,14 @@
-{-# LANGUAGE CPP                  #-}
-{-# LANGUAGE ConstraintKinds      #-}
-{-# LANGUAGE DataKinds            #-}
-{-# LANGUAGE FlexibleContexts     #-}
-{-# LANGUAGE FlexibleInstances    #-}
-{-# LANGUAGE ScopedTypeVariables  #-}
-{-# LANGUAGE TemplateHaskell      #-}
-{-# LANGUAGE TypeFamilies         #-}
-{-# LANGUAGE TypeOperators        #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE CPP                   #-}
+{-# LANGUAGE ConstraintKinds       #-}
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
 
 -- | Generalizes all the data needed to make code generation work with
 -- arbitrary programming languages.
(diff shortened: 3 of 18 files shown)


servant-foreign/src/Servant/Foreign.hs
```diff
@@ -0,0 +1,409 @@
...
+  | Flag
+  | List
+  deriving (Eq, Show)
+
+data QueryArg = QueryArg
+  { _argName :: Arg
+  , _argType :: ArgType
+  } deriving (Eq, Show)
+
+data HeaderArg = HeaderArg
```
jkarni: While we're at it, could we remove records for sum types?

jkarni:
https://github.com/swagger-api/swagger-codegen might provide good inspiration. I'm going to take a deeper look at it and report back.

servant-foreign/src/Servant/Foreign.hs
```diff
@@ -0,0 +1,409 @@
...
+-- | Attempts to reduce the function name provided to that allowed by @'Foreign'@.
+--
+-- Here we are making an assumption that js identifiers are common enough.
+-- https://mathiasbynens.be/notes/javascript-identifiers
+-- Couldn't work out how to handle zero-width characters.
+-- TODO: compare it with other generated languages(such as ruby via lackey)
+-- and generalize.
+-- 
+-- @TODO: specify better default function name, or throw error?
+toValidFunctionName :: String -> String
```
jkarni: This will probably need to be provided by the specific language library, no?

servant-foreign/src/Servant/Foreign.hs
```diff
@@ -0,0 +1,409 @@
...
+
+captureArg :: Segment -> Arg
+captureArg (Segment (Cap s) _) = s
+captureArg                  _  = error "captureArg called on non capture"
+
+defReq :: Req
+defReq = Req defUrl "GET" [] False []
+
+type family Elem (a :: *) (ls::[*]) :: Constraint where
+  Elem a '[] = 'False ~ 'True
```
jkarni: We should probably just remove this `False ~ True` line. It just leads to bad error messages.

servant-foreign/src/Servant/Foreign.hs
```diff
@@ -0,0 +1,409 @@
...
+  , _queryStr :: [QueryArg]
+  } deriving (Eq, Show)
+
+defUrl :: Url
+defUrl = Url [] []
+
+type FunctionName = [String]
+type Method = String
+
+data Req = Req
```
jkarni: We might want to add a `_reqDoc` field (for docstrings).

jkarni: And a (list of?) content-type.

jkarni:
This is a great idea. It'd be interesting to have another language besides JS to test our abstractions against. I wanted to try a template-based approach, and would be happy to provide some python templates. On the other hand maybe the true test is codegen for a typed language.
@tfausak might have interesting suggestions.

servant-foreign/src/Servant/Foreign.hs
tfausak: I agree. This line has already tripped me up.

tfausak:
I don't have any suggestions right now, but I'm glad that @dredozubov factored out the common code here! I was a little frustrated that I had to roll my own representation of the API (my `HasCode` and `HasRuby` classes) when I knew that `servant-js` already did that. This is definitely a step in the right direction, and I would happily change `lackey` over to use `servant-foreign`.

servant-foreign/src/Servant/Foreign.hs
dredozubov: Yes, let's remove it, completely missed it while factoring the code.

servant-foreign/src/Servant/Foreign.hs
dredozubov: It's used by `Show` instance of `HeaderArg`, so i've decided to leave it as it is. The name can be a little misguiding, so we might want to change that.

alpmestan:
@dredozubov That was quick! I'll take a close look tomorrow, but this is exciting. Btw, I suggested `servant-foreign` in another issue as a name for this but if anyone has got a name that would be clearer for stating the purpose of the package, feel free to suggest it!

servant-foreign/src/Servant/Foreign.hs
dredozubov: Multirecords are particularly terrible with standard haskell record accessors, but we use `lens`! It provides prisms, enabling us to do stuff like `headerArg ^? headerPattern`, returning `Maybe String` or getters like `headerArg ^. headerArgName`, unifying field getters and making code slightly more laconic. So... we sure can, but i don't really see the reason to.

dredozubov:
@alpmestan i had options like `servant-client-common` and `servant-generator-common` in mind and i found your version a lot better!

tfausak:
:+1: for `servant-foreign`

kosmikus:
FWIW, I'd actually prefer something like `servant-client-common`, `servant-client-base` or perhaps `servant-foreign-client` over simply `servant-foreign`.

alpmestan:
Had a quick skim over the changes, looks good so far! Just two comments:

- Would you mind raising the version to 0.5? So that it's in sync with all the other packages and it's easy for everyone to figure version bounds for servant packages: same major version for all.
- Shouldn't the generator options (even the "common" variant) stay in *servant-js*? While some similar functionality will probably be needed by any package that generates client functions in another language, some stuffs are also JS specific and won't carry over to other languages smoothly.

servant-foreign/src/Servant/Foreign.hs
jkarni: Should we not use `show` internally then, and leave the derived `Show` instance?

dredozubov: This seems like a cleaner solution with a more ugly `Show` instance, i'm in favor of this.

servant-foreign/src/Servant/Foreign.hs
alpmestan: +1! While in JS land we don't do anything fancy content-type wise, some other languages might want more control in that area.

jkarni: We might also need a type representation (`Rep`?) if we want to clients in typed languages.

servant-js/servant-js.cabal
```diff
@@ -76,12 +74,12 @@ test-suite spec
...
-    , hspec >= 2.1.8
-    , hspec-expectations
-    , language-ecmascript >= 0.16
-  default-language: Haskell2010
+  build-depends:     base == 4.*
+                   , hspec >= 2.1.8
+                   , hspec-expectations
+                   , language-ecmascript >= 0.16
+                   , lens
+                   , lens >= 4
```
alpmestan: `lens` appears twice there, just a typo I guess =)

servant-foreign/src/Servant/Foreign.hs
alpmestan: How would one use `Rep` or any kind of description of a haskell type for/from another language? I think the most reasonable approach is to just let the target language's code make sure the data is in the right shape, with whatever tools/constructs they have to help them. I might be wrong though, woud be glad to be proven (=P) wrong on this one.

servant-foreign/src/Servant/Foreign.hs
dredozubov: In the end i moved the code from this instance to the `toJSHeader` function in `servant-js`. Makes more sense this way.

servant-foreign/src/Servant/Foreign.hs
dredozubov: I'm not sure about the docstrings. Are you implying that we can generate the comment in auto-generated code this way?

servant-js/servant-js.cabal
dredozubov: good catch!

servant-foreign/src/Servant/Foreign.hs
jkarni: Yes, but as you mentioned on IRC it's best if we worry about that (and `Rep`) later.

codedmart:
FWIW, I like @kosmikus suggestion of `servant-client-common`.

dredozubov:
Should we vote on the name? :) I'm okay with either `servant-client-common` or `servant-foreign`.

alpmestan:
I'm not a big fan of `servant-client-common`, because it might convey a non-existant relationship between `servant-client` and this package. The only thing they have in common is that it's for getting client functions. Also I usually don't find `-common` to be a good thing to put in package names, as it doesn't say anything about what's in there.

codedmart:
@alpmestan fair enough I just don't think `servant-foreign` doesn't convey well, but this is just my opinion.

alpmestan:
I agree with you =)

servant-js/src/Servant/JS/Internal.hs
```diff
@@ -152,55 +102,29 @@ toValidFunctionName (x:xs) = [setFirstChar x] <> filter remainder xs
...
-hasArgs :: Segment -> Bool
-hasArgs s = isCapture s || hasMatrixArgs s
-
-matrixArgs :: Segment -> [MatrixArg]
-matrixArgs (Segment _ ms) = ms
-
-captureArg :: Segment -> Arg
-captureArg (Segment (Cap s) _) = s
-captureArg                  _  = error "captureArg called on non capture"
+toJSHeader :: HeaderArg -> String
```
freezeboy: I like this abstraction but I don't really get where the ReplaceHeaderArg come from ?

freezeboy:
I really like this patch, I had this in mind, but not enough time over past weeks, glad to see someone found the time !

jkarni:
No consensus, so as the author of the PR, let's let @dredozubov decide.

dredozubov:
Let's keep `servant-foreign` then, i find `servant-client-common` a little bit too similar to `servant-client` and totally unrelated at the same time.
@jkarni @alpmestan I fixed a bunch of issues you've mentioned. Anything else i should do in that particular PR(let's keep this PR light on refactoring existing code!)?

servant-foreign/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+0.1
+-----
```
alpmestan: `0.5` =)

servant-foreign/servant-foreign.cabal
```diff
@@ -0,0 +1,27 @@
+name:                servant-foreign
+version:             0.5.0.0
```
alpmestan: I usually prefer `0.5` and then only use the next two numbers when doing minor updates to the library. But that's just my opinion.

servant-foreign/servant-foreign.cabal
```diff
@@ -0,0 +1,27 @@
...
+synopsis:            Abstract all necessary info for client code-generation in arbitrary programming language.
```
alpmestan: How about:

> Helper types and functions for generating client functions for servant APIs in any programming language

servant-foreign/servant-foreign.cabal
```diff
@@ -0,0 +1,27 @@
...
+description:         Consolidates such projects as servant-js and lackey.
```
alpmestan: How about:

> Helper types and functions for generating client functions for servant APIs in any programming language
> .
> This package provides types and functions that collect all the data needed to generate client functions in the programming language of your choice. This effectively means you only have to write the code that "pretty-prints" this data as some code in your target language.
> .
> See the *servant-js* module for an example.

?

servant-js/servant-js.cabal
```diff
@@ -37,13 +37,13 @@ library
...
                        Servant.JS.Axios
+                       Servant.JS.Internal
                        Servant.JS.JQuery
                        Servant.JS.Vanilla
-                       Servant.JS.Internal
   build-depends:       base >=4.5 && <5
                      , charset
                      , lens >= 4
-                     , servant == 0.5.*
+                     , servant-foreign >= 0.1
```
alpmestan: `>= 0.5`

alpmestan:
I made a couple of comments again. Aside from that looks good! Except that we will have to document `servant-foreign` at some point, maybe even consider adding a new page to the tutorial, explaining how folks can write their own `servant-<some language>` packages. But that's a separate issue, I guess =)

servant-js/servant-js.cabal
jkarni: `== 0.5.*` please?

servant-js/servant-js.cabal
```diff
@@ -59,16 +59,16 @@ executable counter
...
-    , warp
+  build-depends:    aeson
+                  , base
+                  , filepath
+                  , lens >= 4
+                  , servant
+                  , servant-server
+                  , servant-js
+                  , stm
+                  , transformers
```
jkarni: We've taken to upper and lower bounds. Lower bounds because that's the only honest thing to do. Upper bounds because Snoyman et al. notify us when something outside of them is release, making it very easy to maintain them.

dredozubov: got it

servant-foreign/servant-foreign.cabal
jkarni: Our `bump-version` script forces everything to be consistent, even to extra 0s. As does our upload script, I believe. So yeah, 0.5 is better, but it'll be fixed sooner or later.

servant-js/servant-js.cabal
dredozubov: Oops!

servant-foreign/CHANGELOG.md
dredozubov: changed it for consistency

servant-foreign/servant-foreign.cabal
dredozubov: changed it for consistency

servant-foreign/servant-foreign.cabal
dredozubov: :+1 like it, updated. (Description too)

dredozubov:
@alpmestan @jkarni updated, thanks for your suggestions and quick feedback :heart:

servant-js/servant-js.cabal
alpmestan: oh yeah @jkarni's right, `== 0.5.*` =)

alpmestan:
I made one last minor comment about the cabal description. Aside from this, I'm happy with this PR. Thanks for taking the time to do this, I know it's not the most fascinating thing but it'll significantly help folks write client function generation packages.

dredozubov:
@alpmestan I had to fix one more thing anyway - synopsis needed shortening to make `cabal check` happy, so we can avoid any issues with hackage later.

alpmestan:
Right, good idea. LGTM.

@jkarni Any other comment? Feel free to merge if not.

jkarni:
Thanks! This was a great idea.

PR #233, opened 2015-09-22
Fixed javascript for tutorial/T9

sguy:


alpmestan:
Are those names the one generated with *servant-js* currently? If yes I'll merge this PR. Since we've done quite a few changes in *servant-js* with @freezeboy, this subtle problem might have gone unnoticed indeed.

sguy:
Yes.  I was working through the tutorial last night and I had to make this
case change to get everything to work.

alpmestan:
Merged, thanks!

freezeboy:
You simply mean the api.js was not up to date ?
It's possible, as it is generated by the runtime I may have forgotten to commit it.
other .js files may have the same problem (and I won't have access to my ssh key until next week, so feel free to add them).

alpmestan:
No it's just that there's been a subtle change in the function names as we generate them by default, so the Javascript code that uses it in the tutorial had to be updated.

PR #237, opened 2015-09-24
Test Hackage breakage

sol:


sol:
This is mainly to experiment with a new [`zalora/tinc`](https://github.com/zalora/tinc feature).

Don't merge yet!

jkarni:
neat!

dredozubov:
@sol can you elaborate a bit more? Is it supposed to cache shared(between packages) extra-dependencies with stack or am i misunderstanding this?

sol:
@dredozubov Kind of, but it does not use `stack`.  It uses [`zalora/tinc`](https://github.com/zalora/tinc) for dependency caching and also tests Hackage breakage.

sol:
@jkarni From my perspective this is good for production now.  Currently I test against the latest versions on Hackage with GHC 7.8.4 and 7.10.2.  It's even faster than our current `stack`-based setup.

In addition we could test against Stackage.  If you people are interested in using `tinc` on travis I'm happy to add this.

jkarni:
Will this use the latest versions of packages even if the cache already has an acceptable one?

sol:
Yes.

You may want to have a look at the tinc READ ME.

jkarni:
Perfect! This is exactly what we need!

sol:
You may want to hit the rebuild button on Travis to populate the cache. The slowest build wins, so dependencies for 7.8.4 should already be cached. With the next build 7.10.2 will take longer and "win" when uploading the cache.

BTW, I had to disable -Werror due to aeson-0.10.* issues...

PR #238, opened 2015-09-24
Simplify ToSample (remove redundant second parameter)

### LINKED ISSUE #211: ToSample grievances

fizruk:
Closes #211.

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -403,13 +403,13 @@ class HasDocs layout where
 -- 'toSample': it lets you specify different responses along with
 -- some context (as 'Text') that explains when you're supposed to
 -- get the corresponding response.
-class ToSample a b | a -> b where
-  toSamples :: Proxy a -> [(Text, b)]
-  default toSamples :: (Generic a, Generic b, GToSample (Rep a) (Rep b)) => Proxy a -> [(Text, b)]
+class ToSample a where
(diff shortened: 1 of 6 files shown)


servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -403,13 +403,13 @@ class HasDocs layout where
 -- 'toSample': it lets you specify different responses along with
 -- some context (as 'Text') that explains when you're supposed to
 -- get the corresponding response.
-class ToSample a b | a -> b where
-  toSamples :: Proxy a -> [(Text, b)]
-  default toSamples :: (Generic a, Generic b, GToSample (Rep a) (Rep b)) => Proxy a -> [(Text, b)]
+class ToSample a where
```
jkarni: ln 397 above (docs) also needs to be updated.

jkarni: as does `src/Servant/Docs.hs:81`.

jkarni:
Thanks! I still can't believe how silly this was before, and I'm glad you caught it!
There are just a couple of docs that still need updating.

PR #242, opened 2015-09-28
Pass in Manager as argument to 'client'

### LINKED ISSUE #176: Client functions with Manager param
I'd like to open a discussion for allowing custom connection manager for servant-client. Currently there is an [internal global manager](https://github.com/haskell-servant/servant/blob/master/servant-client/src/Servant/Common/Req.hs#L133) established via unsafePerformIO which is used for everything. My guess it's for convenience. I'd like to propose at least allowing to specify Manager for `client` function, but would be even better to be able to specify it per function as #174 asks for BaseUrl.

jkarni:
Fixed #176 .

### DIFF
--- a/servant-client/CHANGELOG.md
+++ b/servant-client/CHANGELOG.md
@@ -3,6 +3,7 @@ HEAD
 
 * Support for the `HttpVersion`, `IsSecure`, `RemoteHost` and `Vault` combinators
 * Added support for `path` on `BaseUrl`.
+* `client` now takes an explicit `Manager` argument.
 
 0.4.1
 -----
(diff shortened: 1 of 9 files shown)


jkarni:
Don't have much patience to pattern match twice on client functions, so I went with `unsaferPerformIO` in the examples. Of dubious pedagogical value, but no different than what we were doing before.

alpmestan:
Well, at least it will now be obvious that it is the "easy but dirty" way, and give the flexibility to do things in a cleaner, more configurable way. This does give us client-function debugging as well right by tweaking the `Manager` (to print the request whenever it sends one), right?

alpmestan:
The tutorial will need to be updated for this, btw.

jkarni:
I'll keep a list of things I know have to be updated for the tutorial (doing so now would be confusing).

zudov:
:100

PR #243, opened 2015-10-02
Change servant-js and servant-foreign to use text

arianvp:
Changes servant-js and servant-foreign to use Text instead of String


arianvp:
I'd like to know how others feel about this change.  The pragmatic me feels that String should be avoided at all cost and that Text is a lot cleaner. But wondering what other people think of it.

servant-js/test/Servant/JSSpec.hs
```diff
@@ -81,12 +85,16 @@ spec = describe "Servant.JQuery" $ do
     axiosSpec
     --angularSpec    AngularCustom
 
+
```
jkarni: Maybe signatures here?

servant-js/src/Servant/JS/Internal.hs
```diff
@@ -87,7 +88,34 @@ defCommonGeneratorOptions = CommonGeneratorOptions
 -- Couldn't work out how to handle zero-width characters.
 --
 -- @TODO: specify better default function name, or throw error?
-toValidFunctionName :: String -> String
+toValidFunctionName :: Text -> Text
+-- @TODO: Cons text
```
jkarni: What's this TODO about?

servant-js/src/Servant/JS/Internal.hs
```diff
@@ -87,7 +88,34 @@ defCommonGeneratorOptions = CommonGeneratorOptions
...
+                    , Set.titlecaseLetter
+                    , Set.modifierLetter
+                    , Set.otherLetter
+                    , Set.letterNumber ]
+    remainderOK   = firstLetterOK
+               <> [ Set.nonSpacingMark
+                  , Set.spacingCombiningMark
+                  , Set.decimalNumber
+                  , Set.connectorPunctuation ]
+{-
```
jkarni: We can remove the comment, no?

servant-js/test/Servant/JSSpec.hs
arianvp: Yes will add them.   They're not exported though as they're just internal helper functions.  But I guess it's prettier to add a signature to them

servant-js/src/Servant/JS/Internal.hs
arianvp: The `Cons text` TODO should be removed. it concerns a piece of code that doesn't exist anymore.

The other `TODO` as already there in the original version so I left it untouched.

servant-js/src/Servant/JS.hs
```diff
@@ -142,6 +145,8 @@ writeJSForAPI :: (HasForeign api, GenerateList (Foreign api))
               -> JavaScriptGenerator -- ^ js code generator to use (angular, vanilla js, jquery, others)
               -> FilePath -- ^ path to the file you want to write the resulting javascript code into
               -> IO ()
+
+-- TODO Data.Text
```
jkarni: And this one?

servant-examples/digest-auth/digest-auth.hs
```diff
@@ -0,0 +1,33 @@
+{-#LANGUAGE DataKinds, TypeOperators, OverloadedStrings #-}
+module Main where
```
arianvp: I accidentally added this file in one of the commits.   I'll remove it again as it's not suppoted to be in this pull.

servant-js/src/Servant/JS/Internal.hs
```diff
@@ -87,7 +88,34 @@ defCommonGeneratorOptions = CommonGeneratorOptions
...
+      setFirstChar x `T.cons` T.filter remainder xs 
+    Nothing -> "_"
+  where
+    setFirstChar c = if firstChar c then c else '_'
+    firstChar c = prefixOK c || any (Set.member c) firstLetterOK
+    remainder c = prefixOK c || any (Set.member c) remainderOK
+    -- Valid prefixes
+    prefixOK c = c `elem` ['$','_']
+    -- Unicode character sets
+    firstLetterOK = [ Set.lowercaseLetter
```
jkarni: (I know this wasn't introduced in your change but) I find it a little more intuitive to do a union of sets here and below (and probably more efficient), and then get rid of the `any`s above

jkarni:
LGTM

dredozubov:
+1 on this

jkarni:
@arianvp could you add a line to the changelog?

PR #244, opened 2015-10-04
Remove redundant build from .travis.yml file.

arianvp:
The `build` in the `.travis.yml` file is redundant. `stack test` is just
an alias for `stack build --test`.

Also in the old way of doing things,  `-Werror` is actually not set when
running tests. So it is ignored for  test files..

Now also test files need to have no warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,7 @@ install:
 
 script:
   - stack setup
-  - stack build --ghc-options "-Werror"
-  - stack test
+  - stack build --test --ghc-options "-Werror"
 
 cache:
   directories:


alpmestan:
LGTM.

soenkehahn:
As far as I know `stack test` *is* an alias for `stack build --test`, but it only builds the test suite, not the libraries or executables.

soenkehahn:
Both `stack build` and `stack test` offer the flag `--pedantic` which passes `-Wall -Werror` to `ghc`.

arianvp:
I'm pretty sure   `stack build --test`  also builds the libraries and executables.

[From the documentation](https://github.com/commercialhaskell/stack/blob/master/doc/build_command.md#target-syntax)
> This is where the --test and --bench flags come into play. If the package is a local package, then all of the test suite and benchmark components are selected to be built, respectively. _In any event, the library and executable components are also selected to be built._

soenkehahn:
Yes, `stack-0.1.4.0` (what we're using) seems to be doing what you say, sorry for the noise. LGTM.

PR #245, opened 2015-10-04
Fix servant-client tests to properly use hspec-discover.

arianvp:
For some reason hspec-discover wasn't enabled for servant-client,
causing test cases to be skipped and fasely reporting servant-client as
test-passing whilst it was actually supposed to fail the tests.

For example, we redefined BaseUrl recently to have a path field, but the
tests don't reflect that yet but passed anyway. Because the test case
wasn't even discovered and run.

### DIFF
--- a/servant-client/src/Servant/Common/BaseUrl.hs
+++ b/servant-client/src/Servant/Common/BaseUrl.hs
@@ -32,12 +32,19 @@ data BaseUrl = BaseUrl
   , baseUrlHost   :: String   -- ^ host (eg "haskell.org")
   , baseUrlPort   :: Int      -- ^ port (eg 80)
   , baseUrlPath   :: String   -- ^ path (eg "/a/b/c")
-  } deriving (Show, Eq, Ord, Generic)
+  } deriving (Show, Ord, Generic)
+
+instance Eq BaseUrl where
+    BaseUrl a b c path == BaseUrl a' b' c' path'
+        = a == a' && b == b' && c == c' && s path == s path'
+        where s ('/':x) = x
+              s x       = x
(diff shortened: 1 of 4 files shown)


arianvp:
Please do not merge yet untill I also fixed the  actual failing tests.

arianvp:
https://github.com/arianvp/servant/commit/717b18df4e759c30765bdf56f709b0050c4220d4 seems to have changed   `Either` to `MonadThrow m`.  Breaking the test suite

alpmestan:
I would too but some folks didn't, and `Either` still works, right?

arianvp:
Yes the current tests should work with `Either`. I could also rewrite the tests to work with `MonadThrow`

jkarni:
I'm (mildly) in favour of keeping `MonadThrow` (and `MonadIO` etc., generally speaking). It's pretty rare (except in our test suite!) that we need annotations, and from my own perspective the more general type is often convenient.

PierreR:
> What was the reason for this change?. Personally I'd prefer Either here.

I believe `MonadThrow` is better for the reasons expressed [here](https://www.fpcomplete.com/user/commercial/content/exceptions-best-practices).

This is also consistent with what `http-client` is doing:
https://hackage.haskell.org/package/http-client-0.4.24/docs/Network-HTTP-Client.html#v:parseUrl

arianvp:
Okay seems reasonable. I'll update the tests accordingly.

arianvp:
@jkarni I've merged the 245 branch into this one to keep everything in one place.

jkarni:
LGTM

PR #246, opened 2015-10-05
added status code 202 to servant-client POST responses

ericnething:
When using the Plivo API to send a text message through SMS, the response has status code 202 "Accepted", which results in servant throwing an error because it does not recognize this as a valid success code. I added status code 202 to Servant.Client for Post responses to fix this problem.


freezeboy:
Why not consider that any code between 200 and 300 is "valid" ? that's what I understand from the description of these codes.

jkarni:
:+1: on accepting all 200s.

alpmestan:
Agreed.

jkarni:
Created https://github.com/haskell-servant/servant/issues/247, merging this.

PR #248, opened 2015-10-07
178: Raw & Enter

jkarni:
#178 with minor fixes.

### DIFF
--- a/servant-js/examples/counter.hs
+++ b/servant-js/examples/counter.hs
@@ -43,7 +43,7 @@ type TestApi = "counter" :> Post '[JSON] Counter -- endpoint for increasing the 
           :<|> "counter" :> Get '[JSON] Counter -- endpoint to get the current value
 
 type TestApi' = TestApi
-           :<|> Raw -- used for serving static files
+          :<|> Raw IO Application -- used for serving static files
 
 -- this proxy only targets the proper endpoints of our API,
 -- not the static file serving bit
--- a/servant-client/CHANGELOG.md
+++ b/servant-client/CHANGELOG.md
@@ -2,6 +2,7 @@ HEAD
 ----
 
 * Use the `text` package instead of `String`.
+* Update `instance HasClient Raw` for servant Raw changes
 * Support for the `HttpVersion`, `IsSecure`, `RemoteHost` and `Vault` combinators
 * Added support for `path` on `BaseUrl`.
 * `client` now takes an explicit `Manager` argument.
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -347,10 +347,10 @@ instance (KnownSymbol sym, HasClient sublayout)
 
 -- | Pick a 'Method' and specify where the server you want to query is. You get
 -- back the full `Response`.
-instance HasClient Raw where
-  type Client Raw = H.Method -> ExceptT ServantError IO (Int, ByteString, MediaType, [HTTP.Header], Response ByteString)
+instance HasClient (Raw m a) where
+  type Client (Raw m a) = H.Method -> ExceptT ServantError IO (Int, ByteString, MediaType, [HTTP.Header], Response ByteString)
 
-  clientWithRoute :: Proxy Raw -> Req -> BaseUrl -> Manager -> Client Raw
+  clientWithRoute :: Proxy (Raw m a) -> Req -> BaseUrl -> Manager -> Client (Raw m a)
   clientWithRoute Proxy req baseurl manager httpMethod = do
     performRequest httpMethod req baseurl manager
 
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -40,7 +40,7 @@ import qualified Network.HTTP.Client        as C
 import           Network.HTTP.Media
 import           Network.HTTP.Types         (Status (..), badRequest400,
                                              methodGet, ok200, status400)
-import           Network.Socket
+import           Network.Socket             hiding (Raw)
 import           Network.Wai                (Application, responseLBS)
 import           Network.Wai.Handler.Warp
 import           System.IO.Unsafe           (unsafePerformIO)
@@ -101,8 +101,8 @@ type Api =
   :<|> "param" :> QueryParam "name" String :> Get '[FormUrlEncoded,JSON] Person
   :<|> "params" :> QueryParams "names" String :> Get '[JSON] [Person]
   :<|> "flag" :> QueryFlag "flag" :> Get '[JSON] Bool
-  :<|> "rawSuccess" :> Raw
-  :<|> "rawFailure" :> Raw
+  :<|> "rawSuccess" :> Raw IO Application
+  :<|> "rawFailure" :> Raw IO Application
   :<|> "multiple" :>
             Capture "first" String :>
             QueryParam "second" Int :>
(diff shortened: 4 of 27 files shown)


codedmart:
@jkarni Do you want me to rebase this against master? Is this good to go other then that?

codedmart:
@jkarni rebased master so this can be merged. Thanks for your help!

servant-js/examples/counter.hs
```diff
@@ -43,7 +43,7 @@ type TestApi = "counter" :> Post '[JSON] Counter -- endpoint for increasing the
           :<|> "counter" :> Get '[JSON] Counter -- endpoint to get the current value
 
 type TestApi' = TestApi
-           :<|> Raw -- used for serving static files
+          :<|> Raw IO Application -- used for serving static files
```
jkarni: Is `IO` the right `m` for `Raw`? Shouldn't it be `ExceptT ServantErr IO`? It won't matter in these cases, since we don't `enter`, but maybe this is confusing. Which brings up to questions:

- Does leaving it polymorphic work?
- Do we want a type synonym for `Raw (ExceptT ServantErr IO)`?

alpmestan:
Have the haddocks for `Raw` been really updated yet? I'd like for there to be something that doesn't leave the reader with the following thought? "Wait... what? How do I pick my `m`? What's the point of using something else than `Application` for `a`?

alpmestan:
Other than that, LGTM.

codedmart:
@jkarni @alpmestan I will try to finish this up in the next few days.

servant-js/examples/counter.hs
codedmart: @jkarni Have you had much more thought on this? @alpmestan Any thoughts or opinions?

codedmart:
@jkarni @alpmestan I rebased master again so this should be up to date assuming I didn't miss anything and travis is happy.

servant-js/examples/counter.hs
alpmestan: I'm sorry but I don't have any good answer. I still think we shouldn't ship this because it's a true hack and I'm sincerely unable to explain what `m` people should pick. :(

alpmestan:
If we are to ship this, I think we really ought to provide a little more guidance on what roles `m` and `a` play here, and on how folks should pick `m` and `a`. Some of this guidance might admittedly belong to the tutorial, but the haddocks would probably leave them very confused in the current state. The three of us probably are the only 3 persons to know what's going on around the new `Raw`.

codedmart:
@alpmestan I agree the docs/tutorial are needed. I was just getting the PR up to date before I jumped onto that.

codedmart:
@jkarni @alpmestan Is the ready to merge now or is there more that needs to be done.

codedmart:
@jkarni any more comments or thoughts on this?

jkarni:
It's unfortunately not `IO`, but `ExceptT ServantErr IO`, for `Raw`...

codedmart:
cant any m be used there? Would you rather add a type synonym?

jkarni:
It can only if `enter` is not actually used on it. 

The issue is that we can't provide the type synonym in `servant` itself, since `ServantErr` is defined in `servant-server`. 

@codedmart sorry this PR keeps running into snags...

codedmart:
@jkarni I understand. I am more pushing now because @imalsogreg needs this for servant-snap. I will be back on Monday(11/30). If you are on IRC maybe we can try to tie it up or send it to it's grave :smile:, deal?

jkarni:
I'm on IRC now

codedmart:
Sorry can't get on now. On my iPad.

imalsogreg:
Hmm, interesting problem.  Seems like there are a couple options:

 1. Change the documentation `Raw IO Application` -> `Raw (EitherT ServantErr IO) Application`
This forces all non-server interpretations to import servant-server to get at the `ServantErr`. 
 2. In examples add type variables to the site APIs in the tutorial. `type MyAPI m = Raw m (m ())`
In this case, docs/javascript could handle the `MyAPI m` by simply ignoring the `Raw m a` part? Decouples the server details from the API. I did this in greet.hs... would want to make sure it actually works in the wild.
 3. Further push away from having `Raw` combinators in APIs that are shared between interpretations. Add `Raw` only to the server-specific API. Does this invalidate the original reason for this pull request?

Thanks for your thoughts on this. If we have to kill the PR it wouldn't be the end of the world for `servant-snap` necessarily. `servant-snap` might just not be able to handle `Raw`s.. which would be somewhat Ok since any action that would have been a `Raw` endpoint originally could instead just be a handler in snap but outside of servant world.

jkarni:
Short update: after some discussion on IRC with @codedmart and @imalsogreg, we found out we could have:

```
type RawM x = Raw (forall m. m) x
```
And then with an extra constraint to the `Raw` `HasServer` instance, users need not worry about picking an `m :: * -> *` themselves - type inference works its magic.

Crucially, however, users would need to use `ImpredicativeTypes`...

jkarni:
Here are some solutions with different trade-offs:

- [Losing the `m` parameter in the API](https://github.com/haskell-servant/servant/compare/jkarni/another178?expand=1)
- [With ImpredicativeTypes](https://github.com/haskell-servant/servant/compare/master...jkarni/178-wip?expand=1)

Both basically do what we need - allow `enter` on `Raw`, and don't require users to be aware of the first type param of `Raw`. The first one won't do what @imalsogreg needs for `servant-snap`, however. The second one requires users to include the `ImpredicativeTypes` pragma.

kosmikus:
I've unfortunately still not managed to catch up with the discussion here, but IMHO `ImpredicateTypes` should be an absolute no-go. Until this language extension is fixed, it is considered to be barely supported and incredibly fragile. Requiring it to use Servant (or at least `Raw`) would not make me feel comfortable.

alpmestan:
Agreed, I would feel better if we decide to stay away from `ImpredicativeTypes`. It's a shame though that the (only?) other solution gets in the way of *servant-snap*. Anyone's got a third idea?

imalsogreg:
@kosmikus The gist of the discussion is that there are two reasons we would like the `Raw` combinator to have a pair of type parameters: `Raw m a`

 1. This allows `Raw` endpoints to be used as part of an API that gets `enter`ed
 1. The `HasServer` instance for `Raw` can only choose an associated type for `ServerT` if the underlying handler monad is either fixed (as in wai) or passed as a type argument a la `Raw m a` (needed for servant-snap).
 1. Documentation or client interpretations might be able to incorporate that added type information in their output.
`
Adding type parameters to the combinator has some drawbacks. The extra parameters need to be introduced and explained in the tutorial. Adding an `m` parameter might encourage users to mention a concrete server monad in their API definition, in which case their API-defining package would incur a dependency on `servant-server`, `wai`, `snap`, or `happstack`. So, we would need to advise users to add a type parameter to their API definition, for passing through to their own uses of Raw to avoid having to pick a concrete server at the time of designing the API.

@jkarni's experiments yesterday lead to a solution for the awkwardness of supplying that `m` variable when most users won't need it - hide it behind that `forall` using ImpredicativeTypes. But it sounds like we may not want to go too far in that direction.

Here are the ways forward I'm aware of:

 1. Keep the `Raw m a` spelling of the raw combinator, encourage users who want to separate their API from a concrete server to say `type MyAPI m = SomeRoutes :<|> Raw m ()`. So, everything in @jkarni's work so far, minus the `type RawX a = Raw (forall m. m) a` part. API's that don't need any raw endpoints don't need to change anything.
 1. Drop this PR, investigate other ways of supporting `Raw` in `servant-snap` (I think it may be possible - maybe by an `m` param to the `HasServer` class in `servant-snap`?), or just tell `servant-snap` users not to use `Raw` - just use `snap`s routing machinery instead. Forbid the use of `Raw` in an `enter`ed API (or find some other way to do it?).

codedmart:
@jkarni Just getting this caught up with master. Still need to add more docs and discuss any other changes or if we want this.

jkarni:
The name of this branch is messing up RTD (RTD thinks it's a version number). Closing so I can delete/rename

PR #250, opened 2015-10-07
Use http-api-data

### LINKED ISSUE #213: Use path-pieces package?
Is there any reason not to use [`path-pieces`](http://hackage.haskell.org/package/path-pieces) package instead of [`Servant.Common.Text`](http://hackage.haskell.org/package/servant-0.4.4.2/docs/Servant-Common-Text.html)?

I am currently using `persistent` with `servant` and have to manually define instances like this for almost every entity I have (the `PathPiece` instances are provided automatically by `mkPersist`):

```haskell
instance FromText (Key User) where
  fromText = fromPathPiece

instance ToText (Key User) where
  toText = toPathPiece
```

Also, we might use `MultiPathPiece` class for a corresponding combinator (to capture the rest of the path).

fizruk:
Closes #213.


servant-server/servant-server.cabal
```diff
@@ -48,6 +48,7 @@ library
       , attoparsec         >= 0.12 && < 0.14
       , bytestring         >= 0.10 && < 0.11
       , containers         >= 0.5  && < 0.6
+      , http-api-data
```
jkarni: Can we put an upper bound? Our general policy so far has been to do that, since we get conveniently notified by stackage maintainers when they're too restrictive.

servant/servant.cabal
```diff
@@ -53,6 +52,7 @@ library
     , bytestring == 0.10.*
     , bytestring-conversion == 0.3.*
     , case-insensitive >= 1.2
+    , http-api-data
```
jkarni: Upper bound here as well (in fact this is the more important one, since other packages depend on this one).

servant-server/servant-server.cabal
fizruk: Do you want upper bounds for `servant-server` only or `servant` and `servant-client` as well?

stack-ghc-7.8.4.yaml
```diff
@@ -21,4 +21,5 @@ extra-deps:
 - socket-io-1.3.3
 - stm-delay-0.1.1.1
 - control-monad-omega-0.3.1
+- http-api-data-0.1.1.1
```
jkarni: This will be on stackage soon?

fizruk: Oh, it's already there :)

fizruk: I am not sure I can remove that line from `stack-ghc-7.8.4.yaml` though.

jkarni:
Thanks!

servant-server/servant-server.cabal
alpmestan: I'd say both, because some folks only use *one* of those two packages.

fizruk: @alpmestan already done in 2a7d110c175bda47a1aa79c29caa37b301d1d1f0.

Issue #113, opened 2015-06-04
Refactor Matrix Params


jkarni:
There are currently a couple of problems with matrix params:

  * We do a lot of work handling URLs that's specific to matrix params, even when the combinator that's doing the work is never `MatrixParam` itself. We've verified that the performance penalty of this is significant
  * Matrix params influence the ability of other combinators to use the semicolon character in URLs. 
  * Relatedly, `Raw` endpoints get a modified `pathInfo` that is almost certainly not what we want
 
EDIT: A couple of more issues -
   * Writing combinators is more complex since you have to check whether the path starts with ';'
   * "a;b" and "a/;b" currently end up being treated the same

Both could be addressed by making `MatrixParam` a combinator function rather than a combinator - e.g. `WithMatrix ("name", String) <comb>` where `<comb>` is either a capture or a path literal. 

I'm going to get started on this, but wanted to hear opinions on the concrete syntax. Some options:

   1. `WithMatrix '[("name", Type), ... ] <comb> `
   2. `WithMatrix '["name" := Type, ...] <comb>`
   3. As in `1` or `2`, but the first argument is not a list, s.t. multiple `WithMatrix` must be used.
   4. Both `WithMatrices` and `WithMatrix` (i.e. 1-2 and 3).

I'm also interested in hearing thoughts about the name. I personally like (kosmikus' idea of) using `With` to indicate that this is not a standalone combinator. But of course we could continue using `MatrixParam` and/or `MatrixParams`.

jkarni:
Additionally, changing `MatrixParam` in these ways is also justified from a user's perspective: query params are semantically attached to the entire URL, whereas matrix params are semantically attached to a particular path segment.

alpmestan:
Agreed that this should be changed, even just for the unnecessary work done on *all requests*. @dlarsson may have some comments on this proposal, being the original author of the matrix params combinators. I personally like either 1/ or 2/ but I haven't used matrix params in a work project yet, so... =)

dlarsson:
This sounds like a good idea. Indeed, matrix params are a attached to a path segment, so there's several benefits with this change. It makes the semantical connection much more explicit, the implementation much more local, and not everyone has to pay the price.

I don't have a strong opinion about the various solutions. The "With" prefix sounds like a good idea though. Since they're somewhat similar to query parameters, I guess one could argue that the syntax to define them should be similar too, e.g. option 3.

jkarni:
[...]

PR #252, opened 2015-10-08
Remove matrix

jkarni:


jkarni:
Still need to go through this more carefully myself, but there it is.

kosmikus:
What's the history here? Why are we doing this? I'm not overly fond of matrix parameters myself, but I haven't seen a discussion or decision to actually remove them.

alpmestan:
https://github.com/haskell-servant/servant/issues/113 ?

alpmestan:
What's not in the issue I linked to has mostly been discussed on IRC.

kosmikus:
@alpmestan Hmm, perhaps I'm misunderstanding. That issue seems to propose a refactoring, whereas the title of this PR looks like it's aiming at removal. Can you clarify?

alpmestan:
Yeah sorry, my answer was a bit too succint.

The idea is that the current matrix param code has got a few problems and incurs a runtime cost on *every single request*. Taking care of that requires some work and nobody but @dlarsson seems to be willing to take care of it, except that Daniel has been quite busy. The idea is to remove `MatrixParam`s until someone can nicely integrate them with everything else without the current problems.

If you're opposed to removing them and/or want to "fix" the current issues, now would be a good time to let your voice be heard =)

kosmikus:
I'm not opposed to removing them.

dlarsson:
I've started working on it, but yes, I'm not able to spend too much time on
it at the moment. Hopefully we can get them back in when I have a fix

2015-10-13 11:39 GMT+02:00 Andres Löh <notifications@github.com>:

jkarni:
@dlarsson Cool - it'd definitely be better to have them. I think the litmus test for correctness here is that implementing them shouldn't require changing the core functions (i.e., it should in theory be possible to have them defined in a separate library).

alpmestan:
LGTM.

dmjio:
:+1:

PR #254, opened 2015-10-12
fmapRouter.

fisx:
this function is a waste product of something we ended up doing differently, but @jkarni and I but like it a lot.  If you want to have it, I can write the test code.  If you think we should wait for use case, feel free to close this pr.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/Router.hs
+++ b/servant-server/src/Servant/Server/Internal/Router.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveFunctor #-}
+
 module Servant.Server.Internal.Router where
 
 import           Data.Map                                   (Map)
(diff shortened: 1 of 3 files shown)


alpmestan:
Yeah, could be useful. Would you mind adding some haddocks for it? And fixing [this error](https://travis-ci.org/haskell-servant/servant/jobs/84935157#L235) spotted by travis. Just an `#ifdef`'d `import Applicative` to add =)

fisx:
I'm making a mess again...  (-:

I'll squash this tomorrow if travis and you guys have no more complaints.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -1,25 +1,40 @@
...
+  | LeafRouter    a
       -- ^ to be used for routes that match an empty path
   | Choice        Router Router
       -- ^ left-biased choice between two routers
+  deriving Functor
+
+-- | Apply a function to the result of a router in functor style.  The result contains the failure
+-- cases so one use case is to turn failures into middleware response values with appropriate status
+-- codes, message bodies, etc.
+tweakResponse :: (RouteResult Response -> RouteResult Response) -> Router -> Router
```
jkarni: We should make sure to export this from a non-internal module.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -1,25 +1,40 @@
...
   | DynamicRouter (Text -> Router)
       -- ^ first path component used for lookup and removed afterwards
-  | LeafRouter    RoutingApplication
+  | LeafRouter    a
       -- ^ to be used for routes that match an empty path
   | Choice        Router Router
       -- ^ left-biased choice between two routers
+  deriving Functor
+
+-- | Apply a function to the result of a router in functor style.  The result contains the failure
```
jkarni: This haddock isn't very clear to me. "Functor style" doesn't seem helpful, and it's not clear to me why *middleware response values* means, or why only failures are mentioned. Why not just "Apply the transformation to the response of a `Router`"?

alpmestan:
LGTM.

kosmikus:
Even if it's too late, LGTM as well :)

PR #256, opened 2015-10-21
"Required" combinator and parse error propagation

### LINKED ISSUE #241: Include "mandatory QueryParam" (QueryParams, QueryFlag as well) in the next release
Over the past few weeks, several users mentionned that they would like a variation of `QueryParam`s that makes them mandatory. That is, instead of handing a `Maybe a` to the handlers (hence giving Nothing when the query string param is absent or can't be decoded), this variation would hand an `a` directly and just error out if the parameter is absent or can't be decoded.

@kosmikus also mentionned it would be nice to differentiate between "not there" and "can't be decoded", even though this one is less of a problem because you can always just get a `Text` to your handler and let it decide what to do when decoding fails. Whereas today the only way to get a "mandatory `QueryParam`" is to write a new combinator for it.

I see two ways to do this:
- Path of least code breakage: introduce new combinators, "mandatory" variants of the existing ones. Regarding the naming convention, we could add a `'` to the name or some other thing.
- Change the combinators to something like:

``` haskell
data ParamType = Optional | Mandatory
data QueryParam (name :: Symbol) a (paramtype :: ParamType)
```

and have the `HasServer`/`Server` instances dispatch on `paramtype` to get the current behavior or the "mandatory" one.

fizruk:
Closes #241:
- introduces `Required` combinator (if parameter is absent — throw 404 Not Found);
- propagates `Header`, `QueryParam` and `QueryParams` parse errors by throwing 400 Bad Request;
- treats query parameter without value as invalid and throws 400 Bad Request.

If this looks ok, I will write some tests/examples.

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE TypeOperators        #-}
 {-# LANGUAGE UndecidableInstances #-}
 #if !MIN_VERSION_base(4,8,0)
+{-# LANGUAGE DeriveDataTypeable   #-}
 {-# LANGUAGE OverlappingInstances #-}
 #endif
 
@@ -22,6 +23,7 @@ module Servant.Server.Internal
 
 #if !MIN_VERSION_base(4,8,0)
 import           Control.Applicative         ((<$>))
+import           Data.Traversable            (traverse)
 #endif
 import           Control.Arrow               (left)
 import           Control.Monad.Trans.Except  (ExceptT)


jkarni:
No way this is going to build without large revisions. Shall I close for now?

Also, I think I prefer the approach we went with in `Verbs`, where e.g. `QueryParam` gets another parameter indicating whether it's required or not.

phadej:
I needed some notion of `Required` myself. And like @jkarni's idea to add a parameter to `QueryParam`, like

```hs
type QueryParam = QueryParam' Required
```

alpmestan:
Yes, this is probably the sanest solution.

acidghost:
What is the status on this PR? Is it going to get merged?

I have the following API
```haskell
type XHookSecHeader = Header "X-Hook-Secret" String
type XHookSigHeader = Header "X-Hook-Signature" String
type ServerAPI =
    "webhooks" :>
        (       XHookSigHeader :> ReqBody '[JSON] Events :> Post '[JSON] String
        :<|>    XHookSecHeader :> Post '[JSON] (Headers '[XHookSecHeader] NoContent)
        )
```
that should reply to requests from [Asana's API](https://asana.com/developers/api-reference/webhooks): the first route expects a `XHookSigHeader` and JSON content, the second route expects the `XHookSecHeader` and no content.
With the current API description unfortunately only the first handler is ever called (even for requests with a `Content-Type: application/json`).

I think that forcing
- the headers to be present or
- the absence of the `Content-Type` header for requests without a `ReqBody`

would help `servant` better distinguish the two cases (and allow me to continue development).

Do you have any other suggestion / workaround to help me with my problem?

phadej:
I think that this PR as is won't be merged, but will work on the solution to the problem after 0.10

acidghost:
@phadej thanks for the quick reply!

Do you have any temporary workaround to suggest for my use case?

phadej:
Needs rebase (and re-review, as I completely forgot about this). Re-tagging for 0.13

phadej:
Superseded by #873

PR #261, opened 2015-11-04
servant-js: export CommonGeneratorOptions constructors

mattjbray:
Allow `defCommonGeneratorOptions` to be customized using record update syntax.

Without the re-export, we can't do things like:

```haskell
import qualified Servant.JS as SJS

myOptions =
  SJS.defCommonGeneratorOptions
    { SJS.moduleName = "myModule"
    , SJS.urlPrefix = "http://localhost"
    }
```


jkarni:
Thanks!

PR #262, opened 2015-11-04
Fix camelCase

mattjbray:
Previous behaviour was a bit shouty (and dashes aren't allowed in JS variable names):

```haskell
camelCase ["one", "two", "thirty-three"] -- => "oneTWOTHIRTY-THREE"
```

New behaviour:

```haskell
camelCase ["one", "two", "thirty-three"] -- => "oneTwoThirtythree"
```


jkarni:
Thanks!

PR #265, opened 2015-11-16
Foreign internal module

fisx:
Hei everybody,

I have run into two issues with servant-foreign:

1. `Elem` is not exported and I wanted to write an `HasForeign` instance for a variant of the `Post` combinator.
2. `FunctionName` is exported abstract, but I want to construct function names outside of `Foreign`.

I can easily work around both these issues for now, but I will probably run into others, so I moved `Foreign` to `Foreign.Internal` with no export list, and kept the restricted export list in a `Foreign` hull module.

what do you think?
cheers/


alpmestan:
LGTM. @jkarni?

fisx:
Is anything blocking this, btw?

fisx:
Hey, I just realized I have a merge button on this repo!  :-]

What is the policy here?  Is one "LGTM" enough for me to merge it alone?

codedmart:
@fisx I have never officially discussed this with @alpmestan or @jkarni, but unless they explicitly tell me to merge I leave that up to them. Just my process.

fisx:
@codedmart: good process.  I'll adopt that.

alpmestan:
Yeah usually we just try to have @jkarni and myself look at each PR. Waiting on Julian for this one =)

jkarni:
On my phone, so can't do due diligence here, but isn't there an Elem from the servant package that  does basically the same thing and that we should be using instead? 

And what do you mean by exported abstract? Isn't it a type synonym?

Other than that, I'm in favour of having the Internal module.

christian-marie:
+1 I think .Internal modules are fine -- if required. Especially for "official" servant packages, where we're going to have to fix breakages anyway.

jkarni:
Ok, I just checked and it doesn't seem like we're exporting `Elem` from `servant`, so this LGTM. I think maybe we should consider not re-exporting `Servant.API`, but that's not a point about this particular PR. Going to merge if no one objects.

fisx:
@jkarni: you are right, `FunctionName` can't have its constructors exported because it's a `type`.  But the PR is still useful for the first reason IMO.

jkarni:
Thanks

PR #269, opened 2015-11-21
Introduce servant purescript

### LINKED ISSUE #268: Adopt servant-purescript
@fisx has just made servant-purescript compile with the latest changes: https://github.com/anchor/servant-purescript/pull/3

Anchor is no longer using purescript, and most of our original development team there has moved on.

This leads to my question: Do we wish to adopt this repository? I believe it's quite similar to (and based off) servant-js.

I don't think it utilizes servant-foreign yet, though. I haven't checked.

poke: @jkarni, @alpmestan, @soenkehahn, @kosmikus

fisx:
Fixes #268

See also: https://github.com/anchor/servant-purescript/pull/3

I'm still working on this, but you may have an opinion on what's there so far.


servant-purescript/HLint.hs
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE PackageImports #-}
+module HLint.HLint where
+
+import           "hint" HLint.Builtin.All
```
jkarni: Can we fold this into the top level `HLint.hs` that we already have?

servant-purescript/src/Servant/PureScript.hs
```diff
@@ -0,0 +1,229 @@
...
+        typeSig :: ST
+        typeSig = T.concat
+            [ fname
+            , " :: forall eff. "
+            , T.intercalate " -> " typedArgs
+            , if null args then "" else " -> "
+            , "Affjax eff Foreign"
+            ]
+
+        typedArgs :: [ST]
```
jkarni: What's the right way of dealing with types here? Could [this](https://github.com/philopon/haskell-purescript-binding) be it? 

We haven't really figured out the best way of doing codegen for typed languages in servant-foreign yet, and this may be a good opportunity, though purescript is a bit of a special example.

servant-purescript/HLint.hs
fisx: probably.  i'll look into this.

servant-purescript/src/Servant/PureScript.hs
fisx: interesting link.  do you have a strong opinion on whether to sort this out now or do it incrementally, in later PRs?  i think i'm ok with both, but i'm not sure how much time I will have for this, so the former approach may take some time.

jkarni: My general concern is that it's much easier to talk about design and general directions (and reach a good solution) before things go into the repo than after. After merging, we tend to look the other way since cleaning up is much less rewarding than getting a new feature in (and because of backwards compatibility concerns).

jkarni: Actually, I take it back. This isn't about code quality (where punting is generally harmful) but about features, and in that context it makes sense to work incrementally. So (in this respect) let's go with it.

fisx: Cool.  (-:

I still want to look into these first, though:

1. Util module content has moved to purescript-globals, so Util can be removed.
1. Duplicate headers.
1. Correct content-type header.
1. camel-casing.

If anybody is waiting for this to become available, please poke me!

jkarni:
#277 introduces a mechanism for typed clients which seems very relevant

dredozubov:
#277 was merged, so i guess it's even more relevant now

servant-purescript/src/Servant/PureScript.hs
afcondon: I'm very interested in this - Purescript is right in the sweet spot between generated jQuery and generated GHCjs.

eskimor:
Awesome! Now I don't have to do it by myself! :-) 
I might be just stupid - but where can I find the repo this pull request came from?

Best regards,
Robert

fisx:
> Awesome! Now I don't have to do it by myself! :-)

you might, though!  (:

I've been meaning to get this in shape for review, but it doesn't look like it will move to the top of my list any time soon. But I'm happy to discuss this, and will try to respond more swiftly now!

> I might be just stupid - but where can I find the repo this pull request came from?

This is originally work from anchor (referenced in the description of this PR), who are not interested in maintaining it any more.  I've copied the code into my servant fork on zerobuzz, and that is where this PR is coming from.  Not sure that answers your question?

eskimor:
Thank you! That helped, found it. Don't know why I did not find it last time.
https://github.com/zerobuzz/servant/tree/introduce-servant-purescript.

My web-app's backend is written with servant, frontend is going to be purescript - it is very likely that this would come in handy. If it does not take too much time (I am on a tight schedule, unfortunately) I will definitely help out/ take over.

codedmart:
I might have a need for this soon and will try and pitch in as well.

fisx:
So many users!  I'm excited.  (-:

The application I had was trivial enough for it to make sense to do it without this, but I'm also expecting to needing this in the future.  Let's see who runs out of patience first --

jkarni:
As discussed in the [mailing list](https://groups.google.com/d/msg/haskell-servant/eHWiGBSn5nY/7ZalNAAVEAAJ) and #268, we settled on a policy of having codegen libs be in separate repos. I created a `servant-purescript` package in this org where this should be moved to.

eskimor:
@codedmart @fisx  - you might want to have a look at: https://github.com/eskimor/servant-purescript

I am currently implementing an example (central-counter) in order to have a real application for testing the library and make it work. (It compiles already and the generated code looks like it could work.)

PR #273, opened 2015-11-26
wip: add support for ghcjs to servant-client

soenkehahn:
Please, don't merge yet, wip!


soenkehahn:
@jkarni: Was there a specific use-case for de447dfe22ca080dd2ae6d898f08faea0a1b649f? Personally I liked it better with the global variable. Furthermore in `ghcjs` that argument doesn't make sense and will just be ignored.

jkarni:
@soenkehahn the idea was just to be able to configure the manager (there were requests to do so for, for example, request logging, on IRC).

soenkehahn:
I see. Well, then I'll ignore the argument in the case of `ghcjs` for now. Not sure what else to do.

jkarni:
We could try to come up with a datatype for configuration that makes sense for both ghc and ghcjs contexts. E.g.:

```
data OurManager = OurManager 
    { responseTimeout :: Int
    , modifyRequest   :: Request -> IO Request
    ...
    }
```

But ignoring the manager in ghcjs for now seems sensible.

arianvp:
@soenkehahn  I see you're doing your own  FFI calls to XHR. These are exposed in a nice API in the latest ghcjs-base https://github.com/ghcjs/ghcjs-base/blob/master/JavaScript/Web/XMLHttpRequest.hs

soenkehahn:
@arianvp: Thanks for the pointer. Maybe I'm wrong, but at first glance it looks like the version in `ghcjs-base` wouldn't work with `node` so we wouldn't be able to run the test-suite. Ideally that should be fixed upstream of course.

I'll be focusing on getting travis to work and then any changes will be much easier to make.

servant-client/servant-client.cabal
```diff
@@ -46,9 +56,14 @@ library
...
     , transformers-compat
+    , case-insensitive
+  if impl(ghcjs)
+    build-depends:
+        ghcjs-base
+      , ghcjs-prim
   hs-source-dirs: src
   default-language: Haskell2010
-  ghc-options: -Wall
+  ghc-options: -Wall -Werror
```
jkarni: `-Werror` on travis but not the package seems more sensible (in fact I think hackage doesn't even let you upload packages with `Werror`).

servant-client/test/Servant/Client/TestServer/GHCJS.hs
```diff
@@ -0,0 +1,34 @@
...
+
+import           Servant.Common.BaseUrl
+
+buildTestServer :: IO ()
+buildTestServer = do
+  process <- spawnProcess "./test/ghcjs/build-test-server.sh" []
+  ExitSuccess <- waitForProcess process
+  return ()
+
+withTestServer :: Application -> String -> (BaseUrl -> IO a) -> IO a
```
jkarni: This definitely warrants a comment.

servant-client/test/Servant/Client/TestServer.hs
```diff
@@ -0,0 +1,12 @@
+{-# LANGUAGE CPP                    #-}
+
+module Servant.Client.TestServer (
+  buildTestServer,
+  withTestServer,
+)where
+
```
jkarni: Here seems like a good way of explaining how the tests are run in each situation. Then the actual modules can just point here.

servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
```diff
@@ -0,0 +1,151 @@
...
+
+-- * initialization
+
+initXhr :: IO JSXMLHttpRequest
+initXhr = do
+  lib <- requireXMLHttpRequestClass
+  newXMLHttpRequest lib
+
+foreign import javascript unsafe
+  -- branching between node (for testing) and browsers
```
jkarni: I'd make the fact that this must be runnable on node and browsers more prominent (a lot of what goes on in this module happens because of that).

servant-client/src/Servant/Client/PerformRequest/Base.hs
```diff
@@ -0,0 +1,63 @@
...
+    , responseBody              :: ByteString
+    }
+  | ConnectionError
+    { connectionError           :: SomeException
+    }
+  deriving (Show, Typeable)
+
+instance Exception ServantError
+
+parseHeaders :: String -> ResponseHeaders
```
jkarni: Isn't this only used in GHCJS?

servant-client/servant-client.cabal
```diff
@@ -46,9 +56,14 @@ library
     , text
     , transformers
     , transformers-compat
+    , case-insensitive
```
jkarni: If we can in fact move `parseHeaders` to GHCJS-specific stuff, this could also go into the `impl(ghcjs)` branch below (I think).

.travis-ghcjs.sh
```diff
@@ -0,0 +1,12 @@
+#/usr/bin/env bash
+set -ev
+
+# tinc
+#cabal exec which hspec-discover
+#barf
+cabal install hspec-discover --prefix $HOME/.local
```
jkarni: instead of having a separate file, we could just have one `.travis.sh` file and branch on the `GHCJS` env variable.

.travis-ghc.sh
```diff
@@ -0,0 +1,6 @@
+#/usr/bin/env bash
```
jkarni: Probably this should go in the `scripts/` dir?

.travis-ghc.sh
```diff
@@ -0,0 +1,6 @@
...
+set -ev
+
+for package in $(cat sources.txt); do
```
jkarni: Using the `SOURCES` array from `scripts/common/lib.sh` has the advantage of making the script runnable from any dir (and also is more consistent with what we do elsewhere).

servant-client/servant-client.cabal
```diff
@@ -28,6 +28,16 @@ library
     Servant.Client
     Servant.Common.BaseUrl
     Servant.Common.Req
+     -- fixme: shouldn't be part of the public API:
+    Servant.Client.PerformRequest.Base
+  other-modules:
```
jkarni: I'm generally in favour of exporting all modules, but naming them `Internal.Something`.

servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
```diff
@@ -0,0 +1,151 @@
...
+-- * performing requests
+
+-- Performs the xhr and blocks until the response was received
+performXhr :: JSXMLHttpRequest -> Request -> IO ()
+performXhr xhr request = do
+  waiter <- newEmptyMVar
+  callback <- onReadyStateChange xhr $ do
+    state <- readyState xhr
+    case state of
+      4 -> putMVar waiter ()
```
jkarni: Magic number

servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
```diff
@@ -0,0 +1,151 @@
...
+  callback <- asyncCallback action
+  js_onReadyStateChange xhr callback
+  return callback
+foreign import javascript safe "$1.onreadystatechange = $2;"
+  js_onReadyStateChange :: JSXMLHttpRequest -> Callback (IO ()) -> IO ()
+
+foreign import javascript unsafe "$1.readyState"
+  readyState :: JSXMLHttpRequest -> IO Int
+
+openXhr :: JSXMLHttpRequest -> String -> String -> Bool -> IO ()
```
jkarni: I slightly prefer having `data Async = Async | Sync` over `Bool` in the last arg.

jkarni: Although only async is really being used?

servant-client/test/ghcjs/build-test-server.sh
```diff
@@ -0,0 +1,10 @@
+#/usr/bin/env bash
+
+set -o errexit
+
+cd test/ghcjs
```
jkarni: Could this also be folded into the `.travis.sh` file (i.e. have the test server be built *before* the tests are run)?

servant-client/test/ghcjs/package.yaml
```diff
@@ -0,0 +1,25 @@
+name: testServer
```
jkarni: I like `hpack`, but we should probably be consistent, and either switch everything to it, or nothing.

servant-client/test/ghcjs/run-tests.sh
```diff
@@ -0,0 +1,11 @@
+#/usr/bin/env bash
+
+# this script has to be executed from the 'servant-client' directory
```
jkarni: Also not sure why this is a separate script...

servant-client/test/ghcjs/testServer.hs
```diff
@@ -0,0 +1,27 @@
...
+{-# LANGUAGE TypeOperators #-}
+
+import           Network.Wai.Handler.Warp
+import           Servant
+import           System.IO
+
+import           Servant.Client.TestServer.GHC
+import           Servant.ClientSpec
+
+main :: IO ()
```
jkarni: maybe mention somewhere how this works?

jkarni:
Really awesome that the tests work (and very nifty way of getting them to)!

jkarni:
Fixes #51

yogsototh:
Hi! I have a runtime bug for which I really don't know how to resolve. I believe I made a somehow minimal project showing the bug. It is composed of a backend with servant and the front-end uses servant-client.

At runtime I get an error in `rts.js` 

    ReferenceError: Can't find variable: h$hsnet_getaddrinfo
    h$runSyncrts.js:11517
    crts.js:6376

Here is my minimal project:

<https://github.com/yogsototh/ghcjs-servant-client-bug>

As far as I could understand the bug, I believe this is the best place to post the problem.

arianvp:
@luite This probably means some code (probably in `Network`) is trying to call C FFI right?

arianvp:
@yogsototh  I'm trying to reproduce but your ghcjs build doesn't seem to exist anymore:
https://github.com/yogsototh/ghcjs-servant-client-bug/blob/master/ui/stack.yaml#L9

yogsototh:
@arianvp sorry my bad. I updated my repo to link to the right ghcjs repo. It should work now. Thanks.

soenkehahn:
@yogsototh: If I see it right it's not apparent from your repo which version of `servant-client` you're using. You can put github repos (plus commits) in the stack file to make this fully reproducible.

Also note that `servant-client` on master does not support `ghcjs` yet. This here is the PR to add this support. (I hope I'll be working on it now.)

yogsototh:
@soenkehahn ah yes sorry, I believe I put some hint hidden in my README.

I cloned the servant repository and I use the commit `5e215cef6834575c59354d5b1f1cce4d47f9aaa0`:

~~~
cd .../projects/
git clone https://github.com/haskell-servant/servant.git
cd servant/
git checkout 5e215cef6834575c59354d5b1f1cce4d47f9aaa0
~~~

My bad, I believe it should have worked on master.
I'll try with this PR as soon as possible and I'll get back. Thanks a lot and sorry I haven't understood correctly where the fix were.

servant-client/servant-client.cabal
soenkehahn: Removed it. We don't have `-Werror` on travis, I think we also have some warnings in some servant packages. (https://github.com/haskell-servant/servant/pull/273)

soenkehahn:
@yogsototh: No problem. But that explains the error message. If you try to use `ghcjs` to build libraries that use unsupported system calls or c libraries you can get these kind of error messages. So `servant-client` was trying to use the normal networking library.

yogsototh:
I merged master in this PR branch and now it seems to work but not as I would have expected.
I have no more runtime error, I just have a: `thread blocked indefinitely in an MVar operation` and apparently no POST request was sent.

Sorry to have mistaken the branch. I don't know what is wrong actually, the fact to have merged this PR with master might be the origin of the problem I'll inspect further using the branch. I couldn't do it immediately, I'll do this tonight.

Thanks and sorry again!

soenkehahn:
I get that `MVar` message too, but apart from that it works. I'm getting these messages since I started using `releaseCallback`, IIRC. (https://github.com/haskell-servant/servant/pull/273/files#diff-8e451bd14810307c0246f6abbf091384R63)

As you can probably see, this PR still needs some work. :)

yogsototh:
Apparently I fixed the problem for my use case (putMVar was called three times) by changing this:

~~~
git diff                                                                                      client-ghcjs-wip ✱
diff --git a/servant-client/src/Servant/Client/PerformRequest/GHCJS.hs b/servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
index 84afdbb..d4177c1 100644
--- a/servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
+++ b/servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
@@ -54,7 +54,9 @@ performXhr xhr request = do
   callback <- onReadyStateChange xhr $ do
     state <- readyState xhr
     case state of
-      4 -> putMVar waiter ()
+      4 -> do
+          toPut <- isEmptyMVar waiter
+          when toPut (putMVar waiter ())
       _ -> return ()
   openXhr xhr (cs $ method request) (toUrl request) True
   setHeaders xhr (requestHeaders request)
~~~

servant-client/test/Servant/Client/TestServer/GHCJS.hs
soenkehahn: I hope I made this much clearer now by refactoring the code and adding this comment: https://github.com/haskell-servant/servant/commit/8f7dedd90cc92a90688ffb41c851f65cc9b94e1d

servant-client/test/Servant/Client/TestServer.hs
soenkehahn: Hmm, now I added the comment to `Servant.ClientSpec.withTestServer`. @jkarni: Would you prefer it here?

servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
soenkehahn: Are you saying that we should emphasize more that this module works in browsers **and** in node?

I think this declaration (`requireXMLHttpRequestClass`) is the only place where anything happens differently. `xhr2` (the javascript library) is meant to provide a drop-in-replacement for xhr functionality in browsers.

servant-client/src/Servant/Client/PerformRequest/Base.hs
soenkehahn: Yes, you got me. ;)

I thought since it's just a pure function without additional dependencies I don't have to bother to CPP-exclude it. Should I put it into the ghcjs-specific module?

servant-client/servant-client.cabal
soenkehahn: Ok, I'll try this out.

soenkehahn: Turned out to be a bit more complicated: https://github.com/haskell-servant/servant/commit/d2d25954c3bbd04a9bddadf8d3abd8a57488aa33.
Now the test-suite doesn't depend on the library but uses the files from `src` directly. I don't know how to do it differently without putting `parseHeaders` into the public api.

soenkehahn:
@yogsototh: Very good, thanks! I'll incorporate that at some point.

soenkehahn:
Btw, we (@jkarni and I) decided that we're going to work on releasing `0.5` before getting this PR done.

jkarni:
@soenkehahn can we close this while it's not yet ready?

soenkehahn:
@jkarni: We can, but I do like to have a place on github to discuss this attempt. Would you prefer to close this PR and open an issue, `servant-client: support ghcjs`?

jkarni:
Yup

soenkehahn:
Oh, we have that already.

Closing in favor of #51/until ready to merge.

alpmestan:
Why close a WIP pull request?

Issue #239, opened 2015-09-25
Empty route with headers not possible?


edsko:
A route of type 

``` haskell
Post '[] ()
```

works fine, but a route of type

``` haskell
Post '[] (Headers '[Header "X-An-Int" Int] ())
```

yields

```
    Could not deduce (Servant.API.ContentTypes.IsNonEmpty '[])
      arising from a use of ‘serve’
```

jkarni:
Yeah, I guess we're missing a few overlapping instances for `Headers ls ()`. Should be a simple fix, but I really want to focus on the HTTP errors work first, so I might not have time to fix it this weekend.

jkarni:
(And thanks for the report!)

edsko:
Sure, I'm not in any way blocked on this.

jkarni:
It's easy enough to fix this by adding new instances, but we'd be going further down the route of exponential blow-up in instances. Not sure if this is the right ticket for it, but definitely something this brings up...

fizruk:
@jkarni I think this is also closed by #276 and `NoContent` solution.

PR #276, opened 2015-11-28
Consolidate verbs

### LINKED ISSUE #274: Consolidate http verbs?
We currently have each verb as a separate datatype, and consequently we also need separate instances. This means more work to (a) write the instances; (b) test them; (c) make verb combinators with different status codes. Something that kosmikus (and I think others) have proposed is having a type synonym or newtype (the latter with GND). Turns out GND doesn't work with classes with associated type synonyms. So what do we think of having:

```
data Verb (verb :: Method) (status :: Nat) ctypes a

type Get ctypes a = Verb GET 200 ctypes a
...
```
Instead? In this proposal, we would give up having automatic 204s for `()`, but personally I prefer making it easier to return 204s manually rather than trying to be smart but giving users less control. 

Alternatively, we could make `status` polykinded, and provide a few datatypes that encode more complex ways of figuring out the status (e.g., one for the current behaviour, where only `()` is 204, one for `Nothing` also being 204, etc.). I kind of like this idea, but it would introduce a little more complexity into the proposal.

Note that any of the proposals above suffer from the consequence that some type errors would not show `Get`, but instead `Verb GET`.

jkarni:
Fixes #274, #239 .

The main change is that now we have a single leaf combinator:

```
 data Verb (method :: k1) (statusCode :: Nat) (contentTypes :: [*]) a 
```
And the old ones are just type synonyms over this.

The behaviour of `()` also changes, so that it is no longer treated specially. Currently all empty responses return 204, but I think we might be better off always returning 200.


jkarni:
Any thoughts about switching everything to 200?

fizruk:
I am okay with 200 by default.

Although, maybe we should provide more synonyms then for variations of methods.
We might want to borrow [some documentation from HTTP](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html) to tell users which methods/codes they should use. E.g.:

```haskell
-- | POST 200 OK.
-- 
-- Should be used when request does not result
-- in a resource that can be identified by a URI.
-- The response should include an entity that describes the result.
type Post = Verb POST 200

-- | POST 201 Created.
-- 
-- Should be used when a new resource is created.
-- The response should contain an entity which describes
-- the status of the request and refers to the new resource,
-- and a Location header.
type Created = Verb POST 201
```

I also think it's a good thing that `()` will no longer have a special meaning.

It would still be nice to have a "no content" combinator. Maybe something like this:

```haskell
-- | 204 No Content.
--
-- Should be used if the action has been enacted
-- but the response does not include an entity.
data NoContent (method :: k1)
```

jkarni:
@fizruk thanks - good suggestions!

We could be a little more aggressive about good behaviour with the type synonyms. For example, 

```
type Created contentTypes a = Verb POST 201 contentTypes (Headers '[Header "Location" Link] a)
```

What do people think? Is it worth a sacrificing a little consistency for this? Obviously other `Verb` synonyms can always be defined, so neither options really limits what's possible, only settles what's easy.

@fizruk why a `NoContent` datatype? Why not just something like:

```
type PostNoContent contentTypes = Verb 'POST 204 contentTypes ()
``` 

Or the like?

fizruk:
The problem with enforcing `Location` header is that currently nested `Headers` don't work IIUC.
So if I wanted to add another header like in `Created '[JSON] (Headers [...] a)` it won't compile.
Please, correct me if I'm wrong.

I don't insist on `NoContent` datatype, we can make it a synonym, yes.

Does "no content" actually need content types?
Can't we just make it empty list, since there is no content?

Also, `()` is not exactly "no content". E.g. there is a `ToJSON` instance for `()`.

I would be okay with a dedicated `NoContent` unit type however:

```haskell
data NoContent = NoContent

type PostNoContent = Verb 'POST 204 '[] NoContent
```

jkarni:
> The problem with enforcing Location header is that currently nested Headers don't work IIUC.

Very fair - I had just come across this when writing down the new combinator. We could add an extra param for the headers *inside* the list, but that would look very weird.

> Does "no content" actually need content types?

No, it's just somewhat inconvenient to implement it without...

I take your point re: `NoContent` over `()`. The only annoying thing is `Control.Monad.void` and the like not leading directly to no content...

fizruk:
> The only annoying thing is Control.Monad.void and the like not leading directly to no content...

OTOH `return NoContent` in a handler is not hard to write and is kind of documenting it locally.

jkarni:
Ok, agreed - `NoContent` it is.

jkarni:
Back to:
>  The problem with enforcing Location header is that currently nested Headers don't work IIUC.

Is it worth creating a type family to make this work?

jkarni:
OK, I added the type synonyms, and removed the override-to-204-on-no-body behaviour. Status codes now are always, in case of success, what the verb combinator says they are.

fizruk:
> Is it worth creating a type family to make this work?

Can you elaborate on how you're going to make this work with type family?

jkarni:
> Can you elaborate on how you're going to make this work with type family?
```
type family WithHeader a b where
    WithHeader (Header h v) (Headers xs a) = Headers ((Header h v) ': xs) a
    WithHeader (Header h v) a              = Headers '[Header h v] a

x :: WithHeader (Header "test" Int) (Headers '[Header "a" Int] String)
x = addHeader 5 $ addHeader 3 "hi" :: Headers '[Header "test" Int, Header "a" Int] String

y :: WithHeader (Header "test" Int) String
y = addHeader 5 "hi" :: Headers '[Header "test" Int] String
```

jkarni:
OTOH I'm now thinking this might be too much - a comment suggesting `addHeader link` seems simpler and good enough.

jkarni:
Rebased. The only open question, in my mind, is whether we want a `AnyContentType` content type. This makes everything a little uglier, since we can't associate media types directly with content type types, and the internals also suffer. I personally don't think it's worth it.

fizruk:
> OTOH I'm now thinking this might be too much

Agree, after all Location header SHOULD be there, it is not mandatory.

> whether we want a AnyContentType content type

Do folks want it? I am not sure what that would mean for server/client though. Would it be that server can generate **any content-type, demanded by client** in `Accept`? Or would it be that **server can respond with any content type** regardless of `Accept`? In any scenario how server/client will encode/decode *any* content or should the encoding/decoding be left to the application then?

jkarni:
`AnyContentType` would really only make sense for `NoContent`. But the link between the two is a bit loose, which I dislike.

jkarni:
Any further thoughts? Should I merge?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -90,7 +90,7 @@ type TestHeaders = '[Header "X-Example1" Int, Header "X-Example2" String]
 
 type Api =
        "get" :> Get '[JSON] Person
-  :<|> "deleteEmpty" :> Delete '[] ()
+  :<|> "deleteEmpty" :> Delete '[JSON] ()
```
fizruk: Shouldn't this now be `DeleteNoContent '[]`?

jkarni: Well, not sure about the `'[]` - we still need explicit content-types even for `NoContent`. But this brought up a couple of other issues - give me a few min.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -176,95 +180,33 @@ methodRouterHeaders method proxy status action = LeafRouter route'
...
-instance OVERLAPPING_
-         HasServer (Delete ctypes ()) where
-
-  type ServerT (Delete ctypes ()) m = m ()
+  type ServerT (Verb method status ctypes a) m = m a
 
-  route Proxy = methodRouterEmpty methodDelete
+  route Proxy = methodRouter method (Proxy :: Proxy ctypes) status
+    where method = reflectMethod (Proxy :: Proxy method)
+          status = toEnum . fromInteger $ natVal (Proxy :: Proxy status)
```
fizruk: What is `toEnum` converting to here? Wouldn't `fromInteger` be enough?

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -130,15 +130,21 @@ captureSpec = do
 
 
 type GetApi = Get '[JSON] Person
-        :<|> "empty" :> Get '[] ()
-        :<|> "post" :> Post '[] ()
+        :<|> "empty" :> Get '[JSON] ()
+        :<|> "emptyWithHeaders" :> Get '[JSON] (Headers '[Header "H" Int] ())
+        :<|> "post" :> Post '[JSON] ()
```
fizruk: I believe these also should use `*NoContent` combinators.

jkarni:
Hm, just realized `*NoContent` doesn't currently work nicely with headers.

jkarni:
The simplest solution would be to have `*NoContentWithHeaders ctyps hdrs`.

servant-server/src/Servant/Server/Internal.hs
jkarni: We need a `Network.HTTP.Types.Status`, which doesn't have a `Num` instance (but does have an `Enum` instance).

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -307,6 +304,28 @@ instance MimeRender OctetStream ByteString where
 instance MimeRender OctetStream BS.ByteString where
     mimeRender _ = fromStrict
 
+-- | A type for responses with content-body.
```
soenkehahn: `s/with/without`?

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -307,6 +304,28 @@ instance MimeRender OctetStream ByteString where
...
+data NoContent = NoContent
+  deriving (Show, Eq, Read)
+
+instance FromJSON NoContent where
+    parseJSON _ = return NoContent
```
soenkehahn: We should have a test that makes sure that `aeson` doesn't crash before it calls `parseJSON`. In the `NoContent` case we don't have an input that's a valid json value, so `aeson` cannot construct the `Value` that it would pass to `parseJSON`. (Hopefully laziness saves us here, but we need a test.)

servant/src/Servant/API/Verbs.hs
```diff
@@ -0,0 +1,166 @@
...
+
+import           Data.Typeable             (Typeable)
+import           GHC.Generics              (Generic)
+import           GHC.TypeLits              (Nat)
+import           Network.HTTP.Types.Method (Method, StdMethod (..),
+                                            methodDelete, methodGet, methodHead,
+                                            methodPatch, methodPost, methodPut)
+import           Servant.API.ContentTypes  (NoContent(..))
+
+-- | @Verb@ is a general type for representing HTTP verbs/methods. For
```
soenkehahn: Maybe `verbs (aka methods)`?

servant/src/Servant/API/Verbs.hs
```diff
@@ -0,0 +1,166 @@
...
+-- resource should be given in the @Location@ header field.
+--
+-- If the resource cannot be created immediately, use 'PostAccepted'.
+--
+-- Consider using 'Servant.Utils.Links.safeLink' for the @Location@ header
+-- field.
+
+-- | 'POST' with 201 status code.
+--
+type Created contentTypes a = Verb 'POST 201 contentTypes a
```
soenkehahn: Maybe `PostCreated`, just to be consistent (and explicit)?

servant/src/Servant/API/Verbs.hs
```diff
@@ -0,0 +1,166 @@
...
+
+-- ** 204 No Content
+--
+-- Indicates that no response body is being returned. Handlers for these should
+-- return 'NoContent', possibly with headers.
+--
+-- If the document view should be reset, use @205 Reset Content@.
+
+-- | 'GET' with 204 status code.
+type GetNoContent contentTypes noContent = Verb 'GET 204 contentTypes noContent
```
soenkehahn: What about:
``` haskell
type GetNoContent contentTypes = Verb 'GET 204 contentTypes NoContent
```
?

It's less consistent with the other type synonyms, but would disallow to actually return content. (Of course you could still do it with your own type synonym if you really wanted.)

servant/src/Servant/API/Verbs.hs
```diff
@@ -0,0 +1,166 @@
...
+-- ** 206 Partial Content
+--
+-- Indicates that the server is delivering part of the resource due to a range
+-- header in the request.
+--
+-- For more information, see <https://tools.ietf.org/html/rfc7233#section-4.1
+-- RFC7233 Section 4.1>
+
+-- | 'GET' with 206 status code.
+type GetPartialContent contentTypes noContent = Verb 'GET 205 contentTypes noContent
```
soenkehahn: `s/205/206`

servant/src/Servant/API/Verbs.hs
```diff
@@ -0,0 +1,166 @@
...
+
+
+class ReflectMethod a where
+    reflectMethod :: proxy a -> Method
```
soenkehahn: Should this be `Proxy`? I know that this polymorphic proxy is also sometimes used intentionally, but in `servant` so far we use `Proxy`, don't we?

servant/src/Servant/Utils/Links.hs
```diff
@@ -74,7 +74,9 @@
 -- >>> safeLink api bad_link
 -- ...
 --     Could not deduce (Or
---                         (IsElem' (Delete '[JSON] ()) (Get '[JSON] Int))
+--                         (IsElem'
+--                            (Verb 'Network.HTTP.Types.Method.DELETE 200 '[JSON] ())
```
soenkehahn: I would re-export all the `StdMethod`s (`DELETE`, `GET`, etc.) to avoid the fully-qualified identifiers in type errors and thereby make them better readable.

servant-client/src/Servant/Client.hs
```diff
@@ -503,3 +406,20 @@ instance HasClient api => HasClient (IsSecure :> api) where
...
+
+   instance (..., cts' ~ (ct ': cts)) => ... cts' ...
+
+It may seem to make more sense to have:
+
+   instance (...) => ... (ct ': cts) ...
+
+But this means that if another instance exists that does *not* require
+non-empty lists, but is otherwise more specific, no instance will be overall
+more specific. This in turns generally means adding yet another instance (one
```
soenkehahn: `s/turns/turn`

servant-client/src/Servant/Common/Req.hs
```diff
@@ -151,23 +151,26 @@ performRequest reqMethod req reqHost manager = do
...
       unless (status_code >= 200 && status_code < 300) $
         throwE $ FailureResponse status ct body
-      return (status_code, body, ct, hrds, response)
+      return (status_code, body, ct, hdrs, response)
 
 
 performRequestCT :: MimeUnrender ct result =>
-  Proxy ct -> Method -> Req -> BaseUrl -> Manager -> ExceptT ServantError IO ([HTTP.Header], result)
+  Proxy ct -> Method -> Req -> BaseUrl -> Manager
+  -> ExceptT ServantError IO ([HTTP.Header], result)
```
soenkehahn: Indent this line by two more spaces?

servant-client/src/Servant/Common/Req.hs
```diff
@@ -151,23 +151,26 @@ performRequest reqMethod req reqHost manager = do
...
+    Right val -> return (hdrs, val)
 
-performRequestNoBody :: Method -> Req -> BaseUrl -> Manager -> ExceptT ServantError IO ()
-performRequestNoBody reqMethod req reqHost manager =
-  void $ performRequest reqMethod req reqHost manager
+performRequestNoBody :: Method -> Req -> BaseUrl -> Manager
+  -> ExceptT ServantError IO [HTTP.Header]
+performRequestNoBody reqMethod req reqHost manager = do
+  (_status, _body, _ct, hdrs, _response) <- performRequest reqMethod req reqHost manager
+  return hdrs
```
soenkehahn: Before this change you couldn't return headers when returning `()`?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -283,7 +283,7 @@ failSpec = beforeAll (startWaiApp failServer) $ afterAll endWaiApp $ do
           _ -> fail $ "expected InvalidContentTypeHeader, but got " <> show res
 
 data WrappedApi where
-  WrappedApi :: (HasServer api, Server api ~ ExceptT ServantErr IO a,
+  WrappedApi :: (HasServer (api :: *), Server api ~ ExceptT ServantErr IO a,
```
soenkehahn: Is this kind-annotation better or needed?

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -314,7 +316,7 @@ postSpec = do
 type PutApi =
        ReqBody '[JSON] Person :> Put '[JSON] Integer
   :<|> "bla" :> ReqBody '[JSON] Person :> Put '[JSON] Integer
-  :<|> "empty" :> Put '[] ()
+  :<|> "empty" :> Put '[JSON] ()
```
soenkehahn: Shouldn't this be `Put '[] NoContent` now? (Or `Put '[JSON] NoContent`?)

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -505,7 +501,7 @@ responseHeadersSpec :: Spec
 responseHeadersSpec = describe "ResponseHeaders" $ do
   with (return $ serve (Proxy :: Proxy ResponseHeadersApi) responseHeadersServer) $ do
 
-    let methods = [(methodGet, 200), (methodPost, 201), (methodPut, 200), (methodPatch, 200)]
+    let methods = [(methodGet, 200), (methodPost, 200), (methodPut, 200), (methodPatch, 200)]
```
soenkehahn: If this only maps to `200` we can refactor this. Just have a list of methods.

soenkehahn:
I've taken a look at all the changes and I think that everything looks good. Overall it's a nice change, I like how much code can be deleted.

:+1:

servant-server/test/Servant/ServerSpec.hs
fizruk: Or maybe we should expand the test suite with other status codes instead.

servant/src/Servant/API/Verbs.hs
fizruk: It was like, but @jkarni realized it disallowed to easily add headers to `NoContent` then.
With explicit `noContent` argument it is possible to add headers like this:

```haskell
GetNoContent '[JSON] (Headers '[Header "X-Custom-Header" CustomData] NoContent)
```

soenkehahn: Ah, makes sense.

servant/src/Servant/API/Verbs.hs
jkarni: Good question. Maybe we should discuss. I now prefer the polymorphic version, but I'd love to hear what others think. Also, there is the issue of consistency.

soenkehahn: I'm not very passionate, except about using the same everywhere in `servant`. I feel `Proxy` is easier to understand. And `proxy` is only more convienient _sometimes_.

I vote for `Proxy` now but I won't stop anyone changing this everywhere in `servant` later.

servant/src/Servant/API/ContentTypes.hs
jkarni: Turns out `aeson` does crash with no content. However, there is an overlapping instance for `NoContent` in `servant-client`, in which we don't even try parsing the body. I don't know if that's good enough.

servant/src/Servant/Utils/Links.hs
jkarni: Wow, nice - this makes the error go back to the type synonym!

servant-client/src/Servant/Common/Req.hs
jkarni: Nope - this fixes #239, which is that issue.

servant-server/test/Servant/ServerSpec.hs
jkarni: Actually, should I consolidate all the tests as well? I don't think it any longer makes sense to have separate tests for each method, since the code paths are all the same.

servant-server/test/Servant/ServerSpec.hs
jkarni: Good idea.

jkarni: This requires some refactoring of the test suite, which would be useful now in quite a few of the tests. I'm inclined to say that this should be part of the PR (we can considerably reduce the testing logic too now), but it'll take a while longer.

servant-server/test/Servant/ServerSpec.hs
soenkehahn: Yes, please.

servant-client/test/Servant/ClientSpec.hs
jkarni: It is needed (I needed to turn on `PolyKinds` somewhere). We could put it in `HasServer` itself, I guess?

jkarni:
Modulo the `aeson` `NoContent` and `PolyKinds` note above (which I don't know what we think of) I think this is done.

soenkehahn:
Nice!

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -689,124 +674,37 @@ instance (KnownSymbol sym, ToCapture (Capture sym a), HasDocs sublayout)
...
-    where hdrs = allHeaderToSample (Proxy :: Proxy ls)
-          endpoint' = endpoint & method .~ DocGET
-          action' = action & response.respBody .~ take _maxSamples (sampleByteStrings t p)
-                           & response.respTypes .~ allMime t
-                           & response.respHeaders .~ hdrs
-          t = Proxy :: Proxy cts
-          p = Proxy :: Proxy a
-
-instance (KnownSymbol sym, HasDocs sublayout)
-      => HasDocs (Header sym a :> sublayout) where
```
luigy: Was this removal intentional?

jkarni: No, nice catch. Fixed in #324

PR #277, opened 2015-11-28
Type information in servant-foreign.

MaxOw:
I'm creating [C# backend](https://github.com/MaxOw/servant-sharp) to servant and need access to types. All this is basically a `HasForeignType` class and all the piping around it. Then when creating backend you just provide equivalent backend language types for all haskell types that you are interested in, by providing instances to that class. Specifically in dynamic languages you just need to provide empty, catch all, instance.

### DIFF
--- a/servant-foreign/src/Servant/Foreign/Internal.hs
+++ b/servant-foreign/src/Servant/Foreign/Internal.hs
@@ -130,11 +130,6 @@ type family Elem (a :: *) (ls::[*]) :: Constraint where
 -- > -- instances.
 -- > data LangX
 -- >
--- > -- If the language __X__ is dynamically typed then you only need
--- > -- a catch all instance of a form
--- > instance HasForeignType LangX a where
--- >    typeFor _ _ = empty
--- >
 -- > -- Otherwise you define instances for the types you need
 -- > instance HasForeignType LangX Int where
 -- >    typeFor _ _ = "intX"
@@ -150,8 +145,20 @@ type family Elem (a :: *) (ls::[*]) :: Constraint where
 -- >              => Proxy api -> [Req]
 -- > getEndpoints api = listFromAPI (Proxy :: Proxy LangX) api
 --
+-- > -- If language __X__ is dynamically typed then you can use
+-- > -- a predefined NoTypes parameter
+-- > getEndpoints :: (HasForeign NoTypes api, GenerateList (Foreign api))
+-- >              => Proxy api -> [Req]
+-- > getEndpoints api = listFromAPI (Proxy :: Proxy NoTypes) api
+-- >
+--
 class HasForeignType lang a where
     typeFor :: Proxy lang -> Proxy a -> ForeignType
+
+data NoTypes
+
+instance HasForeignType NoTypes a where
+    typeFor _ _ = empty
 
 class HasForeign lang (layout :: *) where
   type Foreign layout :: *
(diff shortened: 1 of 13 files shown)


jkarni:
Wonderful! We've been bumping up against the question of typed client libraries in the past few days, and this is great.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -120,6 +122,9 @@ type family Elem (a :: *) (ls::[*]) :: Constraint where
   Elem a (a ': list) = ()
   Elem a (b ': list) = Elem a list
 
+class HasForeignType a where
+    typeFor :: Proxy a -> ForeignType
```
jkarni: One issue with this is that there'll have to be orphan instances. Also, the class `HasForeignTypes` will mean different things in different contexts, and things will only work if instances for different languages aren't ever in scope at the same time. Would it not perhaps be better to have another type parameter?

```
class HasForeignType typ lang where
     typeFor :: Proxy lang -> Proxy a -> ForeignType
```

That way each package can declare its own `lang` datatype (e.g. `data JSLang`) in its own package, and different instances won't conflict across languages.

MaxOw:
Funny thing is I've actually thought about that, but I've somehow assumed that instances specific to a backend would be hidden internally. But yeah, being explicit is better.

servant-foreign/src/Servant/Foreign/Internal.hs
MaxOw: Funny thing is I've actually thought about that, but I've somehow assumed that instances specific to a backend would be hidden internally. But yeah, being explicit is better.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -120,142 +122,175 @@ type family Elem (a :: *) (ls::[*]) :: Constraint where
   Elem a (a ': list) = ()
   Elem a (b ': list) = Elem a list
 
-class HasForeign (layout :: *) where
+class HasForeignType lang a where
```
jkarni: Could you add a short haddock description (in particular mentioning how dynamically typed languages should deal with this)?

jkarni: In fact, maybe having the return type be `Maybe ForeignType` would be clearer?

jkarni:
Could you add a test checking that `Req` gets the right `ForeignType`? Aside from that, and the small comments above, LGTM - thanks!

alpmestan:
Yeah, I'd love to see some haddocks to describe how one can use `HasForeignType`. Aside from that, LGTM as well.

MaxOw:
Ok, Added tests and documentation. I think that's all.

alpmestan:
So we'll have you *and* @dredozubov to maintain *servant-foreign*? Great =)

The PR now looks ready to me. @jkarni ?

dredozubov:
I have one concern regarding this PR - i'd like to see something dealing with defaults for scripting languages and such. Maybe default method implementation?

@alpmestan haha, seems legit. On the serious note - i can help with maintaining this package, you've got a whole package zoo by now.

MaxOw:
@dredozubov How about something like this?

dredozubov:
I think it's looking good. :)

MaxOw:
Come on. Now you're just bikeshedding. :)

jkarni:
For what it's worth, Dynamic would conflict with Data.Dynamic.

dredozubov:
@MaxOw good job! Thanks for the PR!

PR #281, opened 2015-12-02
Servant dates

axman6:
Adds types for `CaptureTime` and `QueryParamTime(s)`, with implementations of `HasServer`. Some more tests are needed, but this should do for a start. Let me know what further work is needed.


alpmestan:
Why not a single type:

``` haskell
data Time (fmt :: Symbol) a
```

and then a `FromText` instance that uses `parseTime`?

axman6:
That's a much better idea, I'll implement that soonish.

axman6:
I forgot that GHC 7.8 has different `time` package than 7.10, I'll make this work shortly.

axman6:
Turns out I can't get GHC 7.8 to run on El Capitan, is anyone else able to add the necessary `CPP` and .cabal file changes to get this working on 7.8 too? Should be mostly mechanical, replacing `parseTimeM True` with `parseTime`, and checking gcc versions in the cabal file to choose which time packages to use.

christian-marie:
Has anything changed that will cause it not to build? You could try loosening the bound and seeing if travis just works.

axman6:
I'll see if I can find some code of ours which deals with the time issues which happened between 7.8 and 7.10.

axman6:
```haskell
{-# LANGUAGE CPP #-}
... 
#if MIN_VERSION_time(1,5,0)
import           Data.Time                    (ZonedTime, zonedTimeToUTC)
import           Data.Time.Format             (defaultTimeLocale, parseTimeM)
#else
import           Data.Time                    (ZonedTime, parseTime,
                                               zonedTimeToUTC)
import           System.Locale                (defaultTimeLocale)
#endif
...
#if MIN_VERSION_time(1,5,0)
 parseTime = parseTimeM True
#endif
```

Since I can't test this myself at the moment, these aren't the exact changes needed, but the necessary one should be very similar.

servant/src/Servant/API.hs
```diff
@@ -83,6 +87,7 @@ import           Servant.API.ResponseHeaders (AddHeader (addHeader),
                                               HList (..), Headers (..),
                                               getHeadersHList, getResponse)
 import           Servant.API.Sub             ((:>))
+import           Servant.API.Times           (FTime(..), toProxy, getFormat, renderTime, parseTime)
```
jkarni: Does `toProxy` belong in the API?

axman6: I can see it being useful when using the Time type, but it probably only belongs if the type were going to be used more broadly than just servant. Happy for it to be removed/to remove it.

axman6: it could also have a nicer name, like `formatProxy` (which is less likely to conflict with any other packages).

jkarni:
Hm, maybe @fizruk wants to chip in. `http-api-data` is getting more date/time [instances](https://github.com/fizruk/http-api-data/pull/22/files) (though I like the ability to choose the format).

fizruk:
I like the `FTime` approach and it does not conflict with `http-api-data` bringing date/time instances.
So why not have both?

axman6:
The format the time is specified in really is part of the API, which is why I think it's good to have it available in the route/type/documentation automatically.

axman6:
Anyone know why this build is failing? Seems `stack` isn't being used for the Travis build?

luigy:
I think you should relax time bounds to >= 1.4 for 7.8.4 compat

axman6:
Yeah there's a bit more to it than that, but I've almost finished fixing it. time pre gcc 7.10 was a pain -_-

servant/src/Servant/API.hs
jkarni: I'm inclined to say we should remove it. It's available in `Times.hs` if really needed.

servant/src/Servant/API/Times.hs
```diff
@@ -0,0 +1,105 @@
...
+module Servant.API.Times
+    ( FTime(..)
+    , toProxy
+    , getFormat
+    , renderTime
+    , parseTime
+    ) where
+
+import           Data.Typeable (Typeable)
+import           GHC.TypeLits  -- (Symbol)
```
jkarni: Remove or uncomment? We tend to prefer explicit imports, though.

servant/src/Servant/API/Times.hs
```diff
@@ -0,0 +1,105 @@
...
+import           Web.HttpApiData
+import qualified Data.Time.Format as T
+#if !MIN_VERSION_time(1,5,0)
+import           System.Locale as T
```
jkarni: `qualified` (or another import with explicit imports)

servant/src/Servant/API/Times.hs
```diff
@@ -0,0 +1,105 @@
...
+    where
+#if !MIN_VERSION_time(1,5,0)
+        res = case T.parseTime T.defaultTimeLocale fmt str of
+#else
+        res = case T.parseTimeM False T.defaultTimeLocale fmt str of
+#endif
+            Nothing -> Left . pack $ "Could not parse time string \"" ++ str ++ "\" with format \"" ++ fmt ++ "\""
+            Just t -> Right (FTime t)
+
+        fmt = getFormat (toFTimeTy res)
```
jkarni: Could this just be `getFormat (Proxy :: Proxy (FTime format t)` (with scoped type variables)? (And then no need for the `undefined` `toFTimeTy`.)

servant/src/Servant/API/Times.hs
```diff
@@ -0,0 +1,105 @@
...
+    toQueryParam = toQueryParam . renderTime
+
+
+instance (KnownSymbol format, T.ParseTime t) => FromHttpApiData (FTime format t) where
+    parseUrlPiece   = parseUrlPiece   >=> parseTime
+    parseHeader     = parseHeader     >=> parseTime
+    parseQueryParam = parseQueryParam >=> parseTime
+
+
+toProxy :: FTime format t -> Proxy format
```
jkarni: If `toProxy` is not exported, this could be folded into `getFormat`.

servant/src/Servant/API.hs
jkarni: (Or manually defined)

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -194,6 +198,55 @@ captureSpec = do
...
+        response <- get "/datetime/2015-12-02%2012:34:56+1000"
+        liftIO $ simpleBody response `shouldBe` (fromString . show $ ftime)
+
+
+      it "returns 404 if the decoding fails" $ do
+        get "/notAnInt" `shouldRespondWith` 404
+
+    with (return (serve
+        (Proxy :: Proxy (Capture "datetime" (FTime TimeFormatWSpace UTCTime) :> Raw))
+        (\ (FTime day )request_ respond ->
```
jkarni: `) `, not ` )`

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -63,6 +63,7 @@ import           Servant.Server.Internal.RoutingApplication
 import           Servant.Server.Internal.ServantErr
 
 
+
```
jkarni: Extra newlines?

servant/src/Servant/API/Times.hs
```diff
@@ -0,0 +1,105 @@
...
+                       (readParen (i > 1)
+#if !MIN_VERSION_time(1,5,0)
+                                  (T.readsTime T.defaultTimeLocale fmt)
+#else
+                                  (T.readSTime False T.defaultTimeLocale fmt)
+#endif
+                                  str
+                        )
+
+            toFTimeTy :: [(FTime format t, String)] -> FTime format t
```
jkarni: Using scoped type variables again seems nicer than `undefined`

jkarni: Or for that matter, `fst . head`

servant/src/Servant/API/Times.hs
```diff
@@ -0,0 +1,105 @@
...
+-- as specified in 'Data.Time.Format'.
+--
+-- Example:
+-- >>>            -- GET /events/:date
+-- >>> type MyApi = "events" :> Capture "date" (FTime "%Y-%m-%d" Day) :> Get '[JSON] [Event]
+newtype FTime (format :: Symbol) t = FTime {getFTime :: t}
+    deriving (Typeable, Eq, Ord)
+
+instance (KnownSymbol format, T.FormatTime t) => Show (FTime format t) where
+    showsPrec i t = showParen (i > 1) (\str -> renderTime t ++ str)
```
jkarni: Is there a reason not to use the generic `Read/Show`? My sense is that we *want* it to differ from the underlying show.

jkarni:
This has been open for a while. Shall we close until further notice? I think @soenkehahn and I at least have also moved towards preferring things to go into separate packages (though I don't know what everyone else thinks).

axman6:
Hey, sorry about this dropping off my radar, we ended up just embedding it in our app. I'm not sure that the work here is big enough to warrant it's own package, and having instances for times seems fairly fundamental to me (basically all our services use times in their API).

I'll get back to this this weekend and clean things up and we'll see if you're happy with it being in `servant` after that.

servant/src/Servant/API/Times.hs
axman6: I wanted the Read and Show instances to use the format that their type specified, also happens to make playing with it in say GHCi a bit easier (and should keep the `show . read` identity)

servant/src/Servant/API/Times.hs
axman6: In this case, despite having the KnownSymbol format constraint, replacing this with symbolVal (Proxy :: Proxy format) doesn't type check so I've left this as is for now. (push to come)

axman6:
Ok, looks like I did the wrong thing on my branch when trying to bring things up to 0.8 - I'll close this and submit a new PR.

axman6:
See #583 for new PR.

PR #282, opened 2015-12-02
RawServer datatype, for uniformity of Enter

avieth:
The commit message:

> Previously, ServerT Raw m ~ Application. Seems reasonable, but has the
> unfortunate consequence of making Enter useless for Raw routes.
> Attempting to solve the Enter class constraint for a Raw route will run
> up to IO ResponseReceived, the tail end of the Wai Application type,
> in practical use ultimately demanding something along the lines of:
> 
>   Enter (IO ResponseReceived) (yourMonad :~> EitherT ServantErr IO) (IO ResponseReceived)
> 
> There's no need to use Enter on a Raw route anyway, I know, but with
> this change, the programmer can treat Raw routes and non-Raw routes
> uniformly with respect to Enter.

I'd like to see this merged because I think we ought to be able to use `enter` on any route, even if doing so has no effect (as is the case on `Raw`).

### DIFF
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -40,6 +40,7 @@ library
     Servant.Server.Internal.Router
     Servant.Server.Internal.RoutingApplication
     Servant.Server.Internal.ServantErr
+    Servant.Server.Internal.RawServer
     Servant.Utils.StaticFiles
   build-depends:
         base               >= 4.7  && < 5
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -55,6 +55,7 @@ import           Servant.API.ResponseHeaders (GetHeaders, Headers, getHeaders,
 import           Servant.Server.Internal.Router
 import           Servant.Server.Internal.RoutingApplication
 import           Servant.Server.Internal.ServantErr
+import           Servant.Server.Internal.RawServer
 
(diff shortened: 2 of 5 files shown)


imalsogreg:
We've got some similar work going on at #248 and https://github.com/haskell-servant/servant/tree/jkarni/another178 - it seems like 'another178' does what you need?

jkarni:
This might be a much simpler solution than `178` (which is basically becoming this anyhow). The only question is whether we want to add the other (`*`-kinded) param for documentation.

fizruk:
What's the status of this?

@jkarni do you want this merged? This closes #178, #248 and #157, right?

@avieth do you need help fixing the builds?

avieth:
> @avieth do you need help fixing the builds?

I haven't even looked into why it's failing, since I've no idea whether this will be approved for merge.

phadej:
@alpmestan what do you think about this? Adding `RawServer` with my new `Enter` approach would be quite trivial; but drift us even further from "Generic" `Enter`

alpmestan:
This needs to be updated, but I'd otherwise be happy to make `Enter` play well with `Raw` endpoints. It would indeed make us drift away from a generic, ideal `Enter`. On the other hand, we don't have any such implementation yet, so let's not block any related work on it. If one day we find a nice approach, then we can reconsider the case of `Raw` endpoints, but for now LGTM, modulo rebasing against HEAD.

phadej:
Closing in favor of #748

PR #283, opened 2015-12-03
File extensions

axman6:
Adds the ability to capture `Text` file names with an extension specified in the type.

``` haskell
           -- GET /file/:filename.png
type MyApi = "file" :> Capture "filename" (Ext "json") :> Get '[JSON] Text

-- when accessing "/file/foo.json", prints:
--     foo.json
--     foo
handler :: Ext "json" -> ExceptT ServantErr IO Text
handler t@(Ext n) = return (renderExt t <> "\n" <> n)
```

We serve several server generated CSV and PNG files which currently have URLs which do not contain the file extension, which confuses some browsers, despite the correct content type.

I tried using something more generic than Text for the file name, but it proved to be not very useful - supporting a generic `From/ToHttpApiData` type only makes sense sometimes.

### DIFF
--- a/servant/src/Servant/API/FileExtension.hs
+++ b/servant/src/Servant/API/FileExtension.hs
@@ -1,0 +1,88 @@
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE PolyKinds          #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+
+module Servant.API.FileExtension
+    ( Ext(..)
+    , getExt
+    , parseExt
+    ) where
+
+import           Data.Typeable (Typeable)
+import           GHC.TypeLits  -- (Symbol)
+import           Web.HttpApiData
+import           Data.Text (Text, pack, stripSuffix)
+import           Data.Proxy
+import           Data.Maybe (catMaybes)
+import           Control.Monad ((>=>))
+
+-- | A wrapper around a time type which can be parsed/rendered to with `format',
+-- as specified in 'Data.Time.Format'.
+--
+-- Example:
+-- >>>            -- GET /file/:filename.png
+-- >>> type MyApi = "file" :> Capture "filename" (Ext "png") :> Get '[JSON] Text
+newtype Ext (ext :: Symbol) = Ext {getFileName :: Text}
+    deriving (Typeable, Eq, Ord)
+
+instance (KnownSymbol ext) => Show (Ext ext) where
+    showsPrec i tt@(Ext t) = showsPrec i (mappend t . pack $ '.':getExt tt)
+
+instance (KnownSymbol ext) => Read (Ext ext) where
+    readsPrec i str = res
+        where
+            -- res :: [(Ext ext t, String)]
+            res = catMaybes . map f . readsPrec (i+1) $ str
+
+            f :: (Text,String) -> Maybe (Ext ext,String)
+            f (t,s) = fmap (\txt -> (Ext txt,s)) $ stripSuffix (pack $ '.':ext) t
+
+            toExtTy :: [(Ext ext, String)] -> Ext ext
+            toExtTy _ = undefined
+
+            ext = getExt (toExtTy res)
+
+
+
+instance (KnownSymbol ext) => ToHttpApiData (Ext ext) where
+    toUrlPiece   = toUrlPiece   . show
+    toHeader     = toHeader     . show
+    toQueryParam = toQueryParam . show
+
+
+instance (KnownSymbol ext) => FromHttpApiData (Ext ext) where
+    parseUrlPiece   = parseUrlPiece   >=> parseExt
+    parseHeader     = parseHeader     >=> parseExt
+    parseQueryParam = parseQueryParam >=> parseExt
+
+
+toExtProxy :: Ext ext-> Proxy ext
+toExtProxy _ = Proxy
+
+getExt :: KnownSymbol ext => Ext ext -> String
+getExt t = symbolVal (toExtProxy t)
+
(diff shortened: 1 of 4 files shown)


freezeboy:
Why not a composition using (:>) operator ? this would allow the use of this extension with other stuff than Capture ? (I have no example in mind right now)

axman6:
This can also be used in query parameters and headers through the use of the `To/FromHttpApiData`, not just `Capture`s. Maybe I've misunderstood your point @freezeboy.

axman6:
Also `:>` is used to split the URL path by `/`'s not full stops, so having something like: `Capture "file name" Text :> ".png" ... ` matches things like `foo/.png`.

dmjio:
@axman6 why not just make your own combinator for this? 
```haskell
data Ex (fileName :: Symbol) (extension :: Symbol)
instance (KnownSymbol fileName, KnownSymbol extension, HasServer api) => 
     HasServer (Ex fileName ext :> api) where
 .... 
```

axman6:
The file name is the part were trying to capture, so it can't be a static part of the type. The helper allows you to handle requests for any file with a given extension instead of having to leave the extension off the URL.

axman6:
This implementation gets a lot of benefits for free, it can be used in several places in the URL and request without having to implement HasServer or anything else.

dmjio:
@axman6 Ah I see, of course the filename has to be dynamic =) I guess then you could in theory do something like this:
```haskell
data Ex (extension :: Symbol)
instance (KnownSymbol extension, HasServer api) => HasServer (Ex fileName ext :> api) where
 ... capture filename in path, ensure it matches extension ....
```

axman6:
Sure, but this already does that when used with Capture, ensures the extension is present and strips it (but it's recoverable because the type is known)

freezeboy:
@axman6 I only read your description quckly, now I reviewed your commit and I understand better what you call an extension. I though you added a 'parameter' to Capture only, I like your implementation. sorry for the noise

jkarni:
I see the point. It's a little annoying that content-types are not sufficient in all cases, but I guess we have to come to terms with that. @axman6 do you have a reference for when browsers misbehave w.r.t. content-types (but not extensions)?

(Also, you'll have to add a CPP'd import of `mappend` to satisfy the 7.8 build.)

servant/src/Servant/API/FileExtension.hs
```diff
@@ -0,0 +1,90 @@
...
+
+import           Data.Typeable (Typeable)
+import           GHC.TypeLits  -- (Symbol)
+import           Web.HttpApiData
+import           Data.Text (Text, pack, stripSuffix)
+import           Data.Proxy
+import           Data.Maybe (catMaybes)
+import           Control.Monad ((>=>))
+
+-- | A wrapper around a time type which can be parsed/rendered to with `format',
```
jkarni: `Time`? This seems like it's from the other PR.

axman6: Whoops, yes, they we're developed at the tame time and I guess the two branches weren't kept as sterile as they should have been =)

jkarni:
Thanks for the fixes.

General questions I should have asked earlier:

- The more general combinator is, I would think, a Regex one (which however we probably wouldn't want in the main repo since it would add annoying dependencies). Is there a reason to prefer this over a Regex combinator?
- There is no connection between the content-type and the extension. In particular, we can't accept multiple extensions that correspond to content-types that can be served. This seems like it would be confusing, but a clearer sense of when browsers ignore content-types (but not extensions) would be helpful.
- When should this be used? (Related to the above.)

phadej:
From what I can see, this functionality doesn't need to leave in `servant`. 

cc @alpmestan, I guess this one could be used as an example of writing your own combinators.

FWIW, it might make sense to setup `servant-contrib` package (not to be released, if we don't want) to accumulate this kind of *useful, but not enough* combinators / stuff. Also to act as wider "what I will break with this change" -bench.

alpmestan:
Yes, servant-contrib/extras sounds like something we have to put somewhere on the roadmap now. Maybe `servant-multipart` could be merged into it? I don't know really. I'm not fond of -contrib/-extras packages but don't see a better solution here. It's however tough to decide what should go in there and what should be released in an individual package.

phadej:
Yes, it's hard to draw a line. I'd keep multipart as a separate one. But `Ext` is too small to justify a package. OTOH, if we go with @jkarni idea of regexp captures than, `servant-regex-capture` might make sense as a single package. Tough choices.


```hs
-- | A type for quick'n'dirty capture / query param matching.
--
-- @n@ stands for capture groups to extract
--
-- ==== Example
--
-- @
-- type API = "file" :> Capture "filename" (RE "(.*)\\.json") :> Get '[JSON] MyType
-- @
newtype RE n = RE { matchGroups :: Vec n Text }
```

basvandijk:
I need something like this however for my use case I would prefer that Ext is polymorphic in the filename as in:

```haskell
newtype Ext (ext :: Symbol) a = Ext {getBase :: a}
```

```haskell
instance (KnownSymbol ext, ToHttpApiData a) => ToHttpApiData (Ext ext a) where ...
```

So that I can define the following for example:

```haskell
type MyApi = "file" :> Capture "filename" (Ext "jpg" UTCTime) :> Raw
```

Does that make sense?

phadej:
@basvandijk have you seen: https://github.com/haskell-servant/servant-contrib/blob/master/servant-contrib/src/Servant/Contrib/RegexHttpApiData.hs

Consensus here is, that we are unluckily to add this combinator to `servant` as it's a) quite easy to make/copy&paste yourself b) many choices to do.

basvandijk:
@phadej thanks for the heads up. I'll copy&paste&adapt the code then.

phadej:
FWIW: it's not even `servant` related, as defining your `Ext` or `RE` needs only `http-api-data`.

phadej:
Closing; the regex code is in `servant-contrib`

domenkozar:
Just as a reference for anyone else ending up here: another option is to use `newtype` businiess-like wrappers and define your own url parsing as per http://haskell-servant.readthedocs.io/en/stable/tutorial/Server.html#the-fromhttpapidata-tohttpapidata-classes

PR #284, opened 2015-12-03
Relax upper version bound for http-api-data ...

kantp:
... in order to include http-api-data-0.2.1, which is currently in LTS
Haskell (lts-3.16).


jkarni:
Thanks (and hey!)!

Issue #285, opened 2015-12-05
VanillaJS javascript generator doesn't set "Accept" header correctly


utdemir:
It currently sends "text/plain", and this results on "invalid media type" error on servant endpoint generated by the same API.

arianvp:
Could you elaborate a bit here?  The generator generates client-side code and HTTP Clients don't set the `Content-Type` header. They set the `Accept` header.   Could you point out what is exactly going wrong where?

utdemir:
@arianvp Here is a minimal working example:

```haskell
module Main where
--------------------------------------------------------------------------------
import           Control.Monad.Trans.Class
import           Data.Proxy
import           Data.Text
import           Network.Wai.Handler.Warp
import           Servant
import           Servant.JS
--------------------------------------------------------------------------------

type API = "print" :> ReqBody '[JSON] Int :> Get '[JSON] ()

api :: Proxy API
api = Proxy

server :: Server API
server = lift . print

main :: IO ()
main = do
  putStrLn . unpack $ jsForAPI api vanillaJS
  run 2222 $ serve api server
```
Now try requesting "getPrint(12)" with the generated API:

![2015-12-06-130809_462x305_scrot](https://cloud.githubusercontent.com/assets/928084/11613019/6fbce244-9c1a-11e5-8630-1cb8e1c9c613.png)

![2015-12-06-130758_546x441_scrot](https://cloud.githubusercontent.com/assets/928084/11613020/6fbd9784-9c1a-11e5-8863-efc4ba9ceb5a.png)

As you can see, the server gives "HTTP415 Unsupported Media Type" error. But you're right, it looks like "Accept" header, not "Content-Type" header is wrong. It should be "json" instead of "html/xml" there I think. 

JQuery code generator works here because of the line here, I think: https://github.com/haskell-servant/servant/blob/master/servant-js/src/Servant/JS/JQuery.hs#L67

arianvp:
It's actually because jQuery automatically infers the `dataType` parameter and the `accepts` parameter is based on that. http://api.jquery.com/jQuery.ajax/.  The `contentType` parameter is only used in `POST`s.

Setting the accepts header explicitly would probably fix this.  (It's also cleaner to explicitly set it in the JQuery code)

Furthermore, maybe this is a good time to discuss supporting more than just JSON. Because we could probably infer other  clients that communicate in other content-types too.  We don't really have a mechanism for this in place at current time but I think it would be neat to explore.

[...]

PR #287, opened 2015-12-06
WIP: Fix issue #285

arianvp:
This should fix issue #285 . 

I'm wondering if we could write some sort of regression test suit using `nodejs` or http://phantomjs.org/ to catch these kind of bugs.


alpmestan:
We definitely could should have more tests for this yes =)

Also, if some API doesn't support JSON then the devs wouldn't be able to use `servant-js`. If anyone wants to try and make more content types work, that would be very appreciated!

jkarni:
Should we merge this for now and add tests as a separate story?

arianvp:
Yes seems good.

jkarni:
Thanks!

Issue #127, opened 2015-06-18
servant-docs: pretty-printing content


mf59816:
The content in the servant documentation is printed out with `MimeRender`, which is intentionallly whitespace-free.  In documentation, we would like to call pretty-printing variants of the underlying functions, though.

I think I have a work-around for this that works reasonably well: https://github.com/liqd/thentos/issues/155

I am opening this ticket to express that this should be fixed in servant-docs propertly.  As something close to my work-around, I could see a patch that provides an extra function

```haskell
docs' :: HasDocs layout => [(M.MimeType, ByteString -> ByteString)] -> Proxy layout -> API
```

That allows people to pass their own pretty-printers explicitly.  If you would like me to I am happy to provide a pull request, but this is clearly not in the spirit of servant.  Perhaps you have a better idea?

alpmestan:
Another thing you could do is define a `PrettyJSON` content type that uses `aeson-pretty` or something to encode values in a human-friendly JSON format. You can then substitute `PrettyJSON` for `JSON` in your API type (programmatically -- with a type families -- would be the best option I think) and use *that* API type to generate your docs.

``` haskell
data PrettyJSON

instance Accept PrettyJSON where
  accept _ = "application" // "json"

instance ToJSON a => MimeRender PrettyJSON a where
  mimeRender _ val = encodePretty val

type family Prettify api :: *
-- define it so that it looks for Get, Post, etc but also ReqBody, and replaces
-- any occurence of "JSON" by "PrettyJSON" in the content type lists.

prettified :: Proxy api -> Proxy (Prettify api)
prettified Proxy = Proxy

prettyDocs :: Proxy api -> API
prettyDocs = docsFor . prettified
```

jkarni:
I like both suggested approaches quite a bit.

jkarni:
Done in #289 .

tfc:
Hey there,

i got this to work with `docs` and `docsWithIntros`, but not with `docsWith` where i also provide extrainfos etc.: The compiler does then complain that it can't match the `JSON` with the `PrettyJSON` part.
I am not educated enough about type families etc., is there a fix/workaround for that?

alpmestan:
@tfc Is there any chance you could reproduce your problem on a self-contained example? Or share a link to the code in question if it's publicly available?

tfc:
@alpmestan i figured it out an hour ago.

[...]

PR #289, opened 2015-12-10
Add 'pretty' to servant-docs

mitchellwrosen:
Here's one solution to #127. 

Adds one function `pretty :: Proxy api -> Proxy (Pretty api)` to get pretty-printed JSON in generated documentation.

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -26,6 +26,8 @@ import           Control.Arrow              (second)
 import           Control.Lens               (makeLenses, over, traversed, (%~),
                                              (&), (.~), (<>~), (^.), (|>))
 import qualified Control.Monad.Omega        as Omega
+import           Data.Aeson                 (ToJSON(..))
+import           Data.Aeson.Encode.Pretty   (encodePretty)
 import           Data.ByteString.Conversion (ToByteString, toByteString)
 import           Data.ByteString.Lazy.Char8 (ByteString)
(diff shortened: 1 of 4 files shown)


jkarni:
Neat. The only issue I see here is that the `Pretty` type family is not open, so people can't add another pretty-printer for other content-types. That said, they *can* just writer another type family, and moreover JSON prettifying probably covers the vast majority of cases.

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -367,6 +369,37 @@ docsWith opts intros (ExtraInfo endpoints) p =
 docsWithIntros :: HasDocs layout => [DocIntro] -> Proxy layout -> API
 docsWithIntros intros = docsWith defaultDocOptions intros mempty
 
+-- | Prettify generated JSON documentation.
```
jkarni: Could this go in a separate module (e.g. `Docs/Internal/Pretty.hs`), and rexported? I tend to think `Internal.hs` is already getting too big, and this is pretty self-contained.

mitchellwrosen:
Re: open type family, I believe I wouldn't be able to have a catch-all clause at the bottom like

```haskell
type instance Pretty x = x
```

because this would overlap with other instances. So, I think there'd be many more boilerplate instances that cover the other types that people put in their servant API types. And now that I think about it, I'm not sure it'd even be possible to write an instance for all types of kind Symbol, for example.

I could be wrong, type programming is not my forte :)

jkarni:
Fair points! Thanks!

PR #304, opened 2015-12-22
MimeUnrender and MimeRender instances for Cassava

luke-clifton:
This allows the same API type to be used for `serve` and `client`.


alpmestan:
As evidenced by [this](https://travis-ci.org/haskell-servant/servant/jobs/98254860) build failure, you need to do a conditional import of `Control.Applicative`. Once the build is fixed, LGTM. @jkarni ?

jkarni:
Thanks!

PR #307, opened 2015-12-26
Authentication (v3.11)

jkarni:
This allows combinator instances to receive dynamic data that isn't
        constrained by the interface of Delayed etc.


jkarni:
Not ready yet - just putting it here for reference.

jkarni:
Rebased against #276 (which should probably be merged/reviewed first).

servant-server/src/Servant/Server/Internal/Config.hs
```diff
@@ -0,0 +1,76 @@
...
+
+instance Eq (Config '[]) where
+    _ == _ = True
+instance (Eq a, Eq (Config as)) => Eq (Config (a ' : as)) where
+    ConsConfig x1 y1 == ConsConfig x2 y2 = x1 == x2 && y1 == y2
+
+instance NFData (Config '[]) where
+    rnf EmptyConfig = ()
+instance (NFData a, NFData (Config as)) => NFData (Config (a ': as)) where
+    rnf (x `ConsConfig` ys) = rnf x `seq` rnf ys
```
soenkehahn: I **think** you need another ```... `seq` ()``` at the end.

servant-server/src/Servant/Server/Internal/Config.hs
```diff
@@ -0,0 +1,76 @@
...
+    ConfigEntry a >>= f = f a
+
+-- | The entire configuration.
+data Config a where
+    EmptyConfig :: Config '[]
+    ConsConfig :: x -> Config xs -> Config (x ': xs)
+
+instance Eq (Config '[]) where
+    _ == _ = True
+instance (Eq a, Eq (Config as)) => Eq (Config (a ' : as)) where
```
soenkehahn: Weird space: `a ' : as`.

jkarni:
I quite like the general auth! (Though I still think it should be a separate PR :) )

soenkehahn:
I think that just adding config should be a separate PR. I created https://github.com/haskell-servant/servant/tree/shahn/config to work on the config stuff (without the auth stuff on top of it).

aaronlevin:
@jkarni @soenkehahn so, one thing about General auth is that the `HasClient` instance in `servant-client` is difficult to write because we don't know what data is required to add to a request (i.e. in the old version I had a class like `AuthenticateRequest` with a method `authReq :: a -> Req -> Req`).

One solution would be to add another type param:

```haskell
data AuthProtect (tag :: k) (reqData :: *) (usr :: *)
```

I feel good about this, but wanted your thoughts. Usually the types that appear in a servant `API` are reflective of server handlers, but this would mean the API type refers to both the client and server:

```haskell
type API = AuthProtect "foo" OauthToken User :> "api" :> ...
```

Thoughts?

aaronlevin:
nm, I think I have a solution.

aaronlevin:
So, I kind of had a revelation of sorts. For all these auth combinators (including `BasicAuth`), we were placing `servant-server` specific types in them. For example:

```haskell
type API = BasicAuth "foo-realm" TypeOnlyServantServerCaresAbout :> "api" :> ...
```

So, I removed the `user` type from `servant-server` and instead we have an open type family that server authors use. 

I came on this idea because I needed `servant-client` specific types and didn't have access to them. Unfortunately, in the `servant-client` case, I had to use a data family.

jkarni:
> So, I kind of had a revelation of sorts. For all these auth combinators (including BasicAuth), we were placing servant-server specific types in them.

Good point. In theory it's possible to supply types from the `Config`, but I'll need to think a little more about whether that's a good idea.

servant-server/src/Servant/Server/Internal/Auth.hs
```diff
@@ -0,0 +1,75 @@
...
+mkAuthHandler :: (r -> ExceptT ServantErr IO usr) -> AuthHandler r usr
+mkAuthHandler = AuthHandler
+
+-- | The result of authentication/authorization
+data BasicAuthResult usr
+  = Unauthorized
+  | BadPassword
+  | NoSuchUser
+  | Authorized usr
+  deriving (Eq, Show, Read, Generic, Typeable, Functor)
```
fizruk: Should this be under `-- * Basic Auth` section?

aaronlevin: Thanks. Will update when `Config` settles!

fizruk:
What is the current approach to optional authentication?
I mean some API endpoints might not require authentication but depend on whether user has been identified or not.

Would it probably make sense to explicitly separate authentication and authorization?
Authentication combinators will then always wrap authentication result in `Maybe`
(or in some more complicated type to remember authentication error).
Might look like this:

```haskell
type MyAPI = AuthProtect "cookie-auth" :> RequireAuth "cookie-auth" :> ...

handler :: Maybe User -- came from AuthProtect
        -> User       -- came from RequireAuth
        -> ...
handler _ user = ...
```

Or (if `Required` combinator ever comes through) like this:

```haskell
type MyAPI = Required (AuthProtect "cookie-auth") :> ...

handler :: User -> ...
handler user = ...
```

I also think some solution to this problem was previously present in this PR in the form of `AuthPolicy`.
Not sure if it is/was entirely relevant.

aaronlevin:
@fizruk the current approach is to separate `authentication` from `authorization`.

The `AuthPolicy` was not related to this current concern. It was basically allowing downstream handlers to explicitly handle authentication errors, rather than `servant` throwing an `err401` or `err403`. Basically your handler became a function of `Either e User` instead of `User`, but we don't need this anymore as the general auth combinator is much more flexible.

aaronlevin:
@fizruk we will tackle `authorization` once the dust settles on authentication, but I have a feeling it will be very similar.

fizruk:
> the current approach is to separate `authentication` from `authorization`.

It seems that currently to have an optional authentication I have to copy the `AuthProtect` code with altered instances (to return/allow `Nothing` and not reject the request when authentication fails).
That is, as I see it right now authentication in this PR *is also authorization* since it does not allow unauthenticated users to access the API behind `AuthProtect`.

aaronlevin:
@fizruk you could easily create an `AuthProtect`'d endpoint that returns a `Maybe User`.

aaronlevin:
@fizruk `AuthProtect` is very general. You (currently) use a type family to define the return type in the `HasServer` handler (which could be `User` or `Maybe User` or `Either UnauthError User`), and then supply a function `Request -> ExceptT IO ServantErr (Maybe User)` to `Config`.

fizruk:
@aaronlevin right, I forgot it is a general combinator!

Still, I don't think `AuthProtect` should allow rejecting requests upon failed authentication.
You might want to reject request when authentication failed due to invalid authentication params (e.g. parse error), but not when user simply did not provide authentication data (just like with `QueryParam`).

Also my point about code duplication is still true for `BasicAuth`.

aaronlevin:
@fizruk if the user does not provide authentication data, then they should get a `403 Forbbiden` response. I don't disagree with what you're saying, but I do think we can separate authentication and authorization.

Authorization can be very nuance and complicated. Some authorization schemes involve roles, and more involved ones specify per-route access (for example, a user should only be able to access their own urls at `/api/:user_id`).

So, to me Authentication is about verifying the identity of a client/user. Authorization is about determining whether that user is able to access the resource or not. In the case of basic authentication these are very much intertwined, but in the case of our `AuthProtect` combinator, it's geared towards extracting the identity of a client and not determining whether that user has access to a path.

Adding authorization will require similar steps to authentication:

- updating `Delayed` to run authorization code *after* authentication code
- providing route information to authorization code (not sure how to do this yet)
- introducing combinators to run authorization (we may have multiple authorizations per an authentication scope)
- more that I'm not thinking about.

fizruk:
> if the user does not provide authentication data, then they should get a 403 Forbbiden response. I don't disagree with what you're saying, but I do think we can separate authentication and authorization.

Not necessarily, whether I want anonymous users to access API or not is a matter of authorization, not authentication. I think we should draw that line between authentication and authorization firmly now, when we do not have any code to break and users to confuse.

> Authorization can be very nuance and complicated. Some authorization schemes involve roles, and more involved ones specify per-route access (for example, a user should only be able to access their own urls at /api/:user_id).

Right, but we should not hardwire some (even little) authorization mechanics into authentication combinators just because we don't know what authorization will look like.

> So, to me Authentication is about verifying the identity of a client/user.

Exactly! And anonymous user is still a user. If he/she did not provide any authentication credentials, we have nothing to verify!

>  Authorization is about determining whether that user is able to access the resource or not.

This can be applied to anonymous users as well. Some resources might be accessed by any user and some — only by authenticated ones.

One might argue that is some API is available for anonymous users, it should not be behind any authentication combinator. But this is not true. The information about the resource might be different for an anonymous user. E.g. search results for logged user may be different from the search results for anonymous user because we don't have a search history for an anonymous user to personalize search results. But we still want to provide search results.

Just to repeat myself: anonymous users should not be rejected by any authentication combinator!

P.S. When we decide this little question, we should add documentation on why authentication/authorization is designed as it is.

aaronlevin:
@fizruk one subtlety is that the availability of authentication data varies across authentication schemes. For example, in `OAuth` you have a token, in `Basic` you have a username/password, in `JWT` you have a jwt token, in `HMAC` you have a signature, etc. so, to have a general way of saying "authentication data is not present" is challenging.

What can be done is the following two steps:

### Step 1: Amend Authentication

- we can re-add the policy code. This would allow authentication combinators to be tagged by policy.
- we can add a `NoAuthenticationPresent` datatype
- for combinators using `Policy` they can return `Either NoAuthenticatoinPresent User` or `User` depending on the policy

However, I don't see any way of amending the general `Auth` combinator to do this since the guard is so broad (it's just `Request -> ExceptT IO ServantErr a`).

### Step 2: Introduce Authorization

For authorization we need the path info when handling the request. I think this is a broader change and will require a few iterations. I think we need to do this in the next version and not in this version.

fizruk:
@aaronlevin I agree that detecting no authentication data is scheme-dependent.

> we can re-add the policy code

I would leave policy out until we decide about the authorization.

> However, I don't see any way of amending the general Auth combinator to do this since the guard is so broad (it's just `Request -> ExceptT IO ServantErr a`).

How about altering the guard itself to be `Request -> ExceptT ServantErr IO (Maybe a)`?
IIUC this would mean exactly that detecting "no authentication data present" (returning `Nothing`) is scheme-specific.
Alternatively we can use `Either NoAuthenticatoinPresent` or a custom data type like this instead of `Maybe`:

```haskell
data AuthIdenity user
  = NoAuthData
  | AuthIdentity user
```

> For authorization we need the path info when handling the request. I think this is a broader change and will require a few iterations. I think we need to do this in the next version and not in this version.

I am not sure what you mean by "path info", but I agree that this can be handled after authentication has been merged.

aaronlevin:
@fizruk what I mean by path info is the following: suppose we have a request to `/api/:user_id/game/:game_id/score` . Once we authenticate we'll have a `User` object. However, in order to know if this is a valid URL for the request, we need to know the `game_id` *and* the `user_id` to verify it's the same user making the request and that the user has access to the game.

Further, we won't actually extract that data until the route is fully matched, so the full request context will have to work it's way into `Delayed`. It's doable, but not trivial.

aaronlevin:
@fizruk my personal vote, much to your chagrin, is that we need to cut the `0.5` release and I don't want to spend hours analysing what clients *may* want insofar as an API. As it stands, we have reached our goals, which were a simple example (basic authentication) of what a combinator might look like, and we have a really broad / general combinator for those that want super customized / weird authentication schemes.

The key to unlocking tihs was `Config`, and now that we've almost settled on this, the above examples will empower people to start developing their own combinators, and eventually we'll settle on a design that makes sense.

/cc @jkarni @soenkehahn

aaronlevin:
@fizruk I want to be clear, in case it didn't come across, that I appreciate your feedback and challenges on the design! It's really important, but I think we have a decent starting point and, being that `servant` is still only `v0.5`, users should still expect some fluctuations in the api design as we move towards `1.0` :)

fizruk:
> @fizruk what I mean by path info is the following: suppose we have a request to `/api/:user_id/game/:game_id/score` . Once we authenticate we'll have a `User` object. However, in order to know if this is a valid URL for the request, we need to know the `game_id` *and* the `user_id` to verify it's the same user making the request and that the user has access to the game.

Right, in general authorization might use some heavy business logic and will need many of the request parameters. So it probably should be evaluated right before the handler.

> Further, we won't actually extract that data until the route is fully matched, so the full request context will have to work it's way into Delayed. It's doable, but not trivial.

I see. We can probably do better with something like `Context` as I have mentioned in https://github.com/haskell-servant/servant/pull/321#issuecomment-170625407.
But implementation details are still not obvious.

> @fizruk my personal vote, much to your chagrin, is that we need to cut the 0.5 release and I don't want to spend hours analysing what clients may want insofar as an API. As it stands, we have reached our goals, which were a simple example (basic authentication) of what a combinator might look like, and we have a really broad / general combinator for those that want super customized / weird authentication schemes.

I understand your position and agree that it has been too long and something should already be released. Besides having authentication combinators allow anonymous users is an arguably a minor issue, which can be easily fixed in the next release.

In fact I would really like to see minor releases more frequently, especially since `servant` is still `0.*`.
But that's a topic for another discussion :)

> @fizruk I want to be clear, in case it didn't come across, that I appreciate your feedback and challenges on the design! It's really important, but I think we have a decent starting point and, being that servant is still only v0.5, users should still expect some fluctuations in the api design as we move towards 1.0 :)

Okay then, I suggest we move this authorization-related discussion into a separate issue(s).

aaronlevin:
@soenkehahn how did you fix the failing Travis stuff?

soenkehahn:
IIRC by allowing both `deepseq-1.3.*` and `1.4.*`. However in `shahn/config2` this dependency is removed anyways.

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -56,25 +49,54 @@ newtype PublicData = PublicData { somedata :: Text }
...
+-- | Our API, with auth-protection
+type API = "private" :> AuthProtect "cookie-auth" :> PrivateAPI
+      :<|> "public"  :> PublicAPI
+
+-- | A value holding our type-level API
 api :: Proxy API
 api = Proxy
 
+-- | We need to specify the data returned after authentication
+type instance AuthReturnType (AuthProtect "cookie-auth") = User
```
aaronlevin: @jkarni @soenkehahn I wouldn't mind getting some feedback on this design decision. The auth combinators don't carry any `-server` or `-client` specific data type in them. Therefore, we use the type family `AuthReturnType` to specify what would have usually been encoded as `usr`.

I ended up having to do this kind of thing to support `servant-client`. The previous datatype for auth, as an example, was: `AuthProtect (tag :: Symbol) usr` and when implementing the client I had to allow users to define a type they would use to add request data to the request. My two options were: use a type family to specify the data type or add something to `AuthProtect` like: `AuthProtect (tag :: symbol) clientData usr`

I ended up choosing the type family as it made for a cleaner API. But then I thought we should apply the same principal to `-server`.

I then removed any `usr` types from both `AuthProtect` and `BasicAuth`. Would love your thoughts on this.

soenkehahn: For `BasicAuth` you don't need a `usr` parameter or type families, right? Since we know what parameter types are going to be used for the server and the client.

aaronlevin: We allow for some more flexibility with Basic Auth, so that once you check the username/password, you can return a specific `User` datatype. I can remove that functionality if needed. If that functionality is removed, then we don't need the type family.

aaronlevin: (* correction: if that functionality is removed, we won't need the type family for BasicAuth. We will need it still for `AuthProtect`)

aaronlevin: To be more clear. The choice is between:

```haskell
type API = AuthProtect "foo" ClientData ServerData :> "private" :> Get '[JSON] Account
```

and

```haskell
type API = AuthProtect "foo" :> "private" :> Get '[JSON] Account

-- for the server
type instance AuthReturnType (AuthProtect "foo") = ServerData
-- for the client
type instance AuthClientData (AuthProtect "foo") = ClientData
```

soenkehahn: Good question. I think if the general `AuthProtect` is meant as a nice example for how to write auth combinators, I would keep it as simple as possible and not use type families and just hard-code some example types in the `HasServer` and `HasClient` instances. Is it?

soenkehahn:
I would prefer one PR for `BasicAuth` alone. This PR has gotten somewhat messy, even without having both the general auth combinator and `BasicAuth`. @aaronlevin: do you want to open a new one?

soenkehahn:
(Having separate and cleaner PRs for `BasicAuth` and `AuthProtected` would also make it easier to discuss design choices.)

aaronlevin:
@soenkehahn sure, I can do that tonight.

My only reluctance with putting the types in `AuthProtect` is that it forces api authors to think about data specific and relevant only for clients. They can always put `()` in there or paramaterize their API with a type. Though, perhaps that's less cognitive overload than a type family.

aaronlevin:
closed in favour of #341

PR #310, opened 2015-12-27
Fix #294

### LINKED ISSUE #294: VanillaJS js generator doesn't handle "204 No content" answer from server
When the server returns with "204 No content" (this happens when API is defined as `Get '[JSON] ()`), and therefore with an empty request body, the generated JS client tries to `JSON.parse('')` and gives syntax error.

Example:

```
{-# LANGUAGE DataKinds     #-}                                                                                                                                  

module Main where                                                                                                                                               

--------------------------------------------------------------------------------                                                                                
import           Control.Monad.Trans.Class
import           Data.Proxy
import           Data.Text
import           Network.Wai.Handler.Warp
import           Servant
import           Servant.JS
--------------------------------------------------------------------------------                                                                                

type API = "print" :> Capture "num" Int :> Get '[JSON] ()

api :: Proxy API
api = Proxy

server :: Server API
server = lift . print

main :: IO ()
main = do
  putStrLn . unpack $ jsForAPI api vanillaJS
  run 2222 $ serve api server
```

On this case, server correctly returns "204 Not modified" without a response body, but the client fails:

2015-12-16-115210_875x368_scrot

arianvp:
Fixes #294 

I still need to test if it actually works now. But I think it should fix it.


servant-js/src/Servant/JS/Vanilla.hs
```diff
@@ -37,10 +37,13 @@ generateVanillaJSWith opts req = "\n" <>
  <> (if isJust (req ^. reqBody) then "  xhr.setRequestHeader(\"Content-Type\",\"application/json\");\n" else "")
  <> "  xhr.onreadystatechange = function (e) {\n"
  <> "    if (xhr.readyState == 4) {\n"
+ <> "      if (xhr.status == 204) {\n"
```
jkarni: Or 205.

arianvp:
Okay this can be merged I think.

jkarni:
Great, thanks!

PR #315, opened 2016-01-04
Remove Overlapping pragma noise using CPP

luigy:
:lipstick: looks nicer


jkarni:
Looks like we'll need to add ` include/*.h` to all the `extra-source-files`.

jkarni:
(And thanks!)

PR #320, opened 2016-01-07
add servant/.ghci

soenkehahn:


jkarni:
Why only `servant/`?

soenkehahn:
Because I was lazy. :|

PR #321, opened 2016-01-08
Shahn/config

soenkehahn:


aaronlevin:
any idea why the build is failing? I've not used `tinc` before and the error is really opaque. It only fails on `7.8.4`.

soenkehahn:
@aaronlevin: I suspect that there's a dependency that doesn't build cleanly with `ghc-7.8` anymore. But that's just guesswork.

aaronlevin:
@jkarni @soenkehahn Can we rename the `.:` operator to `.:.`? As is, it conflicts with `Data.Aeson.(.:)` and since many apis these days encode data in JSON, qualifying the import has been annoying.

soenkehahn:
If we can make that newtype idea work, we might also be able to use `:.`.

soenkehahn:
This is still WIP. We need to:
 - [ ] try out whether the newtype idea works out,
 - [ ] figure out what best to export from which module.

aaronlevin:
What's the `newtype` proposal? When I did the van Laarhoven Free Monad stuff, I just indexed the HList by the type you want to find. So, for exmaple:

```haskell
myConfig :: Config '[ BasicAuth "foo" , Char ]
myConfig = mkBasicAuth foo bar .:. 'x' .:. EmptyConfig
```

But if you want to have two `Char`s in your `Config`, then you need to newtype wrap.

```haskell
newtype OtherChar = OtherChar { unChar :: Char }

myConfig :: Config '[ BasicAuth "foo" , Char , OtherChar ]
myConfig = mkBasicAuth foo bar .:. 'x' .:. OtherChar 'y' .:. EmptyConfig
```

I haven't tested if this works, but it's what I did with the van laarhoven stuff and that seemed to work fine.

soenkehahn:
We want to allow to have the same type (modulo newtype wrapping) twice in the list and use it for the same combinator. So there has to be either
- a class like `IsBasicAuthConfig Newtype` for every combinator or
- combinators use something like `Coercible Newtype Originaltype`.

But yes, I hope this'll just work fine.

aaronlevin:
@soenkehahn so you want to use the same type twice in the list *without* having to newtype wrap?

soenkehahn:
No, **with** newtype wrapping. But the unwrapping in the combinators needs to use a class.

aaronlevin:
@soenkehahn oh, I see. You don't want clients to have to unpack the newtype themselves? Do we anticipate this (having two of the same types in Config) being a common case?

soenkehahn:
No, I don't think it'll be that common.

aaronlevin:
@soenkehahn so why not just force the clients to unpack their newtypes on their own? That would obviate any need for weird typeclasses and we wouldn't need to tag all the types in `Config`. This is the solution I came to with the van Laarhoven stuff.

soenkehahn:
Here's an example of the problem we're trying to solve: Someone wants to use two `BasicAuth` combinators in different places in one API. They go against different user dbs, so we have to somehow pass different values from the `Config` to the two combinators. So `BasicAuth` has to somehow allow to specify which one you want in both places. Currently (in this branch) that's being done by tags. But I'd like to try out to do so by newtypes. But then `BasicAuth` has to unwrap the newtypes. (I hope that clarifies the issue a bit.)

I hope I can come up with an implementation for this today. Then you can take a look. I'll make sure to include a test-case for the exact problem I have in mind.

Btw, I'm not convinced yet that the newtypes thing is better than tagging or yet a different approach. We'll have to see.

aaronlevin:
@soenkehahn I made a branch with the changes I was suggesting:

https://github.com/haskell-servant/servant/blob/aaronlevin/config/servant-server/src/Servant/Server/Internal/Config.hs

Here are the corresponding tests which use the `newtype`ing:

https://github.com/haskell-servant/servant/blob/aaronlevin/config/servant-server/test/Servant/Server/Internal/ConfigSpec.hs

As far as I can tell, the only downside is that when you use `getConcifgEntry` you need to specify the type, as you can see here:

https://github.com/haskell-servant/servant/blob/aaronlevin/config/servant-server/test/Servant/Server/Internal/ConfigSpec.hs#L34

I'm not 100% sold on this idea, but it is very simple.

soenkehahn:
I just pushed some test-cases. Here's the one I was thinking about: https://github.com/haskell-servant/servant/blob/shahn/config/servant-server/test/Servant/Server/UsingConfigSpec.hs#L73

aaronlevin:
@soenkehahn ok, so taking from your tests, see below (I couldn't commit this because it would require a change to `route`):

```haskell
{-# LANGUAGE CPP                        #-}
{-# LANGUAGE DataKinds                  #-}
{-# LANGUAGE FlexibleInstances          #-}
{-# LANGUAGE FunctionalDependencies     #-}
{-# LANGUAGE GADTs                      #-}
{-# LANGUAGE KindSignatures             #-}
{-# LANGUAGE MultiParamTypeClasses      #-}
{-# LANGUAGE ScopedTypeVariables        #-}
{-# LANGUAGE PolyKinds                  #-}
{-# LANGUAGE TypeOperators              #-}
{-# LANGUAGE UndecidableInstances       #-}

#include "overlapping-compat.h"

module Servant.Server.UsingOtherConfigSpec where

import           Network.Wai
import           Test.Hspec (Spec, describe, it)
import           Test.Hspec.Wai

import           Servant
import           Servant.Server.Internal.Config
import           Servant.Server.Internal.RoutingApplication


-- * Config

-- | The entire configuration.
data Config2 a where
    EmptyConfig2 :: Config2 '[]
    ConsConfig2 :: x -> Config2 xs -> Config2 (x ': xs)

instance Show (Config2 '[]) where
  show EmptyConfig2 = "EmptyConfig2"
instance (Show a, Show (Config2 as)) => Show (Config2 (a ': as)) where
  showsPrec outerPrecedence (ConsConfig2 a as) =
    showParen (outerPrecedence > 5) $
      shows a . showString " .:. " . shows as

instance Eq (Config2 '[]) where
    _ == _ = True
instance (Eq a, Eq (Config2 as)) => Eq (Config2 (a ': as)) where
    ConsConfig2 x1 y1 == ConsConfig2 x2 y2 = x1 == x2 && y1 == y2

(.:.) :: x -> Config2 xs -> Config2 (x ': xs)
e .:. cfg = ConsConfig2 e cfg
infixr 5 .:.

class HasConfig2Entry (cfg :: [*]) (val :: *) | cfg -> val where
    getConfig2Entry :: Config2 cfg -> val

instance OVERLAPPABLE_
         HasConfig2Entry xs val => HasConfig2Entry (notIt ': xs) val where
    getConfig2Entry (ConsConfig2 _ xs) = getConfig2Entry xs

instance OVERLAPPABLE_
         HasConfig2Entry (val ': xs) val where
    getConfig2Entry (ConsConfig2 x _) = x

-- * Custom combinator

data CustomCombinator (a :: *)


instance forall subApi (c :: [*]) a .
  (HasServer subApi) =>
  HasServer (CustomCombinator a :> subApi) where

  type ServerT (CustomCombinator a :> subApi) m =
    a -> ServerT subApi m
  type HasCfg2 (CustomCombinator a :> subApi) c =
    (HasConfig2Entry c a, HasCfg2 subApi c)

  route Proxy config delayed =
    route subProxy config (fmap (inject config) delayed :: Delayed (Server subApi))
    where
      subProxy :: Proxy subApi
      subProxy = Proxy

      inject config f = f (getConfig2Entry config :: a)

-- * API

data CustomConfig = CustomConfig String

type OneEntryAPI =
  CustomCombinator CustomConfig :> Get '[JSON] String

testServer :: Server OneEntryAPI
testServer (CustomConfig s) = return s

oneEntryApp :: Application
oneEntryApp =
  serve (Proxy :: Proxy OneEntryAPI) config testServer
  where
    config :: Config '[CustomConfig]
    config = CustomConfig "configValue" .:. EmptyConfig



type OneEntryTwiceAPI =
  "foo" :> CustomCombinator CustomConfig :> Get '[JSON] String :<|>
  "bar" :> CustomCombinator CustomConfig :> Get '[JSON] String

oneEntryTwiceApp :: Application
oneEntryTwiceApp = serve (Proxy :: Proxy OneEntryTwiceAPI) config $
  testServer :<|>
  testServer
  where
    config :: Config '[CustomConfig]
    config = CustomConfig "configValueTwice" .:. EmptyConfig

-- newtype wrap to provide two differnet configs.
newtype WrappedCustomConfig = WrappedCustomConfig {unWrapCustomConfig ::  CustomConfig }

type TwoDifferentEntries =
  "foo" :> CustomCombinator CustomConfig        :> Get '[JSON] String :<|>
  "bar" :> CustomCombinator WrappedCustomConfig :> Get '[JSON] String

twoDifferentEntries :: Application
twoDifferentEntries = serve (Proxy :: Proxy TwoDifferentEntries) config $
  testServer :<|>
  (testServer . unWrapCustomConfig)
  where
    config :: Config '[CustomConfig, WrappedCustomConfig]
    config =
      CustomConfig "firstConfigValue" .:.
      WrappedCustomConfig (CustomConfig "secondConfigValue") .:.
      EmptyConfig

-- * tests

spec :: Spec
spec =
  describe "using Config in a custom combinator" $ do
    with (return oneEntryApp) $
      it "allows to retrieve a ConfigEntry" $
        get "/" `shouldRespondWith` "\"configValue\""

    with (return oneEntryTwiceApp) $
      it "allows to retrieve the same ConfigEntry twice" $ do
        get "/foo" `shouldRespondWith` "\"configValueTwice\""
        get "/bar" `shouldRespondWith` "\"configValueTwice\""

    with (return twoDifferentEntries) $
      it "allows to retrieve different ConfigEntries for the same combinator" $ do
        get "/foo" `shouldRespondWith` "\"firstConfigValue\""
        get "/bar" `shouldRespondWith` "\"secondConfigValue\""
```

aaronlevin:
oh, nm, let me push to my own branch.

soenkehahn:
It's working in my branch with newtypes now. I'll rename a few things...

aaronlevin:
@soenkehahn for comparison: https://github.com/haskell-servant/servant/blob/aaronlevin/config/servant-server/test/Servant/Server/UsingOtherConfigSpec.hs

soenkehahn:
But if the combinator wants to do something meaningful with the config entry (not just pass it to the handler) then that doesn't work like this. But I see now that my test-case didn't really demonstrate that. So the problem is this: What if the combinator needs the config entry to be of a certain type? E.g. some db connection? And the additional argument to the handler is also a fix type (e.g. some user id)? I don't see how you could work around a class like this: https://github.com/haskell-servant/servant/blob/shahn/config/servant-server/test/Servant/Server/UsingConfigSpec/CustomCombinator.hs#L22

soenkehahn:
I'm wondering if the newtype thing is the best idea. There are a few disadvantages:

- The types are quite difficult to understand. And some of the complexity stems from use-cases where we feel that they are not very common.
- When writing custom combinators it's easy to forget to allow for duplicate use. So while the newtypes approach allows for writing combinators that can be used multiple times with different config values, it's much more straight-forward (and easier) to write combinators that don't support that use-case.

I'm especially worried because we want to use this to pass config values to Auth combinators. When picking the right configuration values for authentication it's very crucial to get it 100% right. I feel it should be easier to understand how everything works.

I don't have a good solution though.

(cc: @jkarni, @aaronlevin)

aaronlevin:
@soenkehahn yeah, so the downside to the `newtype` technique is that if you want to support multiple configurations your type combinator (e.g. `CustomCombinator`) must contain a type param where you will have your configurable types and their newtype wrappers.

The downside to the current technique is `Proxy` and `ConfigEntry` noise (I think renaming that to `TaggedEntry` might be better). I can't think of any other downsides.

tbh, I'm OK with either. I don't like the `Proxy` noise, but I far prefer it over random typeclassing with `Coercible` or something. I do like the simplicity of my solution, but I think if people are rolling out their own combinators, passing "tags" as "values" (via `Proxy`) as opposed to tagging your type combinators, might be a simpler mental model.

soenkehahn:
@aaronlevin: I still don't understand how your approach solves the problem I'm trying to solve. (If it did, it would be great!) In your branch (`aaronlevin/config`) currently the `CustomCombinator` cannot do anything with the config entry, because in the instance for `HasServer` it's just a polymorphic `a`. But what if the combinator wants to have e.g. an `MVar UserName User` and then pass a `User` to the handler? While still allowing the newtype wrapping to distinguish between different config entries.

aaronlevin:
@soenkehahn simple, just wrap `(a, a -> b)`:

See: https://github.com/haskell-servant/servant/blob/aaronlevin/config/servant-server/test/Servant/Server/UsingOtherConfigSpec.hs#L47

and here is a server: https://github.com/haskell-servant/servant/blob/aaronlevin/config/servant-server/test/Servant/Server/UsingOtherConfigSpec.hs#L118

I even used a wonky `MVar User` just for you ;)

alpmestan:
Could we index the config by name (`Symbol`) instead of by type?

soenkehahn:
@aaronlevin: Thanks for using an `MVar`. :) Yeah, that works, but `CustomArrow` is not really all that simple anymore.

@alpmestan: We did have an indexing mechanism before. Every `ConfigEntry` was tagged with a type (not restricted to `Symbols` though). It created a lot of noise.

This newtype-based idea is the attempt to come up with something where
- It's in principle possible to write combinators that allow distinguishing of config values,
- but where -- in the common case where a user of a combinator doesn't need that -- they don't have to think about it. 

I think it comes closer to that than either the tagged version or @aaronlevin's idea.

soenkehahn:
I now implemented a `Config` with *optionally* tagged entries:

- When you don't need to distinguish config entries you just write `value :. EmptyConfig`. No type annotations needed. (Except when `value` is polymorphic itself.)
- When you *do* want to distinguish config entries, you can tag them: `(Tag value :: Tagged "foo" ValueType) :. EmptyConfig`.

For complete examples take a look at the tests.

I think so far this is the best design. @aaronlevin, @jkarni, @alpmestan: Any comments?

(@alpmestan: Thanks for suggesting `Symbol`s. That gave me the idea of optional tagging.)

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -27,9 +27,10 @@ data AuthProtected
 
 instance HasServer rest => HasServer (AuthProtected :> rest) where
   type ServerT (AuthProtected :> rest) m = ServerT rest m
+  type HasCfg (AuthProtected :> rest) c = HasCfg rest c
 
-  route Proxy subserver = WithRequest $ \ request ->
-    route (Proxy :: Proxy rest) $ addAcceptCheck subserver $ cookieCheck request
+  route Proxy p subserver = WithRequest $ \ request ->
+    route (Proxy :: Proxy rest) p $ addAcceptCheck subserver $ cookieCheck request
```
fizruk: Shouldn't this be `cfg` instead of `p`?

aaronlevin:
@soenkehahn is it possible to get rid of the `Proxy` in `getConfigEntry`?  It would mean code would potentially change from:

```haskell
getConfigEntry (Proxy :: Proxy "bar") config
```

to

```haskell
getConfigEntry config :: MyBar
```

or

```haskell
getConfigEntry config :: Tagged "foo" Int
```

(if I understand things correctly)

aaronlevin:
Otherwise: :+1:

fizruk:
IIUC `Config` allows also writing to by custom combinators, right?
That is a combinator might substitute config when passing it to the sub API handler.

I think it can be quite useful at least for authorization purposes:

```haskell
type MyAPI = AuthProtect "cookie-auth" :> "admin" :> RequireRole 'Admin :> AdminAPI
```

If `Config` for this API contains a `Maybe User` entry with `Nothing` as initial/default value, then `AuthProtect` can "write" `Just user` before passing config further down the API.
`RequireRole` combinator might then process the user in config and check his/her role to allow only admin access to this endpoint.

A more interesting approach would be if we change `Config` type along the way.
In the above API `AuthProtect` might pass a config with `User` rather than `Maybe User`, changing the type of the config. Furthermore, `AuthProtect` might **add `User` as a new entry** to the config!

This is nice because we don't have to have `Nothing`s all over the initial config (which is passed to `serve`). The even nicer thing is that `RequireRole` will now *expect* a `User` and **won't typecheck** if we forget to add an authentication combinator before it.

[`spock`](http://hackage.haskell.org/package/Spock) (another Haskell web framework) has a very similar idea of `Context`s (well explained in [this post](https://www.spock.li/2015/08/23/taking_authentication_to_the_next_level.html)).
Perhaps we should seek wisdom from our colleagues' work!

I think nothing about `Config` has to be changed in the code, the ideas I have expressed should be possible to implement with the current code.

If you however find these ideas interesting, I think #307 might be refactored a bit to account for possible authorization combinators.

aaronlevin:
@fizruk the `Config` param allows static knowledge to be accessible in the `HasServer` instances. Not sure it makes sense to modify the config as a request is processed. 

However, it would definitely be interesting for an `Authentication` combinator to produce a `Maybe User` for a handler, and then a `Authorization` combinator amend that to just `User` (and throwing the right error in the case when authorization fail). Good idea.

fizruk:
> the `Config` param allows static knowledge to be accessible in the `HasServer` instances. Not sure it makes sense to modify the config as a request is processed.

Then maybe we should add `Context` for that sole purpose?

> However, it would definitely be interesting for an `Authentication` combinator to produce a `Maybe User` for a handler, and then a `Authorization` combinator amend that to just `User` (and throwing the right error in the case when authorization fail).

How would `Authorization` combinator amend the handler parameter? I can only think of something similar to `Required` combinator (as in `RequireRole Admin (AuthProtect "cookie-auth")`). This can be a solution, but it is quite limited compared to `Context`-based approach, since it requires all the authorization combinators to be wrapped around an authentication combinator. With `Context`-based approach we can ask for authentication once and then require different authorization levels for different subAPIs:

```haskell
type MyAPI = AuthProtect "cookie-auth" :>
  (    RequireAuth :> -- RequireAuth turns Maybe User into User
       (    "admin" :> RequireRole 'Admin :> AdminAPI  -- API for admins
       :<|> "users" :> UserAPI  -- API for any logged in users
       )
  :<|> "public" :> PublicAPI    -- API for everyone,
                                -- responses might still depend on
                                -- whether the user is logged in
  )
```

I imagine there might be more complex authorization combinators for real world business logic, which would not (easily) allow for a wrapped `Required`-style authorization combinators:

```haskell
type ResourceAPI
  = AuthProtect "cookie-auth"  -- optional authentication
                               -- (puts Maybe User into Context)
  :> RequireAuth  -- makes sure authentication succeeded
                  -- (replaces Maybe User with User)
  :> "resource"
  :> Capture "resource_id" ResourceId  -- puts resource ID in Context
                                       -- we can tag it with "resource_id"
                                       -- or even (Capture "resource_id")
                                       -- so that it can be disambiguated later
  :> RequirePermission 'READ ResourceId  -- extract resource ID from Context
                                        -- and check that user has READ permissions
                                        -- for this particular resource
  :> Get '[JSON] Resource
```

servant-server/src/Servant/Server/Internal/Config.hs
```diff
@@ -0,0 +1,59 @@
...
+  showsPrec outerPrecedence (a :. as) =
+    showParen (outerPrecedence > 5) $
+      shows a . showString " :. " . shows as
+
+instance Eq (Config '[]) where
+    _ == _ = True
+instance (Eq a, Eq (Config as)) => Eq (Config (a ': as)) where
+    x1 :. y1 == x2 :. y2 = x1 == x2 && y1 == y2
+
+newtype Tagged (tag :: Symbol) a = Tag a
```
jkarni: Any reason for not using `Data.Tagged`?

servant-server/test/Servant/Server/UsingConfigSpec/CustomCombinator.hs
```diff
@@ -0,0 +1,36 @@
...
+-- the config.
+module Servant.Server.UsingConfigSpec.CustomCombinator where
+
+import           Servant
+import           Servant.Server.Internal.Config
+import           Servant.Server.Internal.RoutingApplication
+
+data CustomCombinator (tag :: k)
+
+instance forall subApi (c :: [*]) tag .
```
jkarni: I don't think you need foralls for ScopedTypeVariables in instance declarations.

servant-server/test/Servant/Server/UsingConfigSpec/CustomCombinator.hs
```diff
@@ -0,0 +1,36 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | This is a custom combinator for S.S.UsingConfigSpec. It's split up into
+-- its own module to be able to test how exactly module import work when using
```
jkarni: Huh? I can't tell from the test what about module import is being tested.

servant-server/src/Servant/Server/Internal/Config.hs
```diff
@@ -0,0 +1,59 @@
...
+    EmptyConfig :: Config '[]
+    (:.) :: x -> Config xs -> Config (x ': xs)
+infixr 5 :.
+
+instance Show (Config '[]) where
+  show EmptyConfig = "EmptyConfig"
+instance (Show a, Show (Config as)) => Show (Config (a ': as)) where
+  showsPrec outerPrecedence (a :. as) =
+    showParen (outerPrecedence > 5) $
+      shows a . showString " :. " . shows as
```
jkarni: What about a `Read` instance?

servant-server/src/Servant/Server/Internal/Config.hs
```diff
@@ -0,0 +1,59 @@
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DeriveFoldable             #-}
+{-# LANGUAGE DeriveFunctor              #-}
```
jkarni: Seems like a lot of these are not being used.

servant-server/test/Servant/Server/UsingConfigSpec.hs
```diff
@@ -0,0 +1,71 @@
...
+      (Tag 'y' :: Tagged "bar" Char) :.
+      EmptyConfig
+
+-- * tests
+
+spec :: Spec
+spec = do
+  describe "using Config in a custom combinator" $ do
+    with (return oneEntryApp) $ do
+      it "allows to retrieve a ConfigEntry" $ do
```
jkarni: `to retrieve` ==> `retrieving`

servant-server/src/Servant/Server/Internal/Config.hs
jkarni: And less importantly, `Ord`. 

There is, I think, also only one sensible `Monoid` instance (delegating to all inner monoids), and that may serve some purpose to someone, but may not :). There is also a `Bounded` instance that makes sense, but I can't imagine a use-case for it.

servant-server/src/Servant/Server/Internal/Config.hs
jkarni: If there is a reason, you should derive a whole lot more instances.

servant-server/src/Servant/Server/Internal/Config.hs
jkarni: And `From/ToJSON`, now that I think of it. Those are very useful instances!

servant-server/servant-server.cabal
```diff
@@ -48,6 +49,7 @@ library
       , attoparsec         >= 0.12 && < 0.14
       , bytestring         >= 0.10 && < 0.11
       , containers         >= 0.5  && < 0.6
+      , deepseq            == 1.4.* || == 1.3.*
```
jkarni: We can remove this, I think

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -331,14 +332,14 @@ headerSpec = describe "Servant.API.Header" $ do
...
+    with (return (serve headerApi EmptyConfig expectsInt)) $ do
+        let delete' x = Test.Hspec.Wai.request methodDelete x [("MyHeader" ,"5")]
 
         it "passes the header to the handler (Int)" $
             delete' "/" "" `shouldRespondWith` 200
 
-    with (return (serve headerApi expectsString)) $ do
-        let delete' x = Test.Hspec.Wai.request methodDelete x [("MyHeader", "more from you")]
+    with (return (serve headerApi EmptyConfig expectsString)) $ do
+        let delete' x = Test.Hspec.Wai.request methodDelete x [("MyHeader" ,"more from you")]
```
jkarni: huh? Was this me?

servant-server/src/Servant/Server/Internal/Config.hs
soenkehahn: No, no reason. I'm just not familiar with it.

servant-server/test/Servant/Server/UsingConfigSpec/CustomCombinator.hs
soenkehahn: I wanted to put
- writing the combinator and
- using the combinator

into different modules to be able to look at what imports are needed for what. Mostly I wanted to make sure you don't need to import an internal module for using a combinator with a config.

soenkehahn:
@fizruk: I wasn't aware of the idea that combinators could put things into the config. Very interesting. I'll see whether I can add a test-case for that.

soenkehahn:
So putting entries into the `Config` in combinators *works*, but you need two instances to make the optional tagging work. :(

servant-server/test/Servant/Server/UsingConfigSpec.hs
```diff
@@ -0,0 +1,105 @@
...
+
+injectApp :: Application
+injectApp = serve (Proxy :: Proxy InjectAPI) config $
+  (\ s -> return s) :<|>
+  (\ s -> return ("tagged: " ++ s))
+  where
+    config = EmptyConfig
+
+spec2 :: Spec
+spec2 = do
```
jkarni: Any reason to have each spec call the next one, rather than have a top-level spec that calls all of them (as we do elsewhere)? Also, the names are a bit generic...

jkarni:
The subconfig idea is very interesting! However, it doesn't seem to fit very nicely into some cases - if some sub-api needs a specific (named) typ1, and that same sub-api needs the general (unnamed) typ2, but the complement of the sub-api needs the named version of typ2, it seems like things get a bit nasty. In essence, subconfig is applied to a whole subtree rather than individual nodes, and that's sometimes not ideal.

soenkehahn:
@jkarni: Yeah, I see that problem too. However I don't think it'll be *that* problematic. You can still factor out the common `Config` that is needed both in the top-level `Config` and in the `SubConfig`. And just include it in both by name. (And even use type aliases for the subconfig so that you don't have to repeat yourself in the types either.)

But more importantly: What I like about this approach is that we have untagged `Config`s, which makes them much nicer to work with in the common case (where you don't need to distinguish between config entries with the same type). Also, the `SubConfig` idea is completely optional. And we might also come up with alternatives to `SubConfig` in the future. (For example we could come up with a combinator that expands a subconfig into the parent config, to address @jkarni's concern.) So my proposal is to go with this for now and gather some experience on what use-cases actually come up for this in practice.

Another disadvantage is that `SubConfig` doesn't really have much significance for the other interpretations, e.g. `servant-client` or `servant-docs`. Then again tags also have the same disadvantage.

jkarni:
What about an instance for `SubConfig x :> ...` and for `SubConfig x y :> ...` (or different combinators) for affecting a sub-api and a single node? The name perhaps would have to change.

fizruk:
@jkarni as I understand named config/context entries are expected to be rare.
Most API endpoints should not contain too many path elements for a conflict to arise.

> if some sub-api needs a specific (named) typ1, and that same sub-api needs the general (unnamed) typ2, but the complement of the sub-api needs the named version of typ2, it seems like things get a bit nasty

You always get the closest entry. That means that combinators that *write* to config/context can override that entry. It should be pretty straightforward to understand from the API declaration which config entry is used where.

@soenkehahn I think we might still need config/context to provide nice client functions with `servant-client`.

fizruk:
@jkarni @soenkehahn I am sorry, I missed the `SubConfig` type in the code, so I assumed you're talking about contexts.

servant-examples/auth-combinator/auth-combinator.hs
soenkehahn: This example file wasn't using the new Config mechanism at all. Now it is: 9079c55. (And I called it `config`.)

servant-server/src/Servant/Server/Internal/Config.hs
soenkehahn: Now that we have a `SubConfig` I would actually prefer to stick with that.

servant-server/test/Servant/Server/UsingConfigSpec/CustomCombinator.hs
soenkehahn: True. Fixed here: eb20151. I wanted to put that kind signature somewhere for clarity, but I put it further down.

servant-server/src/Servant/Server/Internal/Config.hs
soenkehahn: For the sake of moving forward, I think this should be in a separate PR. (And personally I don't really care about these instances that much, so I'd like to leave them to someone else.)

Also I wonder whether the `Monoid` instance might cause more harm than good, because it doesn't behave like the instance for normal lists.

servant-server/src/Servant/Server/Internal/Config.hs
soenkehahn: Yup: eb286d4.

servant-server/src/Servant/Server/Internal/Config.hs
jkarni: OK. I'll implement `Read` and `From/ToJSON` soon then in a separate PR - those I think are important.

servant-server/test/Servant/Server/UsingConfigSpec.hs
soenkehahn: 4a2a27d

servant-server/servant-server.cabal
soenkehahn: Good catch: 7de2f79

servant-server/test/Servant/ServerSpec.hs
soenkehahn: Don't know. 94dbfd4

soenkehahn:
@jkarni: Is this roughly what you meant: d8db0fa ?

jkarni:
> @jkarni: Is this roughly what you meant: d8db0fa ?

Yes!

soenkehahn:
So my proposal is to put `SubConfig` and `Descend` (and their `HasServer` instances) into a separate module, and not to export them in `Servant` or `Servant.Server`. For two reasons:

- That way we can experiment a bit with the best way to access sub-configs.
- We also don't want to bother people that just want to use a `Config` in a simple way with these things.

And maybe rename the combinators.

Any opinions?

Apart from that this seems ready to merge from my side. @jkarni, @aaronlevin, @fizruk: Any comments or objections? Does this provide everything we need for authentication?

jkarni:
I think `Descend` and `SubConfig` together are wonderful - I'd expect quite a lot of things to become possible with all of this machinery. I don't think we need anything else. 

One thing to note is that `SubConfig`, as far as I can tell, does not provide any functionality that `Descend` doesn't already provide. It just provides a different syntax. I'm not sure how everyone feels about this. I think I personally prefer just one.

If just one combinator, maybe `WithConfig`? The names aren't all that clear...

soenkehahn:
I'm happy to have just one.

In my mind these `SubConfig`s are really somewhat like hierarchical namespaces or scopes. I think `WithConfig` doesn't really convey this very nicely.

There's also the fact that we're kind of misusing the combinator type `SubConfig` for the actual type of the sub-config (similarly to how we kind of misuse the combinator type `:<|>` also for the actual server and client types). Don't know if that's the best idea either.

aaronlevin:
lgtme :+1: 

I just have one question / anecdote about the `HasConfigEntry` class, but I don't think it should be a blocker. Right now, we have no `FunctionalDependencies` on that class. This is obviously working fine and we have some nice tests. However, the other day I was working on the van Laarhoven stuff again and ran into some type inference issues where I had to add functional dependencies that looked like:

```haskell
class HasConfigEntry (config :: [*]) (val :: *) | config -> val where
    getConfigEntry :: Config config -> val
```

So my question is just: how confident are we about this class? Will it hurt to add functional dependencies? I just added it to test it out and everything worked (after adding `-XFunctionalDependencies` and `-XUndecidableInstances`), but I'm curious about your thoughts.

jkarni:
@aaronlevin I don't understand how that fundep gets accepted - surely `config` doesn't determine the `val`?

aaronlevin:
@jkarni well, remember `config = [Type1, Type2, Type3]` so it's fair to say that `[Type1, Type2, Type3]` determines `Type2` (for example). You need to turn on `-XUndecidableInstances` for this to work, but it allows type checking some more abstract types that you are not able to do without it.

The examples I have are all based on the van Laarhoven stuff, so I can't easily port it over. For example, if you wanted the effect `State s a m` in the van Laarhoven free monad (i.e. one paramaterized over `s`, `a`, *and* `m`), you need to include the functional dependencies, otherwise you'll need to provide call-site type specification whenever you use `s` or `a` (e.g. `get :: Free effects MyState Int m` or something).

aaronlevin:
tbh, I don't fully understand functional dependencies, so I could be out to lunch with that suggestion! I do know I was able to compile things I wasn't previously and they worked.

servant-server/src/Servant/Server/Internal/Config.hs
```diff
@@ -0,0 +1,57 @@
...
+    x1 :. y1 == x2 :. y2 = x1 == x2 && y1 == y2
+
+class HasConfigEntry (config :: [*]) (val :: *) where
+    getConfigEntry :: Config config -> val
+
+instance OVERLAPPABLE_
+         HasConfigEntry xs val => HasConfigEntry (notIt ': xs) val where
+    getConfigEntry (_ :. xs) = getConfigEntry xs
+
+instance OVERLAPPABLE_
```
jkarni: `OVERLAPPING_`, not `OVERLAPPABLE_`, I think.

jkarni:
> @jkarni well, remember config = '[Type1, Type2, Type3] so it's fair to say that '[Type1, Type2, Type3] determines Type2 (for example).

How so? It could be any of Type1, 2 or 3, so it isn't determined by/deducible from `config`. In other words, given `config`, you can't tell me what `val` is. 

I don't really understand why it compiled, and maybe we should open a GHC ticket for it. It's pretty egregious in this case: the `HasConfigEntry` instances only differ by the fact that one requires `val` to be the head of the `config` list, so the two instances shouldn't be allowed to exist at all because the overlap is matching, if the fundep holds.

soenkehahn:
I would have had the same objection as @jkarni, but enabling `UndecidableInstances` helps and everything still compiles. Turns out I also don't fully understand functional dependencies, apparently.

I'd like to push this to master soon. Do we agree with adding just one combinator for accessing subconfigs? If yes, we still have to:

- [ ] Decide how to call that combinator,
- [ ] decide which combinator to use (`SomeName (name :: Symbol) (subConfig :: [*]) subApi` or `SomeName (name :: Symbol) (SubConfig :: [*]) :> subApi`) and
- [ ] write some documentation about how to use the config.

Help with these things would be appreciated (especially making decisions).

aaronlevin:
@soenkehahn does `name` need to be `Symbol` or can we just have it be of kind `k`, ala `(name :: k)`?

servant-server/src/Servant/Server/Internal/Config.hs
soenkehahn: Like this: 4f78e48 ?

jkarni:
Ok, [this](http://lpaste.net/149494) doesn't compile for the reasons I would have expected, and using `OVERLAPPABLE` doesn't change this. What in the world is going on? They look exactly the same.…

soenkehahn:
@aaronlevin: I think it could be polykinded. But I hope that we will get better error messages and maybe some other advantages from the additional kind-safety if we restrict it to `Symbol`.

aaronlevin:
@jkarni try adding the kind specifications, `class BreakFunDeps (a :: [*]) (b :: *) | a -> b` (sorry, not at my normal computer right now)

jkarni:
> decide which combinator to use (SomeName (name :: Symbol) (subConfig :: [*]) subApi or SomeName (name :: Symbol) (SubConfig :: [*]) :> subApi) 

I don't fully get this.

I am in favour of a single combinator. Re: name - `NamedConfig`? That way we can still "abuse", as you said, the combinator as we do with `:<|>`.

soenkehahn:
> I don't fully get this.

I want to know whether we want to have the one that's currently called `SubConfig` or the one that's currently called `Descend` (in the test cases). I. e. do we want to have the birdface in the instance or not? 

(I'm talking just about the API specification language here, independent of whether we re-use the combinator type for the values.)

jkarni:
Ah, I see. I think not having the birdface is clearer - it feels a little more explicit - but I can go either way.

OTOH I'm no longer so sure having two different combinators is worse...

All of this is to say I'm actually quite happy with any of the three possibilities.

soenkehahn:
But I like `NamedConfig`. I'm going to go with that, barring other proposals.

soenkehahn:
Created another PR where everything is sqashed: #327

soenkehahn:
Closing in favor of #327.

PR #329, opened 2016-01-14
server: tiny refactoring

soenkehahn:
I think this is *slightly* easier to understand. (Suggested by @aaronlevin.)

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -66,7 +66,7 @@ import           Servant.Server.Internal.ServantErr
 class HasServer layout where
   type ServerT layout (m :: * -> *) :: *
 
-  route :: Proxy layout -> Delayed (Server layout) -> Router
+  route :: Proxy layout -> Delayed (ServerT layout (ExceptT ServantErr IO)) -> Router
 
 type Server layout = ServerT layout (ExceptT ServantErr IO)
 


jkarni:
I disagree, but don't feel strongly either way.

alpmestan:
I honestly don't know what's nicer for a baginner, could you maybe ask a couple of users? (mailing list / IRC)

soenkehahn:
https://groups.google.com/d/msg/haskell-servant/dNIXIodUVlk/FZ8rmzZTDwAJ

bugthunk:
My thoughts, for what it's worth: If `Server layout` is used several times elsewhere in the code, the type alias should stay. Otherwise, drop it. I don't think newbness has much to do with it since `type` should be known even to them, otherwise you're not even on newbie stage - you're just getting started.

soenkehahn:
@bugthunk: This is not about dropping the type alias but about just not using it in the class definition. `Server` will still exist.

I think we shouldn't drop the type alias, as it would be a hugely breaking change.

bugthunk:
That means it's used elsewhere in the code, and should be kept :) Keep it in the class def as well. The alias is a good unclutterer of types, and if one does not understand it it could be argued that the training is needed. Problem solved!

fizruk:
I think `ServerT` is a very confusing name per se. Because it is **not a monad transformer**.
I was confused by that several times when I first encountered servant.

This confusion does not appear when using `Server` because with no `T` at the end I am not inclined to think it has to do anything with monads/monad transformers.

jkarni:
Yeah, `ServerT` was a bad choice. That said, it might not be worth changing, in my opinion.

soenkehahn:
Ok, I'll close this then.

codedmart:
@jkarni @fizruk I see your point. What would you change it to though?

jsdw:
ServerF (for Family)?

Issue #333, opened 2016-01-16
Make it easier to test that we have instances for core combinators for all interpretations


soenkehahn:
I imagine we could have an API that contains all core combinators at least once. In the different interpretations we would just add a test case where we use an instance for that big API. To ensure that all interpretations are there.

I want this for #327.

Should I put this into `Servant.API.Internal.Test` (or something like that) and expose it in the package? If we don't want to add an exposed module, I wouldn't be sure how to do it.

(cc: @jkarni)

jkarni:
Great idea. Maybe CPP the module so that by default it is not there, if that doesn't get too complicated?

fizruk:
Alternatively, maybe we can collect all sorts APIs for testing in a separate package, like `servant-samples`/`servant-api-samples`/`servant-test-samples`?

Or maybe `servant-hspec`/`hspec-servant` with not only common API samples, but also common `hspec` machinery (I suspect there is some, didn't check).

soenkehahn:
@jkarni: I don't want to tinker with `CPP` unless I absolutely have to.

@fizruk: A separate package seems a bit too heavy-weight just for one module with one type alias.

I think if we mark this as an `Internal` module there is no real harm in just exposing this in `servant`.

fizruk:
@soenkehahn the idea is that this separate package would declare all the useful sample APIs for testing, not just the mega API. E.g. we can extract APIs from [`ServerSpec.hs`](https://github.com/haskell-servant/servant/blob/master/servant-server/test/Servant/ServerSpec.hs) and [`ClientSpec.hs`](https://github.com/haskell-servant/servant/blob/master/servant-client/test/Servant/ClientSpec.hs).

I think having test APIs unified is a good thing, because then we know that for this set of APIs all of the packages actually work. Also for any new `servant-*` package will then use this set of APIs rather than invent its own, likely incomplete set.

I also think that some `hspec` machinery specific to `servant` might get factored out too.

soenkehahn:
@fizruk: Thanks for clarifying. I think that's a reasonable idea. I **do** think that having the API types next to the test cases is nice and that wouldn't be the case if they all get factored out into a separate package. I'm not very passionate about that, though.

[...]

PR #334, opened 2016-01-16
Mega api

soenkehahn:
This PR adds tests that we have instances for all core interpretations and all core combinators.

Can someone who knows more about `servant-foreign` and `servant-js` help with the instances for `Raw`, please? `servant-js` fails for `Raw`. That would be awesome. (Uncomment this line and fix the errors: https://github.com/haskell-servant/servant/blob/mega-api/servant/src/Servant/API/Internal/Test/ComprehensiveAPI.hs#L22) (cc: @jkarni @MaxOw @arianvp @dredozubov)

### DIFF
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -49,8 +49,11 @@ import           Test.HUnit
 import           Test.QuickCheck
 
 import           Servant.API
+import           Servant.API.Internal.Test.ComprehensiveAPI
 import           Servant.Client
 import           Servant.Server
+
+_ = client comprehensiveAPI
 
 spec :: Spec
 spec = describe "Servant.Client" $ do
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -53,6 +53,7 @@ import           Servant.API.Internal.Test.ComprehensiveAPI
 import           Servant.Client
 import           Servant.Server
 
+-- This declaration simply checks that all instances are in place.
 _ = client comprehensiveAPI
 
 spec :: Spec
--- a/servant-docs/test/Servant/DocsSpec.hs
+++ b/servant-docs/test/Servant/DocsSpec.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE DataKinds             #-}
 {-# LANGUAGE DeriveGeneric         #-}
-{-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE TypeOperators         #-}
@@ -18,7 +18,23 @@ import           GHC.Generics
 import           Test.Hspec
 
 import           Servant.API
+import           Servant.API.Internal.Test.ComprehensiveAPI
 import           Servant.Docs.Internal
+
+-- * comprehensive api
+
+_ = docs comprehensiveAPI
+
+instance ToParam (QueryParam "foo" Int) where
+  toParam = error "unused"
+instance ToParam (QueryParams "foo" Int) where
+  toParam = error "unused"
+instance ToParam (QueryFlag "foo") where
+  toParam = error "unused"
+instance ToCapture (Capture "foo" Int) where
+  toCapture = error "unused"
+
+-- * specs
 
 spec :: Spec
 spec = describe "Servant.Docs" $ do
--- a/servant-docs/test/Servant/DocsSpec.hs
+++ b/servant-docs/test/Servant/DocsSpec.hs
@@ -23,6 +23,7 @@ import           Servant.Docs.Internal
 
 -- * comprehensive api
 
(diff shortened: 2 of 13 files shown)


fizruk:
Related to #333.

jkarni:
What should be the instance for `Raw`?

soenkehahn:
We still don't have `Raw` working with `servant-js`, but I think this shouldn't block this PR. (I opened an issues instead: #339.)

I'd appreciate reviews.

servant-mock/src/Servant/Mock.hs
```diff
@@ -166,4 +175,5 @@ instance (Arbitrary a, Arbitrary (HList hs))
       => Arbitrary (HList (Header h a ': hs)) where
   arbitrary = HCons <$> fmap Header arbitrary <*> arbitrary
 
-
+instance Foldable (Headers headerTypes) where
```
jkarni: Is this needed? Why here?

servant/src/Servant/API/Internal/Test/ComprehensiveAPI.hs
```diff
@@ -0,0 +1,32 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TypeOperators #-}
+
```
jkarni: Maybe a comment, saying something like:

```
-- | This is a module containing an API with all `Servant.API` combinators. It is used for testing only (in particular, checking that instances exist for the core servant classes for each combinator), and should not be imported.
```

servant-mock/src/Servant/Mock.hs
jkarni: It also breaks on 7.8 (`Foldable` needs to be imported).

jkarni:
Small comments, but besides them +1.

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -49,9 +49,12 @@ import           Test.HUnit
 import           Test.QuickCheck
 
 import           Servant.API
+import           Servant.API.Internal.Test.ComprehensiveAPI
 import           Servant.Client
 import           Servant.Server
 
+_ = client comprehensiveAPI
```
fizruk: I would like a comment here. Something like

```haskell
-- This declaration simply checks that all instances are in place.
```

servant-docs/test/Servant/DocsSpec.hs
```diff
@@ -18,8 +18,24 @@ import           GHC.Generics
 import           Test.Hspec
 
 import           Servant.API
+import           Servant.API.Internal.Test.ComprehensiveAPI
 import           Servant.Docs.Internal
 
+-- * comprehensive api
+
+_ = docs comprehensiveAPI
```
fizruk: A similar comment here too.

servant-js/test/Servant/JSSpec.hs
```diff
@@ -29,6 +30,12 @@ import qualified Servant.JS.JQuery            as JQ
 import qualified Servant.JS.Vanilla           as JS
 import           Servant.JSSpec.CustomHeaders
 
+-- * comprehensive api
+
+_ = (jsForAPI comprehensiveAPI vanillaJS :: Text)
```
fizruk: A similar comment here too. Parenthesis seem redundant.

servant-mock/src/Servant/Mock.hs
soenkehahn: Good catch, that one isn't needed. I'll remove it.

servant-mock/test/Servant/MockSpec.hs
```diff
@@ -0,0 +1,81 @@
...
+import           Servant.API
+import           Test.Hspec hiding (pending)
+import           Test.Hspec.Wai
+import           Test.QuickCheck
+
+import           Servant
+import           Servant.API.Internal.Test.ComprehensiveAPI
+import           Servant.Mock
+
+_ = mock comprehensiveAPI
```
fizruk: A similar comment here, too.

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -60,6 +61,9 @@ import           Servant.Server.Internal.Router
                                             (tweakResponse, runRouter,
                                              Router, Router'(LeafRouter))
 
+-- * comprehensive api test
+
+_ = serve comprehensiveAPI (error "unused") (error "unused")
```
fizruk: A similar comment here.
Also, do we actually need two last arguments here? Can't it be just `_ = serve comprehensiveAPI`?

fizruk:
Aside from a few comments: :+1:

servant/src/Servant/API/Internal/Test/ComprehensiveAPI.hs
soenkehahn: Yes, added comments everywhere.

servant-client/test/Servant/ClientSpec.hs
soenkehahn: I added comments everywhere.

servant-server/test/Servant/ServerSpec.hs
soenkehahn: Yes, fixed.

soenkehahn:
Thanks for the quick review. I'll merge after travis passes.

PR #341, opened 2016-01-18
Authentication (v4.0)

aaronlevin:


aaronlevin:
FYI: `server-mock` is failing on this branch.

soenkehahn:
@aaronlevin: This is difficult to review with the config changes in between. You could create a PR against `shahn/config2`.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -10,11 +10,13 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE UndecidableInstances       #-}
```
aaronlevin: Not sure how contentious this is. I needed to use this to get around the whole "Constraint is no smaller than instance head". The full error is here:

```
Internal.hs|458 col 10 error| Constraint is no smaller than the instance head                                                                                                                 
||   in the constraint: HasConfigEntry       
||                        config (BasicAuthCheck (AuthReturnType (BasicAuth realm)))
|| (Use UndecidableInstances to permit this) 
|| In the instance declaration for           
||   ‘HasServer (BasicAuth realm :> api) config’
```

.gitignore
```diff
@@ -25,3 +25,6 @@ Setup
 .stack-work
 shell.nix
 default.nix
+tags
+TAGS
+codex.tags
```
soenkehahn: This should go into your global gitignore file.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -151,36 +154,71 @@ toApplication ra request respond = do
...
-          -> IO (RouteResult b)
-          -> (a -> b -> RouteResult c)
-          -> Delayed c
+data Delayed c = forall captures auth body. Delayed
+  { capturesD :: IO (RouteResult captures)
+  , methodD   :: IO (RouteResult ())
+  , authD     :: IO (RouteResult auth)
+  , bodyD     :: IO (RouteResult body)
+  , serverD   :: (captures -> auth -> body -> RouteResult c)
+  }
```
soenkehahn: There is a syntax for GADTs with selectors. (See somewhere here: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/data-type-extensions.html#gadt-style.) I would prefer that.

aaronlevin: pushing a change now.

servant-server/src/Servant/Server/Internal/Auth.hs
```diff
@@ -0,0 +1,77 @@
...
+data BasicAuthResult usr
+  = Unauthorized
+  | BadPassword
+  | NoSuchUser
+  | Authorized usr
+  deriving (Eq, Show, Read, Generic, Typeable, Functor)
+
+-- * Basic Auth
+
+newtype BasicAuthCheck usr = BasicAuthCheck
```
jkarni: Maybe a note about timing attacks.

servant-server/src/Servant/Server/Internal.hs
jkarni: I think this particular case is okay, but one big issue is that `UndecidableInstances` is a module pragma, and we won't notice anymore when new undecidable instances come in with this.

servant-server/src/Servant/Server/Internal/Auth.hs
```diff
@@ -0,0 +1,77 @@
...
+  { unBasicAuthCheck :: BS.ByteString  -- Username
```
jkarni: Since there's a `BasicAuthData` datatype for client, probably moving it in here and using it here too makes sense?

aaronlevin: yeah, i will move into `servant-server`.

I was thinking it actually should belong in `servant` but that project seems really minimal and type combinator driven so I didn't want to pollute that.

aaronlevin: oh, nm, I forgot that `servant-client` only depends on `servant-server` for the tests. I will move into core.

jkarni: Sorry, meant moving it to `servant`.

servant-server/src/Servant/Server/Internal/Auth.hs
aaronlevin: added one here and in `servant`.

servant-server/src/Servant/Server/Internal/Auth.hs
aaronlevin: done.

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -148,6 +154,46 @@ failServer = serve failApi EmptyConfig (
...
+
+serverConfig :: Config '[ BasicAuthCheck ()
+                        , AuthHandler Request ()
+                        ]
+serverConfig = basicAuthHandler :. authHandler :. EmptyConfig
+
+authServer :: Application
+authServer = serve authAPI serverConfig (const (return alice) :<|> const (return alice))
+
+{-
```
jkarni: what's this?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -148,6 +154,46 @@ failServer = serve failApi EmptyConfig (
...
+        if username == "servant" && password == "server"
+        then return (Authorized ())
+        else return Unauthorized
+  in BasicAuthCheck check
+
+authHandler :: AuthHandler Request ()
+authHandler =
+  let handler req = case lookup "AuthHeader" (requestHeaders req) of
+        Nothing -> throwE (err401 { errBody = "Missing auth header"
+                                  , errReasonPhrase = "denied!"
```
jkarni: Don't override the reason phrase - they're pretty standardized

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -9,56 +9,39 @@
...
+-- | A user type that we "fetch from the database" after
+-- performing authentication
+newtype User = User { unUser :: Text }
+
+
+-- | A method that, when given a password, will return a User.
+-- This is our bespoke (and bad) authentication logic.
+lookupUser :: ByteString -> ExceptT ServantErr IO User
+lookupUser cookie =
+  if cookie == "good password"
```
jkarni: The logic here doesn't make much sense. Looking up the user by password?

servant-examples/auth-combinator/auth-combinator.hs
```diff
@@ -69,28 +52,54 @@ newtype PublicData = PublicData { somedata :: Text }
...
+serverConfig :: Config (AuthHandler Request User ': '[])
+serverConfig = authHandler :. EmptyConfig
+
+-- | Our API, where we provide all the author-supplied handlers for each end point.
+-- note that 'prvdata' is a function that takes 'User' as an argument. We dont' worry
+-- about the authentication instrumentation here, that is taken care of by supplying
+-- configuration
 server :: Server API
-server = return prvdata :<|> return pubdata
+server = prvdata :<|> return pubdata
```
jkarni: maybe `privdata` rather than `prvdata`?

servant-examples/basic-auth/basic-auth.hs
```diff
@@ -0,0 +1,109 @@
...
+-- | our API
+type API = "public"  :> PublicAPI
+      :<|> "private" :> BasicAuth "foo-realm" :> PrivateAPI
+
+-- | a value holding a proxy of our API type
+api :: Proxy API
+api = Proxy
+
+-- | a value holding a proxy of our basic auth realm.
+authRealm :: Proxy "foo-realm"
```
jkarni: This can be removed.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -450,6 +455,28 @@ instance HasServer api config => HasServer (HttpVersion :> api) config where
...
+
+  route Proxy config subserver = WithRequest $ \ request ->
+    route (Proxy :: Proxy api) config (subserver `addAuthCheck` authCheck request)
+    where
+       realm = BC8.pack $ symbolVal (Proxy :: Proxy realm)
+       basicAuthConfig = getConfigEntry config
+       authCheck req = runBasicAuth req realm basicAuthConfig
+
+-- | General Authentication
+instance (HasServer api config, HasConfigEntry config (AuthHandler Request (AuthReturnType (AuthProtect tag)))) => HasServer (AuthProtect tag :> api) config where
```
jkarni: This line is huge. Please try to keep lines *around* 80 chars (a little more is fine).

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -48,21 +47,26 @@ import           Servant.API                ((:<|>) (..), (:>), Capture, Delete,
...
+import           Servant.Server             (ServantErr (..), Server, err401, err404,
+                                             serve, Config((:.), EmptyConfig))
 import           Test.Hspec                 (Spec, context, describe, it,
                                              shouldBe, shouldContain)
+import qualified Test.Hspec.Wai as THW
 import           Test.Hspec.Wai             (get, liftIO, matchHeaders,
                                              matchStatus, request,
                                              shouldRespondWith, with, (<:>))
 
+import           Servant.API.Auth (BasicAuthData(BasicAuthData))
```
jkarni: alignment

servant/src/Servant/API/Auth.hs
```diff
@@ -0,0 +1,34 @@
...
+-- timing attacks.
+--
+-- In Basic Auth, username and password are base64-encoded and transmitted via
+-- the @Authorization@ header. Handshakes are not required, making it
+-- relatively efficient.
+data BasicAuth (realm :: Symbol)
+  deriving (Typeable)
+
+-- | A simple datatype to hold data required to decorate a request
+data BasicAuthData = BasicAuthData { basicAuthUsername :: !ByteString
```
jkarni: Good call on strictness annotations.

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -528,6 +533,53 @@ miscCombinatorSpec = with (return $ serve miscApi EmptyConfig miscServ) $
...
+        it "returns with 401 with bad password" $ do
+          get "/basic" `shouldRespondWith` 401
+        it "returns 200 with the right password" $ do
+          THW.request methodGet "/basic" [("Authorization","Basic c2VydmFudDpzZXJ2ZXI=")] "" `shouldRespondWith` 200
+
+      context "Custom Auth Protection" $ do
+        it "returns 401 when missing headers" $ do
+          get "/auth" `shouldRespondWith` 401
+        it "returns 200 with the right header" $ do
+          THW.request methodGet "/auth" [("Auth","secret")] "" `shouldRespondWith` 200
```
jkarni: I like the qualified import, but the full `Test.Hspec.Wai` is used in several places and should be updated for consistency.

jkarni:
Tests should be added to `ErrorSpec` to verify that we have the right behaviour w.r.t. to error priorities and backtracking.

servant-examples/basic-auth/basic-auth.hs
```diff
@@ -0,0 +1,109 @@
...
+
+import           Data.Aeson                   (ToJSON)
+import           Data.Proxy                   (Proxy (Proxy))
+import           Data.Text                    (Text)
+import           GHC.Generics                 (Generic)
+import           Network.Wai.Handler.Warp     (run)
+import           Servant.API                  ((:<|>) ((:<|>)), (:>), BasicAuth,
+                                               Get, JSON)
+import           Servant.API.Auth             (BasicAuthData(BasicAuthData))
+import           Servant.Server               (AuthReturnType, BasicAuthResult (Authorized, Unauthorized), Config ((:.), EmptyConfig),
```
jkarni: Line too long

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -450,6 +455,28 @@ instance HasServer api config => HasServer (HttpVersion :> api) config where
   route Proxy config subserver = WithRequest $ \req ->
     route (Proxy :: Proxy api) config (passToServer subserver $ httpVersion req)
 
+-- | Basic Authentication
+instance (KnownSymbol realm, HasServer api config, HasConfigEntry config (BasicAuthCheck (AuthReturnType (BasicAuth realm))))
```
jkarni: Ditto on line length.

servant-client/test/Servant/ClientSpec.hs
aaronlevin: I am partly doing an end-to-end test here, to ensure that the way the serve modify's the response is verified by the client. I can change this to only test the body.

servant-examples/auth-combinator/auth-combinator.hs
aaronlevin: The cookie contains the password? It's such a minimal example. anyway, what I cna do instead is create a fake database in a `Map String User` or something. That might be clearer.

servant-client/test/Servant/ClientSpec.hs
aaronlevin: updated.

servant-client/test/Servant/ClientSpec.hs
aaronlevin: removed.

aaronlevin:
@jkarni added tests in `ErrorSpec.hs`

aaronlevin:
@jkarni @soenkehahn what do you need from me to push this forward?

jkarni:
I decided to take a look at the idea I mentioned earlier, of having `ServerT` parametrized by config as well, so that we don't need the type families (which I still strongly feel contradict part of the spirit of `Config`). You can see it [here](https://github.com/haskell-servant/servant/commit/895c95d1a1b1). Just a proof-of-concept, but it works. 

This would considerably reduce the difficulty involved in getting auth to work (no need to do anything besides put the `AuthCheck` in the config), and also make them overridable by the same mechanism as we have already. 

I still remain in favour of separate PRs for BasicAuth and the rest, and indeed for not having anything besides basic auth for now while we gather experience reports. But if we choose to do it otherwise, I also wonder whether BasicAuth should be a separate combinator, rather than part of the more general auth. Having more instances is bad for maintainability; and we're not testing how convenient the more general auth is.

aaronlevin:
@jkarni ok, I will separate into two PRs. However, I feel strongly about having some solution for a pragmatic way to plug in ad-hoc authentication. You might just have to put your foot down (and I'm ok with that), but it feels like a lot of wasted work to only launch with `Basic`. I also think the ad-hoc authentication provides great documentation for how to use the new `Config`/`Context` stuff.

I'll play around with `GetConfigByConstructor`. I'm not so certain it will work in all the cases, but I can see.

rvion:
@jkarni I have been silently following the discussion, but I agree with @aaronlevin that having ad-hoc authentication support would really nice.

alpmestan:
+1 with @aaronlevin here, even though I'd understand if a majority of folks prefer to just ship `BasicAuth` in 0.5.

codedmart:
I haven't been following all the code for auth in a while. While @jkarni's proof of concept sounds awesome. I do feel for @aaronlevin and anyone else involved in all the work that has been done for auth. I do feel it would have been better to provide something simpler earlier and get user feedback before doing too much work. But as long as this auth stuff doesn't cause any extra work if a user decides not to use it why not include it at this point and get user feedback from it.

aaronlevin:
I also believe that a simple and pragmatic ad-hoc solution will not be a too-far deviation from the `Basic` stuff (or at least this has been my experience thus far with the several different authentication implementations).

jkarni:
OK, I think we should maybe include both, and maybe even in 0.5, but have big warning labels that the API is unstable. @soenkehahn what do you think?

arianvp:
As long as the Auth stuff doesn't interfere with other parts too much (which it currently doesn't). I think it's okay to release it with the warning labels, instead of waiting for it to stabilize.  It would help us gain more feedback on this stuff.

freezeboy:
I agree with the fact that 0.5 also add many other things, if we know authentication still needs work maybe we could release the rest ?

dzhus:
Does anybody have an estimate on when 0.5 might get released?

rvion:
I'm wondering the very same thing right now :)

alpmestan:
@jkarni has started working on release notes. :)

aaronlevin:
Closing in favour of:
- https://github.com/haskell-servant/servant/pull/375
- https://github.com/haskell-servant/servant/pull/376

Issue #133, opened 2015-06-19
File upload combinator


alpmestan:
Just creating this issue to put some code I have written up here for discussion.

``` haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeOperators #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeSynonymInstances #-}
module Files where

import Control.Monad.IO.Class
import Control.Monad.Trans.Either
import Control.Monad.Trans.Resource
import Data.ByteString.Lazy (ByteString)
import Network.Wai
import Network.Wai.Handler.Warp (run)
import Network.Wai.Parse
import Servant
import Servant.Server.Internal

-- Backends for file upload: in memory or in /tmp ?

data Mem
data Tmp

class KnownBackend b where
  type Storage b :: *

  withBackend :: Proxy b -> (BackEnd (Storage b) -> IO r) -> IO r

instance KnownBackend Mem where
  type Storage Mem = ByteString

  withBackend Proxy f = f lbsBackEnd

instance KnownBackend Tmp where
  type Storage Tmp = FilePath

  withBackend Proxy f = runResourceT . withInternalState $ \s ->
    f (tempFileBackEnd s)

-- * Files combinator, to get all of the uploaded files

data Files b

instance (KnownBackend b, HasServer api) => HasServer (Files b :> api) where
  type ServerT (Files b :> api) m =
    [File (Storage b)] -> ServerT api m

  route Proxy subserver req respond = withBackend pb $ \b -> do
    (_, files) <- parseRequestBody b req
    route (Proxy :: Proxy api) (subserver files) req respond

    where pb = Proxy :: Proxy b

type FilesMem = Files Mem
type FilesTmp = Files Tmp

-- test

type API = "files" :> FilesTmp :> Post '[JSON] ()
      :<|> Raw

api :: Proxy API
api = Proxy

server :: Server API
server = filesHandler :<|> serveDirectory "."

  where filesHandler :: [File FilePath] -> EitherT ServantErr IO ()
        filesHandler = liftIO . mapM_ ppFile

        ppFile :: File FilePath -> IO ()
        ppFile (name, fileinfo) = do
          putStrLn $ "Input name: " ++ show name
          putStrLn $ "File name: " ++ show (fileName fileinfo)
[...]

PR #343, opened 2016-01-19
Allow IO when unrendering content types

fizruk:
This makes path to `multipart/form-data` content type, as mentioned in https://github.com/haskell-servant/servant/issues/133#issuecomment-151204917.

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -22,7 +22,7 @@ module Servant.Server.Internal
   , module Servant.Server.Internal.ServantErr
   ) where
 
-import           Control.Monad.Trans.Except (ExceptT, runExceptT)
+import           Control.Monad.Trans.Except (runExceptT)
 import           Control.Monad.Trans        (liftIO)
 import qualified Data.ByteString            as B
 import qualified Data.ByteString.Char8      as BC8
@@ -43,7 +43,6 @@ import           Network.Wai                (Application, Request, Response,
                                              responseLBS, vault)
 import           Prelude                    ()
 import           Prelude.Compat
-import           System.IO.Unsafe           (unsafeInterleaveIO)
 import           Web.HttpApiData            (FromHttpApiData)
 import           Web.HttpApiData.Internal   (parseHeaderMaybe,
                                              parseQueryParamMaybe,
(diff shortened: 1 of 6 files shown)


servant/src/Servant/API/ContentTypes.hs
```diff
@@ -195,19 +197,19 @@ instance OVERLAPPABLE_
...
 class Accept ctype => MimeUnrender ctype a where
-    mimeUnrender :: Proxy ctype -> ByteString -> Either String a
+    mimeUnrender :: Proxy ctype -> ByteString -> ExceptT String IO a
 
 class AllCTUnrender (list :: [*]) a where
     handleCTypeH :: Proxy list
                  -> ByteString     -- Content-Type header
                  -> ByteString     -- Request body
-                 -> Maybe (Either String a)
+                 -> Maybe (ExceptT String IO a)
```
jkarni: Is the reason we're keeping the `Maybe` that we want to be able to not do IO?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -400,8 +403,9 @@ instance ( AllCTUnrender list a, HasServer sublayout
         -- http://www.w3.org/2001/tag/2002/0129-mime
         let contentTypeH = fromMaybe "application/octet-stream"
                          $ lookup hContentType $ requestHeaders request
-        mrqbody <- handleCTypeH (Proxy :: Proxy list) (cs contentTypeH)
-               <$> lazyRequestBody request
+        lbody <- lazyRequestBody request
+        mrqbody <- traverse (liftIO . unsafeInterleaveIO . runExceptT) $
```
jkarni: This is going to need some thought. I'm not sure it's safe. We do the `bodyCheck` possibly multiple times per request, and probably having the side-effect happen multiple times is not what people expect.

jkarni: Story changed somewhat with #357

soenkehahn:
@fizruk: This PR is a bit old and stale. Should probably be rebased on current `master`.

Also I'm not clear on the motivation. Was this to support file-uploads?

jkarni:
> Also I'm not clear on the motivation. Was this to support file-uploads?

Yes.

I think this is still the right approach for that, especially since I think it's easier to do with recent changes. I'll take a look at this PR today, and try to rebase it.

jkarni:
Done. Tests pass and we no longer need `unsafeInterleaveIO`. That said, I think this needs a more precise set of tests about the functionality in `servant-server` (i.e., when the file is written, what errors are thrown if that fails, etc.). And the easiest way of writing those tests, to my mind, is writing the multipart content-type. So at the risk of scope creep, I'd wait on merging this.

jkarni:
Taking a further look, it seems like it makes the most sense to keep the new `MultiPartFormData` combinator as a separate package, since the most natural way to define the instances is to use `wai-extra`, which is not a dependency I'm happy with `servant` having.

Another point to note is that I don't immediately see a nice way of having configurable [options](https://hackage.haskell.org/package/wai-extra-3.0.16.1/docs/Network-Wai-Parse.html#t:ParseRequestBodyOptions) - e.g. file size limits. If we make this *not* specific to `MultiPartFormData` - e.g., if we have the maximum request body size in the `Context` - it would be nicer.

For now I'm going to use the default configuration options.

jkarni:
For reasons mentioned in #133, I'm in favor of closing this. @fizruk agreed?

phadej:
ping @fizruk, can this be closed?

phadej:
Closing this.

Issue #121, opened 2015-06-14
import jsForAPI docs


bwo:
it seems as if this should work:

```haskell
type StaticAPI = "static" :> Raw
type TagsAPI = "pagetags" :> ReqBody '[PlainText] T.Text :> Post '[JSON] ApiResponse

type API = StaticAPI :<|> TagsAPI

data ApiResponse -- blah blah

server' :: Server TagsAPI
server' = undefined -- blah blah
api' :: Proxy TagsAPI
api' = Proxy

apijs = jsForAPI api'
```

But it doesn't:

```
src/Main.hs:123:9:
    Couldn't match type ‘'False’ with ‘'True’
    In the expression: jsForAPI api'
    In an equation for ‘apijs’: apijs = jsForAPI api'
```

The documentation doesn't really indicate what has to be the case for this to work.

christian-marie:
Hmmn. That's a terrible error message. I'll see if I can get that not unifying as opposed to trying to match 'True and 'False.

From there, we can work out what's actually wrong, but, at an absolute glance I suspect it's may be related to posting PlainText, which may or may not be supported in this context.

I agree that it should work, or at least provide an error message that makes sense.

bwo:
Yeah, I worked it out eventually, it looks like only JSON is supported.

christian-marie:
On Sun, Jun 14, 2015 at 06:43:35PM -0700, bwo wrote:
> Yeah, I worked it out eventually, it looks like only JSON is supported.

This is worth documenting, by the way.

Is there some piece of documentation that you've read along the way where we
could have told you about this limitation?

I still think the type family here can be cleaned up, and is worth doing.

bwo:
I had been going by the tutorial, so I would think that http://haskell-servant.github.io/tutorial/javascript.html is a good place for it. 

[This](https://github.com/haskell-servant/servant-jquery/blob/master/src/Servant/JQuery/Internal.hs#L241) appears to be the only instance for `HasJQ` for `Post`s.

alpmestan:
Yeah for now we just support JSON. I would gladly take any patches that lets one configure which content type to get the data under. =) I just wanted a quick solution to get javascript functions for free and have never bothered investigating anything beyond my needs. I don't really know how we would go about handling various others, with jquery/JS.

tfausak:
[...]

PR #345, opened 2016-01-20
Servant.API.TypeLevel

### LINKED ISSUE #305: IsSubAPI constraint
I am using this currently to [tag sub APIs](https://github.com/fizruk/servant-swagger/blob/e2423d04c80fcc6475d53d2145c39dc5fbe7ad25/src/Servant/Swagger/Internal.hs#L44) in `servant-swagger`, but I believe this belongs in `servant`:

``` haskell
-- | Check whether @sub@ is a sub API of @api@.
type family IsSubAPI sub api :: Constraint where
  IsSubAPI sub api = AllIsElem (EndpointsList sub) api

-- | Check that every element of @xs@ is an endpoint of @api@.
type family AllIsElem xs api :: Constraint where
  AllIsElem '[] api = ()
  AllIsElem (x ': xs) api = (IsElem x api, AllIsElem xs api)

-- | Apply @(e :>)@ to every API in @xs@.
type family MapSub e xs where
  MapSub e '[] = '[]
  MapSub e (x ': xs) = (e :> x) ': MapSub e xs

-- | Append two type-level lists.
type family AppendList xs ys where
  AppendList '[]       ys = ys
  AppendList (x ': xs) ys = x ': AppendList xs ys

-- | Build a list of endpoints from an API.
type family EndpointsList api where
  EndpointsList (a :<|> b) = AppendList (EndpointsList a) (EndpointsList b)
  EndpointsList (e :> a)   = MapSub e (EndpointsList a)
  EndpointsList a = '[a]
```

I am not sure where to put it though.

fizruk:
Closes #305.

I need to write some tests, otherwise, should be complete.

### DIFF
--- a/servant/src/Servant/API/TypeLevel.hs
+++ b/servant/src/Servant/API/TypeLevel.hs
@@ -1,0 +1,78 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+module Servant.API.TypeLevel where
+
+import GHC.Exts(Constraint)
+import Servant.API.Capture ( Capture )
+import Servant.API.ReqBody ( ReqBody )
+import Servant.API.QueryParam ( QueryParam, QueryParams, QueryFlag )
+import Servant.API.Header ( Header )
+import Servant.API.Verbs ( Verb )
+import Servant.API.Sub ( type (:>) )
+import Servant.API.Alternative ( type (:<|>) )
+
+-- * API predicates
+
+-- | You may use this type family to tell the type checker that your custom
+-- type may be skipped as part of a link. This is useful for things like
+-- 'QueryParam' that are optional in a URI and do not affect them if they are
+-- omitted.
+--
+-- >>> data CustomThing
+-- >>> type instance IsElem' e (CustomThing :> s) = IsElem e s
+--
+-- Note that 'IsElem' is called, which will mutually recurse back to `IsElem'`
+-- if it exhausts all other options again.
+--
+-- Once you have written a HasLink instance for CustomThing you are ready to
+-- go.
+type family IsElem' a s :: Constraint
+
+-- | Closed type family, check if @endpoint@ is within @api@.
+-- Uses @'IsElem''@.
+type family IsElem endpoint api :: Constraint where
+    IsElem e (sa :<|> sb)                   = Or (IsElem e sa) (IsElem e sb)
+    IsElem (e :> sa) (e :> sb)              = IsElem sa sb
+    IsElem sa (Header sym x :> sb)          = IsElem sa sb
+    IsElem sa (ReqBody y x :> sb)           = IsElem sa sb
+    IsElem (CaptureAll z y :> sa) (CaptureAll x y :> sb)
+                                            = IsElem sa sb
+    IsElem (Capture z y :> sa) (Capture x y :> sb)
+                                            = IsElem sa sb
+    IsElem sa (QueryParam x y :> sb)        = IsElem sa sb
+    IsElem sa (QueryParams x y :> sb)       = IsElem sa sb
+    IsElem sa (QueryFlag x :> sb)           = IsElem sa sb
+    IsElem (Verb m s ct typ) (Verb m s ct' typ)
+                                            = IsSubList ct ct'
+    IsElem e e                              = ()
+    IsElem e a                              = IsElem' e a
+
+-- * Helpers
+
+-- ** Lists
+
+type family IsSubList a b :: Constraint where
+    IsSubList '[] b          = ()
+    IsSubList (x ': xs) y    = Elem x y `And` IsSubList xs y
+
+type family Elem e es :: Constraint where
+    Elem x (x ': xs) = ()
+    Elem y (x ': xs) = Elem y xs
+
+-- ** Logic
+
+-- | If either a or b produce an empty constraint, produce an empty constraint.
+type family Or (a :: Constraint) (b :: Constraint) :: Constraint where
+    -- This works because of:
+    -- https://ghc.haskell.org/trac/ghc/wiki/NewAxioms/CoincidentOverlap
+    Or () b       = ()
+    Or a ()       = ()
+
+-- | If both a or b produce an empty constraint, produce an empty constraint.
+type family And (a :: Constraint) (b :: Constraint) :: Constraint where
+    And () ()     = ()
+
--- a/servant/src/Servant/API/TypeLevel.hs
+++ b/servant/src/Servant/API/TypeLevel.hs
@@ -1,29 +1,51 @@
-{-# LANGUAGE CPP #-}
-{-# LANGUAGE DataKinds #-}
-{-# LANGUAGE KindSignatures #-}
-{-# LANGUAGE PolyKinds #-}
-{-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeOperators #-}
-{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP                   #-}
+{-# LANGUAGE ConstraintKinds       #-}
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE KindSignatures        #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE PolyKinds             #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-|
+This module collects utilities for manipulating @servant@ API types. The
+functionality in this module is for advanced usage.
+
+The code samples in this module use the following type synonym:
+
+> type SampleAPI = "hello" :> Get '[JSON] Int
+>             :<|> "bye" :> Capture "name" String :> Post '[JSON, PlainText] Bool
+
+-}
 module Servant.API.TypeLevel where
 
-import GHC.Exts(Constraint)
-import Servant.API.Capture ( Capture, CaptureAll )
-import Servant.API.ReqBody ( ReqBody )
-import Servant.API.QueryParam ( QueryParam, QueryParams, QueryFlag )
-import Servant.API.Header ( Header )
-import Servant.API.Verbs ( Verb )
-import Servant.API.Sub ( type (:>) )
-import Servant.API.Alternative ( type (:<|>) )
+
+import           GHC.Exts                (Constraint)
+import           Servant.API.Alternative (type (:<|>))
+import           Servant.API.Capture     (Capture, CaptureAll)
+import           Servant.API.Header      (Header)
+import           Servant.API.QueryParam  (QueryFlag, QueryParam, QueryParams)
+import           Servant.API.ReqBody     (ReqBody)
+import           Servant.API.Sub         (type (:>))
+import           Servant.API.Verbs       (Verb)
 #if MIN_VERSION_base(4,9,0)
-import GHC.TypeLits (TypeError, ErrorMessage(..))
+import           GHC.TypeLits            (TypeError, ErrorMessage(..))
 #endif
+
+
 
 -- * API predicates
 
 -- | Flatten API into a list of endpoints.
+--
+-- >>> :t showType @(Endpoints SampleAPI)
+-- ...
+-- ... :: Proxy
+-- ...  '["hello" :> Verb 'GET 200 '[JSON] Int,
+-- ...    "bye"
+-- ...    :> (Capture "name" String
+-- ...        :> Verb 'POST 200 '[JSON, PlainText] Bool)]
 type family Endpoints api where
   Endpoints (a :<|> b) = AppendList (Endpoints a) (Endpoints b)
   Endpoints (e :> a)   = MapSub e (Endpoints a)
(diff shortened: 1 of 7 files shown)


servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,121 @@
...
+  AppendList '[]       ys = ys
+  AppendList (x ': xs) ys = x ': AppendList xs ys
+
+type family IsSubList a b :: Constraint where
+    IsSubList '[] b          = ()
+    IsSubList (x ': xs) y    = Elem x y `And` IsSubList xs y
+
+type family Elem e es :: Constraint where
+    Elem x (x ': xs) = ()
+    Elem y (x ': xs) = Elem y xs
```
jkarni: The similar version in [-foreign](https://github.com/haskell-servant/servant/blob/357cc839b65601975cb186a3f00a729f7f3bb2de/servant-foreign/src/Servant/Foreign/Internal.hs#L122) can be removed in favour of this one, I think.

fizruk: Done.

jkarni:
Also fixes #121

jkarni:
LGTM

codedmart:
I don't know much about type families yet, but also LGTM. :+1:

aaronlevin:
LGTM :+1:

fizruk:
FYI: some more cool stuff is being [embedded in `servant-swagger`](https://github.com/haskell-servant/servant-swagger/blob/swagger2-2.0/src/Servant/Swagger/Internal/TypeLevel/API.hs#L68-L82).

dredozubov:
The only issue i have with this is style: indentation is very inconsistent, sometimes i see 2 spaces, sometimes 4. Can i propose 2 spaces for the whole repo? We have to deal with huge type declarations and systematically having code closer to the left side will do some good IMHO.

servant/src/Servant/Utils/Links.hs
```diff
@@ -86,6 +85,8 @@
 --  bad_link under api after trying the open (but empty) type family
 --  `IsElem'` as a last resort.
 module Servant.Utils.Links (
+  module Servant.API.TypeLevel,
```
phadej: shouldn't this be at the bottom of export list?

fizruk:
@dredozubov I am in favour of 2 spaces too.

It has also been brought to our attention that [`singletons`](https://hackage.haskell.org/package/singletons) implements a lot of type-level machinery in their `Data.Promotion.Prelude` modules. I am currently considering adding `singletons` as a dependency of `servant` to reuse these.

servant/src/Servant/Utils/Links.hs
dredozubov: there'll be some warning if anything will be shadowed afaik

dredozubov:
@fizruk I personally feel like adding singletons to the dependency list is like opening a gateway to hell. In my experience it quickly escalates to the longer compile-time and awkward(IMHO) type signatures, because "we already have singletons as a dep, why not do X?". To make picture fuller, there is also https://hackage.haskell.org/package/type-fun

fizruk:
@dredozubov that is my only concern so far, yes.

arianvp:
What is holding us back currently to merge this? Except from the discussion whether we want to add an extra dependency for type-level utility functions?

fizruk:
@arianvp tests are still missing and there are also few things in [`Servant.Swagger.Internal.TypeLevel`](http://hackage.haskell.org/package/servant-swagger-1.0.3/docs/Servant-Swagger-Internal-TypeLevel.html) that we might want to move here too. But that can be done in a separate PR.

I am currently overloaded at work so I won't be able to work on this in the next couple of weeks.

kosmikus:
Do I misunderstand this PR? This isn't adding new functionality, but moving scattered functions to a common module, right? If so, I don't think we should require new tests for this to go in. Also, adding additional functions from elsewhere should be a separate PR.

And I'd vote against adding a dependency on singletons if it can be avoided.

As far as this is just a refactoring (which I thought it was), I'm happy for it to be merged.

fizruk:
@kosmikus there is new functionality here, e.g. [`Endpoints`](https://github.com/haskell-servant/servant/pull/345/files#diff-a20697aa772f592c27457c5013a42c02R20) and [`IsSubAPI`](https://github.com/haskell-servant/servant/pull/345/files#diff-a20697aa772f592c27457c5013a42c02R59) and a few other type families which came from (an older version of) [`Servant.Swagger.Internal.TypeLevel.API`](http://hackage.haskell.org/package/servant-swagger-1.0.3/docs/Servant-Swagger-Internal-TypeLevel-API.html).

Over the time `servant-swagger` gained more type-level stuff (like [`BodyTypes`](http://hackage.haskell.org/package/servant-swagger-1.0.3/docs/Servant-Swagger-Internal-TypeLevel-API.html#t:BodyTypes)). I had a few more type families in mind (e.g. `ParamTypes` to extract types for `ToParamSchema` testing), but never had time to add those.

My idea was to figure out all useful types in `servant-swagger` and then move them here.

kosmikus:
I see. But then, why should all this be added to the `servant` package? I can kind of understand it for the things that are needed by the core servant machinery itself. But if it becomes more like "let's add all type families that anyone could ever need", it feels like it'd be better in a separate `servant-typelevel` package to me.

fizruk:
@kosmikus well, maybe. But if we go with `servant-typelevel` then some things have to stay in `servant` (e.g. ones used by `Servant.Utils.Links`) while others move to `servant-typelevel`. There aren't that many type-level functions left, so I did not consider an option of `servant-typelevel`.

Most of the type-level code is actually very general and not related to `servant` per se. Hence the `singletons` discussion. `servant-swagger` has some general type-level code too. It is not clear where all this code belongs.

I also think type-level functions like `IsSubAPI` and `BodyTypes` are essential for servant-based documentation and testing:
- `IsSubAPI` (used in `servant-swagger`) is a "more general" version of `IsIn` used in `servant-docs`; it is very useful when modifying documentation for a selected part of an API;
- `BodyTypes` is useful to extract a list of all API types used for request/response bodies filtered by content-type; it enables things like [`Servant.Swagger.Test`](http://hackage.haskell.org/package/servant-swagger-1.0.3/docs/Servant-Swagger-Test.html), but is not limited to that.

IMO, these are not just some type functions "anyone could ever need", but rather "anyone building a servant-*something* library should probably use". Think `servant-docs`, `servant-quickcheck`, etc.

In any case, feel free to kill this PR and introduce another one which would just put existing things in one place. We can discuss all the extra type-level functions in a separate issue/PR.

kosmikus:
I definitely agree that there are probably a whole lot of useful type-level functions in the context of servant, and that it makes sense to provide them, rather than letting every user of servant reinvent them.

So for the time being, I'm happy to go with the approach proposed here. It just think we should keep an eye on how much we'll accumulate over time. As long as there are no new external dependencies (such as singletons) introduced by this effort, I think we're probably ok.

jkarni:
I just rebased against master. @haskell-servant/maintainers can I merge? As far as I can tell we have two +1s from before, unless something changed.

soenkehahn:
I'm happy to give this another review before merging.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,123 @@
...
+module Servant.API.TypeLevel where
+
+import GHC.Exts(Constraint)
+import Servant.API.Capture ( Capture, CaptureAll )
+import Servant.API.ReqBody ( ReqBody )
+import Servant.API.QueryParam ( QueryParam, QueryParams, QueryFlag )
+import Servant.API.Header ( Header )
+import Servant.API.Verbs ( Verb )
+import Servant.API.Sub ( type (:>) )
+import Servant.API.Alternative ( type (:<|>) )
```
soenkehahn: Whitespaces are inconsistent.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,123 @@
...
+
+-- | You may use this type family to tell the type checker that your custom
+-- type may be skipped as part of a link. This is useful for things like
+-- @'QueryParam'@ that are optional in a URI and do not affect them if they are
+-- omitted.
+--
+-- >>> data CustomThing
+-- >>> type instance IsElem' e (CustomThing :> s) = IsElem e s
+--
+-- Note that @'IsElem'@ is called, which will mutually recurse back to @'IsElem''@
```
soenkehahn: Where is `IsElem` called?

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,123 @@
...
+                                            = IsElem sa sb
+    IsElem sa (QueryParam x y :> sb)        = IsElem sa sb
+    IsElem sa (QueryParams x y :> sb)       = IsElem sa sb
+    IsElem sa (QueryFlag x :> sb)           = IsElem sa sb
+    IsElem (Verb m s ct typ) (Verb m s ct' typ)
+                                            = IsSubList ct ct'
+    IsElem e e                              = ()
+    IsElem e a                              = IsElem' e a
+
+-- | Check whether @sub@ is a sub API of @api@.
```
soenkehahn: Maybe we should include a comment that this only checks for equality on types that are being serialized to JSON. And not for some form of sub-typing with regards to the JSON encoding.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,123 @@
...
+type family Or (a :: Constraint) (b :: Constraint) :: Constraint where
+    -- This works because of:
+    -- https://ghc.haskell.org/trac/ghc/wiki/NewAxioms/CoincidentOverlap
+    Or () b       = ()
+    Or a ()       = ()
+
+-- | If both a or b produce an empty constraint, produce an empty constraint.
+type family And (a :: Constraint) (b :: Constraint) :: Constraint where
+    And () ()     = ()
+
```
soenkehahn: Trailing newline.

soenkehahn:
Currently, we only have indirect tests. Can we have some direct tests for the `TypeLevel` module itself, please?

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,123 @@
...
+-- | Closed type family, check if @endpoint@ is exactly within @api@.
+-- We aren't sure what affects how an endpoint is built up, so we require an
+-- exact match.
+type family IsIn (endpoint :: *) (api :: *) :: Constraint where
+    IsIn e (sa :<|> sb)                = Or (IsIn e sa) (IsIn e sb)
+    IsIn (e :> sa) (e :> sb)           = IsIn sa sb
+    IsIn e e                           = ()
+
+-- | Check whether @sub@ is a sub API of @api@.
```
soenkehahn: We have the exact same documentation for `IsSubAPI` and `IsStrictSubAPI`. Currently I also don't understand the difference. (In set theory a strict subset is a subset that is not equal to the superset. But that doesn't seem to be the case here.)

Can we have:

- test cases demonstrating the difference and
- comments explaining the difference, please?

jkarni: Test cases == doctests, I imagine?

soenkehahn: Haven't thought much about where to put these tests. I was thinking about `hspec` tests, but `doctest` is a good idea. Whatever works, as long as we have tests. :)

servant/src/Servant/API/TypeLevel.hs
jkarni: I think a few things here are not used in the repo. Should I remove them all?

servant/src/Servant/Utils/Links.hs
soenkehahn: @dredozubov: I don't fully understand your comment. What is shadowed by what here? In any case I would expect the order of the export list to be irrelevant.

servant/src/Servant/API/TypeLevel.hs
fizruk: I am ok with removing something that's not used currently. We can always add more things later.

However, keep in mind that `TypeLevel` is not meant as an internal module. Rather I hoped to remove type-level stuff from `servant-swagger` since it can be useful for other packages such as `servant-docs` or `servant-quickcheck`.

jkarni:
Added loads of doctests and other documentation.

soenkehahn:
Can we also squash the commits before merging, please?

jkarni:
Yup. Got some more work to do first though.

On Thu, Aug 25, 2016 at 08:24:45AM -0700, Sönke Hahn wrote:
> Can we also squash the commits before merging, please?
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/haskell-servant/servant/pull/345#issuecomment-242427979

--
Julian K. Arni
Haskell Consultant, Turing Jump
https://turingjump.com

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,239 @@
...
+
+-- | Flatten API into a list of endpoints.
+--
+-- >>> :t showType @(Endpoints SampleAPI)
+-- ...
+-- ... :: Proxy
+-- ...  '["hello" :> Verb 'GET 200 '[JSON] Int,
+-- ...    "bye"
+-- ...    :> (Capture "name" String
+-- ...        :> Verb 'POST 200 '[JSON, PlainText] Bool)]
```
soenkehahn: I wouldn't use `...` and just match `ghc`'s indentation.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,239 @@
...
+import           GHC.TypeLits            (TypeError, ErrorMessage(..))
+#endif
+
+
+
+-- * API predicates
+
+-- | Flatten API into a list of endpoints.
+--
+-- >>> :t showType @(Endpoints SampleAPI)
```
soenkehahn: I think `showType` is confusing. You could do `:kind! Endpoints SampleAPI`.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,239 @@
...
+-- $setup
+-- >>> :set -XTypeApplications
+-- >>> :set -XPolyKinds
+-- >>> import Data.Proxy
+-- >>> import Servant.API
+-- >>> data OK ctx = OK deriving (Show)
+-- >>> let ok :: ctx => OK ctx; ok = OK
+-- >>> let showType :: Proxy a ; showType = Proxy
+-- >>> type SampleAPI = "hello" :> Get '[JSON] Int :<|> "bye" :> Capture "name" String :> Post '[JSON, PlainText] Bool
+-- >>> let sampleAPI = Proxy :: Proxy SampleAPI
```
soenkehahn: I would consider including this in the first doctest snippet at the top of the module. It's a bit noisy, but I would prefer to have a fully functioning example at the top, without relying on readers to find this `$setup` section.

servant/src/Servant/API/TypeLevel.hs
soenkehahn: @jkarni: I think you misunderstood my comment. The haddock comment seems to be saying that `IsElem'` is implemented in terms of `IsElem`. (Which then in turn calls back into `IsElem'`.) But I don't see where this call is being made.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,239 @@
...
+-- ...
+-- ... :: Proxy
+-- ...  '["hello" :> Verb 'GET 200 '[JSON] Int,
+-- ...    "bye"
+-- ...    :> (Capture "name" String
+-- ...        :> Verb 'POST 200 '[JSON, PlainText] Bool)]
+type family Endpoints api where
+  Endpoints (a :<|> b) = AppendList (Endpoints a) (Endpoints b)
+  Endpoints (e :> a)   = MapSub e (Endpoints a)
+  Endpoints a = '[a]
```
soenkehahn: Can we have a test that shows that `Endpoints` distributes combinators correctly in an API like this:

``` haskell
type Api = "commonPath" :> (subApiA :<|> subApiB)
```

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,239 @@
...
+-- >>> ok @(IsElem (Get '[JSON] Int) (Header "h" Bool :> Get '[JSON] Int))
+-- OK
+--
+-- >>> ok @(IsElem (Get '[JSON] Int) (ReqBody '[JSON] Bool :> Get '[JSON] Int))
+-- OK
+--
+-- *N.B.:* @IsElem a b@ can be seen as capturing the notion of whether the URL
+-- represented by @a@ would match the URL represented by @b@, *not* whether a
+-- request represented by @a@ matches the endpoints serving @b@ (for the
+-- latter, use 'IsIn').
```
soenkehahn: Can we have a more descriptive name than `IsElem` for this then, please? `HasLink`, or `ContainsUrl` or something like that?

soenkehahn:
So `Lax` means that we're considering the given `servant` endpoint as a description of a url? And `Strict` means that we are talking about real sub-apis, i.e. including request bodies and headers?

If yes, we should name all these things differently, IMO.

servant/src/Servant/API/TypeLevel.hs
jkarni: In the example (CustomThing)? I guess maybe it's best to just remove that comment altogether?

servant/src/Servant/API/TypeLevel.hs
jkarni: I feel like that'd be testing indentation, and might break more easily across GHC versions?

servant/src/Servant/API/TypeLevel.hs
jkarni: Good call.

servant/src/Servant/API/TypeLevel.hs
jkarni: `ContainsURL` and `HasLink` would probably require switching the order of the arguments. Should I do that, or something like `IsURLOf` or `IsURLIn`?

jkarni: Or `URLMatches`, or `URLBelongsTo`

jkarni:
> So Lax means that we're considering the given servant endpoint as a description of a url? And Strict means that we are talking about real sub-apis, i.e. including request bodies and headers?

Yes

> If yes, we should name all these things differently, IMO.

Agreed. How about `URLMatches` and `EndpointMatches`? We'd have to be consistent about usage of `endpoint` to mean "everything", though.

More generally, what actually *is* the correct terminology to distinguish between these things?

servant/src/Servant/API/TypeLevel.hs
jkarni: I don't know how to get that to work. It seems like named chunks in that intro section don't mean the same thing?

servant/src/Servant/API/TypeLevel.hs
soenkehahn: I just think it's confusing in the documentation. Other than that, I would run the doctests with one `ghc` version only. (As discussed before.)

servant/src/Servant/API/TypeLevel.hs
soenkehahn: Ah, now I get it. Yeah, I think this needs more explanation to be really useful.

servant/src/Servant/API/TypeLevel.hs
phadej: Added chunk to the export list.

servant/src/Servant/API/TypeLevel.hs
phadej: Done

servant/src/Servant/API/TypeLevel.hs
phadej: Let's open new issue for the renaming, to get this merged,

servant/src/Servant/API/TypeLevel.hs
phadej: Used `Refl` to make GHC decide type-equality

phadej:
is @fizruk happy with the current state of this PR?

As I pointed above, name bike-shedding can be done later.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -0,0 +1,263 @@
...
+#if MIN_VERSION_base(4,9,0)
+  -- Note [Custom Errors]
+  ElemGo x '[] orig       = TypeError ('ShowType x
+                                 ':<>: 'Text " expected in list "
+                                 ':<>: 'ShowType orig)
+#else
+  ElemGo x '[] orig       = ElemNotFoundIn x orig
+#endif
+
+-- ** Logic
```
fizruk: Is it worth having sections in both code and export list?
Also this one is missing in the export list.

phadej: I usually keep some section indicators in the code as well, so it's organised as well. (Sometimes sections in code are in the different order though, thanks to Template Haskell).

Added Logic subsection to the export list.

fizruk:
Ok, I'm happy with the current state of this PR. 👍

PR #347, opened 2016-01-20
CaptureAll combinator

### LINKED ISSUE #257: Combinator for capturing all remaining segments
It'd be nice to have a combinator for capturing all remaining segments in a URL, say for serving some sort of filesystem-like interface. 

It's easy enough to write such a combinator. Here's one I wrote with the help of @alpmestan: 

```haskell
data CaptureAll (sym :: Symbol) deriving (Typeable)

instance (KnownSymbol capture, HasServer sublayout)
         => HasServer (CaptureAll capture :> sublayout) where

  type ServerT (CaptureAll capture :> sublayout) m = [Text] -> ServerT sublayout m

  route Proxy subserver request respond =
    let request' = request { pathInfo = [] }
        segments = processedPathInfo request
    in
    route (Proxy :: Proxy sublayout) (subserver segments) request' respond
```

However, AFAICT, it's not possible to write a `HasLink` instance, which means one can't make safe links for the API.

fizruk:
Closes #257.

Need to add instances for `servant-foreign` and `servant-doc`.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/Router.hs
+++ b/servant-server/src/Servant/Server/Internal/Router.hs
@@ -13,8 +13,8 @@ type Router = Router' RoutingApplication
 
 -- | Internal representation of a router.
 data Router' a =
-    WithRequest   (Request -> Router)
-      -- ^ current request is passed to the router
+    WithRequest   (Request -> (Request, Router))
+      -- ^ current request is passed to the router and can be changed
   | StaticRouter  (Map Text Router)
       -- ^ first path component used for lookup and removed afterwards
   | DynamicRouter (Text -> Router)
@@ -34,29 +34,20 @@ tweakResponse f = fmap (\a -> \req cont -> a req (cont . f))
 --
 --   * Two static routers can be joined by joining their maps.
 --   * Two dynamic routers can be joined by joining their codomains.
---   * Two 'WithRequest' routers can be joined by passing them
---     the same request and joining their codomains.
---   * A 'WithRequest' router can be joined with anything else by
---     passing the same request to both but ignoring it in the
---     component that does not need it.
 --
 choice :: Router -> Router -> Router
 choice (StaticRouter table1) (StaticRouter table2) =
   StaticRouter (M.unionWith choice table1 table2)
 choice (DynamicRouter fun1)  (DynamicRouter fun2)  =
   DynamicRouter (\ first -> choice (fun1 first) (fun2 first))
-choice (WithRequest router1) (WithRequest router2) =
-  WithRequest (\ request -> choice (router1 request) (router2 request))
-choice (WithRequest router1) router2 =
-  WithRequest (\ request -> choice (router1 request) router2)
-choice router1 (WithRequest router2) =
-  WithRequest (\ request -> choice router1 (router2 request))
 choice router1 router2 = Choice router1 router2
 
 -- | Interpret a router as an application.
 runRouter :: Router -> RoutingApplication
 runRouter (WithRequest router) request respond =
-  runRouter (router request) request respond
+  runRouter subRouter subRequest respond
+  where
+    (subRequest, subRouter) = router request
 runRouter (StaticRouter table) request respond =
   case pathInfo request of
     first : rest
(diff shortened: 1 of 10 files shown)


fizruk:
Looks like `servant-foreign` needs an adjustment to fit `CaptureAll`. Would be nice to get some help.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -34,29 +34,20 @@ tweakResponse f = fmap (\a -> \req cont -> a req (cont . f))
...
---   * A 'WithRequest' router can be joined with anything else by
---     passing the same request to both but ignoring it in the
---     component that does not need it.
 --
 choice :: Router -> Router -> Router
 choice (StaticRouter table1) (StaticRouter table2) =
   StaticRouter (M.unionWith choice table1 table2)
 choice (DynamicRouter fun1)  (DynamicRouter fun2)  =
   DynamicRouter (\ first -> choice (fun1 first) (fun2 first))
-choice (WithRequest router1) (WithRequest router2) =
```
jkarni: Hmm, I suspect this is a performance regression; we no longer would have the nice trie we currently do. I wonder if the `Request -> (Request, Router)` idea will work after all. @kosmikus might be able to chime in.

arianvp:
I'll see if I can find some time to look at servant-foreign for this. Will be after next week though

servant-server/src/Servant/Server/Internal/Router.hs
fizruk: I wonder if this smart `choice` constructor actually does any noticeable optimization.
If not, I would remove it entirely.

jml:
Hi! Is there anything I can do to help this fix land?

jml:
I've prepared a version of this which fixes the conflicts: 
https://github.com/haskell-servant/servant/compare/master...jml:capture-all-257 by merging in master.

It compiles, but the `mock` tests fail, and maybe others do too:

```
    [1 of 2] Compiling Servant.MockSpec ( test/Servant/MockSpec.hs, .stack-work/dist/x86_64-osx/Cabal-1.22.5.0/build/spec/spec-tmp/Servant/MockSpec.o ) [Servant.API changed]

    /Users/jml/src/servant/servant-mock/test/Servant/MockSpec.hs:26:5:
        No instance for (HasMock
                           (ReqBody '[JSON] Int :> GET) '[NamedContext "foo" '[]])
          arising from a use of ‘mock’
        In the expression:
          mock comprehensiveAPI (Proxy :: Proxy '[NamedContext "foo" '[]])
        In a pattern binding:
          _ = mock
                comprehensiveAPI (Proxy :: Proxy '[NamedContext "foo" '[]])
```

I would fix them, but I really do need to head in to the office. 

I haven't addressed the issues re `foreign` and `doc`, nor have I changed the overall approach, because I need a bit more context to grok @jkarni's comment.

Hope this helps. If I don't hear anything I'll keep plugging away.

jml:
Test failure was bad merge. Fixed.

jkarni:
I guess this needs to be rebased. If the `choice` changes don't degrade performance, this LGTM.

jml:
@jkarni even if we don't yet have `Foreign` support?

Also, how would I test for performance degradations? Is there a benchmark?

jkarni:
After discussion with @kosmikus on IRC, it seems like the best way of changing the new [`Router'`](https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal/Router.hs#L31) to allow `CaptureAll` is to have:

```
data CaptureType t where
   Single :: CaptureType Text
   All :: CaptureType [Text]

data Router' env a =
   ... 
   CaptureRouter (CaptureType t) (Router' (t, env) a)
   ...
```

jml:
OK. I'm going to try that in a different branch (rebasing this is too hard for my very small brain).

jml:
Some other notes from discussion with @jkarni and @kosmikus 

* `CaptureAll` should capture an empty list
* In `choice`, could use the fact that `CaptureAllRouter r` ~= `Choice (leafRouter r) (CaptureSingleRouter (CaptureAllRouter r))`
* There's no real distinction between two constructors in Router' or one with an extra tag, since uniform treatment in `choice` is not possible

soenkehahn:
Fixed through #519.

PR #350, opened 2016-01-24
RFC for embedded static resource support

wuzzeb:
Servant already uses `wai-app-static` for the very simple `serveDirectory` which just produces a `Server Raw`.  This isn't ideal because the routes for individual files are not represented in the type.  Also, `wai-app-static` has the more sophisticated ability to embed resources into the executable at compile time (full disclosure, I was the original author of this code in `wai-app-static` although since then several people have contributed to `wai-app-static`).

This pull request is my RFC for an approach to extend `serveDirectory` to be able to specify individual resources in the API type for better exposure of the static resources that are available.  Along the way, I also add support for embedding the static content at compile time.

```
type MyAPI = "static" :> "js" :> "bootstrap.js" :> EmbeddedContent "application/javascript"
        :<|> "static" :> "css" :> "bootstrap.css" :> EmbeddedContent "text/css"
        :<|> "static" :> "css" :> "mysite.css" :> EmbeddedContent "text/css"
```

At the moment I added all this into a new project, but that was just for ease of testing.  If you like this approach, some of the types should be moved into Servant.API, other code merged into servant-server, and perhaps the more exotic generators such as compiling with lesscss and ghcjs moved into a new project.  I can morph this code around once I get some feedback.

I went through about 4 iterations on the design before I found one I was happy with.  My first iteration used `wai-app-static` directly but it had to do A LOT with template haskell and as always using as little template haskell as possible is good.  I then realized that we only need a few internal functions from `wai-app-static`, so the later evolutions of the design don't depend on `wai-app-static` anymore.  I then went through several iterations on how the combinators look.  I have started using it in my own servant-based application and the combinators are working well, but I'm still open to improvements.  

I wrote haddock documentation, so for more details you should see the documentation, especially the haddock comment in Servant.Server.Embedded.hs

### DIFF
--- a/servant-static/Setup.hs
+++ b/servant-static/Setup.hs
@@ -1,0 +1,2 @@
+import Distribution.Simple
+main = defaultMain
--- a/servant-static/src/Servant/Server/Embedded.hs
+++ b/servant-static/src/Servant/Server/Embedded.hs
@@ -1,0 +1,142 @@
+-- | A module to embed static content such as javascript and CSS into the executable at compile time
+-- so that it does not need to be distributed along with the server.  In addition, this module
+-- supports processing of these resources before they are embedded, such as javascript or CSS
+-- minification.  Finally, there is a development mode which will recompute each resource on every
+-- request, so allow a simple browser refresh to reload potentially changed javascript or CSS.
+--
+-- To use this module, use 'EmbeddedContent' in your servant API definition.  For example,
+--
+-- >type MyAPI = "static" :> "js" :> "bootstrap.js" :> EmbeddedContent "application/javascript"
+-- >        :<|> "static" :> "css" :> "bootstrap.css" :> EmbeddedContent "text/css"
+-- >        :<|> "static" :> "css" :> "mysite.css" :> EmbeddedContent "text/css"
+--
+-- Then, decide on a generator for each 'EmbeddedContent'.  There are several generators which embed
+-- files directly, minifiy files, and use 3rd party tools like less and postcss.  You can also
+-- easily create your own generators.  Each generator is passed to 'embed' and will produce a
+-- haskell variable of type 'EmbeddedEntry'.  For example,
+--
+-- >#if DEVELOPMENT
+-- >#define DEV_BOOL True
+-- >#else
+-- >#define DEV_BOOL False
+-- >#endif
+-- >
+-- >embed DEV_BOOL
+-- >    [ embedFileWith uglifyJs "bootJs" "node_modules/bootstrap/dist/js/bootstrap.js"
+-- >    , embedFile "bootCss" "node_modules/bootstrap/dist/css/bootstrap.min.css"
+-- >    , embedWithPostCSS "mysiteCss" "css/mysite.css"
+-- >    ]
+--
+-- The above template haskell splice will produce the following three variables automatically (you
+-- do not need to enter anything extra):
+-- 
+-- >bootJs :: EmbeddedEntry "application/javascript"
+-- >bootCss :: EmbeddedEntry "text/css"
+-- >mysiteCss :: EmbeddedEntry "text/css"
+--
+-- These 'EmbeddedEntry's are used to create the server for the 'EmbeddedContent' endpoints.
+--
+-- >staticServer :: Server MyAPI
+-- >staticServer = bootJs :<|> bootCss :<|> mysiteCss
+--
+-- If the DEVELOPMENT define is true (I suggest you use a cabal flag), on each
+-- request the server will recompute the resource.  This means that the file will be reloaded from
+-- disk or postcss will be re-executed on each request.  Thus when the DEVELOPMENT flag is true, a
+-- browser refresh will reload and recompute the resources from disk.
+--
+-- When the DEVELOPMENT define is false, instead at compile time the resource will be loaded, the
+-- processing will occur, it will be compressed with gzip, and finally the resulting bytes will be
+-- embedded directly into the executable.  The server will then return this embedded content on each
+-- request without computing anything or loading anything from disk.
+--
+-- In addition, when DEVELOPMENT is false, the server will use etags, 304 not modified responses,
+-- and potentially Cache-Control headers to reduce the need for the client to re-request these
+-- resources.  The server will always use etags to return 304 not modified responses.  By default,
+-- these etags are the md5 hash of the content.  So for example if bootstrap is updated to a new
+-- version, the file @node_modules\/bootstrap\/dist\/js\/bootstrap.js@ will change and so the etag
+-- will be different.  Thus when the client re-requests the resource, the etag the client sends will
+-- differ from the server and so the server will return the new content.
+--
+-- Using just etags still requires the client to send a request for each resource and for the
+-- server to respond with 304 not modified the vast majority of the time.  To mitigate that, a
+-- Cache-Control header can be configured to be used to tell the client to not re-request the
+-- resource.  In this module, such a Cache-Control header is controlled by an etag query parameter
+-- on the URL.  If the client requests the resource via a URL @\/static\/js\/bootstrap.js@, no
+-- Cache-Control header is sent because when a new version of the server is released the client
+-- might need to re-download the bootstrap.js.  If instead the client requested the
+-- resource via the URL @\/static\/js\/bootstrap.js?etag=123456789@ and the etag is correct, a
+-- Cache-Control header is set to tell the client to cache the resource for one year.  When a new
+-- version of the server is released with an updated bootstrap version, the etag will change and as
+-- long as the new server uses an HTML script tag referring to a URL with the new etag, the client
+-- will download the new bootstrap version because the URL has changed.
+--
+-- The calculated etag is stored inside the 'EmbeddedEntry' created by template haskell and can be
+-- passed to 'safeLink' in order to create a link which includes the correct etag.  The function
+-- 'embeddedLink' is a simple wrapper around 'safeLink' which extracts the etag from the
+-- 'EmbeddedEntry'.
+--
+-- >bootstrapJsLink :: URI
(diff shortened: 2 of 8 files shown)


3noch:
:+1: Really cool feature!

jkarni:
Wow, great work!

An issue with having it as a separate package is that, if it's to have instances for all the core classes, it'll need to depend on all the core packages, making e.g. `wai` a transitive dependency of client code that uses it. 

One thing we've been talking about is starting a `servant-contrib` set of packages. @haskell-servant/maintainers would we want it there or in the core packages? I personally think this will be useful enough to have in the core ones. On the other hand, it's perhaps good practice to start the `-contrib` packages and have new combinators go there, and only then move them, so we have (a) some time to be more aggressive about release cycles and breaking API changes as we try things out, and (b) give people who are writing other interpreter classes (e.g. servant-swagger) some time to know what's in the pipeline.

utdemir:
+1 for me for self-contained web apps.

But, I think the part minifiers, compressors, CSS transpilers etc. belong to a different package, as you said. 

And maybe it's worth to talk about the general idea of "serving static files" here. For example, on our API's, we were serving the JS client of our server on some endpoint. But in order to set `Content-Type` header for them, we had to define the type as `Raw`:

```
myApi :: "client.js" :> Raw
serve :: Application
serve req resp = resp $ responseLBS status200 [("Content-type", "text/javascript")]
   (BL.fromStrict . T.encodeUtf8 $ jsForApi myApi)
```

In order to simplify this, I wrote these instances (simplified):

```
data ContentType (typeName :: Symbol) (subtypeName :: Symbol)

instance Accept (ContentType typeName subtypeName) 
instance MimeRender (ContentType t s) ByteString 
instance MimeRender (ContentType t s) Text 
...
```

And then I was able to define API's like this

```
myApi :: "client.js" :> Get '[ContentType "application" "javascript"] Text
serve :: EitherT ServantError IO Text
serve = return $ jsForAPI myApi
```

I told you about this, because I think the TH file embedding also fits here as a separate type, something like:

```
instance MimeRender (ContentType t s) EmbeddedContent where
myApi :: "somejs" :> Get '[ContentType "application" "javascript"] EmbeddedContent
serve = embedFile "./somejs.js"
```

Maybe this RFC can cover serving static content (but not from an external file) with a specific content-type too.

soenkehahn:
@wuzzeb: I very much like the idea of embedding static files into the executable. I've been using that feature from `wai-app-static` and it's awesome.

I do wonder however: what is the advantage of having all the static files represented in the type of the API? What `serveDirectory` was aimed at is use-cases that follow this pattern:

``` haskell
type MyJsonRestAPI =
  ... -- lots of JSON REST routes

type MyServerAPI =
  MyJsonRestApi :<|>
  Raw -- to serve some html, css and javascript that uses the API
```

I don't see how this would benefit from having all the static files explicitly represented in the type? I.e. I probably don't need to be able to generate type safe client functions to fetch all the static assets individually. Also I don't want all the css files to show up in generated documentation, be it by `servant-docs` or `servant-swagger`. Am I overlooking an important use-case?

These static file combinators would also force you to manually keep them in sync with the files on your disk. What I like very much about `serveDirectory` is that you can just throw a new file in the directory and it'll be automatically served.

Generally I'm wondering if it wouldn't be good to solve these things purely in `wai`. `serveDirectory` does not depend on anything in `servant` at all. (The implementation is just based on `wai-app-static` and although the type is given as `FilePath -> Server Raw` that's just fancy talk for `FilePath -> Application`.) Maybe it should be moved to `wai-app-static` (and re-exported in `servant`?). And maybe we do want something like `serveDirectoryEmbedded`. That way not only `servant` users, but all `wai` users would benefit.

soenkehahn:
What I could imagine is a combinator like this:
``` haskell
type MyApi =
  MyJsonRestApi :<|>
  "static" :> StaticAssets

instance HasServer StaticAssets where
  type Server StaticAssets = StaticAssetsConfig
  route = ...

data StaticAssetsConfig
  = FromDisk FilePath
  | Embedded Embedded -- from wai-app-static
```

If I think about the different interpretations, that feels like the right amount of information that should be in the API specification. So generated documentation can say something like:
```
/static (and subpaths) - static assets like html, css, javascript, images, etc.
```
And we could have:
``` haskell
instance HasClient StaticAssets where
  type Client StaticAssets = FilePath -> ExcepT ServantError IO ByteString
```
So that you can easily generate a client function that allows to fetch assets like this: `getStaticAssets "index.html"`.

jkarni:
One obvious advantage is type-safe links to static assets.

Maybe the solution is explicit routes, but with TH to generate them from a directory?

soenkehahn:
But where would you want to have type-safe links to static assets? I guess if you're generating html files within haskell... Personally I would prefer to have this functionality in a templating library or an html generation library and de-coupled from `servant`. Maybe that also already exists.

codedmart:
I tend to agree with @soenkehahn on this. Awesome work from @wuzzeb, but I don't see the advantages of this yet to be in core. Seems more like a specified use case that would be great as a contrib lib that someone can pull in when needed.

wuzzeb:
Sorry, was busy the past few days.

My first approach was much closer to `serveDirectory`.  I didn't call it `serveDirectoryEmbedded` but that was essentially what it was.  I still have the code around in my git history (not on github, for github I did a rebase) so we could go to it if you like.  The main reason I changed to include it all in the route is ETags and the HasLink instance.  

For static resources, the recommendation is to tell the browser to cache them and then change the URL used to access.  For example, see [here for example](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#defining-optimal-cache-control-policy).  This means that when the server refers to a static resource, it must change the URL each time the content changes.  `wai-app-static` does that via including the ETag as a query-param.  What I mean is the server must generate a link such as

```
<link rel="stylesheet" href="/static/bootstrap.css?etag=12345678"/>
```

and the etag must change each time the content changes.  So in order for the server to generate such links, it needs to know the computed etag.  With my first approach which used `serveDirectoryEmbedded`, the template haskell had to create a variable containing a string with the etag, but then that string had to be combined with the full path.  Bringing each resource out into the actual route allows `HasLink` instances which easily allow the server to generate the correct link with the correct etag.

jkarni:
Changing the url on file changes for better caching would be amazing - I've been impressed by that feature of yesod/wai-app-static, and would be pretty happy to see it in `servant` too.

soenkehahn:
@wuzzeb: Btw, I would love to see something like `serveDirectoryEmbedded` as a contribution to `wai-app-static`.

soenkehahn:
It seems to me that perfect caching (with `etag` query params, like @wuzzeb describes) is ideally done in a separate library on top of `wai`. I think the fact that `HasLink` does not provide perfect caching out of the box is not worth the complexity at the type-level.

I'm in favor of not merging (and closing) this PR. 

If however `HasLink` doesn't play nice with an external library that supports perfect caching, then we should fix that, of course.

rashadg1030:
Will this ever be merged into master? Seems like something I could use on a project I'm working on.

servant-static/Setup.hs
```diff
@@ -0,0 +1,2 @@
+import Distribution.Simple
+main = defaultMain
```
tchoutri, line 2: Unneeded, this is the default

servant-static/src/Servant/Server/Embedded.hs
```diff
@@ -0,0 +1,142 @@
...
+-- request, so allow a simple browser refresh to reload potentially changed javascript or CSS.
+--
+-- To use this module, use 'EmbeddedContent' in your servant API definition.  For example,
+--
+-- >type MyAPI = "static" :> "js" :> "bootstrap.js" :> EmbeddedContent "application/javascript"
+-- >        :<|> "static" :> "css" :> "bootstrap.css" :> EmbeddedContent "text/css"
+-- >        :<|> "static" :> "css" :> "mysite.css" :> EmbeddedContent "text/css"
+--
+-- Then, decide on a generator for each 'EmbeddedContent'.  There are several generators which embed
+-- files directly, minifiy files, and use 3rd party tools like less and postcss.  You can also
```
jhrcek, line 14:
```suggestion
-- files directly, minify files, and use 3rd party tools like less and postcss.  You can also
```

servant-static/src/Servant/Server/Embedded.hs
```diff
@@ -0,0 +1,142 @@
+-- | A module to embed static content such as javascript and CSS into the executable at compile time
+-- so that it does not need to be distributed along with the server.  In addition, this module
+-- supports processing of these resources before they are embedded, such as javascript or CSS
+-- minification.  Finally, there is a development mode which will recompute each resource on every
+-- request, so allow a simple browser refresh to reload potentially changed javascript or CSS.
```
jhrcek, line 5:
```suggestion
-- request, to allow a simple browser refresh to reload potentially changed javascript or CSS.
```

servant-static/src/Servant/Server/Embedded.hs
```diff
@@ -0,0 +1,142 @@
...
+--
+-- >staticServer :: Server MyAPI
+-- >staticServer = bootJs :<|> bootCss :<|> mysiteCss
+--
+-- If the DEVELOPMENT define is true (I suggest you use a cabal flag), on each
+-- request the server will recompute the resource.  This means that the file will be reloaded from
+-- disk or postcss will be re-executed on each request.  Thus when the DEVELOPMENT flag is true, a
+-- browser refresh will reload and recompute the resources from disk.
+--
+-- When the DEVELOPMENT define is false, instead at compile time the resource will be loaded, the
```
jhrcek, line 47:
```suggestion
-- When the DEVELOPMENT define is false, instead the resource will be loaded at compile time, the
```

jhrcek (commented):
Just few suggestions from reading the docs..

PR #351, opened 2016-01-24
servant-foreign: don't constrain ForeignType to Text

purcell:
With these changes, we allow a user-specified type to represent the foreign type of Haskell types encountered in the API. This lets users map Integer, Date etc. to representations other than Text, and have those representations available in the returned list of Req.

For example, we might want to map a type which has an instance of Generic to both a foreign type name and a class declaration for that foreign type such that it can encode/decode itself to JSON. The previous limitation to a single Text output prevented this case.

This is all implemented by explicitly parameterising various types with the desired output type: for `NoTypes`, this is `()`. For typed languages, the implementor might want it to be a custom record type containing necessary "import" directives, the foreign type name, and other information about the encountered Haskell type. All tests still pass as far as I can tell. The `(Text, Text)` tuple for which `Arg` was an alias has been promoted to a abstract data type: this could be undone if preferred.

See https://github.com/purcell/servant-elm/issues/6 for some related discussion -- this change was prompted by an inability to use the existing code to generate strongly typed Elm bindings for Servant without needing to substantially duplicate the code in `servant-foreign`.

Posting this here for concrete discussion, happy to rework if necessary.

### DIFF
--- a/servant-foreign/src/Servant/Foreign/Internal.hs
+++ b/servant-foreign/src/Servant/Foreign/Internal.hs
@@ -27,15 +27,6 @@ newtype FunctionName = FunctionName { unFunctionName :: [Text] }
 
 makePrisms ''FunctionName
 
-newtype ForeignType f = ForeignType { unForeignType :: f }
-
-deriving instance Show f => Show (ForeignType f)
-deriving instance Eq f => Eq (ForeignType f)
-deriving instance IsString f => IsString (ForeignType f)
-deriving instance Monoid f => Monoid (ForeignType f)
-
-makePrisms ''ForeignType
-
 newtype PathSegment = PathSegment { unPathSegment :: Text }
   deriving (Show, Eq, IsString, Monoid)
 
@@ -43,7 +34,7 @@ makePrisms ''PathSegment
 
 data Arg f = Arg
   { _argName :: PathSegment
-  , _argType :: ForeignType f }
+  , _argType :: f }
 
 deriving instance Eq f => Eq (Arg f)
 deriving instance Show f => Show (Arg f)
@@ -130,8 +121,8 @@ data Req f = Req
   { _reqUrl        :: Url f
   , _reqMethod     :: HTTP.Method
   , _reqHeaders    :: [HeaderArg f]
-  , _reqBody       :: Maybe (ForeignType f)
-  , _reqReturnType :: ForeignType f
+  , _reqBody       :: Maybe f
+  , _reqReturnType :: Maybe f
   , _reqFuncName   :: FunctionName
   }
 
@@ -140,8 +131,8 @@ deriving instance Show f => Show (Req f)
 
 makeLenses ''Req
 
-defReq :: Req Text
-defReq = Req defUrl "GET" [] Nothing (ForeignType "") (FunctionName [])
+defReq :: Req ftype
+defReq = Req defUrl "GET" [] Nothing Nothing (FunctionName [])
 
 -- | To be used exclusively as a "negative" return type/constraint
 -- by @'Elem`@ type family.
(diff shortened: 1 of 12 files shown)


purcell:
Note that it would also have been possible to use the `lang` parameter as the type of the translated foreign values, such that `NoTypes` would become `()`. I chose not to do this, because I could imagine several different backends using `Text`, but I guess that users could tag their return types, e.g. `data JavaType = JavaType Text`...

jkarni:
I see the basic idea. 

> For typed languages, the implementor might want it to be a custom record type containing necessary "import" directives, the foreign type name, and other information about the encountered Haskell type.

It seems like import directives might come up elsewhere. I'm wondering whether having `HasForeign`, `HasForeignType` etc shouldn't have state. @MaxOw and @dredozubov know much more about `servant-foreign` than I, though, so I defer to them.

dredozubov:
I like the general idea, but i'm currenly on vacation and can't seem to find the time to properly review the request, sorry. I'll try to get back to it later.

purcell:
Yep, in the PR I referenced previously, @MaxOw was very helpful and committed some changes into a branch in his fork which would basically support my use case. Those changes sketched out a whole new set of `RenderForeignType` instances and an extra `typesFromAPI` entry point.

But that approach effectively has to destructure a Servant API type in the exact same way that the `listFromAPI` machinery does, and it seemed to me that not constraining `ForeignType` to `Text` would have the same effect, although it would slightly change the existing API.

My plan is that with these changes I've outlined, the generation of a full set of code for a typed language would involve not just mapping the `[Req ftype]` returned by `listFromAPI` into a bunch of `Text` code fragments, but rather folding all that info into a collection of concrete `ftype`s mentioned in each `Req ftype` tree of nested datatypes, together with a chunk of endpoint-specific info for each `Req` in the list. And the folded information would then be used to generate code for imports, code to declaration and JSON-serialise non-built-in datatypes, and code for each `Req` endpoint.

My disclaimer is that I haven't actually written that client code yet. An alternative might even be to allow an arbitrary monadic context for `listFromAPI` and `HasForeignType` etc. so that the types encountered while destructuring the API type can be directly collected as desired.

dredozubov:
@purcell got back from vacation, sorry for the long wait. I'm still not completely sure how the servant-foreign should change to address typed languages, so i'm thinking of writing a prototype of codegen tool for some typed language now(probably rust, just to check it out) to experience it myself.

purcell:
@dredozubov Yep, good plan! I'll be keen to see how you get on. :-)

purcell:
@dredozubov I see you started to do some work on a basic version of `servant-rust` - yay! I don't think you've yet encountered the case which stumped me, namely where the types in the Servant API don't already match native types in Rust, particularly where those types are compound types like JSON records. Looking forward to seeing how you get on. :-)

dredozubov:
@purcell At this point i feel like i should properly convey my hesitation to you. :) I see the rationale, but i slightly dislike the current design. It misses either functional dependency or an associated type family in the `HasForeignType` class itself. Associated type family seems like a better choice, because the notion of having instances of `HasForeignType X FType1 Z` and `HasForeignType X FType2 Z` simultaneous feels awkward. Another thing i don't particularly like is `ftype` type parameters to a lot of types, it probably should may be existentially quantified instead. Also state.

purcell:
Yep, I was far from convinced that this was the best approach, but the other proposed alternative felt odd to me in other ways, which is why I went as far as filing this PR. I'd be really happy if you were to find a better approach that what I've sketched out here!

mattjbray:
Hi @dredozubov, @purcell. I don't know if this is of use to anyone, but I went ahead and refactored my own version of `servant-elm` based on this proposal, just to see what it would look like. [Here's the code](https://github.com/mattjbray/servant-elm/blob/4752452ca447c260b6a0156f7be6f2d50c7c8423/src/Servant/Elm/Foreign.hs).

This allows me to keep track of everything I need to generate the Elm functions:

* type names to use in the function signatures (and their definitions)
* JSON decoder names (and their definitions)
* JSON encoder names (and their definitions)

I understand the current PR is unlikely to be merged, but I thought it might be useful to have a concrete use case to take into account.

purcell:
@mattjbray Oh, cool. That looks pretty tidy.

dredozubov:
@mattjbray @purcell Ok, i'm convinced if you both support this change. If someone is willing to rebase this PR over the latest changes https://github.com/haskell-servant/servant/pull/378(which'll make servant-foreign and libraries using it more type-safe), i'll merge it.

purcell:
Done! :-)

- Now that the foreign type is a type parameter, I've removed the `ForeignType` newtype wrapper again, and also the unused `HasNoForeignType` alias.
- The "null request" defined by the existing `defReq` code seems a bit hokey, because it forces `reqReturnType` to be a `Maybe`, when (I think) it really doesn't need to be.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -79,47 +80,59 @@ data ArgType
...
-  { _reqUrl        :: Url
+data Req f = Req
+  { _reqUrl        :: Url f
   , _reqMethod     :: HTTP.Method
-  , _reqHeaders    :: [HeaderArg]
-  , _reqBody       :: Maybe ForeignType
-  , _reqReturnType :: ForeignType
+  , _reqHeaders    :: [HeaderArg f]
+  , _reqBody       :: Maybe f
+  , _reqReturnType :: Maybe f
```
dredozubov: i agree, having `Maybe` here feels pretty wonky

dredozubov:
LGTM this.

Is anyone else want to share an opinion?

servant-foreign/src/Servant/Foreign/Internal.hs
purcell: On reflection, previously it was getting set to `""` by `defReq`, which is fairly equally odd.

dredozubov:
@purcell thanks for working on it!

purcell:
Yay! Thanks for your support.

apatil:
Awesome! Looking forward to trying this out.

Issue #338, opened 2016-01-18
Use Read the Docs?


fizruk:
I have started a branch ([fizruk/read-the-docs](https://github.com/haskell-servant/servant/tree/fizruk/read-the-docs)), organizing servant documenation for [readthedocs.org](https://readthedocs.org).

So far I've borrowed a config from Stack and put `README.md` and `CONTRIBUTING.md` links.
Here's how it looks: http://haskell-servant.readthedocs.org/en/fizruk-read-the-docs.

I find Read the Docs somewhat more convenient for documentation/tutorials than [haskell-servant.github.io](http://haskell-servant.github.io):
- it builds and updates automatically;
- it can serve multiple versions of documentation (so people can lookup documentation for `0.4` as well as for `HEAD`);
- it has search and better navigation.

Perhaps, [haskell-servant.github.io](http://haskell-servant.github.io) should only retain blog posts which do not exactly fit as documentation.

@jkarni, @alpmestan waiting for your approval to continue work on this.

jkarni:
@soenkehahn is working on getting our tutorial etc. to be literate markdown ([here](https://github.com/haskell-servant/haskell-servant.github.io/pull/37), I believe it's pretty much done), so we can test that everything compiles. That's pretty useful, I think. Could we get that behaviour with RTD?

fizruk:
@jkarni I don't see why not, we can simply use `.md` links to `.lhs` files (which are markdown literate haskell files IIUC).

soenkehahn:
@fizruk: Sounds great. But yes, this should be coordinated with https://github.com/haskell-servant/haskell-servant.github.io/pull/37.

Is your plan to move the tutorial and other things from https://github.com/haskell-servant/haskell-servant.github.io (except the blog posts) into the `servant` repo, directory `doc` then?

fizruk:
@soenkehahn yes, that would be the plan.

The idea is that tutorial and explanatory articles actually belong in servant repo, not servant's website:
- it should be in sync with the repo's content (at least at release points);
- it would really help to be able to review built documentation before the release;
- I think it would help servant's users to look at the documentation/tutorial relevant to their servant version (I don't expect everyone to migrate fast with every new version of servant).

Also, if we compile code in documentation together with the servant code itself, we would notice breakages in documentation early and it won't be such a daunting task to fix/update the docs.

soenkehahn:
@fizruk: Sounds all very good to me.

[...]

PR #355, opened 2016-01-25
Tutorial

soenkehahn:
These are the two tutorial branches merged. There's still lots of work to do.

I wrote a small script to convert from birdtrack syntax to backtrack-markdown-syntax, since sphinx doesn't understand `.lhs`: `servant-examples/tutorial/convert.hs`. I haven't committed any of the changes it produces, since I thought that would be more confusing than helpful at this point.

One problem is that the sphinx syntax highlighting for haskell seems to choke on fancier features like `TypeOperators`. :( I'm currently out of ideas, what to do about that.

@jkarni @fizruk


soenkehahn:
Fixes #338.

soenkehahn:
@jkarni: you added a top-level stack file. Is the plan to migrate travis to `stack` then? If not, I think the the stack file should be inside `doc/tutorial`.

soenkehahn:
Fixes #360.

soenkehahn:
I'd appreciate reviews. Apart from that, this is ready to be merged. @haskell-servant/maintainers

jkarni:
> @jkarni: you added a top-level stack file. Is the plan to migrate travis to stack then? If not, I think the stack file should be inside doc/tutorial.

@soenkehahn Not sure what you mean. It wasn't my intention to migrate to stack. However. I am in favour of not having a separate stack file for this, since:
   - This is what we've been doing with all other packages
   - This way it's easier to check locally (for those using stack) that the tutorial also compiles fine (I'd `stack test` to basically be all I have to run to know everything is fine, and CI should pass, so that it's intuitive for people what they have to do to know that everything worked)

jkarni:
I'm going to rebase against master, and also squash some of my commits. Force pushes coming up in the next ~1h.

jkarni:
LGTM (after CI)

PR #357, opened 2016-01-26
Remove memoReqBody.

### LINKED ISSUE #356: Remove memoReqBody
Currently `memoReqBody` keeps the entire request body in memory. This was originally there because, prior to @kosmikus 's work, we *had* to backtrack requests against the API. We decided to keep it there because it makes the semantics of `ReqBody` somewhat nicer and we weren't sure how things would play out, but I think by now everyone agrees having the request in memory is not acceptable.

I think this should mean that bad bodies are `FailFatal`, so that if two `ReqBody`s occur as in:

```
type API = ReqBody '[JSON] Int :> Post ..
          :<|> ReqBody '[JSON] Bool :> Post ...
```

The second one never gets tried.

jkarni:
Fixes #356

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -158,7 +158,7 @@ methodCheck method request
 acceptCheck :: (AllMime list) => Proxy list -> B.ByteString -> IO (RouteResult ())
 acceptCheck proxy accH
   | canHandleAcceptH proxy (AcceptHeader accH) = return $ Route ()
-  | otherwise                                  = return $ Fail err406
+  | otherwise                                  = return $ FailFatal err406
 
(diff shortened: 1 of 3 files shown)


codedmart:
This LGTM, I am going to have a need for this soon.

soenkehahn:
I think we should have test cases for:
- streaming working properly and
- an API with overlapping `ReqBody` routes (and the behavior we want for that).

Without that I'm not really convinced this is working.

jkarni:
Added a test for 400 in overlapping bodies. This PR doesn't propose to add streaming - just to remove the memoized body - I'm not sure how to test streaming without implementing it.

soenkehahn:
Very fair. LGTM.

PR #362, opened 2016-01-28
Various small and hopefully uncontroversial changes.

kosmikus:


soenkehahn:
Regarding eta-reducing the verb-synonyms: I'm not sure I find this easier to understand. Of course it's less verbose. (But I could go either way.)

Otherwise LGTM.

kosmikus:
@soenkehahn  I'm generally in favour of eta-reducing type synonyms if in doubt because it actually makes them more flexible, i.e., it allows partial applications of the type synoynm where the eta-expanded version does not. I don't have a use case, but it's somewhat annoying if a user wants a partial application and is prevented for no other reason than the fact that an unnecessarily eta-expanded version has been used.

soenkehahn:
@kosmikus: Good point. I was surprised that this is the actual behavior of type synonyms...

Issue #367, opened 2016-02-03
Issues found on the new tutorial


rwobben:
I did check the new tutorial for errors
First I clone the repo
change to the tutorail repo
did pip install -r requirements..txt 
make html 

And saw these things that did not look well to me 

1) on the servant tutorial page I saw this :    

 [this website’s issue tracker](http://github.com/haskell-servant/haskell-servant.github.io/issues).

   the servant packages: [haskell-servant/servant](https://github.com/haskell-servant/servant)
    the website (including this tutorial): [haskell-servant/haskell-servant.github.io](https://github.com/haskell-servant/haskell-servant.github.io/)
    Feel free to use the issue tracker (or to send PRs!) on the website’s repository to give feedback and suggestions about this tutorial

All links seems not good to me. 

on the Serving a Api I saw this : title: Serving an API toc: true

jkarni:
Fixed in #368 - thanks for the report!

rwobben:
Sorry but all the problems are not solved. 
The last one . A wrong title is not solved.

rwobben:
Found the problem : 

On the wrong place I see this in the code : 

   ---
   title: Serving an API
    toc: true
    ---

On all the other pages I see this : 

# A web API as a type

soenkehahn:
@rwobben: Generally the tutorial needs a read-through before the `0.5` release as some things have changed. I bet there's also more little formatting mistakes like this. Fixing the tutorial is tracked here: #360.

Is everything from this issue solved now?

rwobben:
yes, this can be closed.

PR #368, opened 2016-02-06
Use rst links syntax (it's not a markdown)

phadej:
Resolves https://github.com/haskell-servant/servant/issues/367


jkarni:
Thanks!

PR #369, opened 2016-02-09
Changed so two titles are displayed correctly

rwobben:
Solves #367

### DIFF
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -1,7 +1,4 @@
----
-title: Serving an API
-toc: true
----
+# Serving an API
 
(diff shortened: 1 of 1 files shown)


alpmestan:
LGTM. @soenkehahn can you check this is OK?

soenkehahn:
Can't be worse than before. :)

soenkehahn:
Merged.

soenkehahn:
@rwobben: Thanks for the patch.

Why did you say this was about two titles? I only saw one title changed.

rwobben:
Yes, I thought I was 2 tityles but it was one title two times.
First  I made it a good looking title and later on I deleted the word title in the title so it looks more as the other one.

PR #372, opened 2016-02-11
Make servant-foreign code nicer

dredozubov:
I felt like i have to do this before addressing a few other issues with servant-foreign

Basically it's:
* non-messy imports
* got rid of most long lines (>80 chars)
* prisms for sum types and newtypes(we use lens anyway, so why not)
* consistent indentation

Also i made some changes to demonstrate my point from https://github.com/haskell-servant/servant/pull/345#issuecomment-182807730

[Review on Reviewable](https://reviewable.io/reviews/haskell-servant/servant/372)

### DIFF
--- a/servant-foreign/src/Servant/Foreign.hs
+++ b/servant-foreign/src/Servant/Foreign.hs
@@ -1,36 +1,50 @@
 -- | Generalizes all the data needed to make code generation work with
 -- arbitrary programming languages.
 module Servant.Foreign
-  ( HasForeign(..)
-  , HasForeignType(..)
+  ( ArgType(..)
+  , HeaderArg(..)
+  , QueryArg(..)
+  , Req(..)
   , Segment(..)
   , SegmentType(..)
+  , Url(..)
+    -- aliases
+  , Path
+  , ForeignType
+  , Arg
   , FunctionName
-  , QueryArg(..)
-  , HeaderArg(..)
-  , ArgType(..)
-  , Req
+    -- lenses
+  , reqUrl
+  , reqMethod
+  , reqHeaders
+  , reqBody
+  , reqReturnType
+  , reqFuncName
+  , path
+  , queryStr
+  , argName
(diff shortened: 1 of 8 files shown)


dredozubov:
If you too feel like it looks more readable this way, maybe i'll go through other packages with this PR

fizruk:
Nice! :+1: 

Would be even nicer if you add a proper haddock documentation!
Would be magical if you also document design choices and intended usage!

No pressure though :)

dredozubov:
@fizruk, i plan to do that after i'll properly address https://github.com/haskell-servant/servant/pull/351 ;) I feel like there may be some design changes on its way.

purcell:
This is all great, but it'd be a big job to rebase #351 on top of it. Let me know if you want a hand with anything.

jkarni:
:+1: - @dredozubov  can we merge?

dredozubov:
@jkarni sure, but i also wonder if it's worth doing repo-wise. I'd be more than happy to see it, but it contradicts the 
> Though we aren't sticklers for style
from https://github.com/haskell-servant/servant/blob/master/CONTRIBUTING.md

dredozubov:
@purcell i started doing `servant-rust` to test the design, but my day work consumes the most of my time, so i kinda feel that i don't want to rush #351, it can always land in 0.5.1 or something

purcell:
Fair enough. Do you think it'd be helpful if I try to rebase #351 on top of this now? Or do you already know my approach there wouldn't be optimal?

dredozubov:
@purcell i'll be able to tell a bit later, don't hurry with rebasing it. Are you on irc btw?

purcell:
> i'll be able to tell a bit later, don't hurry with rebasing it. Are you on irc btw?

Ah, no, sorry, but I'm on github *a lot* and happy to discuss. And happy to pop onto IRC at some point if it'd help.

PR #375, opened 2016-02-17
Basic Authentication Support

aaronlevin:
This pull-request takes the basic authentication code from https://github.com/haskell-servant/servant/pull/341 . It uses none of the generalized authentication framework, including the `AuthReturnType` type family used in the previous request. 

This should provide a very simple example of using `Config` in conjunction with a authentication pattern where the only user-defined type are server specific (e.g. `servant-client` only concerns itself with the concrete `BasicAuthData` datatype)

### DIFF
--- a/servant-client/CHANGELOG.md
+++ b/servant-client/CHANGELOG.md
@@ -8,6 +8,7 @@ HEAD
 * Use `http-api-data` instead of `Servant.Common.Text`
 * Client functions now consider any 2xx successful.
 * Remove matrix params.
+* Added support for Basic authentication
 
 0.4.1
 -----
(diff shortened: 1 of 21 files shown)


jkarni:
This LGTM! We're removing `servant-examples` in the tutorial branch though (too many different places for documentation, and that's not the most illuminating one). Though in my opinion we could merge with it, since it'll help figure out how to incorporate it in the tutorial.

arianvp:
Do we want instances for `servant-js` as well? They shouldn't be too hard to add. It took only a few lines of code in the previous version of Basic authentication.

aaronlevin:
@arianvp yes, please!

aaronlevin:
just rebased on top of `master`. can we please merge this? /cc @soenkehahn @arianvp @fizruk

arianvp:
LGTM

aaronlevin:
@arianvp ok, I will rebase on master and then merge the changes.

codedmart:
This LGTM

Issue #377, opened 2016-02-17
Get rid of code duplication in Servant.JS.Internal submodules


dredozubov:
I believe it is a fairly reasonable thing to do and it looks pretty straightforward with the help of the test-suite.

jkarni:
@dredozubov is this still relevant?

dredozubov:
@jkarni i don't really use servant-js nowadays, so i'm probably not the best person to ask anymore. Also judging by the lack of feedback, it's safe to assume that no one really cares for this kind of improvement.

soenkehahn:
Alright, closing.

PR #378, opened 2016-02-17
Utilize newtypes in servant-foreign

dredozubov:
I've started working on https://github.com/dredozubov/servant-rust and it seems like `servant-foreign` is currently stringly-typed. This patch makes use of newtypes and some convenience lenses to make user's lives a bit happier. Also https://github.com/haskell-servant/servant/issues/377

[Review on Reviewable](https://reviewable.io/reviews/haskell-servant/servant/378)

### DIFF
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -26,7 +26,9 @@ source-repository head
   location: http://github.com/haskell-servant/servant.git
 
 library
-  exposed-modules:     Servant.Foreign, Servant.Foreign.Internal
+  exposed-modules:     Servant.Foreign
+                     , Servant.Foreign.Internal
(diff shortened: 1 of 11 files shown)


arianvp:
You'll have to import `Data.Monoid` explicitly for  GHC version 7.8.4 as it's not in Prelude

dredozubov:
@arianvp yep, you're right. Thanks for reminding me btw, totally forgot about it!

phadej:
Should we consider using `base-compat`?

dredozubov:
@phadej it can be a pretty good idea, i've seen similar problem in other PRs as well

soenkehahn:
I'm very much in favor of using `base-compat`.

fizruk:
I am also fine with `base-compat`. :+1:

PR #385, opened 2016-02-18
Make servant compilable with GHC8

phadej:
Part of https://github.com/haskell-servant/servant/issues/384

GHC8.0 is more conservative when deducing context

Issue #384, opened 2016-02-18
Make compilable with GHC 8.0


phadej:
Some of `servant-` packages' dependencies are already GHC8.0 ready, so the next release should either have restricted `base <4.9` or verified by other means to work with `GHC8.0`.

I edited (also) the last package metadata on Hackage: https://hackage.haskell.org/package/servant-0.4.4.6/revisions/

Related https://github.com/haskell-servant/servant/issues/374

jkarni:
Done in #388 (and thanks a lot!)

alpmestan:
Thanks @phadej !

PR #388, opened 2016-02-20
Compatibility with GHC 8.0

bgamari:
This builds with the current state of the `ghc-8.0` branch.

I'm not entirely certain about the `ContentTypes` change. Someone more familiar with this code should have a closer look before merging.


bgamari:
This should address #384.

phadej:
see also https://github.com/haskell-servant/servant/pull/385

kosmikus:
Thanks. Merging this as a partial step toward GHC-8 compatibility that's definitely necessary on the way. Leaving #385 open for now to track CI issues.

PR #389, opened 2016-02-21
Add basic auth to tutorial.

aaronlevin:
This adds basic authentication information to the tutorial. I haven't checked if it compiles. We should probably merge https://github.com/haskell-servant/servant/pull/375 first :)

At least you can read the text now and give me any feedback.

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,1 +1,0 @@
-:set -itest -isrc -packagehspec2


doc/tutorial/Server.lhs
```diff
@@ -1220,6 +1220,112 @@ $ curl http://localhost:8081/b
...
+api = Proxy
+
+-- | 'BasicAuthCheck' holds the handler we'll use to verify a username and password.
+authCheck :: BasicAuthCheck User
+authCheck =
+  let check (BasicAuthData username password) =
+        if username == "servant" && password == "server"
+        then return (Authorized (User "servant"))
+        else return Unauthorized
+  in BasicAuthCheck check
```
3noch: I'm out-of-the-loop enough that this question may be answered somewhere else, but from this particular example I'm not sure how I would write an authorization algorithm that is predicated on any data in the request. I.e. Authorization rules will depend on the end-point but this function only works with `BasicAuthData`. So how could I write:

```haskell
type ViewEntity = "entity" :> Capture "id" String :> BasicAuth "my-app" User :> Get '[JSON] Entity
```

where authorization depends on *which* Entity `id` is being viewed. Authentication can often be answered "globally" (without regard to a particular request), but authorization often requires more information.

3noch: Perhaps this functionality leaves the problem of complex authorization up to the route handler itself? If that's the case, it might be less confusion to just stick with the terms "Authenticated"/"Unauthenticated" in the example.

doc/tutorial/Server.lhs
```diff
@@ -1220,6 +1220,112 @@ $ curl http://localhost:8081/b
...
+{- Sample session
+$ curl -XGET localhost:8080/public
+[{"somedata":"foo"},{"somedata":"bar"}
+$ curl -iXGET localhost:8080/private
+HTTP/1.1 401 Unauthorized
+transfer-encoding: chunked
+Date: Thu, 07 Jan 2016 22:36:38 GMT
+Server: Warp/3.1.8
+WWW-Authenticate: Basic realm="foo-realm"
+$ curl -iXGET localhost:8080/private -H "Authorization: Basic c2VydmFudDpzZXJ2ZXI="
```
3noch: To someone unfamiliar with the Base64 business going on here, it could be helpful to explain it, or just supply auth in the URL: `servant:server@localhost/private`

3noch:
Can't wait to use this!

doc/tutorial/Server.lhs
aaronlevin: this kind of functionality usually exists in `Authorization` and not `Authentication`. Authentication is the act of determining that an entity is the entity they say they are. Access to resources falls within authorization and has not been implemented, though we are thinking about how to do this!

3noch: Cool. Do you think using "Authentication" in your example would make this situation more obvious? Since you use `Authorized` I thought you were actually doing the job of authorization.

aaronlevin: In the case of Basic Authentication, which has been standardized, the two are conflated (so basic authentication *is* authorization by virtue of the fact that it doesn't allow route-based logic outside of its concept of `realm`). Unfortunately we decided to stick with what was stated in the standard and use those names.

aaronlevin:
ugh. please ignore. I messed up all the rebasing.

PR #391, opened 2016-02-22
[RFC] HasBefore: bind an effect before a Server

np:
While writing handlers (assume a result type of the form `ServerT api m`), I wanted to run an effect (assume an effect of type `m a`), bind its result and make it available to all the handlers of the sub-layout `api`.

Since the workarounds looked quite ugly, I'm proposing a type class `HasBefore` (name suggestions are welcome) which features a method `before`:

```
before :: Monad m => Proxy layout -> m a -> (a -> ServerT api m) -> ServerT api m
```

As one can see from its type it `before` is some kind of bind but where `ServerT api m` is not a monad.

The reasons why this is in a different type class than `HasServer` are:

* There is no meaningful instance for `Raw`.
* This is forward compatible.

### DIFF
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -37,6 +37,7 @@ library
     Servant
     Servant.Server
     Servant.Server.Internal
+    Servant.Server.Internal.Before
     Servant.Server.Internal.Config
     Servant.Server.Internal.Enter
     Servant.Server.Internal.Router
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -18,6 +18,9 @@ module Servant.Server
     HasServer(..)
   , Server
 
+  , -- * Before hook for all standard combinators (except Raw)
+    HasBefore(..)
+
     -- * Enter
     -- $enterDoc
 
@@ -86,6 +89,7 @@ module Servant.Server
 import           Data.Proxy                    (Proxy)
 import           Network.Wai                   (Application)
 import           Servant.Server.Internal
+import           Servant.Server.Internal.Before
 import           Servant.Server.Internal.Enter
 
 
--- a/servant-server/src/Servant/Server/Internal/Before.hs
+++ b/servant-server/src/Servant/Server/Internal/Before.hs
@@ -1,0 +1,84 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE PolyKinds           #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE TypeOperators       #-}
+
+module Servant.Server.Internal.Before where
+
+import           Data.Typeable
+import           GHC.TypeLits                (KnownSymbol)
+import           Network.HTTP.Types          hiding (Header, ResponseHeaders)
+import           Servant.API                 ((:<|>) (..), (:>), Capture,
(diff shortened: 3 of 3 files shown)


fisx:
`Delayed` is related, but I think it serves a different purpose.  From what I understand, `Delayed` shuffles effects to the end of an end-point, whereas `Before` will trigger them at arbitrary places in the API tree.

fisx:
Even more related, the more I think about it: `enter`.  It can be used to write a handler for a sub-API in a custom monad and translate that handler into the type of the super-API handler.  I can't see how, but this may be a way to trigger effects half-way down an API tree.

np:
Another way to see it is that this can help making `ServerT api m` more monadic.

As I understand it `ServerT`, is not a monad transformers for two reasons:

(1) it as no return type
(2) it would not make sense for `Raw`

Addressing (1) is a matter of realizing that the return type is always equal to `Void`. Therefore would makes no sense to have `lift` such as:

```
lift :: m Void -> ServerT api m Void
```

As `m Void` is not really useful, however binding an effect before an handler works and make sense.

fizruk:
@np return type is not `Void`. Return type is driven by what `Verb` returns.
E.g. `Post '[JSON] a` has `a` as a return type.

np:
@fizruk you're right, I'm editing my previous comment.

np:
Here is my best attempt so far at using `enter` instead:

```
beforeReader :: forall m api e.
    (Enter (ServerT api (ReaderT e m)) (ReaderT e m :~> m) (ServerT api m), Monad m) =>
    Proxy api -> m e -> ServerT api (ReaderT e m) -> ServerT api m
beforeReader Proxy m f = enter (Nat rp) f
  where
    rp :: forall a. ReaderT e m a -> m a
    rp rm = m >>= runReaderT rm
```

Pro: no new type class

Cons:

* Ugly `Enter` constraint
* You probably get to use `enter` again to work with this `ReaderT`

aaronlevin:
Could you achieve this with `Config`?

np:
I don't see how it could help.

fizruk:
@np theoretically, you can put things into `Config` for the latter combinators or the handler to use.
That would be a more general and nice approach than a custom `HasBefore` type class.
See https://github.com/haskell-servant/servant/pull/321#issuecomment-170625407 for some examples of permissions logic put into the API which rely on writeable `Config`.

You can use `WithNamedConfig` to get value available from the handler.

Unfortunately, you can't do that right now (see #392 for a suggestion to improve the situation).

jkarni:
> While writing handlers (assume a result type of the form ServerT api m), I wanted to run an effect (assume an effect of type m a), bind its result and make it available to all the handlers of the sub-layout api.

This can be done both by using a new combinator, or with `enter`. The one covers cases where the moment in which the action is run matters, and happens before a handler is picked; the other runs after a handler is picked, and is, I think, a much more general `before`. I'm loathe to include yet another solution to something that seems well-covered, but perhaps I'm missing something.

> Cons:
>  - Ugly Enter constraint
>  - You probably get to use enter again to work with this ReaderT

I think if you just write the natural transformation that you want (probably  `\x -> Nat (\r -> x >>= \y -> runReaderT r y)`), compose them with `Control.Category..`, and only run them when actually needed, these won't be serious problems. If they are, please do let us know.

kosmikus:
What is the status of this?

It would help me if I had a more concrete use case for this functionality. Then I could hopefully better judge what the right approach to solve the problem is.

But it seems to me that adding a combinator is not really desirable here, as this only affects (a part of) the handlers for the server, and no other interpretations. Or am I wrong? On the other hand, if this is really doable with just `before` as proposed, it's also unclear whether we need this in servant itself, as it can easil be implemented externally in the (probably rare?) situations where it's needed.

jkarni:
Closing since there's no activity, and it's not yet clear whether we want this. Please re-open if you think closing was a mistake.

Issue #360, opened 2016-01-27
[tutorial] Migrate to current master.


soenkehahn:
The tutorial is currently broken for `servant-0.5`. Before the release, we have to fix that.

We should probably fix #358 first.

soenkehahn:
Should be done in #355.

soenkehahn:
Now the whole tutorial is at least compiling. We should read through it to make sure that the comments are also up-to-date. Here's a checklist that hopefully helps with that:

- [ ] ApiType.lhs
- [ ] Client.lhs
- [ ] Docs.lhs
- [ ] Javascript.lhs
- [ ] Server.lhs

(cc: @jkarni @kosmikus)

rwobben:
Found another problem with the text. 
I did  serving a api -> a first example. 
I copy/paste everything and did a stack build. 

Then I saw this error message : Not in scope: data constructor ‘EmptyConfig’

Which seems to be right. EmptyConfig is never defined.

soenkehahn:
Did you use `servant-0.5`?

rwobben:
maybe a stupid question , but how can I check this ?

fisx:
> maybe a stupid question , but how can I check this ?

```shell
grep ^version: */*.cabal
```

`git show` would also be interesting.

rwobben:
grep ^version: */*.cabal
version:             0.1.0.0

git show does not have output.
I made my own project with stack.I did not clone the servant-example repo

fisx:
sorry, I'm lacking context here.  the version of servant that you are using is named in the cabal file of the servant package.  not sure what cabal file you just grepped there, but it may be just the one from your project.

rwobben:
It's the project version. 
In the cabal version there is only mentioned of servant-server without version numbers

rwobben:
it seems that stack has installed 0.4.4.6 and not 0.5 which I had installed locally .
Anyone a tip how to solve this ?

rvion:
yes, with something like that in your stack.yaml :

```yaml
packages:
- location: http://github.com/....
  subdirs:
  - servant-blaze
  - servant-cassava
  - servant-client
  - servant-docs
  - servant-examples
  - servant-foreign
  - servant-js
  - servant-lucid
  - servant-mock
  - servant-server
```

rwobben:
Then I see this error 

[...]

PR #398, opened 2016-02-26
change a dependency and delete EmptyConfig

rwobben:
I change aeson to aeson-compat because Stack has still 0.9 in the repo.
With this change the tutorial will run without problems.

Second change I made is because I saw a error message. See my remarks on the bug report. 
According to fizruk on irc , servant puts automatical EmptyConfig so we do not have to add this ourself.

With these changes the first example part compiles and runs well on my box

### DIFF
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -29,7 +29,7 @@ import Prelude.Compat
 import Control.Monad.IO.Class
 import Control.Monad.Reader
 import Control.Monad.Trans.Except
-import Data.Aeson
+import Data.Aeson.Compat
 import Data.Aeson.Types
 import Data.Attoparsec.ByteString
 import Data.ByteString (ByteString)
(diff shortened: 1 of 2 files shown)


rwobben:
This solves the problems I described on  #360

arianvp:
You seem to have forgotten to add `aeson-compat` to the cabal file, hence the build is failing currently.

rwobben:
@arianvp  Added it , Thanks for the feedback.

arianvp:
There seems to be a merge conflict with our tutorial branch that we need to resolve first. Could you try pulling in the latest tutorial branch and resolve the conflicts? If you run into trouble, hit me up on IRC

rwobben:
Oke, 

This is the merge conflict : 

    import Control.Monad.Except
    import Control.Monad.Reader
   <<<<<<< HEAD  // our changes 
    import Control.Monad.Trans.Except
    import Data.Aeson.Compat
    =======  
    import Data.Aeson
    >>>>>>> 23d36afbb6ebe5dda2538979f6af16af10419e2f Thier changes 
    import Data.Aeson.Types

arianvp:
I think changing it to:

```
import Control.Monad.Except
import Control.Monad.Reader
import Control.Monad.Trans.Except
import Data.Aeson.Compat
import Data.Aeson.Types
```

should work

rwobben:
This one can be closed. A new PR is made ; #405

PR #399, opened 2016-02-26
Make ServantErr throwable

eskimor:
Being able to throw an error is useful at times.


eskimor:
e.g. to implement something like yesod-persistent's get404. It needs to be an exception to roll back the database transaction, just catching them at the toplevel and delivering them to the user seems reasonable.

fizruk:
Looks good to me! :+1:

@haskell-servant/maintainers any objections?

alpmestan:
LGTM.

soenkehahn:
I would remove the comment. In any case LGTM.

PR #402, opened 2016-03-01
servant-server: add a test-case for streaming request bodies from cli…

soenkehahn:
…ent to server

I thought it would be good to add a test-case, since

- this is a pretty important feature and
- this test was failing with `0.4`. (It works with `0.5` now.)


soenkehahn:
Could someone review this, please? That'd be great.

cc: @haskell-servant/maintainers

alpmestan:
LGTM.

PR #406, opened 2016-03-02
Fix compilation error, spurious Proxy

alanz:


soenkehahn:
I think you're somehow using the wrong version of `servant-mock`.

alanz:
I guess something has changed in the git version. On my local machine, with GHC 7.10.3, and

```
$ ghc-pkg list | grep servant
    servant-0.4.4.6
    servant-mock-0.4.4.6
    servant-server-0.4.4.6
    servant-swagger-1.0.3
```

I get 

```
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help
Prelude> :load "/home/alanz/mysrc/github/haskell-servant/servant/servant-mock/example/main.hs"
[1 of 1] Compiling Main             ( /home/alanz/mysrc/github/haskell-servant/servant/servant-mock/example/main.hs, interpreted )

/home/alanz/mysrc/github/haskell-servant/servant/servant-mock/example/main.hs:23:30:
    Couldn't match type ‘Control.Monad.Trans.Either.EitherT
                           ServantErr IO User’
                   with ‘Proxy t0 -> Server API’
    Expected type: Proxy t0 -> Server API
      Actual type: ServerT
                     API (Control.Monad.Trans.Either.EitherT ServantErr IO)
    The function ‘mock’ is applied to two arguments,
    but its type ‘Proxy API -> Server API’ has only one
    In the second argument of ‘($)’, namely ‘mock api Proxy’
    In the second argument of ‘run’, namely
      ‘(serve api $ mock api Proxy)’
Failed, modules loaded: none.
```

soenkehahn:
Yes, the examples are being kept in sync with the code.

You should be able to use the examples from the branch `0.4`.

alanz:
ok, thanks

PR #413, opened 2016-03-18
Enable Werror on CI.

### LINKED ISSUE #313: fix compilation warnings
And enable `-Werror` on travis.

jkarni:
Fixes #313

### DIFF
--- a/servant-foreign/src/Servant/Foreign/Internal.hs
+++ b/servant-foreign/src/Servant/Foreign/Internal.hs
@@ -7,7 +7,7 @@
 -- arbitrary programming languages.
 module Servant.Foreign.Internal where
 
-import           Control.Lens hiding (cons, List)
+import           Control.Lens (makePrisms, makeLenses, Getter, (&), (<>~), (.~), (%~))
 #if !MIN_VERSION_base(4,8,0)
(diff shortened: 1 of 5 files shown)


jkarni:
Hmm, not sure how to get the warnings from `defer-type-errors` (where we use `should-not-typecheck`) to go away.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -7,7 +7,7 @@
 -- arbitrary programming languages.
 module Servant.Foreign.Internal where
 
-import           Control.Lens hiding (cons, List)
+import           Control.Lens (makePrisms, makeLenses, Getter, (&), (<>~), (.~), (%~))
```
dredozubov: FWIW, i dislike this kind of imports with lens. There are just a few name clashes and you have to import something if you have to use a slightly different operator/combinator.

jkarni: I agree about lens in particular, but we were getting a warning since `List` didn't exist, so it was this or CPP

jkarni:
Closing in favour of #477

PR #418, opened 2016-03-20
Making dependencies known, demonstrating how to split things apart

bitemyapp:
I got a confusing error about EitherT and ExceptT when I followed along with the tutorial without specifying `servant >= 5`. Only reason I figured it out quickly is that I talked to y'all about migrating from EitherT to ExceptT before.


bitemyapp:
Well, the literate Haskell was an unfortunate choice.

bitemyapp:
Oy vey, you're importing it in Docs.lhs too.

I give up.

PR #419, opened 2016-03-20
Improving tutorial

bitemyapp:
Round two.

### DIFF
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -10,7 +10,7 @@ write our first webservice.
 
 ## Setting up our project
 
-Don't forget to look at the [cabal file included with this example project](tutorial.cabal). You can run the main function provided in the following examples within the `Server` module without using a separate executable.
+Don't forget to look at the [cabal file included with this example project](doc/tutorial/tutorial.cabal). You can run the main function provided in the following examples within the `Server` module without using a separate executable.
 
 If you're using Stack, then in your `stack.yaml` you may want to specify version 0.11.1.1 for `aeson`, and version 0.5 for `servant` and `servant-server` in your `extra-deps`.
 
@@ -159,7 +159,7 @@ main = run 8081 app1
 You can put this all into a file or just grab [servant's
 repo](http://github.com/haskell-servant/servant) and look at the
 *doc/tutorial* directory. This code (the source of this web page) is in
-*doc/tutorial/Server.md*.
+*doc/tutorial/Server.lhs*.
 
 If you run it, you can go to `http://localhost:8081/users` in your browser or
 query it with curl and you see:
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -8,7 +8,15 @@ type. Can we have a webservice already?
 Equipped with some basic knowledge about the way we represent APIs, let's now
 write our first webservice.
 
-The source for this tutorial section is a literate haskell file, so first we
+## Setting up our project
+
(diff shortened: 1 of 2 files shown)


doc/tutorial/Server.lhs
```diff
@@ -8,7 +8,15 @@ type. Can we have a webservice already?
 Equipped with some basic knowledge about the way we represent APIs, let's now
 write our first webservice.
 
-The source for this tutorial section is a literate haskell file, so first we
+## Setting up our project
+
+Don't forget to look at the [cabal file included with this example project](doc/tutorial/tutorial.cabal). You can run the main function provided in the following examples within the `Server` module without using a separate executable.
```
soenkehahn: I don't think the link to the cabal file will work on readthedocs.org.

soenkehahn: You write that you can use the `main` function without an executable. Are you talking about ghci? If so, I think this should be mentioned.

doc/tutorial/Server.lhs
```diff
@@ -8,7 +8,15 @@ type. Can we have a webservice already?
...
+
+If you're using Stack, then in your `stack.yaml` you may want to specify version 0.11.1.1 for `aeson`, and version 0.5 for `servant` and `servant-server` in your `extra-deps`.
```
soenkehahn: I'm not sure what you're trying to explain here.

Is this about setting up a new `stack` project for `servant`? (In that case I think this explanation should go somewhere else.)

Or is this about how to set up a stack file so that you can load the tutorial in `ghci`? In that case, you can use the stack file provided in the repo (top-level). So in `doc/tutorial` you can do:

``` bash
chmod go-w .ghci .
stack setup
stack build
stack exec ghci Server.lhs
```

^^^ This information is sorely missing from the tutorial, btw, I think.

bitemyapp: @soenkehahn I think most people (myself included) will setup a new project and type things in from that without necessarily knowing where the `.lhs` files live.

Good feedback, I'll give it another pass when I have time again.

soenkehahn: @bitemyapp: Ah, that makes sense.

Maybe we should encourage people to checkout the `servant` repo and play around with the tutorial in the repo. And explain in detail how to do that in the `README`. This would have the added benefit that it would make it easier for people to contribute something to the tutorial.

On the other hand it's maybe a better learning experience if you actually set up files and write code yourself while reading the tutorial.

@bitemyapp: I can't speak for other maintainers, but I'd be happy to leave that decision up to you. Since you're the one working on this.

Generally it would be nice to have a very minimal `servant` project to get people started. Including cabal file, stack file, test suite, etc.. (This of course has little to do with your PR.)

soenkehahn: Just for reference, here's a ticket for the project template idea: https://github.com/haskell-servant/servant/issues/420.

rwobben:
@soenkehahn  so make for example  first the serving api page and then try to make a api for myself. 
I can do that and put the code in my github or bitbucket page.

rwobben:
Hmm, first problem. I work with Stack and that has still 0.4.7 :(

rwobben:
I do not know if this is the right place to post but I have made my first API with the help of the tutorial. 
What I have changed is making 2 new datatypes. 
One named head Account for looking if it's is a balance account or a profit/losts account.
Second name subAccount so you can group accounts like sales costs that schould hold the total amount  for all the seperate sales costs accounts 

The code can be found here :  https://github.com/rwobben/paytime.git
 @soenkehahn  : I hope this is what you mean with experiment with the tutorial and say how you achieve this 

If not, ask questions and I try to answer them

PR #423, opened 2016-03-23
Capture tweaks

soenkehahn:
@tmbull pointed out that `400` is what he would expect and I think he has a point. Imagine a capture path snippet is being used for an id: If a capture can be successfully parsed, then the handler for that route can e.g. look it up in a database. If it can't be found, the handler has to throw a `404`. If it can be found, it can do whatever it's meant to do. But if the id can't even be parsed from the path snippet, then `400` seems more appropriate than `404`.

(I feel like we discussed this before, but I can't find anything.)


alpmestan:
In theory, that makes sense. In practice, I don't know what status code the majority of web applications use for that situation -- has anyone got a relevant link here? I'm not opposed to this change, but just would like _something_ that backs up this patch based on either some standard or on the behavior of most web frameworks.

tmbull:
The best that I can do is link to the relevant RFC [here](https://tools.ietf.org/html/rfc7231). 

[6.5.1.  400 Bad Request](https://tools.ietf.org/html/rfc7231#section-6.5.1)
```
The 400 (Bad Request) status code indicates that the server cannot or
   will not process the request due to something that is perceived to be
   a client error (e.g., malformed request syntax, invalid request
   message framing, or deceptive request routing).
```

[6.5.4.  404 Not Found](https://tools.ietf.org/html/rfc7231#section-6.5.4)
```
   The 404 (Not Found) status code indicates that the origin server did
   not find a current representation for the target resource or is not
   willing to disclose that one exists.  A 404 status code does not
   indicate whether this lack of representation is temporary or
   permanent; the 410 (Gone) status code is preferred over 404 if the
   origin server knows, presumably through some configurable means, that
   the condition is likely to be permanent.

   A 404 response is cacheable by default; i.e., unless otherwise
   indicated by the method definition or explicit cache controls (see
   Section 4.2.2 of [RFC7234]).
```

My interpretation of this is the 404 indicates that a resource does not exist, while 400 indicates that the request is invalid. If I have a path parameter of type Int, and I receive "foo", that request is invalid in my mind.

Of course these things are pretty loosely defined, and without a doubt there are other frameworks out there that take a different interpretation.

soenkehahn:
@tmbull: My guess is, everyone agrees, that it should be a `400`, if we stick to the standard. But we've been bitten by similar train of thoughts before. (For example returning a `201` on `POST` requests, even though it conforms to the standard, is apparently rather uncommon and can create problems with client software.) So the question is: What's the best thing to do in practice?

@alpmestan:

- ruby on rails: They give you query parameters as strings, so people have to decide for themselves what to do when they can't parse the parameters. It does return `400` on missing parameters.
- django: I haven't found any way to specify a type of a parameter, so I assume they will be unparsed, so it's also up to the programmer to decide what to do on parse errors.

Since web-frameworks in dynamic languages don't seem to be very informative here, I looked at what actual APIs are doing: reddit seems to just ignore invalid query params, so does github. (Of course, I'm not sure, if they do that consistently.)

I'd be in favor of merging this. The standard says so and I didn't find any major framework that enforces anything differently.

soenkehahn:
LGTM.

alpmestan:
Alright, I'm in favor of this patch as well :+1:

PR #425, opened 2016-03-24
fix MVar infinite blocking

nrolland:


servant-client/src/Servant/Client/PerformRequest/GHCJS.hs
```diff
@@ -62,7 +62,9 @@ performXhr xhr request = do
   callback <- onReadyStateChange xhr $ do
     state <- readyState xhr
     case state of
-      4 -> putMVar waiter ()
+      4 -> do
+          toPut <- isEmptyMVar waiter
```
arianvp: A Race condition can still happen between this and `putMVar`. It won't actually fix the problem I think.


I think `tryPutMVar` is more appropriate. It's an atomic version of this behaviour.


Also, is it really a problem in the first place? I can not see a situation here where `putMVar` will actually block as `waiter` will always be empty before `putMVar` is called.   (Because there is only one `putMVar` happening, namely on the `onReadyStateChange` callback)

arianvp:
@yogsototh Could you provide a bit more info about this pull request? I'd love to get the issue you're experiencing resolved.

yogsototh:
It was some time ago. But I made a repository to show the bug. 

https://github.com/yogsototh/ghcjs-servant-client-bug

In my memory there were a runtime crash when waiting for some answer. At least using Safari. The demo should be a minimal form triggering a POST request. 

I'm on mobile right now. I'll could give you more details later if you want.

yogsototh:
Ah yes, the function was called many time with the same state 4.

arianvp:
Also cc @soenkehahn as he was working on this branch

arianvp:
I will have a look at this this weekend.

soenkehahn:
@nrolland: I'd like to delete the branch `client-ghcjs-wip` in favor of `client-ghcjs` (https://github.com/haskell-servant/servant/tree/client-ghcjs). Could you please point this PR at the new branch?

soenkehahn:
Sorry, closing this to be able to delete the `client-ghcjs-wip` branch.

PR #427, opened 2016-03-25
Document http error codes

alexanderkjeldaas:
It's more useful to see some examples when we're trying to see which error to return.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/ServantErr.hs
+++ b/servant-server/src/Servant/Server/Internal/ServantErr.hs
@@ -23,6 +23,13 @@ responseServantErr ServantErr{..} = responseLBS status errHeaders errBody
   where
     status = HTTP.mkStatus errHTTPCode (BS.pack errReasonPhrase)
 
+-- | 'err300' Multiple Choices
+--
+-- Example:
+--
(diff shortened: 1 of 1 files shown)


alpmestan:
Great initiative, examples rarely hurt :+1: LGTM

dredozubov:
> throwErr $ err503 { errBody = "We're rewriting in PHP." }

:+1: 

but srsly, LGTM

servant-server/src/Servant/Server/Internal/ServantErr.hs
```diff
@@ -163,6 +303,13 @@ err413 = ServantErr { errHTTPCode = 413
                     , errHeaders = []
                     }
 
+-- | 'err414' Request-URI Too Large
+--
+-- Example:
+--
+-- > failingHandler :: ExceptT ServantErr IO ()
+-- > failingHandler = throwErr $ err414 { errBody = "Maxiumu length is 64." }
```
3noch: Oops!

alexanderkjeldaas: Fixed. Thanks!

soenkehahn:
LGTM, I'll merge.

PR #430, opened 2016-03-27
cabal: decrease upper bound for base to disallow ghc-8

soenkehahn:
Since `servant` currently doesn't compile with `ghc-8`.

Changing this in `servant` should be enough, since every package in here depends on it.

CC: @phadej

### DIFF
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -46,7 +46,7 @@ library
     Servant.API.WithNamedContext
     Servant.Utils.Links
   build-depends:
-      base >=4.7 && <5
+      base >= 4.7 && < 4.9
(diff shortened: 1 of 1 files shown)


soenkehahn:
Could someone give this a quick review, please? @haskell-servant/maintainers

georgefst:
This might seem pedantic, but this should really have been accompanied by Hackage revisions to older Servant versions. I've just found myself building the ancient `servant-0.2.2` as a transitive dependency because, as far as Cabal is concerned, it's the most recent version which supports `base-4.21`!

tchoutri:
@georgefst nah this is a legitimate comment, thanks for the feedback! :)

Issue #428, opened 2016-03-26
[servant-client-0.5] generating composable client functions


echatav:
Generating composable client functions with `servant-client-0.5` is difficult because of the `Manager` parameter, with the current workaround being the use of `unsafePerformIO`. This is mentioned in the [tutorial](http://haskell-servant.readthedocs.org/en/stable/tutorial/Client.html) and recently on [reddit](https://www.reddit.com/r/haskell/comments/4bw7ju/haskell_bindings_for_the_mondo_api_using_servant05/). I propose using a type family to distribute the `Manager` over alternatives, producing composable clients in the `ReaderT Manager (ExceptT ServantError IO)` monad. Here's a sketch. Let me know what you think.

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeOperators #-}

module Distribute where

import Control.Monad.Except
import Control.Monad.Reader
import Data.Proxy
import Network.HTTP.Client (Manager)
import Servant.API
import Servant.Client

type family (-<>) a x where
  a -<> (b :<|> c) = (a -<> b) :<|> (a -<> c)
  a -<> (b -> c) = b -> (a -<> c)
  a -<> m b = ReaderT a m b

class Distribute a x where
  distribute :: (a -> x) -> (a -<> x)

instance {-# OVERLAPPING #-} (Distribute a b , Distribute a c) => Distribute a (b :<|> c) where
  distribute abc = distribute (altb . abc) :<|> distribute (altc . abc)
    where
      altb (b :<|> _) = b
      altc (_ :<|> c) = c

instance {-# OVERLAPPING #-} Distribute a c => Distribute a (b -> c) where distribute abc = distribute . flip abc

instance {-# OVERLAPPING #-} (a -<> m b) ~ (ReaderT a m b) => Distribute a (m b) where distribute = ReaderT

clients :: (HasClient layout, Distribute Manager (Client layout)) => Proxy layout -> BaseUrl -> (Manager -<> Client layout)
clients proxy baseurl = distribute (client proxy baseurl)

type API1 = "position" :> Capture "x" Int :> Capture "y" Int :> Get '[JSON] (Int,Int)
type API2 = "hello" :> QueryParam "name" String :> Get '[JSON] String
[...]

PR #432, opened 2016-03-28
Manager in client

mbg:
This PR removes the `BaseUrl` and `Manager` parameters from the `client` function, as discussed in #428. In turn, the type of all client functions changes from `ExceptT ServantError IO a` to `ReaderT BaseUrl (ReaderT Manager (ExceptT ServantError IO)`. For convenience, I have defined a type named `ClientM` as an alias for this.

### DIFF
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -28,7 +28,7 @@ import           Control.Applicative        ((<$>))
 import           Control.Arrow              (left)
 import           Control.Concurrent         (forkIO, killThread, ThreadId)
 import           Control.Exception          (bracket)
-import           Control.Monad.Trans.Except (ExceptT, throwE)
+import           Control.Monad.Trans.Except (ExceptT, throwE, runExceptT)
 import           Data.Aeson
 import qualified Data.ByteString.Lazy       as BS
 import           Data.Char                  (chr, isPrint)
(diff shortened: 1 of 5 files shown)


servant-client/test/Servant/ClientSpec.hs
```diff
@@ -208,47 +208,48 @@ sucessSpec :: Spec
 sucessSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
 
     it "Servant.API.Get" $ \(_, baseUrl) -> do
-      let getGet = getNth (Proxy :: Proxy 0) $ client api baseUrl manager
-      (left show <$> runExceptT getGet) `shouldReturn` Right alice
+      let getGet = getNth (Proxy :: Proxy 0) $ client api
+      (left show <$> SCR.runClientM getGet baseUrl manager) `shouldReturn` Right alice
```
soenkehahn: I'm hoping we can now get rid of the `getNth` hack and just declare `getGet`, `getDeleteEmpty`, etc. at the top level. (This shouldn't necessarily be part of this PR. Just saying that I like this change. :) )

soenkehahn:
Awesome, thanks a lot!

Two comments:

- I would have expected something like `type ClientM = ReaderT (BaseUrl, Manager) IO`. Why did you opt for two stacked `ReaderT`s?
- It would be nice to have one test-case that motivates this feature. Where the client functions are actually bound at the top level. To see whether this actually works **and** to document the feature.

Other than that, LGTM. Very readable code.

mbg:
> I would have expected something like type `ClientM = ReaderT (BaseUrl, Manager) IO`. Why did you opt for two stacked `ReaderT`s?

There is no particular reason I chose one over the other, although you might argue that two stacked `ReaderT`s are a bit more modular allow you to write something like the following more easily:

```
foo = runExceptT $ runReaderT manager $ do
  runReaderT someClientFunction (BaseUrl Http "localhost" 8081 "")
  runReaderT someClientFunction (BaseUrl Https "localhost" 8082 "")
```

It might be useful if you wanted to test the same API in that way, but I don't feel strongly about this and would be happy to change it to a single `ReaderT`, parametrised over a pair.

> It would be nice to have one test-case that motivates this feature. Where the client functions are actually bound at the top level. To see whether this actually works **and** to document the feature.

Perhaps I could do this by getting rid of `getNth` and `getLast`? I thought it was already nice to get rid of `System.IO.Unsafe` and `unsafePerformIO` in the tutorial.

soenkehahn:
Seems like a good enough reason for using two stacked `ReaderT`s.

> Perhaps I could do this by getting rid of getNth and getLast?

Yes, that would work.

> I thought it was already nice to get rid of System.IO.Unsafe and unsafePerformIO in the tutorial.

Yes. :)

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -30,17 +30,16 @@ import           Control.Concurrent         (forkIO, killThread, ThreadId)
...
 import           Data.Monoid                hiding (getLast)
 import           Data.Proxy
 import qualified Data.Text                  as T
 import           GHC.Generics               (Generic)
-import           GHC.TypeLits
 import qualified Network.HTTP.Client        as C
 import           Network.HTTP.Media
-import           Network.HTTP.Types         (Status (..), badRequest400,
-                                             methodGet, ok200, status400)
+import qualified Network.HTTP.Types         as HTTP
```
mbg: I had to import this as qualified because the `Header` type is needed in the signature of `getRawSuccess` and `getRawFailure`, but there is already servant's own `Header` type in scope.

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -120,6 +119,53 @@ type Api =
...
+    :<|> getBody
+    :<|> getQueryParam
+    :<|> getQueryParams
+    :<|> getQueryFlag
+    :<|> getRawSuccess
+    :<|> getRawFailure
+    :<|> getMultiple
+    :<|> getRespHeaders
+    :<|> getDeleteContentType = client api
+
```
mbg: I wasn't sure how to go about the formatting of the types and the patterns here (it seems that Haskell is a bit inflexible with line breaks inside of patterns).

soenkehahn: Personally I wouldn't align types (or code in general) at all. I would go with two spaces indentation (although I don't have anything against 4). So I would prefer e.g.

``` haskell
getRawFailure :: HTTP.Method
  -> SCR.ClientM (Int, BS.ByteString, MediaType, [HTTP.Header], C.Response BS.ByteString)
```

Mainly because of this: http://blog.madewithlove.be/post/code-style-options-for-cleaner-diffs/, Tip 2.

`servant` code uses alignment in lots (most?) places, so this is really just my 2 cents.

servant-client/test/Servant/ClientSpec.hs
soenkehahn: You could consider hiding `servant`'s `Header`. Don't know if that'd be better...

soenkehahn:
Reviewed the newest commits. Still LGTM.

servant-client/test/Servant/ClientSpec.hs
3noch: Not to diverge into a flamewar, but it's at least good to realize that aligning things is a trade-off between ugly diffs and "ease of comparison." Which one to choose can often depend on which of the two things we want more. I agree that in this case alignment doesn't really buy anything.

dmjio:
@mbg, great work on this! cc @soenkehahn @3noch, Hope it's not too short notice, would like to throw my two cents in. Would really like it if we could keep the number of constructors minimized, as additional transformer layers can produce considerable heap allocation. I anticipate people will be using this library for many APIs (public / private).

`type Client a = BaseUrl -> Manager -> IO a` 

@mbg, how do you feel about this?

soenkehahn:
@dmjio: I would swap the `BaseUrl` and `Manager` arguments. Since the `Manager` is less likely to change between calls. (#bikeshedding)

@mbg: One thing that's missing from the PR is an entry in the `CHANGELOG.md`.

mbg:
@dmjio The extra heap allocations are a good point against the additional transformer layer. Did you see the reasoning for `ReaderT` in #428, though? I don't think we ever came to a real consensus there. Also, you removed `ExceptT` as well -- was that intentional?


@soenkehahn OK, I will add an entry to `CHANGELOG.md` and fix the formatting in `ClientSpec.hs`

dmjio:
> Also, you removed ExceptT as well -- was that intentional?

whoops, I've updated the example :) 

> Did you see the reasoning for ReaderT in #428, though? 

@mbg, the main questions seems to boil down to this (correct me if I'm wrong), 

> .... but there is a question here about whether we think most people will end up doing something similar (in which case we should probably go with ReaderT out of the box) or not (in which case function types would be preferable).

I would vote for giving the user the choice to add a `ReaderT` on their own volition, rather than imposing additional "machinery" for them. Even though, yes, I would agree that a reader would fit naturally, it would be simpler to not add it (see Occam's razor).

Some other suggestions against `ReaderT` would be:
  - I don't think it's safe to assume everyone will be wrapping things into a `ReaderT`
  - `(->) r a` is equivalent to `Reader r a`, but with less constructors / allocation (true of `ReaderT` as well).
  - Consumer libraries would be simplified in that they don't need to add an additional import statement for `ReaderT`.
  - Removes the need to do `runReaderT` with `runExceptT` on every call.
  - Simpler, more readable for beginners (multiple transformer stacks can be difficult for new comers)

soenkehahn:
@mbg: Like @dmjio I'd also (still) be in favor of function arguments.

Maybe someone from @haskell-servant/maintainers wants to chime in?

(If we're settling on function arguments I'd also like to offer to make the needed changes on top of your commits.)

mbg:
That seems to be the consensus then, unless anyone else chimes in. I'd be happy to implement those changes as well.

alpmestan:
I agree with @dmjio - and you can probably get `ReaderT` based clients with some enter magic?

arianvp:
I'd say LGTM.

soenkehahn:
Very good, I'll merge this. And we should probably make a release (`0.6`).

soenkehahn:
Thanks again @mbg!

joaomilho:
This should also update the example in the README, right?

soenkehahn:
@joaomilho: Which README are you talking about?

PR #439, opened 2016-04-04
sources.txt: prefer servant-server and servant-client

soenkehahn:
I think both for the release script and for CI it makes sense to prefer these
two packages.


alpmestan:
By "prefer", you mean "process them first" ?

soenkehahn:
Yes, exactly. :)

PR #441, opened 2016-04-05
re-export `ClientM` from `Servant.Client`.

soenkehahn:
@mbg, what do you think?


mbg:
Makes sense to me and probably something I should have done when I introduced `ClientM`.

alpmestan:
:+1:

Issue #443, opened 2016-04-06
Missing ReflectMethod instances?


gibiansky:
In `servant-0.6` there is a `ReflectMethod` class that converts from `StdMethod` to `http-types` `Method`. It has instances for PATCH, DELETE, PUT, HEAD, GET, POST, but none of the other `StdMethod` constructors. Is there a reason for this, or is this an oversight? 

I would like Servant to export the same API for all the HTTP methods, so IMHO the `ReflectMethod` instances should be added.

gibiansky:
Implemented this in #444

jkarni:
Closed with #444

PR #444, opened 2016-04-06
Add ReflectMethod instances for OPTIONS, TRACE, and CONNECT

gibiansky:
Addresses #443 by adding a few instances. 

Tested solely by verifying that `stack build` completes without error. (More testing may be required?)


soenkehahn:
Do we have any tests for the other instances? If yes, maybe we can add tests for these too.

Otherwise LGTM.

gibiansky:
I grepped through the `servant/test` directory for `ReflectMethod` and `methodPost` and other similar things and it doesn't seem like there are any tests for the other instances.

Is there anything else I should do in order to get this merged?

alpmestan:
LGTM.

soenkehahn:
Rebased and merged: ba57d20

soenkehahn:
@gibiansky: Thanks a lot!

PR #445, opened 2016-04-06
Auth bug

### LINKED ISSUE #440: Raw gobbles up auth-protected endpoint?
Using the tutorial, if I have this API:

```haskell
type Api = "users" :> BasicAuth "here" Bool :> Get '[JSON] [User]
```

Everything works as expected when browsing with my browser. It asks for creds.

However,

```haskell
type Api = "users" :> BasicAuth "here" Bool :> Get '[JSON] [User] :<|> Raw
```
(and serving a local directory) makes it impossible for me to visit the "users" endpoint. I get `404 File not found`.

Am I doing something wrong?

soenkehahn:
Fixes #440.

Please review.

Also, how do I create basic auth tokens? (https://github.com/haskell-servant/servant/blob/auth-bug/servant-server/test/Servant/ServerSpec.hs#L576)

CC: @aaronlevin @3noch @jkarni @kosmikus @alpmestan

### DIFF
--- a/servant-server/test/Servant/ServerSpec.hs
+++ b/servant-server/test/Servant/ServerSpec.hs
@@ -19,8 +19,10 @@ import           Control.Applicative        ((<$>))
 import           Control.Monad              (forM_, when, unless)
 import           Control.Monad.Trans.Except (ExceptT, throwE)
 import           Data.Aeson                 (FromJSON, ToJSON, decode', encode)
+import qualified Data.ByteString.Base64     as Base64
 import           Data.ByteString.Conversion ()
 import           Data.Char                  (toUpper)
+import           Data.Monoid
 import           Data.Proxy                 (Proxy (Proxy))
 import           Data.String                (fromString)
 import           Data.String.Conversions    (cs)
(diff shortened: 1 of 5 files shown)


servant-server/test/Servant/ServerSpec.hs
```diff
@@ -30,6 +30,7 @@ import           Network.HTTP.Types         (Status (..), hAccept, hContentType,
                                              methodDelete, methodGet,
                                              methodHead, methodPatch,
                                              methodPost, methodPut, ok200,
+                                             imATeaPot418,
```
3noch: How did we ever do without this all this time?

3noch:
@soenkehahn You've officially made my day!

kosmikus:
I am ok with this change. (Generally, I am still wary of anything that makes use of `FailFatal`, because it is inherently non-modular. But currently, this definitely looks like the right thing to do.)

arianvp:
```
token user password = base64 $ user ++ ":" ++ password
```

alpmestan:
@soenkehahn What Arian just described is also [in servant-client](https://github.com/haskell-servant/servant/blob/master/servant-client/src/Servant/Common/BasicAuth.hs).

Regarding the PR, looks good to me!

soenkehahn:
I'm going to merge once CI passes. Thanks for the reviews.

I suspect that the general auth is having the same problem, btw.

aaronlevin:
@soenkehahn thanks for fixing this so quickly. LGTM.

General auth uses `FailFatal`, so it should be ok.

aaronlevin:
(though we should probably add a test case - I can do that now)

aaronlevin:
@soenkehahn i'm going to push to this branch with a change to test Gen Auth.

soenkehahn:
@aaronlevin: Can you use a separate PR, please?

soenkehahn:
@aaronlevin: Sorry, removed your commit from this PR. Please create a new one.

aaronlevin:
@soenkehahn no worries. submitted.

PR #446, opened 2016-04-06
Add test: Gen Auth properly supports Raw endpoints

aaronlevin:
Continuing with the work from https://github.com/haskell-servant/servant/pull/445, this adds an explicit test to ensure General Authentication handles Raw endpoints appropriately.

The previous behaviour was correct, but this test will ensure that any future changes are correct w.r.t. Raw.


alpmestan:
:+1:

arianvp:
LGTM

soenkehahn:
LGTM.

Issue #449, opened 2016-04-07
Debug/trace logging toggle.


alexanderkjeldaas:
I have a need to understand where specifically a request fails.  Often when I have a javascript front-end and a servant back-end, there might be some small JSON encoding issue, or maybe the Content-Type header is set wrong, or similar.

Dumping the request and response is a good start, but it would be nice to get a trace of what happens, especially at points where things can go wrong, and to be able to see what is going wrong in a log.

Another thing that would be good to support in such a mechanism are ekg counters.  In a production setup, individual requests might not be too interesting to look at, but ekg counters that count the same things (number of successful JSON decodes vs failed ones broken down by endpoint for example) would be good to graph.

A third thing that could be plugged in there are performance counters, just doing a `gettimeofday` at that point in time and log/graph it.

A fourth thing that one might want to do, but that would require a different machinery is a chaos monkey, where some requests are asked to fail so check the robustness of the larger system as a whole.

kosmikus:
Dumping the request/response is something that can relatively easily be done on the WAI middleware layer.

A trace of what happens is a very interesting request, and quite doable. But the question is how exactly to do it. Simply adding the machinery for potentially tracing into the main routing machinery might slow it down. We could have a separate `runRoute` function doing the tracing. In any case, how do we parameterize `serve` on whether to do tracing or not?

Also, how do we write the trace? Simply print to `stdout` or `stderr` or use some logging framework?

EKG counters and performance counters sound like things to me again that could be done on WAI level, or is there anything in particular where you need access to info only the Servant layer has?

Four is certainly something for doing on the Servant-level, but it's not exactly clear to me what is desired here? Do you want the server to just fail on random requests, in order to test robustness of a client? Or do you want to test robustness of the server itself somehow?

alexanderkjeldaas:
I had a dump of the request, but in this case I was still scratching my head.  Of course it was a stupid mistake on my side, but confirmation that my handler was actually run, routed to etc. would have shortened the debug cycle (saved me 30 minutes probably).

I think a logging framework should be used.  I don't know what kind of optimizations are in the logging frameworks, but writing to a file descriptor should ideally happen in a separate thread for latency reasons.

For EKG, if I put them at the WAI level, can I get information about which route was matched so the counters can be broken down by handler?

For Chaos Monkey, the idea is to test for robustness in a larger system http://techblog.netflix.com/2011/07/netflix-simian-army.html

rvion:
[...]

PR #451, opened 2016-04-07
Improvements of router merging, visualization and testing

kosmikus:
I wanted to understand my own router machinery better, and write about it and test it. Therefore, I have added code that can display how the generated router actually looks like.

Here is a small example:
```
{-# LANGUAGE DataKinds #-}
import Data.Text.IO as T
import Data.Proxy
import Servant.API
import Servant.Server

type End1 = Get '[JSON] ()
type End2 = Put '[JSON] ()

type API1 =
       "a" :> "b" :> End1
  :<|> "a" :> "c" :> Capture "n" Int :> "d" :> End1

type API2 =
       "b" :> End1
  :<|> "a" :> "d" :> End1

type API3 =
       "a" :> "b" :> End2
  :<|> "a" :> "c" :> Capture "n" Int :> "e" :> End2

type API = API1 :<|> API2 :<|> API3

main :: IO ()
main = T.putStr (layout (Proxy :: Proxy API))
```

This produces:
```
<root>
├─ a
│  ├─ b
│  │  ┆
│  │  ├─•
│  │  ┆
│  │  └─•
│  ├─ c
│  │  └─ <dyn>
│  │     ├─ d
│  │     │  └─•
│  │     └─ e
│  │        └─•
│  └─ d
│     └─•
└─ b
   └─•
```

The output is intended for debugging purposes, and you have to know a little bit of how it works to be able to interpret it. Static dispatch via a finite map is encoded via connected lines (they still display with interruptions on github). Less efficient sequential choices are displayed using dotted lines. Delayed captures are displayed as `<dyn>`. Leaves / endpoints are displayed using dots.

We can see in this example that everything seems to be reordered correctly despite the fact that it is defined in several API fragments, as desired.

I have added a few tests for the reorderings. I have also found a few bugs in reordering where the desired most efficient solution isn't "found", and started to fix them. There are still some issues remaining, which I'm trying to fix asap.

Main questions:

* Is the functionality desirable to have?
* Is the output format sufficient? Can it be improved easily?
* Is it ok to change the `Router` datatype as I already did? How many external libs are we expecting to break?

### DIFF
--- a/servant-server/src/Servant/Server/Internal/Router.hs
+++ b/servant-server/src/Servant/Server/Internal/Router.hs
@@ -1,10 +1,13 @@
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings #-}
 module Servant.Server.Internal.Router where
 
 import           Data.Map                                   (Map)
 import qualified Data.Map                                   as M
+import           Data.Monoid
 import           Data.Text                                  (Text)
+import qualified Data.Text                                  as T
 import           Network.Wai                                (Request, Response, pathInfo)
 import           Servant.Server.Internal.RoutingApplication
 import           Servant.Server.Internal.ServantErr
@@ -13,36 +16,46 @@ type Router = Router' RoutingApplication
 
 -- | Internal representation of a router.
 data Router' a =
-    WithRequest   (Request -> Router)
+    WithRequest   (Request -> Router' a)
       -- ^ current request is passed to the router
-  | StaticRouter  (Map Text Router)
-      -- ^ first path component used for lookup and removed afterwards
-  | DynamicRouter (Text -> Router)
-      -- ^ first path component used for lookup and removed afterwards
-  | LeafRouter    a
-      -- ^ to be used for routes that match an empty path
-  | Choice        Router Router
+  | StaticRouter  (Map Text (Router' a)) [a]
+      -- ^ the map contains routers for subpaths (first path component used
+      --   for lookup and removed afterwards), the list contains handlers
+      --   for the empty path, to be tried in order
+  | DynamicRouter (Text -> Router' a)
+      -- ^ first path component passed to the function and removed afterwards
+  | RawRouter     a
+      -- ^ to be used for routes we do not know anything about
+  | Choice        (Router' a) (Router' a)
       -- ^ left-biased choice between two routers
   deriving Functor
 
--- | Apply a transformation to the response of a `Router`.
-tweakResponse :: (RouteResult Response -> RouteResult Response) -> Router -> Router
-tweakResponse f = fmap (\a -> \req cont -> a req (cont . f))
+-- | Smart constructor for a single static path component.
+pathRouter :: Text -> Router' a -> Router' a
+pathRouter t r = StaticRouter (M.singleton t r) []
+
+-- | Smart constructor for a leaf, i.e., a router that expects
+-- the empty path.
+--
+leafRouter :: a -> Router' a
+leafRouter l = StaticRouter M.empty [l]
 
 -- | Smart constructor for the choice between routers.
 -- We currently optimize the following cases:
 --
---   * Two static routers can be joined by joining their maps.
+--   * Two static routers can be joined by joining their maps
+--     and concatenating their leaf-lists.
 --   * Two dynamic routers can be joined by joining their codomains.
 --   * Two 'WithRequest' routers can be joined by passing them
 --     the same request and joining their codomains.
 --   * A 'WithRequest' router can be joined with anything else by
 --     passing the same request to both but ignoring it in the
 --     component that does not need it.
+--   * Choice nodes can be reordered.
 --
 choice :: Router -> Router -> Router
-choice (StaticRouter table1) (StaticRouter table2) =
-  StaticRouter (M.unionWith choice table1 table2)
+choice (StaticRouter table1 ls1) (StaticRouter table2 ls2) =
+  StaticRouter (M.unionWith choice table1 table2) (ls1 ++ ls2)
 choice (DynamicRouter fun1)  (DynamicRouter fun2)  =
   DynamicRouter (\ first -> choice (fun1 first) (fun2 first))
 choice (WithRequest router1) (WithRequest router2) =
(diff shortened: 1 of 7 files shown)


3noch:
Amazing.

phadej:
Should there be intermediate `data` so `sameStructure` would be a structural equality on it? Would be easier to write different predicates too, probably.

Great idea, anyway!

kosmikus:
@phadej Yes, that's a good idea.

kosmikus:
@phadej I have implemented your suggestion in dea4aafd8e85964cc88f3b2b4ab0d750b9076551.

servant-server/test/Servant/Server/RouterSpec.hs
```diff
@@ -0,0 +1,260 @@
...
+    it "properly reorders permuted static paths in the presence of Raw in middle" $ do
+      permute7 `shouldHaveSameStructureAs` permute8
+    it "properly reorders permuted static paths in the presence of a root endpoint in end" $ do
+      permute9 `shouldHaveSameStructureAs` permute12
+    it "properly reorders permuted static paths in the presence of a root endpoint in beginning" $ do
+      permute10 `shouldHaveSameStructureAs` permute12
+    it "properly reorders permuted static paths in the presence of a root endpoint in middle" $ do
+      permute11 `shouldHaveSameStructureAs` permute12
+    it "properly handles mixing static paths at different levels" $ do
+      level3 `shouldHaveSameStructureAs` level4
```
soenkehahn: I wish this could be a bit easier to read and understand. Sadly, we don't have local type synonyms, so the tests and the types cannot be easily put close to each other.

I don't have a very good idea how to make this better, unfortunately.

Currently the following convention seems to be followed: ```someApi$N `shouldHaveSameStructureAs` someApi$(N + 1)``` for odd Ns. Maybe another convention would be easier. E.g. ```someApi `shouldHaveSameStructureAs` someApi'```. You only would have to come up with unique names for the different test cases.

alpmestan:
I don't think there are many libraries that care about `Router`, except `servant-server` and maybe `servant-snap` (not released yet) but I'm not even sure. @imalsogreg ?

@kosmikus I love the idea. I definitely think this should go into servant-server. I'm wondering whether this could be combined with some more work to give an decent answer to #449: imagine something that shows the path(s) that a request follows during routing, along with data about the request, shows which endpoint is reached (and if none is reached, says routing couldn't succeed) and shows some data about the response body. Anyway, sorry to hijack this thread with this, I just found it interesting that you came up with this right after @alexanderkjeldaas submitted an issue about relating things.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -51,43 +64,139 @@ choice (WithRequest router1) router2 =
...
+    highestPri (Fail _) y = y
+    highestPri x _ = x
 
 -- Priority on HTTP codes.
 --
 -- It just so happens that 404 < 405 < 406 as far as
 -- we are concerned here, so we can use (<).
 worseHTTPCode :: Int -> Int -> Bool
 worseHTTPCode = (<)
+
```
soenkehahn: Trailing newline.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -51,43 +64,139 @@ choice (WithRequest router1) router2 =
...
 runRouter :: Router -> RoutingApplication
 runRouter (WithRequest router) request respond =
   runRouter (router request) request respond
-runRouter (StaticRouter table) request respond =
+runRouter (StaticRouter table ls) request respond =
   case pathInfo request of
-    first : rest
-      | Just router <- M.lookup first table
+    []   -> runChoice ls request respond
+    [""] -> runChoice ls request respond
```
soenkehahn: This is for urls with trailing slashes, right? (Just trying to understand the code.)

soenkehahn:
> Is the functionality desirable to have?

Regarding the visualization: Yes, I think so.
Regarding the fixes to `Router`: Definitely.

> Is the output format sufficient? Can it be improved easily?

I think it would be slightly easier with slashes (`/`) after the path segments. Also, since probably no one expects this format to be stable, we can easily tweak it later.

> Is it ok to change the Router datatype as I already did? How many external libs are we expecting to break?

Not many external libraries, but everyone with a custom combinator would be affected. So ideally we would have a migration guide.

And could you add a section to the changelog, please? :)

Apart from my comments, LGTM.

servant-server/src/Servant/Server/Internal/Router.hs
kosmikus: Yes, I'm going to add a comment.

servant-server/test/Servant/Server/RouterSpec.hs
kosmikus: I will try to think of something.

kosmikus:
@soenkehahn Re changelog, for what upcoming version? 0.7?

kosmikus:
@alpmestan I think the debugging for actual requests is definitely desirable, but mostly separate from this issue, as the code here is completely static and does not really affect `runRouter` at all (except for the changes to `runRouter` implied by changes to the underlying `Router` representation). We could write a version of `runRouter` that emits a trace of what it's doing for a specific request for the other issue. I'd support that, modulo figuring out the details.

soenkehahn:
Usually I put something temporary in the changelog (e.g. `on master`) and then when making a release look at all the changes and consult the PVP to decide what version bump we need.

soenkehahn:
If you think this PR warrants a major version bump, then you can put `0.7` into the changelog.

kosmikus:
@soenkehahn I guess it depends on how strictly we interpret the PVP. If we consider `Router` to be a morally unexported type because it's in `Internal`, then we could get away with a minor bump, as we otherwise just introduce new functions. But in a strict interpretation, we're changing the definition of an exposed datatype, so it would have to be major.

kosmikus:
@soenkehahn Do you think this is better (I'm not saying it's not, but I don't find it an easy choice):
```
<root>
├─ a/
│  ├─ d/
│  │  └─•
│  └─ e/
│     └─•
├─ b/
│  └─ <dyn>/
│     ├─•
│     ┆
│     └─•
├─ c/
│  └─•
┆
└─ <raw>
```

soenkehahn:
@kosmikus:

Re version bump: I think I would go with a major bump. Since this is a breaking change for custom combinators.

Re slashes: I think it's easier to understand if you look at these visualizations for the first time. So I'd prefer it with slashes.

alpmestan:
I think I'd find the following to be the most natural. For some reason, I'm more used to seeing leading slashes than trailing ones.

```
/
├─ /a
│  ├─ /d
│  │  └─•
│  └─ /e
│     └─•
├─ /b
│  └─ /<dyn>
│     ├─•
│     ┆
│     └─•
├─ /c
│  └─•
┆
└─ <raw>
```

kosmikus:
@alpmestan I'm perfectly happy to replace `<root>` with `/`. But in combination with leading slashes, it then looks strange because it feels like there's a double '/' at the beginning. I think trailing slashes are more common, in e.g. the output of `ls -F` or `tree -F`.

kosmikus:
@alpmestan The method isn't information we currently have in the `Router` type. We could add it, but that'd require more refactoring. I'd rather leave this for later :)

codedmart:
For what it is worth. I agree with @soenkehahn:

> > Is the functionality desirable to have?

> Regarding the fixes to Router: Definitely.

This looks like a great change to me. I find @alpmestan idea/thoughts about possible uses for routing interesting as well. LGTM

kosmikus:
Ok, I've added some more comments, in particular to the test cases to hopefully address @soenkehahn's concerns. I've also added a CHANGELOG entry and rebased on master.

I haven't squashed that changes (do you want me to?).

I'd like to propose this for merging now (assuming the CI goes through).

codedmart:
I would vote yes on squashing and merging :+1:

fizruk:
Can someone explain to me what happens when I put a `QueryParam` in front of the whole API:

```haskell
type API = QueryParam "param" String :> (API1 :<|> API2 :<|> API3)
```

Specifically, I have trouble understanding whether `Router` is computed for each request or shared among them?

Also, this visualization does not seem to be always valid since `Text -> Router' a` for `DynamicRouter` allows different router structures for different values of `Text` (in theory). I am not sure any of the existing combinators suffer from this though.

alpmestan:
The debug/logging thing was just a thought I wanted to write down somewhere before I forgot and anything needed to write it should not be considered part of this PR of course, @kosmikus. Sorry I didn't make that clear. Re trailing/leading slashes, well, trailing slashes are common for displaying directories yeah, but HTTP endpoints? Not sure. Either way, I don't feel _that_ strong about this and wouldn't mind leaving the current display as-is.

kosmikus:
@fizruk The `Router` is always computed only once. Look at `serveWithContext` in `Servant.Server`. There is one call to `route` which computes the `Router`, then one call to `runRouter` which runs it. The result of `runRouter` is the logic that is applied for every single request.

Adding the query parameter should not change the (visualized) router layout at all. All the `QueryParam` combinator does is add a delayed check that happens once reaching the endpoint, using `WithRequest`.

Yes, you are right that one can design combinators for which `layout` would crash. This is now documented here: https://github.com/kosmikus/servant/blob/show-router/servant-server/src/Servant/Server.hs#L143

I'm curious to learn whether there are any existing combinators that trigger this.

fizruk:
@kosmikus I know that outermost `Router` is computed once, but I am more concerned with inner `Router`s. E.g. consider this API:

```haskell
type API = Capture "param" Int :> "something" :> Post '[] ()
```

For this API router would essentially be `Text -> Map Text RouterApplication`, right?
My question is whether this `Map` is shared for all `Text` arguments.
In general this is obviously not true for a function with this type, but I suspect that it is true for (existing) servant combinators (since `Map` does not depend on that `Text` value, or does it?).

kosmikus:
@fizruk Oh, now I see what you mean. No, I'm afraid this is currently not shared, and you're right that it sounds as if it should be possible. I'll have to think about this. Anyway, this might be something for a new issue. The current PR isn't making this better or worse in any way.

soenkehahn:
From my side, feel free to merge to `master`.

Re squashing: If lots of your intermediate commits break the tests then I'd be in favor of squashing. To get better `git bisect` behavior. If they don't, then I don't feel strongly either way.

One more note about the release policy -- since this wasn't really followed pre `0.5`: If you merge to `master` expect this to land on hackage soon.

Great work!

kosmikus:
@soenkehahn Ok, squashed and merged. Yes, having releases more often is fine, but please don't release 0.7 before resolving #457 somehow.

soenkehahn:
@kosmikus: Agreed, #457 should be fixed first.

PR #452, opened 2016-04-07
Introduce a `Handler` alias for `ExceptT ServantErr IO`

### LINKED ISSUE #434: Introduce a type alias `Handler`
``` haskell
type Handler = ExceptT ServantErr IO
```

This would make it shorter to write type signatures. It would also make it less painful to replace this with a custom type later on.

lcycon:
Fixes #434

### DIFF
--- a/servant-server/src/Servant/Server/Experimental/Auth.hs
+++ b/servant-server/src/Servant/Server/Experimental/Auth.hs
@@ -12,8 +12,7 @@
 
 module Servant.Server.Experimental.Auth where
 
-import           Control.Monad.Trans.Except                 (ExceptT,
-                                                             runExceptT)
+import           Control.Monad.Trans.Except                 (runExceptT)
 import           Data.Proxy                                 (Proxy (Proxy))
 import           Data.Typeable                              (Typeable)
 import           GHC.Generics                               (Generic)
@@ -28,7 +27,7 @@ import           Servant.Server.Internal                    (HasContextEntry,
 import           Servant.Server.Internal.Router             (Router' (WithRequest))
 import           Servant.Server.Internal.RoutingApplication (RouteResult (FailFatal, Route),
                                                              addAuthCheck)
-import           Servant.Server.Internal.ServantErr         (ServantErr)
+import           Servant.Server.Internal.ServantErr         (ServantErr, Handler)
 
 -- * General Auth
 
@@ -42,11 +41,11 @@ type family AuthServerData a :: *
 --
 -- NOTE: THIS API IS EXPERIMENTAL AND SUBJECT TO CHANGE
 newtype AuthHandler r usr = AuthHandler
-  { unAuthHandler :: r -> ExceptT ServantErr IO usr }
+  { unAuthHandler :: r -> Handler usr }
   deriving (Generic, Typeable)
 
 -- | NOTE: THIS API IS EXPERIMENTAL AND SUBJECT TO CHANGE
-mkAuthHandler :: (r -> ExceptT ServantErr IO usr) -> AuthHandler r usr
+mkAuthHandler :: (r -> Handler usr) -> AuthHandler r usr
 mkAuthHandler = AuthHandler
 
 -- | Known orphan instance.
--- a/doc/tutorial/Authentication.lhs
+++ b/doc/tutorial/Authentication.lhs
@@ -44,7 +44,6 @@ You can use this combinator to protect an API as follows:
 
 module Authentication where
 
-import Control.Monad.Trans.Except       (ExceptT)
 import Data.Aeson                       (ToJSON)
 import Data.ByteString                  (ByteString)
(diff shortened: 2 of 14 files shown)


lcycon:
Ack. This is gonna fail methinks. Ambiguous `Handler`. Not sure why this doesn't happen on my machine..

3noch:
Maybe `HandlerM` to match `ClientM`?

lcycon:
That feels more natural to me, but I was just going off the issue referenced above.

kosmikus:
Sorry for bikeshedding: But if it really is called `ClientM` now (I'm not particularly fond of that name), then I think the corresponding concept for the server should also be called `ServerM`.

lcycon:
You know, the more I look at some of my apps (modified to build against the
head of my branch), the more I like the look of `Handler`.

That said, all but one of my servant using applications use a custom Monad
stack, so I can't say it's a huge difference..

alpmestan:
I like the simplicity of `Handler` but a conflict with an existing type in `Control.Exception` is really problematic. I'm really not fond of `ServerM` either. This is probably going to be confusing, when put next to `Server`/`ServerT`.

soenkehahn:
My vote is (still) for `Handler`.

- I think it's the most fitting.
- I'm also not *that* happy with `ClientM`. I think it shouldn't affect our choice in `servant-server`. We may even want to rename `ClientM`.
- `snap` uses the same name (`Handler`) for something similarish.
- I don't see a big problem with the name collision with `Control.Exception`. (Maybe this warrants a major version bump, since `Control.Exception` is very widely used. But I'm fine with that.)

servant-server/src/Servant/Server/Experimental/Auth.hs
```diff
@@ -1,19 +1,18 @@
...
+{-# LANGUAGE DeriveDataTypeable    #-}
+{-# LANGUAGE DeriveGeneric         #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PolyKinds             #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
```
soenkehahn: This seems like two formatting tools having an edit-war. :) @lcycon, can you remove this, please?

servant-server/src/Servant/Server/Internal/Handler.hs
```diff
@@ -0,0 +1,6 @@
+module Servant.Server.Internal.Handler where
+
+import           Control.Monad.Trans.Except         (ExceptT)
+import           Servant.Server.Internal.ServantErr (ServantErr)
+
+type Handler = ExceptT ServantErr IO
```
soenkehahn: I would consider putting `Handler` into `Servant.Server.Internal.ServantErr`.

soenkehahn:
Apart from my comments, LGTM.

kosmikus:
I can live very well with `Handler` if that's what everyone wants, but I'd like to consider renaming `ClientM` then. There is some value to consistent naming ...

servant-server/src/Servant/Server/Experimental/Auth.hs
lcycon: Ack. I thought I turned that off. Emacs has a mind of its own ;)

servant-server/src/Servant/Server/Internal/Handler.hs
lcycon: I had the same thought, but couldn't decide. I'll move it over.

mbg:
I agree that it makes sense to rename `ClientM` as well to maintain consistency. I didn't think about the naming too much when I introduced it since I was only looking for an alias for the monad stack we had initially in #432.

soenkehahn:
@lcycon: Could you squash your commits and rebase on master, please? After that, I'll merge.

Squashing, since there's a failing commit in between and some whitespace back-and-forth. I'd like to keep that out of the git history, if possible.

lcycon:
@soenkehahn 10-4, gimme a moment

lcycon:
@soenkehahn There ya go

soenkehahn:
@lcycon: I've added something to the changelog: d4c6f67.

Thanks a lot!

3noch:
Did ya'll end up deciding that `ClientM` needs to change too?

soenkehahn:
@3noch: No. I think this should happen in a separate issue or PR.

3noch:
@soenkehahn I was just curious if we need an issue to track that work, if it was desired.

soenkehahn:
@3noch: Feel free to open an issue.

PR #457, opened 2016-04-09
Fix router sharing

kosmikus:
Thanks to @fizruk for exposing a significant flaw in the former `Router` type.

This PR is based on top of #451, assuming that one will get merged, because it affects the same code and would otherwise cause lots of conflicts.

We've previously used functions in the `Router` type to provide information for subrouters. But this accesses the `Request`s too early, and breaks sharing of the router structure in general, causing the `Router` or large parts of the `Router` to be recomputed on every request.

We now do not use functions anymore, and properly compute all static parts of the router first, and gain access to the request only in `Delayed`.

This also turns the code used within `Delayed` into a proper monad now called `DelayedIO`, making some of the code using it a bit nicer.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/Router.hs
+++ b/servant-server/src/Servant/Server/Internal/Router.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Servant.Server.Internal.Router where
 
@@ -8,36 +10,41 @@ import qualified Data.Map                                   as M
 import           Data.Monoid
 import           Data.Text                                  (Text)
 import qualified Data.Text                                  as T
-import           Network.Wai                                (Request, Response, pathInfo)
+import           Network.Wai                                (Response, pathInfo)
 import           Servant.Server.Internal.RoutingApplication
 import           Servant.Server.Internal.ServantErr
 
-type Router = Router' RoutingApplication
+type Router env = Router' env RoutingApplication
 
 -- | Internal representation of a router.
-data Router' a =
-    WithRequest   (Request -> Router' a)
-      -- ^ current request is passed to the router
-  | StaticRouter  (Map Text (Router' a)) [a]
+--
+-- The first argument describes an environment type that is
+-- expected as extra input by the routers at the leaves. The
+-- environment is filled while running the router, with path
+-- components that can be used to process captures.
+--
+data Router' env a =
+    StaticRouter  (Map Text (Router' env a)) [env -> a]
       -- ^ the map contains routers for subpaths (first path component used
       --   for lookup and removed afterwards), the list contains handlers
       --   for the empty path, to be tried in order
-  | DynamicRouter (Text -> Router' a)
-      -- ^ first path component passed to the function and removed afterwards
-  | RawRouter     a
+  | CaptureRouter (Router' (Text, env) a)
+      -- ^ first path component is passed to the child router in its
+      --   environment and removed afterwards
+  | RawRouter     (env -> a)
       -- ^ to be used for routes we do not know anything about
-  | Choice        (Router' a) (Router' a)
+  | Choice        (Router' env a) (Router' env a)
       -- ^ left-biased choice between two routers
   deriving Functor
 
 -- | Smart constructor for a single static path component.
-pathRouter :: Text -> Router' a -> Router' a
+pathRouter :: Text -> Router' env a -> Router' env a
 pathRouter t r = StaticRouter (M.singleton t r) []
 
 -- | Smart constructor for a leaf, i.e., a router that expects
 -- the empty path.
 --
-leafRouter :: a -> Router' a
+leafRouter :: (env -> a) -> Router' env a
 leafRouter l = StaticRouter M.empty [l]
 
 -- | Smart constructor for the choice between routers.
@@ -46,40 +53,27 @@ leafRouter l = StaticRouter M.empty [l]
 --   * Two static routers can be joined by joining their maps
 --     and concatenating their leaf-lists.
 --   * Two dynamic routers can be joined by joining their codomains.
---   * Two 'WithRequest' routers can be joined by passing them
---     the same request and joining their codomains.
---   * A 'WithRequest' router can be joined with anything else by
---     passing the same request to both but ignoring it in the
---     component that does not need it.
 --   * Choice nodes can be reordered.
 --
-choice :: Router -> Router -> Router
+choice :: Router' env a -> Router' env a -> Router' env a
 choice (StaticRouter table1 ls1) (StaticRouter table2 ls2) =
   StaticRouter (M.unionWith choice table1 table2) (ls1 ++ ls2)
-choice (DynamicRouter fun1)  (DynamicRouter fun2)  =
-  DynamicRouter (\ first -> choice (fun1 first) (fun2 first))
-choice (WithRequest router1) (WithRequest router2) =
-  WithRequest (\ request -> choice (router1 request) (router2 request))
-choice (WithRequest router1) router2 =
-  WithRequest (\ request -> choice (router1 request) router2)
-choice router1 (WithRequest router2) =
-  WithRequest (\ request -> choice router1 (router2 request))
+choice (CaptureRouter router1)   (CaptureRouter router2)   =
+  CaptureRouter (choice router1 router2)
 choice router1 (Choice router2 router3) = Choice (choice router1 router2) router3
 choice router1 router2 = Choice router1 router2
 
 -- | Datatype used for representing and debugging the
--- structure of a router. Abstracts from the functions
--- being used in the actual router and the handlers at
--- the leaves.
+-- structure of a router. Abstracts from the handlers
+-- at the leaves.
 --
 -- Two 'Router's can be structurally compared by computing
 -- their 'RouterStructure' using 'routerStructure' and
 -- then testing for equality, see 'sameStructure'.
 --
 data RouterStructure =
-    WithRequestStructure   RouterStructure
-  | StaticRouterStructure  (Map Text RouterStructure) Int
-  | DynamicRouterStructure RouterStructure
+    StaticRouterStructure  (Map Text RouterStructure) Int
+  | CaptureRouterStructure RouterStructure
   | RawRouterStructure
   | ChoiceStructure        RouterStructure RouterStructure
   deriving (Eq, Show)
@@ -87,18 +81,15 @@ data RouterStructure =
 -- | Compute the structure of a router.
 --
 -- Assumes that the request or text being passed
--- in 'WithRequest' or 'DynamicRouter' does not
+-- in 'WithRequest' or 'CaptureRouter' does not
 -- affect the structure of the underlying tree.
 --
-routerStructure :: Router' a -> RouterStructure
-routerStructure (WithRequest f) =
-  WithRequestStructure $
-    routerStructure (f (error "routerStructure: dummy request"))
+routerStructure :: Router' env a -> RouterStructure
 routerStructure (StaticRouter m ls) =
   StaticRouterStructure (fmap routerStructure m) (length ls)
-routerStructure (DynamicRouter f) =
-  DynamicRouterStructure $
-    routerStructure (f (error "routerStructure: dummy text"))
+routerStructure (CaptureRouter router) =
+  CaptureRouterStructure $
+    routerStructure router
 routerStructure (RawRouter _) =
   RawRouterStructure
 routerStructure (Choice r1 r2) =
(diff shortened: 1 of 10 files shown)


alpmestan:
Haven't looked thoroughly through the code, but is there anything to test that we're properly sharing the router now instead of recomputing it?

kosmikus:
@alpmestan Yes, good point. I've been running some ad-hoc benchmarks, and they do indicate that this patch is an overall improvement. This could be made more systematic, though.

I've tested this program:

Library.hs:
```
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE DataKinds, TypeOperators #-}
{-# LANGUAGE NoMonomorphismRestriction #-}
module Library where

import Language.Haskell.TH
import Servant.API

routeN :: Int -> TypeQ
routeN n = [t| "a" :> "b" :> "c" :> $(litT (strTyLit (show n))) :> Get '[JSON] () |]

routes :: Int -> [TypeQ]
routes n = map routeN [0..n - 1]

completeAPI :: Int -> TypeQ
completeAPI n = foldr1 (\ x y -> [t| $x :<|> $y |]) (routes n)

handler = return ()

handlers n = map (const [| handler |]) [0..n - 1]

completeHandlers :: Int -> ExpQ
completeHandlers n = foldr1 (\ x y -> [| $x :<|> $y |]) (handlers n)
```

Main.hs:
```
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE DataKinds, TypeOperators #-}
{-# OPTIONS_GHC -fcontext-stack=1500 #-}
module Main where

import Data.Proxy
import Library
import Servant.API
import Servant.Server

type API = Capture "param" Int :> $(completeAPI 200)

server :: Server API
server _ = $(completeHandlers 200)

main :: IO ()
main = run 8080 (serve (Proxy :: Proxy API) server)
```

And here are some results:

#### servant-0.4
```
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/0
Running 10s test @ http://localhost:8080/0/a/b/c/0
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   202.90us   23.88us   1.65ms   82.89%
    Req/Sec    25.32k     1.54k   28.22k    67.58%
  479030 requests in 10.00s, 38.37MB read
Requests/sec:  47903.64
Transfer/sec:      3.84MB
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/199
Running 10s test @ http://localhost:8080/0/a/b/c/199
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.56ms  274.30us   5.19ms   93.06%
    Req/Sec     2.05k   103.97     2.44k    71.59%
  39000 requests in 10.00s, 3.12MB read
Requests/sec:   3899.87
Transfer/sec:    319.91KB
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/200
Running 10s test @ http://localhost:8080/0/a/b/c/200
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.64ms  299.01us   5.46ms   94.78%
    Req/Sec     1.98k    99.98     2.33k    69.32%
  37791 requests in 10.00s, 4.79MB read
  Non-2xx or 3xx responses: 37791
Requests/sec:   3778.97
Transfer/sec:    490.82KB
```

#### servant-0.5
```
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/0
Running 10s test @ http://localhost:8080/0/a/b/c/0
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.70ms  293.63us   5.45ms   97.60%
    Req/Sec     1.95k   110.36     2.30k    65.79%
  36892 requests in 10.00s, 5.31MB read
Requests/sec:   3689.15
Transfer/sec:    544.01KB
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/199
Running 10s test @ http://localhost:8080/0/a/b/c/199
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.09ms  380.37us   6.35ms   97.78%
    Req/Sec     1.70k   138.65     2.00k    65.28%
  32361 requests in 10.00s, 4.66MB read
Requests/sec:   3235.89
Transfer/sec:    477.17KB
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/200
Running 10s test @ http://localhost:8080/0/a/b/c/200
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.65ms  328.84us   7.44ms   95.28%
    Req/Sec     1.98k   121.26     2.33k    67.02%
  37666 requests in 10.00s, 4.17MB read
  Non-2xx or 3xx responses: 37666
Requests/sec:   3766.41
Transfer/sec:    426.66KB
```

#### servant-0.7 (this PR)
```
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/0
Running 10s test @ http://localhost:8080/0/a/b/c/0
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   286.65us   29.23us   1.53ms   90.10%
    Req/Sec    18.11k     0.98k   20.33k    62.38%
  343300 requests in 10.00s, 49.44MB read
Requests/sec:  34330.36
Transfer/sec:      4.94MB
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/199
Running 10s test @ http://localhost:8080/0/a/b/c/199
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   483.90us   58.32us   1.47ms   60.66%
    Req/Sec    10.77k     1.20k   12.89k    45.68%
  207048 requests in 10.00s, 29.82MB read
Requests/sec:  20705.14
Transfer/sec:      2.98MB
andres@neli:~ 4 wrk http://localhost:8080/0/a/b/c/200
Running 10s test @ http://localhost:8080/0/a/b/c/200
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   196.82us   25.79us   1.66ms   82.33%
    Req/Sec    26.27k     1.63k   29.78k    61.38%
  494679 requests in 10.00s, 54.72MB read
  Non-2xx or 3xx responses: 494679
Requests/sec:  49468.49
Transfer/sec:      5.47MB
```

As you can see, for the given example with the top-level capture, `servant-0.5` (and 0.6) is actually worse than `servant-0.4`. For `servant-0.4`, we get the typical top-to-bottom interpretation, so accessing the first route is actually fast. Accessing the last route or an unfortunate non-existing route is slow, because they all have to be attempted.

For `servant-0.5`, in this scenario, everything is slow. This is because indeed, the top-level capture destroys all sharing, and the routing table is computed for every single request. So it doesn't matter whether the subsequent lookup is fast or not, the computation of the table dominates.

With this PR, we get relatively fast times for all the routes. Interestingly, we get a slower response time for `199` than for `0` (although still much faster than with `servant-0.4` or `servant-0.5`). I've been thinking about this, and I believe this is because while the routing itself (i.e., determining the right handler) is equally fast for all routes, actually running it (extracting it) is not. This phenomenon is also dependent on the top-level capture, which results in all the handlers being parameterized by a common function argument. I'll have to think some more if this slowdown is fixable somehow, but all in all, I think it is harmless compared to the response times we were getting before.

soenkehahn:
@kosmikus: We could change this PR (through the github interface) to be based on your other branch. Then the diff would be much easier to look at. Once the other one is merged, we change it back to be based on `master`. WDYT?

kosmikus:
Here's a link for the relative changes: https://github.com/kosmikus/servant/compare/show-router...kosmikus:fix-router-sharing

@soenkehahn I am not sure how/if I can change the branch this is based on in github to one in my fork and stlil have it show up as a PR for servant. If that's possible, feel free to do so. But I'll now squash and merge the other one anyway.

alpmestan:
@kosmikus Great, that's all I needed to hear =) Re: the slowdown you mention at the end, it looks quite neglectible compared to the performance improvement you're introducing here, so I'm not really worried.

LGTM.

soenkehahn:
@kosmikus: Ah, I thought this branch was in `haskell-servant/servant`, as you have write permissions here. Nevermind then.

Do you want to add something about this change to the changelog?

kosmikus:
One issue to note about this change:

It now becomes quite difficult to actually get hold of the request prior to reaching an endpoint. In other words, the only easy way to use the request is via `DelayedIO`, by adding something to the delayed code to be run if an endpoint is determined to be a candidate. This is desirable, exactly because if the router structure depends on the request, it cannot be shared between requests. It is still possible to get the request if one really needs it (by going via `RawRouter`), but actively discouraged by the design. I'm not currently aware of any code that would be inconvenienced by this, but I'm happy to be pointed to examples so that we can consider them.

@soenkehahn: Yes, I'll update the changelog.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -146,45 +136,46 @@ routerLayout router =
...
     first : rest
       -> let request' = request { pathInfo = rest }
-         in  runRouter (fun first) request' respond
-runRouter (RawRouter app)      request respond = app request respond
-runRouter (Choice r1 r2)       request respond =
-  runChoice [runRouter r1, runRouter r2] request respond
+         in  runRouterEnv router (first, env) request' respond
+runRouterEnv (RawRouter app)         env request respond = app env request respond
+runRouterEnv (Choice r1 r2)          env request respond =
+  runChoice [runRouterEnv r1, runRouterEnv r2] env request respond
```
soenkehahn: I suspect this would be easier to read with a `case` statement for the first argument (`Router env`).

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -95,138 +98,155 @@ toApplication ra request respond = ra request routingRespond
...
+withRequest f = DelayedIO (\ req -> runDelayedIO (f req) req)
+
+-- | A 'Delayed' without any stored checks.
+emptyDelayed :: RouteResult a -> Delayed env a
+emptyDelayed result =
+  Delayed (const r) r r r (\ _ _ _ _ -> result)
+  where
+    r = return ()
+
+instance Functor (Delayed env) where
```
soenkehahn: Wouldn't it be better to keep the instances for `Delayed` next to the datatype?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -95,138 +98,155 @@ toApplication ra request respond = ra request routingRespond
...
+  fmap f Delayed{..} =
+    Delayed
+      { serverD = \ c a b req -> f <$> serverD c a b req
+      , ..
```
soenkehahn: Nice use of `RecordWildCards`! :+1:

alpmestan:
> I'm not currently aware of any code that would be inconvenienced by this, but I'm happy to be pointed to examples so that we can consider them.

I'm happy with this. If this is problematic for some combinator, it probably means there should be another type of "check" in `Delayed`.

soenkehahn:
Maybe we should change the name of `DynamicRouter`. I suspect it was named that way since it depended on the request, but this PR changes that.

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -146,45 +136,46 @@ routerLayout router =
...
 --
-runChoice :: [RoutingApplication] -> RoutingApplication
-runChoice []       _request respond = respond (Fail err404)
-runChoice [r]       request respond = r request respond
-runChoice (r : rs)  request respond =
-  r request $ \ response1 ->
+runChoice :: [env -> RoutingApplication] -> env -> RoutingApplication
+runChoice []      _env _request respond = respond (Fail err404)
+runChoice [r]      env  request respond = r env request respond
+runChoice (r : rs) env  request respond =
```
soenkehahn: I think I would prefer a `case` statement on the list.

soenkehahn:
Apart from my comments, LGTM.

Great work!

kosmikus:
@soenkehahn Do you have a suggestion for a new name for `DynamicRouter`? `SkipRouter`, because it skips a path component until later?

soenkehahn:
@kosmikus: Not really. Maybe `SkipSegmentRouter`? Or just `SegmentRouter`?

alpmestan:
Given that it's only used for `Capture`, how about `CaptureRouter`?

kosmikus:
I'm happy to go with the suggestion of @alpmestan. I originally decided against it because I thought there's nothing capture-specific in the routing code, but looking at it from what it does (drop a path segment and pass it to the delayed part), then any use of it will always either ignore it completely or be somewhat like a capture.

soenkehahn:
Sounds good.

aaronlevin:
Is it specific to only `Capture`? I am trying to think if there is a scenario involving query params where this functionality could be useful, but am not coming up with anything. If not, I like `CaptureRouter` because it's very specific and `DynamicRouter` didn't ever really click with me :)

kosmikus:
Ok, I hope I addressed all comments and concerns with my latest commit.

soenkehahn:
Yes, looks like it. :)

PR #459, opened 2016-04-10
Fix code example in haddock  of servant-mock

joncfoo:
The example in the haddock does not coincide with the example code in servant-mock/example/main.hs

### DIFF
--- a/servant-mock/src/Servant/Mock.hs
+++ b/servant-mock/src/Servant/Mock.hs
@@ -36,7 +36,7 @@
 -- and call 'mock', which has the following type:
 --
 -- @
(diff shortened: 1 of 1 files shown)


soenkehahn:
LGTM.

There's also a `Proxy` missing here: https://github.com/haskell-servant/servant/pull/459/files#diff-3dd101a81e601f585d593000de99549dR39

joncfoo:
@soenkehahn Thanks, fixed.  Btw, should "context" be expanded on?  I don't see it mentioned in the doc.  Do you think "context" should be mentioned in the docs?  I would write something about it but I'm not knowledgeable enough at this moment.

Issue #460, opened 2016-04-11
Should accept mismatches affect routing?


MichaelXavier:
First off, apologies if this is covered somewhere, but I couldn't find it.

I'm porting an existing API to servant and noticed a curious behavior with routing. I've got an endpoint that can return either JSON or CSV content. The CSV API has query params that are relevant only to it and since query params are part of the type, I split the route. Here's an example using plaintext:

```haskell
{-# LANGUAGE DataKinds             #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings     #-}
{-# LANGUAGE TypeOperators         #-}
module Main where

import           Data.Aeson
import           Network.Wai.Handler.Warp
import           Servant

data Bar = Bar

instance ToJSON Bar where
  toJSON Bar = String "Bar"

instance MimeRender PlainText Bar where
  mimeRender _ Bar = "Bar"

type API
    = Get '[JSON] Bar
 :<|> QueryParam "n" Int :> Get '[PlainText] Bar

server :: Server API
server = jsonBar :<|> plainBar

jsonBar = return Bar

plainBar _ = return Bar

api :: Proxy API
api = Proxy

main = run 8000 (serve api server)
```

If you make an unqualified `GET http://localhost:8000`, it returns "Bar" with an `application/json content type`. That seems fine. If you make the same get with `Accept: text/plain`, you'll get a 406 Not Acceptable. I would expect the router logic to:

1. Check for an Accept. If none is provided, simply find the first route that matches (it does this today, but perhaps by coincidence).
2. If 1 or more accepts are supplied, find a route whose content type includes one of the given accept content types.
3. Failing all that, throw the 406.

I don't know the internals but it seems like route matching and accept checking are 2 distinct, sequential steps rather than being 1 or doing any sort of backtracking.

In the mean time, the workaround seems to be to combine routes and ignore irrelevant params, but it isn't clear to me how a server handler can know the media type requested, as all the rendering gets taken care of at the typeclass level. Maybe I could add the accept or media type to the route segments and it will get passed in as an argument?

kosmikus:
For some reason, if the `acceptCheck` fails, `FailFatal` is being used and other routes are no longer tried. I think this is wrong, but then again I'm known for generally opposing the idea of using `FailFatal` in most cases, whereas others think it's a good idea :)

So personally, I think you're right. Your example should work. And it would be a one-line fix to make it work.

[...]

PR #461, opened 2016-04-12
Do the accept check before the body check.

kosmikus:
This is a reasonably simple attempt at fixing #460.
By moving the accept check to a place before the body check,
we can make it recoverable (the body check is irreversible,
so everything done after the body check has to fail fatally).

The advantage is that we can now specify routes offering
different content types modularly. Failure to match one
is not fatal, and will result in subsequent routes being
tried.

The disadvantage is that we hereby bump the error priority
of the 406 status code. If a request contains a bad accept
header and a bad body, we now get 406 rather than 400. This
deviates from the HTTP decision diagram we try to follow,
but seems like an acceptable compromise for now.

### DIFF
--- a/servant-server/CHANGELOG.md
+++ b/servant-server/CHANGELOG.md
@@ -7,6 +7,10 @@
   efficiently. Functions `layout` and `layoutWithContext` have been
   added to visualize the router layout for debugging purposes. Test
   cases for expected router layouts have been added.
+* If an endpoint is discovered to have a non-matching "accept header",
+  this is now a recoverable rather than a fatal failure, allowing
+  different endpoints for the same route, but with different content
+  types to be specified modularly.
 * Export `throwError` from module `Servant`
 * Add `Handler` type synonym
 
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -154,10 +154,17 @@ methodCheck method request
   | allowedMethod method request = return ()
   | otherwise                    = delayedFail err405
 
+-- This has switched between using 'Fail' and 'FailFatal' a number of
+-- times. If the 'acceptCheck' is run after the body check (which would
+-- be morally right), then we have to set this to 'FailFatal', because
+-- the body check is not reversible, and therefore backtracking after the
+-- body check is no longer an option. However, we now run the accept
+-- check before the body check and can therefore afford to make it
+-- recoverable.
 acceptCheck :: (AllMime list) => Proxy list -> B.ByteString -> DelayedIO ()
 acceptCheck proxy accH
   | canHandleAcceptH proxy (AcceptHeader accH) = return ()
-  | otherwise                                  = delayedFailFatal err406
+  | otherwise                                  = delayedFail err406
 
 methodRouter :: (AllCTRender ctypes a)
              => Method -> Proxy ctypes -> Status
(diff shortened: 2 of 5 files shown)


kosmikus:
This is a suggestion, based on discussion with @soenkehahn. I'm not sure if this is the right trade-off to make, so feedback is certainly welcome.

soenkehahn:
This seems like a good idea to me.

I would prefer implementing this by adding a new field to `Delayed` (without having tried it out).

So LGTM.

kosmikus:
I think adding a new field to `Delayed

MichaelXavier:
@kosmikus thanks for the work on this! I'll let others closer to the library weigh in but to me that tradeoff sounds reasonable. It would be one thing if a request had a bad body and it gave a 406 or vice versa, but its hard for me to find fault with there being 2 errors with the request and it choosing one. This may be me but 406 seems like a more specific error code anyways, so the additional specificity seems like a nice outcome.

kosmikus:
I think this should get an explicit ok from at least @jkarni and @alpmestan before we're doing it.

alpmestan:
I have to admit that the argument put forward by @MichaelXavier is quite convincing. I can't think of any problem this might cause, so unless someone else comes up with a good reason not to merge this, I'm all for merging this patch.

jkarni:
I don't consider the HTTP decision diagram @kosmikus mentioned in #460 sacred - its just a good guideline. So I'm quite willing to diverge from it. My one concern was how to make whether a failure is fatal or not principled and clear to end users. "Everything but body check is not fatal" is fairly clear, but seems unprincipled, in that it leaks an implementation detail (that we don't want to keep the body in memory). 

An alternative in my mind is "everything that locates/identifies the resource is non-fatal". Locating a resource is what traditional web frameworks consider routing - url + method - whereas we do a lot more in a way that intermingles with that notion of routing (e.g. checking the body and auth creds). I *wanted* to go in this alternative direction, and in that light accept checks should also be fatal - they have nothing to do with routing, traditionally conceived. But I know @kosmikus prefers the expressiveness we gain from having non-fatal failures, and, unlike with request bodies (keeping them in memory is too high a cost for that expressiveness) I see the point.

So for me either options are fine.

kosmikus:
@jkarni I can see your point, but not why it is "deviating from REST principles". Can you explain that?
Personally, I don't find the rules you give (locating vs rest) any less arbitrary than "everything but the body check". But if there's a good reason why it is not, I'd like to know it (even if it's likely I'll still disagree ...)

I'd love to make the body-check non-fatal as well, but I cannot see a good way to do so. At least not in general. So yes, while I don't like it, I know that I must live with it for now.

BTW, if the goal was to restore the old error priority order, this is clearly possible, but would be a lot of work. But the way I interpret the current discussion, nobody seems to particularly care about that aspect of the change.

jkarni:
Having thought more about it, I take it back. This LGTM.

codedmart:
I like this change as well.

kosmikus:
Ok. @soenkehahn, do you insist on the extra field in `Delayed`, or can I merge this? Do we need a changelog entry for this?

soenkehahn:
No, I don't insist, please merge.

No strong opinion on a changelog entry.

alpmestan:
@kosmikus If you don't mind, I'd like to see a changelog entry about this.

PR #466, opened 2016-04-13
Fix some warnings

rvion:
(I use master version as extra-deps with stack and those warnings were annoying me in ghci)


arianvp:
Looks good. Thanks.   Why weren't these picked up by `-Werror` ?

soenkehahn:
LGTM.

@arianvp: We don't have `-Werror` on CI. There's an issue about that, and a branch, I believe.

PR #474, opened 2016-04-15
Add changelog to be included on hackage

wiz:


kosmikus:
I think this is good. But shouldn't it be done for all packages?

kosmikus:
Actually, `servant-client` seems to be the only other package where it's missing so far ...

kosmikus:
Both the cabal files actually already have an `extra-source-files` field. Could you please join these up, and also add the `README`, just as it is done for `servant-server`? Thanks.

wiz:
Sent cleaner PR in #475. Github online editor - not even once! :worried:

PR #476, opened 2016-04-16
Fix doctests failure when testpack or checkers are installed

trofi:
Test fails as:
  Test suite doctests: RUNNING...

  test/Servant/API/ContentTypesSpec.hs:31:18:
    Ambiguous module name `Test.QuickCheck.Instances':
      it was found in multiple packages:
      checkers-0.4.4@check_A5bAKHstANbBRqwFoOaIKx testpack-2.1.3.0@testp_BjTqfpWNTOG5Lwlc3iqqG9 quickcheck-instances-0.3.12@quick_3Tkh09kYN8p78zxMKFPcZI
  Test suite doctests: FAIL

Fixed by importing 'Test.QuickCheck.Instances' from "quickcheck-instances".

Signed-off-by: Sergei Trofimovich <siarheit@google.com>

### DIFF
--- a/servant/test/Servant/API/ContentTypesSpec.hs
+++ b/servant/test/Servant/API/ContentTypesSpec.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DeriveGeneric         #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PackageImports        #-}
 {-# LANGUAGE PolyKinds             #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 module Servant.API.ContentTypesSpec where
@@ -28,7 +29,7 @@ import           GHC.Generics
 import           Network.URL               (exportParams, importParams)
 import           Test.Hspec
 import           Test.QuickCheck
-import           Test.QuickCheck.Instances ()
+import "quickcheck-instances" Test.QuickCheck.Instances ()
 
 import           Servant.API.ContentTypes
 


arianvp:
I am not sure why you are experiencing this issue. I cannot seem to reproduce this locally. and I'm pretty sure we don't depend on either `testpack` or `checkers`. 

I use the following command to reproduce:

`stack test servant:doctests`

If this is an issue in your own project that depends on servant, I personally think it's not our task to add an exception case using `-XPackageImports` but you would be better off importing `testpack` and `checkers` with `-XPackageImports` in your own package.

trofi:
I don't have "a project" as is. I install all packages globally in the system (in a way distribution provides it).
```
$ ghc-pkg list | wc -l
1513
```
I think you'll be able to reproduce it in a sandbox or user-wide install with simple:
```
cabal install testpack
cabal test # in servant directory
```

The problem here is `doctests` style tests does not care about project's `.cabal` file and does not do proper package resolution for `ghci`. It hopes ghc to find proper packages. `Test.QuickCheck.Instances` happens to be a very popular module name:

https://github.com/haskell-servant/servant/blob/master/servant/test/Doctests.hs#L14

Sometimes people accept patches to add `-hide-all-packages -package=...` (like lens: https://github.com/ekmett/lens/blob/master/tests/doctests.hsc#L62)

Sometimes people like cleaner solution, like this. warp example: https://github.com/yesodweb/wai/commit/5d2863e99e226e306e1416d0078ea9ffa681d204

arianvp:
Personally I feel like it is a non-issue, because testpack isn't in our dependency-list and the test-suite can be run in a sandbox to avoid conflicting with a globally installed testpack. Right?

I might be wrong though. If someone else from @haskell-servant/maintainers could chip in here that'd be great.

Anyhow, it wouldn't do harm to merge.

soenkehahn:
@arianvp: It's a non-issue for you (and for me and maybe for most people), but cabal *does* support installing packages into the user-wide package database. And then we should still allow to run the test-suite.

@trofi: I wonder if this would better be solved when invoking `doctest`. I think there's also a `-hide-package` flag for hiding individual packages. Somehow it feels like selecting packages doesn't belong into the haskell source code. But that's just my gut feel, so I'll leave this up to you.

LGTM.

kosmikus:
I don't understand how doctest works, and I've basically given up running it locally. It fails for me with errors similar to the one described here all the time. There seems to be something fundamentally broken, but I doubt that this particular fix solves the real problem. What happens if other modules are provided by multiple packages? What happens if you have multiple versions of a library available in the package DB that provides a module?

soenkehahn:
`doctest` uses `ghci` under the hood and behaves pretty much like `ghci`. It forwards the command line arguments to `ghci` and reads the `.ghci` file. (Well, it spawns a `ghci` process, which will read that file.) So the rule of thumb is: If you can load a module in `ghci` you can execute it with `doctest`. And I just tried it: if you have `checkers` installed, you can't just say `ghci test/Spec.hs`, it'll fail with an error message similar to the one @trofi quoted.

@kosmikus: So saying that you have given up running `doctest` is also saying that you've given up loading the test-suite into `ghci`. Or do you have a way of doing that?

For some people (for example me) loading the test-suite in `ghci` is a very essential part of their development workflow. Personally I'm not affected by this, as I use either `stack` or `tinc` and then get sandboxed build environments. So I'm not *that* passionate about this PR. However I think we should try to support loading modules into `ghci`.

sol:
@soenkehahn Yes, exactly, except for one thing: `doctest` does not load `.ghci` files.  The reason is that `doctest` uses both `ghci` and the GHC API, and it's not straight forward to feed `.ghci` files to the GHC API.

kosmikus:
@soenkehahn You're right, I don't usually try to load the test suite into ghci. But ironically, I'd expect `cabal repl` for the testsuite to work just fine, because it will probably be invoked with the right flags. Whereas `doctest` is not. I'll have to verify this though.

soenkehahn:
Meanwhile I would tend to merge this. While the situation with `ghci` and `doctest` is not great, this change can't make it worse. And if it helps @trofi and some others to run the tests...

arianvp:
I agree with @soenkehahn . LGTM

PR #477, opened 2016-04-19
travis: enable -Wall -Werror

soenkehahn:


soenkehahn:
Please review. @haskell-servant/maintainers

soenkehahn:
Replaces #413.

servant-js/src/Servant/JS/Internal.hs
```diff
@@ -179,8 +179,8 @@ paramToStr qarg notTheEnd =
            <> "=' + encodeURIComponent("
            <> name
            <> if notTheEnd then ") + '" else ")"
-    Flag   -> name <> "="
-    List   -> name
+    Flag -> name <> "="
+    Servant.Foreign.List -> name
```
kosmikus: This change doesn't seem like an improvement to me. Asymmetrically qualifying native (as in servant-provided) names just because `Lens` exports too many entities?

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -255,7 +255,7 @@ instance
   type Foreign ftype (QueryParams sym a :> sublayout) = Foreign ftype sublayout
   foreignFor lang Proxy Proxy req =
     foreignFor lang (Proxy :: Proxy ftype) (Proxy :: Proxy sublayout) $
-      req & reqUrl.queryStr <>~ [QueryArg arg List]
+      req & reqUrl.queryStr <>~ [QueryArg arg Servant.Foreign.Internal.List]
```
kosmikus: Same as my other comment.

travis.sh
```diff
@@ -6,7 +6,7 @@ for package in $(cat sources.txt) doc/tutorial ; do
   echo testing $package
   pushd $package
   tinc
-  cabal configure --enable-tests --disable-optimization
+  cabal configure --enable-tests --disable-optimization --ghc-options='-Wall -Werror'
```
kosmikus: Should `-Wall` be specified here? It looks it's also specified in the individual cabal files. Furthermore, in some cabal files in some places, there are `-fno-warn-...` flags listed. My understanding is that this top-level `-Wall` would override all that. Do we want to make it impossible for components of individual packages to specify `-fno-warn-...` flags?

If yes, then I think they should be removed there. If not, I think we should pass just `-Werror` at this place and leave it to the cabal files what warnings to enable. The latter has my preference.

servant-js/src/Servant/JS/Internal.hs
soenkehahn: I've change this to use explicit import lists.

servant-foreign/src/Servant/Foreign/Internal.hs
soenkehahn: Changed too.

travis.sh
soenkehahn: Good point. I've removed `-Wall` from the script and cleaned up the `ghc-options` in the individual cabal files. (I've removed all the `-fno-warn-...` flags, since they weren't necessary to get warning free builds right now. But it should be possible to add them back in without being overridden by the travis script.)

servant-server/test/Servant/Server/Internal/ContextSpec.hs
```diff
@@ -1,9 +1,9 @@
 {-# LANGUAGE DataKinds #-}
-{-# OPTIONS_GHC -fdefer-type-errors #-}
+{-# OPTIONS_GHC -fdefer-type-errors -Wwarn #-}
```
jkarni: Aha! Didn't know about this flag.

jkarni:
LGTM

PR #478, opened 2016-04-20
Relocate Enter to `servant` package

amarpotghan:
@soenkehahn and I were pairing on implementing `staticClient` function which will allow providing `BaseUrl` (or any argument in general) statically. After considering a few approaches we realised that `Enter` is already capable of achieving that (with an additional instance). Hence we decided to move `Enter` to `servant` package so that we could use it in `servant-client` without depending on `servant-server`. 

Please let us know your thoughts.

### DIFF
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -43,6 +43,7 @@ import           Servant.Client.Experimental.Auth
 import           Servant.Common.BaseUrl
 import           Servant.Common.BasicAuth
 import           Servant.Common.Req
+import           Servant.Utils.Enter
 
 -- * Accessing APIs as a Client
 
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -43,7 +43,6 @@ import           Servant.Client.Experimental.Auth
 import           Servant.Common.BaseUrl
 import           Servant.Common.BasicAuth
 import           Servant.Common.Req
-import           Servant.Utils.Enter
 
 -- * Accessing APIs as a Client
 
--- a/servant/src/Servant/Utils/Enter.hs
+++ b/servant/src/Servant/Utils/Enter.hs
@@ -8,7 +8,7 @@
 {-# LANGUAGE TypeFamilies           #-}
 {-# LANGUAGE TypeOperators          #-}
 {-# LANGUAGE UndecidableInstances   #-}
-module Servant.Server.Internal.Enter where
+module Servant.Utils.Enter where
 
 import qualified Control.Category            as C
 #if MIN_VERSION_mtl(2,2,1)
--- a/servant-server/CHANGELOG.md
+++ b/servant-server/CHANGELOG.md
@@ -1,3 +1,8 @@
+0.7.1
+------
+
+* Remove module `Servant.Server.Internal.Enter` (https://github.com/haskell-servant/servant/pull/478)
+
 0.7
 ---
 
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -40,7 +40,6 @@ library
     Servant.Server.Internal
     Servant.Server.Internal.BasicAuth
     Servant.Server.Internal.Context
-    Servant.Server.Internal.Enter
     Servant.Server.Internal.Router
     Servant.Server.Internal.RoutingApplication
     Servant.Server.Internal.ServantErr
@@ -101,7 +100,6 @@ test-suite spec
   other-modules:
       Servant.Server.ErrorSpec
       Servant.Server.Internal.ContextSpec
-      Servant.Server.Internal.EnterSpec
       Servant.ServerSpec
       Servant.Server.UsingContextSpec
       Servant.Server.UsingContextSpec.TestCombinators
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -56,7 +56,6 @@ library
       , http-types         >= 0.8  && < 0.10
       , network-uri        >= 2.6  && < 2.7
       , mtl                >= 2    && < 3
-      , mmorph             >= 1
       , network            >= 2.6  && < 2.7
       , safe               >= 0.3  && < 0.4
       , servant            == 0.7.*
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -101,7 +101,7 @@ import           Data.Proxy                    (Proxy)
 import           Data.Text                     (Text)
 import           Network.Wai                   (Application)
 import           Servant.Server.Internal
-import           Servant.Server.Internal.Enter
+import           Servant.Utils.Enter
 
 
 -- * Implementing Servers
--- a/servant-server/src/Servant/Server/Internal/Context.hs
+++ b/servant-server/src/Servant/Server/Internal/Context.hs
@@ -18,7 +18,7 @@ import           GHC.TypeLits
 -- | 'Context's are used to pass values to combinators. (They are __not__ meant
 -- to be used to pass parameters to your handlers, i.e. they should not replace
 -- any custom 'Control.Monad.Trans.Reader.ReaderT'-monad-stack that you're using
--- with 'Servant.Server.Internal.Enter.enter'.) If you don't use combinators that
+-- with 'Servant.Utils.Enter'.) If you don't use combinators that
 -- require any context entries, you can just use 'Servant.Server.serve' as always.
 --
 -- If you are using combinators that require a non-empty 'Context' you have to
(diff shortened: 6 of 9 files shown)


servant-client/test/Servant/ClientSpec.hs
```diff
@@ -53,9 +56,9 @@ import           Test.QuickCheck
 import           Servant.API
 import           Servant.API.Internal.Test.ComprehensiveAPI
 import           Servant.Client
+import qualified Servant.Common.Req                         as SCR
 import           Servant.Server
 import           Servant.Server.Experimental.Auth
-import qualified Servant.Common.Req         as SCR
```
soenkehahn: Above is lots of whitespace change. Can you remove that from the PR, please?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -296,6 +299,22 @@ sucessSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
...
+              :<|> (_ ::  Person -> C.Manager ->  SCR.ClientM Person)
+              :<|> (_ ::  Maybe String -> C.Manager ->  SCR.ClientM Person)
+              :<|> (_ ::  [String] -> C.Manager ->  SCR.ClientM [Person])
+              :<|> (_ ::  Bool -> C.Manager ->  SCR.ClientM Bool)
+              :<|> (_ ::  HTTP.Method -> C.Manager -> SCR.ClientM (Int, BS.ByteString, MediaType, [HTTP.Header], C.Response BS.ByteString))
+              :<|> (_ ::  HTTP.Method -> C.Manager -> SCR.ClientM (Int, BS.ByteString, MediaType, [HTTP.Header], C.Response BS.ByteString))
+              :<|> (_ ::  String -> Maybe Int -> Bool -> [(String, [Rational])] -> C.Manager -> SCR.ClientM (String, Maybe Int, Bool, [(String, [Rational])]))
+              :<|> (_ ::  C.Manager -> SCR.ClientM (Headers TestHeaders Bool))
+              :<|> (_ ::  C.Manager -> SCR.ClientM NoContent) = staticClient baseUrl api
+        (left show <$> runExceptT (getGet' manager)) `shouldReturn` Right alice
```
soenkehahn: So this only works if we provide all the types? It's not enough to provide the type of `getGet'`?

servant-client/src/Servant/Client.hs
```diff
@@ -60,6 +62,24 @@ import           Servant.Common.Req
...
+-- >
+-- > getAllBooks :: Manager -> BaseUrl -> ClientM [Book]
+-- > postNewBook :: Book -> Manager -> BaseUrl -> ClientM Book
+-- > (getAllBooks :<|> postNewBook) = client myApi
+-- >
+-- > getAllBooks' :: Manager -> ClientM [Book]
+-- > postNewBook' :: Book -> Manager -> ClientM Book
+-- > (getAllBooks' :<|> postNewBook') = staticClient (BaseUrl Http "localhost" 8080 "/") myApi
+staticClient :: (Enter (Client layout) BaseUrl ret, HasClient layout) => BaseUrl -> Proxy layout -> ret
+staticClient b = enter b . client
```
soenkehahn: I'm not sure if `staticClient` is really the best name. Does anyone have better ideas?

servant/src/Servant/Utils/Enter.hs
```diff
@@ -39,6 +39,9 @@ instance ( Enter typ1 arg1 ret1, Enter typ2 arg2 ret2
 instance (Enter b arg ret) => Enter (a -> b) arg (a -> ret) where
     enter arg f a = enter arg (f a)
 
+instance Enter (arg -> ret) arg ret where
+    enter arg f = f arg
```
soenkehahn: It feels a bit arbitrary to have an instance like this. Maybe it's better to have some newtype wrapper for the argument?

``` haskell
instance Enter (arg -> ret) (InjectedArgument arg) ret where
  enter (InjectedArgument arg) f = f arg
```

servant-client/test/Servant/ClientSpec.hs
amarpotghan: Sure, will get rid of those

servant-client/test/Servant/ClientSpec.hs
amarpotghan: Yeah unfortunately full type is required here.

servant/src/Servant/Utils/Enter.hs
amarpotghan: Yeah that's a good idea, will change it

servant-client/test/Servant/ClientSpec.hs
jkarni: In theory that's not the case though. Is it the case here because `Enter` is missing a fundep? (Which may not be fixable though - there were some issues with a sensible fundep on GHC 7.6).

servant/src/Servant/Utils/Enter.hs
jkarni: Hm, this still seems like it muddles the purpose of `enter`. `enter` should apply a natural transformation to the return types of functions combined with `:<|>`. I don't think it should take on extra burdens, unless we can very clearly state it's new purpose (without using a conjuction or disjunction in the explanation). There are a lot of similar transformations I can imagine, and if we start including them all as instances for `enter` without a good overarching story, it'll become pretty hard to understand or explain what `enter` does.

I think I myself would prefer code that looks a lot like `enter` but just for what we want to do in the client. It's not like there'd be much duplicated code. Conceptually, if I'm getting this right, we need only two instances - one for recursing over `:<|>` and one for applying the client functions to the baseurl/manager arg.

servant-client/test/Servant/ClientSpec.hs
amarpotghan: Ah, I overlooked it, I was still thinking about `[HasArgument](https://github.com/haskell-servant/servant/blob/mappable/servant/src/Servant/Utils/Map.hs)` class in my head :). `Enter` already has required fundep, it should work without full type. Will try that tomorrow morning.

servant/src/Servant/Utils/Enter.hs
amarpotghan: I don't have strong opinion about this. Just that, we thought we were (almost) re-inventing `Enter` with our `HasArgument` class, but I get your point of doing more than one thing without sensible laws would be vague. I am all in for replicating `Enter`like class for `staticClient` purpose. WDYT @soenkehahn ?

soenkehahn: Yes, I think that's better. I wouldn't include it in `servant` then, though.

@jkarni: @amarpotghan and I tried for some time to come up with a nice general way of mapping over the leafs of a `:<|>`-`:>`-structure, but didn't manage. We hoped that this general mapping function would also allow to implement `staticClient`. Now since it's much less general, I think that this mapping should stay in `servant-client`.

(We could still move `enter` to `servant`, since it's been proven to be useful in e.g. `servant-client`, but that'd be a completely unrelated change. #NoScopeCreep)

amarpotghan:
@soenkehahn @jkarni : I removed `staticClient` stuff from this PR. It only contains moving `Enter` from `servant-server` to `servant`. I will create an another PR for `staticClient` later

jkarni:
I'm in favour of this. FWIW, I thought a little more about the general problem, and I think an option is making the return type of the `client` function a `Reader` monad over `BaseUrl` and `Manager`. Then `enter` would work without modification for this purpose.

(There are merge conflicts from the commits here that have already been merged)

3noch:
@jkarni Could you have an analog to `enter` for client functions? I.e. you provide a transformation from the client function result to some user-defined result type. Then anyone could construct such a `Reader` version of their client functions themselves.

jkarni:
@3noch `enter` itself should work as is for client functions (which is a large part of why this relocation makes sense)

3noch:
@jkarni Oh, awesome!

alpmestan:
Can't we remove the `mmorph` dep from `servant-server` then? Also, this requires updating the tutorial, the section about `enter` (for the import). Apart from that, LGTM!

amarpotghan:
removed the `mmorph` from servant-server. I couldn't find any Enter import
in doc though.

alpmestan:
Nevermind then. LGTM.

alpmestan:
Oh hang on, do we want to export `Servant.Utils.Enter` from some "core" module? Or should users just import it when they need it?

amarpotghan:
actually it's reexported from `Servant.Server`, exporting it from Servant would make more sense?

jkarni:
I'm okay leaving the export situation as is for now. When the uses for `enter` in other situations are worked out, we can export it from `servant`.

jkarni:
LGTM

jkarni:
Oh, could you add the change to the changelogs?

amarpotghan:
@soenkehahn @jkarni I tentatively put changelogs under 0.7.1, not sure if that's the correct version?

jkarni:
@amarpotghan yup - assuming we merge in time for the release (#499).

jkarni:
@alpmestan not clear whether your LGTM still holds?

alpmestan:
Yeah, nevermind my comment about exports, this can be tweaked later, let's not block this patch on some never-ending discussion about the optimal way to export `enter` =)

jkarni:
@amarpotghan thanks!

soenkehahn:
Great, thanks @amarpotghan!

PR #479, opened 2016-04-20
Update examples.md: diener-monad

rubenmoor:


doc/examples.md
```diff
@@ -11,3 +11,7 @@
 
 	Repository for templates for haskell projects, including some templates using
 	**servant**. These templates can be used with `stack new`.
+
+- ** [custom-monad](https://github.com/themoritz/diener)**:
+
+        A custom monad that can replace `IO` in servant applications. It adds among other things logging functionality and a reader (for database conncetions). A full usage example of servant/diener is also provided.
```
jkarni: `conncetions` -> `connections`.
`a reader` -> `a reader monad`?


Could you also line-break at around 80 chars?

doc/examples.md
```diff
@@ -11,3 +11,7 @@
 
 	Repository for templates for haskell projects, including some templates using
 	**servant**. These templates can be used with `stack new`.
+
+- ** [custom-monad](https://github.com/themoritz/diener)**:
```
jkarni: Maybe remove the space between `**` and `[` for consistency?

doc/examples.md
rubenmoor: sry, I oversaw your corrections here ...

jkarni:
LGTM

soenkehahn:
This isn't really an example for a project using `servant`, but more a library to use in conjunction with `servant`. Which means it should definitely be in the documentation, but maybe in another section? Something like `Useful Libraries`?

Otherwise LGTM.

arianvp:
Ahh diener as in the dutch word for servant.

I think putting it in examples for now is fine.  But maybe it's worth creating an issue for " Collect a list of useful libraries"

I'd say LGTM

arianvp:
@soenkehahn are you convinced? I'll merge this in that case

alpmestan:
LGTM, that can be useful for users. Ultimately, we should probably use the wiki for collecting all kinds of examples.

soenkehahn:
@arianvp: Sorry for being unresponsive. Not really convinced, but hey, I wasn't responsive. :)

:+1: for merging.

PR #484, opened 2016-04-25
WIP - Implement HasArgument (a la Enter) for servant-client to supply static arguments to baseUrl

amarpotghan:
Already discussed a bit in #478. 

@jkarni in #478 suggested that client functions return a Reader over BaseUrl and Manager so that we can use `Enter` as is for the purpose of `HasArgument` class. Being able to use `Enter` as is is definitely better solution, hence WIP.

### DIFF
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -63,7 +63,7 @@ import           Servant.Common.Req
 client :: HasClient layout => Proxy layout -> Client layout
 client p = clientWithRoute p defReq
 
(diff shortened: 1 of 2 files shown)


3noch:
It seems there was some discussion in the past regarding wrapping client functions in a custom monad. See this comment and following: https://github.com/haskell-servant/servant/pull/432#issuecomment-202701482

amarpotghan:
Ah I see, thanks for the pointing that out.

jkarni:
Closing (see #599 )

PR #485, opened 2016-04-25
for discussion only

mwotton:
I've had a bit of trouble working out how to use the new general authentication
framework on anything but trivial examples, so I thought it might make sense to
provide a simple example & a failing test of how I would expect to be able to
use it in the hope that someone can point out a flaw in my thinking.

### DIFF
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -395,10 +395,9 @@ genAuthSpec = beforeAll (startWaiApp genAuthServer) $ afterAll endWaiApp $ do
       (left show <$> runExceptT (getProtected authRequest manager baseUrl)) `shouldReturn` Right alice
 
     it "Also works for more complicated apis" $  \(_, baseUrl) -> do
-      let (getProtected :<|> getOtherProtected) = client genAuthAPI2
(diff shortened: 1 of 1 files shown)


mwotton:
thanks to @aaronlevin who's been helping me work through this.

jkarni:
I don't entirely understand what's happening here. Why is this test failing? And what more generally is the pattern?

mwotton:
I'm trying to work out how to protect an entire route with the new
authentication system. It works OK if I protect each :<|> branch
individually, but I can't work out how to do it at the top level and still
unpack the individual routes as API functions.

mwotton:
Entire app, I mean.

mwotton:
So, this makes it work :) may be worth leaving in as an example, I bashed my head against it for a while trying to work out how to get it working.

alpmestan:
It definitely can't hurt to leave it around or include something along those lines in the tutorial, or both!

soenkehahn:
I agree with @alpmestan, that ideally we'd have some better documentation about this, but this should happen in another PR.

Can we close this?

mwotton:
@soenkehahn yep, kill it with fire (but a flagship project showing use of the api would be extra-fantastic).

PR #505, opened 2016-05-10
Re-export Application.

### LINKED ISSUE #462: Re-export `Application` from `Servant` and `Servant.Server`
Since it's the result type of `serve`, I think this would make sense.

jkarni:
Fixes #462

### DIFF
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -41,8 +41,7 @@ import qualified Network.HTTP.Client        as C
 import           Network.HTTP.Media
 import qualified Network.HTTP.Types         as HTTP
 import           Network.Socket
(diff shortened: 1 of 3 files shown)


soenkehahn:
It's a bit buried in the discussions there (#NoMissionCreep), but #462 was about re-exporting `Application` from both `Servant` and `Servant.Server`. So this doesn't fully fix #462.

Still LGTM and could be merged as is (#NoMissionCreep).

jkarni:
Hm, why `Servant.Server` too? My intuition was the `Servant` was where all of the not-directly-`servant-server` stuff got exported.

soenkehahn:
My reasoning was that every module that exports `serve` should also re-export `Application`, because it's its return type.

jkarni:
Done.

alpmestan:
👍

PR #508, opened 2016-05-11
fix stack.yaml

soenkehahn:


soenkehahn:
Btw, this will revert `aeson` to `0.9`. Is there a good reason to diverge from what stackage does?

jkarni:
The original reason for 0.11 was that the tutorial (IIRC) needed some instance from it. But adding it to the stack file alone (without putting a version bound in the tutorial's cabal file) was the wrong thing to do...

soenkehahn:
Well, the tutorial at least compiles with `0.9`...

jkarni:
Then by all means!

jkarni:
@haskell-servant/maintainers can we merge?

alpmestan:
👍

arianvp:
:+1:

Tehnix:
With #520 and #521 merged, the `stack.yaml` update is not needed anymore - I've no idea about the rest though.

soenkehahn:
@Tehnix: Yes, some of the changes were obsolete. Just rebased.

PR #512, opened 2016-05-13
Dry build with stack, to test that yaml files are valid

phadej:
- `stack.yaml` seemed to be broken
- GHC-8.0 situation is a bit sad atm, as hvr's launchpad get updates once in a while. (would work with `--skip-ghc-check`, but not /that/ important atm I guess?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: false
 language: c
 
 env:
-  - GHCVER=7.8.4 CABALVER=1.22
-  - GHCVER=7.10.3 CABALVER=1.22
+  - GHCVER=7.8.4 CABALVER=1.22 STACK_YAML=stack-ghc-7.8.4.yaml
+  - GHCVER=7.10.3 CABALVER=1.22 STACK_YAML=stack.yaml
   - GHCVER=8.0.1 CABALVER=1.24
 
 addons:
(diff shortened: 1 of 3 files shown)


jkarni:
Some overlap with #508 .

For --dry-run, will `stack` try to install a different GHC?

phadej:
@jkarni, no, `stack` won't install anything. It only downloads its own copy of Hackage, but that is cached, and updated only when a package version aren't found in the index.

soenkehahn:
`stack test` doesn't work on current `master`, which this added check fails to detect. My guess is that test dependencies are ignored.

And I still wonder whether it would be better to have a script that runs `stack test` for every stack file that we can run manually. Instead of putting this into the travis file. In the interest of keeping the CI simpler.

soenkehahn:
We now have a `scripts/test-stack.sh` script that tests all the included stack files. It's not executed as part of the CI.

I would prefer to not merge this PR in the interest of keeping the CI simpler. If you want to make absolutely sure, that the stack files work, you have to call the above-mentioned script anyways. (I'm saying, this PR only catches a subset of possible problems and I don't think it's worth the added complexity.)

I would consider switching the whole CI over to `stack` though. I'll open another issue for that.

jkarni:
See also #568

soenkehahn:
#568 is merged, closing.

PR #519, opened 2016-05-26
CaptureAll combinator - take #2

### LINKED ISSUE #257: Combinator for capturing all remaining segments

jml:
Closes #257.

Builds on @fizruk's work in #347, but updated for recent changes to servant, and including support for foreign & docs.

---
This change is [Reviewable](https://reviewable.io/reviews/haskell-servant/servant/519)

### DIFF
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -105,7 +105,7 @@ type Api =
        "get" :> Get '[JSON] Person
   :<|> "deleteEmpty" :> DeleteNoContent '[JSON] NoContent
   :<|> "capture" :> Capture "name" String :> Get '[JSON,FormUrlEncoded] Person
-  :<|> "captureAll" :> CaptureAll "names" String :> Get '[JSON] Person
+  :<|> "captureAll" :> CaptureAll "names" String :> Get '[JSON] [Person]
   :<|> "body" :> ReqBody '[FormUrlEncoded,JSON] Person :> Post '[JSON] Person
   :<|> "param" :> QueryParam "name" String :> Get '[FormUrlEncoded,JSON] Person
   :<|> "params" :> QueryParams "names" String :> Get '[JSON] [Person]
@@ -126,7 +126,7 @@ api = Proxy
 getGet :: C.Manager -> BaseUrl -> SCR.ClientM Person
 getDeleteEmpty :: C.Manager -> BaseUrl -> SCR.ClientM NoContent
 getCapture :: String -> C.Manager -> BaseUrl -> SCR.ClientM Person
-getCaptureAll :: [String] -> C.Manager -> BaseUrl -> SCR.ClientM Person
+getCaptureAll :: [String] -> C.Manager -> BaseUrl -> SCR.ClientM [Person]
 getBody :: Person -> C.Manager -> BaseUrl -> SCR.ClientM Person
 getQueryParam :: Maybe String -> C.Manager -> BaseUrl -> SCR.ClientM Person
 getQueryParams :: [String] -> C.Manager -> BaseUrl -> SCR.ClientM [Person]
@@ -158,7 +158,7 @@ server = serve api (
        return alice
   :<|> return NoContent
   :<|> (\ name -> return $ Person name 0)
-  :<|> (\ (name : _) -> return $ Person name 0)
+  :<|> (\ names -> return (zipWith Person names [0..]))
   :<|> return
   :<|> (\ name -> case name of
                    Just "alice" -> return alice
@@ -255,7 +255,8 @@ sucessSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
       (left show <$> runExceptT (getCapture "Paula" manager baseUrl)) `shouldReturn` Right (Person "Paula" 0)
 
     it "Servant.API.CaptureAll" $ \(_, baseUrl) -> do
-      (left show <$> runExceptT (getCaptureAll ["Paula", "Peta"] manager baseUrl)) `shouldReturn` Right (Person "Paula" 0)
+      let expected = [(Person "Paula" 0), (Person "Peta" 1)]
+      (left show <$> runExceptT (getCaptureAll ["Paula", "Peta"] manager baseUrl)) `shouldReturn` Right expected
 
     it "Servant.API.ReqBody" $ \(_, baseUrl) -> do
       let p = Person "Clara" 42
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -105,6 +105,7 @@ type Api =
        "get" :> Get '[JSON] Person
   :<|> "deleteEmpty" :> DeleteNoContent '[JSON] NoContent
   :<|> "capture" :> Capture "name" String :> Get '[JSON,FormUrlEncoded] Person
+  :<|> "captureAll" :> CaptureAll "names" String :> Get '[JSON] Person
   :<|> "body" :> ReqBody '[FormUrlEncoded,JSON] Person :> Post '[JSON] Person
   :<|> "param" :> QueryParam "name" String :> Get '[FormUrlEncoded,JSON] Person
   :<|> "params" :> QueryParams "names" String :> Get '[JSON] [Person]
@@ -125,6 +126,7 @@ api = Proxy
 getGet :: C.Manager -> BaseUrl -> SCR.ClientM Person
 getDeleteEmpty :: C.Manager -> BaseUrl -> SCR.ClientM NoContent
 getCapture :: String -> C.Manager -> BaseUrl -> SCR.ClientM Person
+getCaptureAll :: [String] -> C.Manager -> BaseUrl -> SCR.ClientM Person
 getBody :: Person -> C.Manager -> BaseUrl -> SCR.ClientM Person
 getQueryParam :: Maybe String -> C.Manager -> BaseUrl -> SCR.ClientM Person
 getQueryParams :: [String] -> C.Manager -> BaseUrl -> SCR.ClientM [Person]
@@ -140,6 +142,7 @@ getDeleteContentType :: C.Manager -> BaseUrl -> SCR.ClientM NoContent
 getGet
   :<|> getDeleteEmpty
   :<|> getCapture
+  :<|> getCaptureAll
   :<|> getBody
   :<|> getQueryParam
   :<|> getQueryParams
@@ -155,6 +158,7 @@ server = serve api (
        return alice
   :<|> return NoContent
   :<|> (\ name -> return $ Person name 0)
+  :<|> (\ (name : _) -> return $ Person name 0)
   :<|> return
   :<|> (\ name -> case name of
                    Just "alice" -> return alice
@@ -249,6 +253,9 @@ sucessSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
 
     it "Servant.API.Capture" $ \(_, baseUrl) -> do
       (left show <$> runExceptT (getCapture "Paula" manager baseUrl)) `shouldReturn` Right (Person "Paula" 0)
+
+    it "Servant.API.CaptureAll" $ \(_, baseUrl) -> do
+      (left show <$> runExceptT (getCaptureAll ["Paula", "Peta"] manager baseUrl)) `shouldReturn` Right (Person "Paula" 0)
 
     it "Servant.API.ReqBody" $ \(_, baseUrl) -> do
       let p = Person "Clara" 42
@@ -351,7 +358,7 @@ failSpec = beforeAll (startWaiApp failServer) $ afterAll endWaiApp $ do
           _ -> fail $ "expected UnsupportedContentType, but got " <> show res
 
       it "reports InvalidContentTypeHeader" $ \(_, baseUrl) -> do
-        let (_ :<|> _ :<|> _ :<|> getBody :<|> _) = client api
+        let (_ :<|> _ :<|> _ :<|> _ :<|> getBody :<|> _) = client api
         Left res <- runExceptT (getBody alice manager baseUrl)
         case res of
           InvalidContentTypeHeader "fooooo" _ -> return ()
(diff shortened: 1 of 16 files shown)


jml:
Any chance I could get a review of this?

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -155,6 +158,7 @@ server = serve api (
        return alice
   :<|> return NoContent
   :<|> (\ name -> return $ Person name 0)
+  :<|> (\ (name : _) -> return $ Person name 0)
```
fizruk: Why not use the whole list? E.g. `\names -> return $ zipWith Person names [0..]`.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -211,6 +211,21 @@ instance (KnownSymbol sym, HasForeignType lang ftype t, HasForeign lang ftype su
         { _argName = PathSegment str
         , _argType = ftype }
 
+instance (KnownSymbol sym, HasForeignType lang ftype [t], HasForeign lang ftype sublayout)
+  => HasForeign lang ftype (CaptureAll sym t :> sublayout) where
+  type Foreign ftype (CaptureAll sym t :> sublayout) = Foreign ftype sublayout
+
+  foreignFor lang Proxy Proxy req =
+    foreignFor lang Proxy (Proxy :: Proxy sublayout) $
+      req & reqUrl . path <>~ [Segment (Cap arg)]
```
fizruk: Doesn't this code mean a single path segment (not _all_ the remaining segments)?

fizruk:
Apart from `servant-foreign` code everything looks good to me! 👍

jml:
Thanks for the review @fizruk! 

I think the `servant-foreign` code is actually OK, but would be super happy if you or someone else could show me why it's wrong (perhaps giving me enough clues so I can write a test that demonstrates it?)

Looking forward to hearing back from you.

servant-foreign/src/Servant/Foreign/Internal.hs
jml: I don't _think_ so, but I have to confess I don't know exactly what `<>~` does. 

AFAICT, all this code has to do is generate signatures of the correct type, rather than capturing values. In which case, it's OK as is, because:

 1. `arg` here is different from the `arg` in `Capture` because here the `ftype` is a proxy around `[t]` rather than `t`.
 2. the test in `ForeignSpec.hs` verifies that it generates valid signatures

But I'll happily admit I don't understand this code very well, so I'd be glad to be corrected.

servant-client/test/Servant/ClientSpec.hs
jml: Good point. Changed as per your suggestion (although I've used parens rather than `$` to be consistent w/ existing code).

jml: Done, although used parens to be consistent w/ existing code.

arianvp:
So I think  it's okay that arg is a `[t]` instead of a `t`.  But I'm not sure if it would break servant-js code. I'd have to check. I'll have a look at it tonight.

jml:
@arianvp Thanks for checking up on this. If what I've got is wrong, then I'd *really* appreciate a unit test that demonstrates it (even if it's just the outline of one), as that will help me understand.

arianvp:
Good point. I will see if I can generate a testcase that can express my 'concern'

jml:
Thanks! Let me know how it goes.

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -702,6 +702,22 @@ instance (KnownSymbol sym, ToCapture (Capture sym a), HasDocs sublayout)
           symP = Proxy :: Proxy sym
 
 
+-- | @"books" :> 'CaptureAll' "isbn" Text@ will appear as
+-- @/books/:isbn@ in the docs.
```
jkarni: Maybe `/books/[:isbn]` or something distinct from a single capture?

jkarni:
Sorry for the delay. This generally looks really good to me. I just had a small note on `-docs`, and `CaptureAll` needs to be added to [`ComprehensiveAPI`](https://github.com/haskell-servant/servant/blob/a35aa161a67f52d93db2277858afe1f761df45bd/servant/src/Servant/API/Internal/Test/ComprehensiveAPI.hs).

arianvp:
Do we also need an -js instance @jkarni? I'll try to add it this weekend

jkarni:
@arianvp that'd be great. I imagine `comprehensiveAPI` will tell us if we are missing any other instances besides.

jml:
Great, thanks. @arianvp I've invited you to collaborate on my fork, in case that helps. 

Am very excited to get this done & landed in master.

arianvp:
Okay so my current problem with this is the following js code is generated for this definition:

```
type GET = Get '[JSON] ()
type API =   CaptureAll "foo" Int :> GET
```

```
var getByFoo = function(foo, onSuccess, onError)
{
  var xhr = new XMLHttpRequest();
  xhr.open('GET', '/' + encodeURIComponent(foo) + '', true);
  xhr.setRequestHeader("Accept","application/json");
  xhr.onreadystatechange = function (e) {
    if (xhr.readyState == 4) {
      if (xhr.status == 204 || xhr.status == 205) {
        onSuccess();
      } else if (xhr.status >= 200 && xhr.status < 300) {
        var value = JSON.parse(xhr.responseText);
        onSuccess(value);
      } else {
        var value = JSON.parse(xhr.responseText);
        onError(value);
      }
    }
  }
  xhr.send(null);
}

```


Current behaviour:

```
getByFoo([1,2,3],  onSuccess, onFailure);    // queries  /'1%2C2%2C3'
```

Expected behaviour:

```
getByFoo([1,2,3], onSuccess, onFailure); // queries  /1/2/3
```

arianvp:
@jml could you try rebase this onto master? I have a feeling the build is broken because of #532   . @phadej can you check if this is true?

jkarni:
I restarted the builds

jkarni:
@arianvp do we need  `foo.map(encodeURIComponent).join('/')`?

arianvp:
@jkarni yes that's what I had in mind

jml:
Rebased.

jml:
Spent 30 mins this morning before work trying to update the JS according to @jkarni's suggestion.

Here are my notes:

* assume that jkarni's suggestion only applies to capture-all / lists. Pretty sure Javascript floats don't implement `.map()`
* the bit that needs to change is `segmentTypeToStr` in `Servant.JS.Internal`
* as currently written, it would need to dispatch based on `argType s`, but I don't really see a way of looking at `s` and telling whether it's a `CaptureAll` or not
* query arguments solve this by having extra information in the argument type, but...
  * if I add a branch to `SegmentType` (e.g. `| CapAll (Arg f)`) then should `isCapture (CapAll x)` be `True` or `False`?
  * similarly, is it still a `SegmentType` (singular) even though it describes multiple segments?
* `typeFor` seems built for what I'm trying to do, but I can't figure out how to use it
* I tried writing a test, but what I really want to test is the `url` helper in the various JS generators
  * it looks the same for all of the implementations, so maybe I should extract it to `Internal.hs` and just write at test for it
 
No new code written.

soenkehahn:
What about *not* writing an interpretation for `servant-foreign` (or `servant-js`)? Not sure, but it seems a bit as if no one participating in this PR really needs this feature, at least not yet. I think that's always a red flag -- writing code that has no clear user yet. Please correct me if my assumption is wrong and someone is looking forward to using this in practice.

Of course it's kind of nice to have every interpretation in the `servant` repo support every combinator in the repo. But we could

1. deviate from that policy or
2. put `servant-foreign` in it's own repo and thereby make it more explicit that these instances are not necessarily provided for everything or
3. put the `CaptureAll` combinator in it's own package. Which brings me to my second comment:

What about putting `CaptureAll` in it's own package? That way it would be very easy to just start out with a few interpretations that you care about and let others already try out the combinator. And gather some feedback. I would prefer that, tbh.

jml:
100% in favour of not having a foreign implementation for this. I only did the work because others seemed to consider it necessary to get it merged.

Re own package:

* I'm already using this code in a fork of servant, so at least for me there'd be minimal testing benefit to having it in a separate package—it works great!
* With servant 0.4, I had this code in my own repo. However, servant 0.5 made it impossible to do this correctly using public APIs, so it had to be added directly. I don't know if recent changes have fixed this. 

My motivation for doing this work (apart from trying to be a good open source citizen) is to make sure that future releases of servant don't break this functionality for me.

jkarni:
Several people, both in person and on IRC, asked about capturing multiple path segments. I've wanted it myself. I'm very much in favor of having this in the repo. Moreover, as @jml mentioned, this can't now be *purely* an external package, because it requires changes to `Router'` etc. 

As for the `js` and `foreign`, I'm slightly in favor of implementing them, since it must be a little annoying for people to have parts of the ecosystem not work, and because it's nice to keep everything tested against `comprehensiveAPI`. But my feelings aren't that strong. 

If we do decide to add it, the name `SegmentType` still seems okay for me. I'm in favor of deprecating `isCapture`, but for now I'd say that `CaptureAll` should count, since the use case is clearly checking whether the path should be added to the argument list.

soenkehahn:
> Several people, both in person and on IRC, asked about capturing multiple path segments. I've wanted it myself.

Yeah, I've never questioned that. I wondered whether there's any users interested in the `servant-js` interpretations. `CaptureAll` is clearly useful.

> I'm very much in favor of having this in the repo.

But why? Why is this worse as a separate package? Discoverability? Or just the dependency problem?

> Moreover, as @jml mentioned, this can't now be purely an external package, because it requires changes to Router' etc.

Well, my vision for having more combinators in external packages is that we would closely work with people writing new combinators and if they need any changes to internals like `Router'` we would *of course* do our best to make these changes. That still doesn't mean that the combinators have to be in this repo.

Regarding `servant-js` instances: I'm convinced that it'll be a strong advantage for the `servant` ecosystem if we make it very easy to 1) write new combinators and 2) write new interpretations. I think we should encourage experimentation and we shouldn't ask people who want to contribute to write a bunch of instances that they don't even care about, just to be able to make a (possibly experimental) combinator or interpretation available publicly for others. So *if* we want to merge this PR I'm strongly against asking @jml for the `servant-js` stuff.

(In any case I'd be in favor of putting `servant-foreign` in its own repo.)

jkarni:
>  Or just the dependency problem?

That, which is to my mind pretty serious. And the fact that it can bitrot more easily. And the fact that it feels strange to have code in this repo (in `Router'`) whose only purpose, to a first approximation, is enabling one particular combinator, but having that combinator in a different repo; it means, for example, that we can't actually functionally test that section of our code. 

> Regarding servant-js instances: I'm convinced that it'll be a strong advantage for the servant ecosystem if we make it very easy to 1) write new combinators and 2) write new interpretations. I think we should encourage experimentation and we shouldn't ask people who want to contribute to write a bunch of instances that they don't even care about, just to be able to make a (possibly experimental) combinator or interpretation available publicly for others. So if we want to merge this PR I'm strongly against asking @jml for the servant-js stuff.

I agree; this is a point I hadn't much considered, but is very important. It'd be nice to document what instances are provided in some way though, to avoid the situation I mentioned. Preferably automatically. Anyhow, I'm okay nixing -foreign and -js for now, and opening another issues about how to keep track of what combinators have what instances.

arianvp:
Okay. so lets resolve these conflicts and merge?

soenkehahn:
> That, which is to my mind pretty serious.

Yes, true. I've created [servant-bazaar](https://github.com/haskell-servant/servant-bazaar) which is one attempt to solve this problem, there may be others. But I agree it's a problem.

> And the fact that it can bitrot more easily.

I think this has a bad and a good side. If a package bitrots, it means it's not being used anymore. And nobody should put in work maintaining it. On the other hand, if we keep merging combinators, they won't bitrot, because we are forced to maintain them. But we don't have a good way of telling whether they are still being used. I'm afraid we'll end up maintaining combinators for a long time, that are not really all that useful to anybody.

> And the fact that it feels strange to have code in this repo (in Router') whose only purpose, to a first approximation, is enabling one particular combinator, but having that combinator in a different repo; it means, for example, that we can't actually functionally test that section of our code.

This doesn't feel strange to me at all. We're writing a plugin system and it's very natural to have features in that system that allow plugins to do certain things, without necessarily having a plugin that uses these features in the same repo. And of course we can test the features by writing test plugins (as in test combinators).

But I'm sorry, I'm straying very far from the issue at hand (whether to merge this PR or not). But my opinion stays the same, I'd be in favor of not merging.

jml:
I've rebased on master, resolving the conflicts but leaving `comprehensiveApi` as it was.

@soenkehahn You make good points: servant ought to be easy to extend externally; more combinators in means more maintenance overhead for the servant team; and one of the many difficulties of open source software is that it's nearly impossible to tell which parts of your software actually get used.

You've got to draw the line somewhere. I just don't think it should be here.

`CaptureAll` should be a part of servant. It offers basic functionality that's frequently asked for that relies on APIs that aren't exposed publicly and have changed significantly since `CaptureAll` was first implemented. Similar packages in other languages provide support for capturing multiple path segments (e.g. flask's [path](http://flask.pocoo.org/docs/0.11/quickstart/#variable-rules)).

soenkehahn:
@jml: Yeah. There's certainly combinators in `servant` that deserve less to be in there than `CaptureAll`. And since @arianvp and @jkarni seem to be convinced already, you got your two :+1:s. So let's go ahead and merge. :)

(Looking forward to using the combinator.)

soenkehahn:
@jml, @arianvp: Would anyone be up for adding something to the changelog about `CaptureAll`?

jml:
@soenkehahn I can write it for you. Better for someone with merge privileges to do it.

```
* Add `CaptureAll` combinator. Use this to capture all of the remaining segments in a URL.
```

Not sure if it should go in servant or servant-server.

soenkehahn:
@jml: Thanks.

PR #520, opened 2016-05-27
Update stack.yaml to LTS 6.0

Tehnix:
All the missing extra-deps are now on Stackage, and the ones that were 
forced to a later version have also been updated, so the LTS should suffice.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -11,11 +11,4 @@ packages:
 - servant-server/
 - doc/tutorial
 extra-deps:
-- base-compat-0.9.0
-- engine-io-wai-1.0.2
-- control-monad-omega-0.3.1
-- should-not-typecheck-2.0.1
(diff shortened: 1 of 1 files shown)


phadej:
Could you update `stack-ghc-8.0.yaml` to use the latest nightly as well?

soenkehahn:
@phadej: I don't think we should ask @Tehnix to do additional work if not necessary. They might not care about ghc-8 at all.

@Tehnix: LGTM as is.

phadej:
@soenkehahn made a https://github.com/haskell-servant/servant/pull/521 for GHC-8.0

soenkehahn:
@Tehnix: Thanks!

Tehnix:
@soenkehahn @phadej Hehe, I definitely wouldn't have minded doing that, but only just saw the updates here just now and can see it's already taken care of :)

PR #523, opened 2016-06-02
Rename type variables 'layout' and 'sublayout' to 'api'

soenkehahn:
I think `api` much more clearly expresses, what this really is.


arianvp:
I'm neutral about this. Not really for it or against it. But LGTM if the rest is fine with it

jkarni:
LGTM

Issue #465, opened 2016-04-13
Add authorization combinators to `Servant.API.Internal.Test.ComprehensiveAPI`


soenkehahn:
cc: @aaronlevin 

This would've revealed #463 earlier.

Issue #365, opened 2016-01-28
Document How To Respond To Requests at "/"


m4dc4p:
Spent the last hour trying to figure out how to serve "Raw" files from requests to the root directory. That is, requests to "http://foo.com/".

I tried `type FileAPI = "" :> Raw`, and then I used `serveDirectory` to respond to requests to the root URL. But my server always gave 404, even if the file was in the directory.

Reading through closed issues, I found #195, which had a signature similar to what I wanted. Based on the comment at  https://github.com/haskell-servant/servant/issues/195#issuecomment-132013000, I figured out the type I wanted was just `type FileAPI = Raw`.

Perhaps this is worth adding to the "tips & tricks" section, calling out in the tutorial, or maybe documenting with the `:>` data type?

alpmestan:
You're right. This piece of information is only given implicitly at the moment, it can't hurt to make it louder and explicit.

taakazazzz:
And if it wasn't "Raw" but `Get '[JSON] [Something]`,how would I do that?


    type PersonAPI = "" :> Get '[JSON] [Person] -- doesn't work
    type PersonAPI = "/" :> Get '[JSON] [Person] -- doesn't work

soenkehahn:
Fixed in #553.

PR #524, opened 2016-06-05
Authorization combinator test

supermario:
I've started looking at  #465 as it had the `newcomer-friendly` tag, but I've either missed something simple or this is much more complicated that it sounds.

- I've extended `ComprehensiveAPI` and get complaints about missing `ToAuthInfo` and `HasContextEntry` instances.
- I think I've roughly grokked the `ToAuthInfo` instance, but unsure where it should go (`Servant/Docs/Internal` where I've got it seems obviously wrong)
- The `HasContextEntry` instance I'm completely lost.

Would appreciate some pointers, I was inspired by [this](https://www.reddit.com/r/haskell/comments/4m5p8k/what_haskell_projects_can_expert_beginners/d3sudo9) but if this is misguided I'm happy to close.

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -517,6 +517,10 @@ class ToCapture c where
 class ToAuthInfo a where
       toAuthInfo :: Proxy a -> DocAuthentication
 
+instance ToAuthInfo (BasicAuth "foo-realm" ()) where
+  toAuthInfo _ = DocAuthentication "herp" "derp"
+
+
 -- | Generate documentation in Markdown format for
 --   the given 'API'.
 markdown :: API -> String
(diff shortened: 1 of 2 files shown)


soenkehahn:
@supermario: Sorry, I added the `newcomer-friendly` flag because I thought it would be a simple one-line change. But it's not and it requires some knowledge about type families, how the authentication combinator works and how `Context` works. Sorry, about that. I've removed that label.

It also seems to me as if there are some issues that are rather easy, but not tagged with `newcomer-friendly`. (For example documentation issues like this one: https://github.com/haskell-servant/servant/issues/365.) So maybe you just want to browse through the whole list...

supermario:
@soenkehahn ok, thanks for letting me know.

I'll try take a look around, but IMO having a `newcomer-friendly` tag and not using will probably be more discouraging than having no `newcomer-friendly` tag at all. 😞

soenkehahn:
@supermario: Yes, it's not ideal. @arianvp: Any comment on this? Should we get rid of the `newcomer-friendly` label?

PR #526, opened 2016-06-14
Fix example `MimeRender` instance

ehamberg:
The `MimeRender` instance in the documentation was incomplete and expected one more argument.


jkarni:
Thanks!

PR #530, opened 2016-07-01
CPP deprecated parseUrl function.

jkarni:


jkarni:
Failing on < 7.10 due to [this](https://github.com/informatikr/aeson-pretty/pull/18)

servant-client/src/Servant/Common/Req.hs
```diff
@@ -129,6 +129,9 @@ reqToRequest req (BaseUrl reqScheme reqHost reqPort path) =
                                               | not . null . reqAccept $ req] }
         toProperHeader (name, val) =
           (fromString name, encodeUtf8 val)
+#if !MIN_VERSION_http_client(0,4,30)
+        parseUrlThrow = parseUrl
+#endif
```
soenkehahn: I think I would've declared this top-level. I think that'd be more intuitive. Not a very strong argument, though.

soenkehahn:
LGTM.

PR #532, opened 2016-07-05
Add upper bounds http-client <0.5

phadej:
I'd like to add bounds for all other dependencies as well.

screen shot 2016-07-05 at 12 24 56

 SItuations like are good example why you should have them. ping @hvr


arianvp:
LGTM

PR #536, opened 2016-07-05
Fix servant-js examples

jkachmar:
The Servant.JS Counter example was broken:

- `axios` was not completely applied; it now uses `defAxiosOptions` similarly to the `angular` example
- `OverloadedStrings` was required, but not included

### DIFF
--- a/servant-js/examples/counter.hs
+++ b/servant-js/examples/counter.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE TypeOperators              #-}
 
 import           Control.Concurrent.STM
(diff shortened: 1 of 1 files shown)


christian-marie:
Thanks.

Could we add this file to the test suite in the cabal file just to have it compile? I think that should work, but am not entirely sure.

arianvp:
@jml could you try rebase this onto master? I have a feeling the build is broken because of #532   . @phadej can you check if this is true?

phadej:
yeah, `http-client-0.5` is picked

phadej:
Also closing/reopening PR will make github to create new merge commit, on current master. So it's not mandatory to rebase. ping @arianvp

jkarni:
I just restarted the builds.

jkachmar:
@arianvp Should I add this to the servant-js test suite and rebase, or does this looks okay as-is?

arianvp:
@jkachmar You could add examples to  `src-dirs`  and to `Other-Modules` for the test suits perhaps. then if there are mistakes in examples, the build at least fails.  

But i'm fine merging as is as well.

soenkehahn:
LGTM as is.

:+1: on adding this to the test-suite, but could be a different PR.

jkarni:
LGTM.

Issue #534, opened 2016-07-05
Add dependency bounds


phadej:
`http-client` issue is a good example why we should have them.

I can make a PR adding them, with lower-bounds from say *Stackage lts-2.22*, and upper-bounds allowing current major versions (if they do work, of course).

soenkehahn:
I'm against adding preemptive upper bounds. While the situation with `http-client-0.5` is a good example for when preemptive upper bounds would be nice, in general I think *not* having upper bounds is the better trade-off.

cc: @haskell-servant/maintainers

jkarni:
@soenkehahn why do you think so? As I see it, with stackage we get told pretty soon whether our upper bounds are ruling out some new version, and without them not only do things sometimes break, but also they remain broken for a while unless we (often @phadej ) painfully manually add upper bounds retrospectively to versions already on hackage.

soenkehahn:
- Having to bump preemptive upper bounds causes additional work.
- I don't really care that much about these broken versions on hackage as long as the newest versions of dependencies are supported. So in the case of the `http-client-0.5` breakage I think we should strive for fixing it fast and uploading a new version to hackage. So that after a `cabal update` it'll build with the newer `http-client`. (And for existing projects I'm using pinned down dependencies, so I won't get a newer version of `http-client` without asking for it.)
- I always felt that compilation errors that are caused by new versions of dependencies are quite easy to understand and work around. (Much easier than the error messages by cabal's constraint solver in case it can't find a valid install plan, for example.)

But I'm not dogmatic about this. If most `servant` maintainers want to add upper bounds I'm perfectly fine with that.

soenkehahn:
Just talked about this with @jkarni. We're planning to cherry-pick a0d5ed9aea10757caafa7d9654cb6f912d75a350 from #538 which adds upper bounds. Thanks @phadej!

soenkehahn:
Here's the PR: #546.

soenkehahn:
Done in #546.

PR #538, opened 2016-07-06
Add bounds in servant, servant-server and servant-client; support http-client-0.5

phadej:


soenkehahn:
I think it would have been nice to have separate PRs for
- fixing `servant` for the new version of `http-client` and
- adding upper bounds for all other deps.

To be able to discuss and merge these things independently.

servant-client/servant-client.cabal
```diff
@@ -35,24 +35,24 @@ library
...
-    , transformers-compat
+      base                  >= 4.7      && < 4.10
+    , aeson                 >= 0.7      && < 0.12
+    , attoparsec            >= 0.12     && < 0.14
+    , base64-bytestring     >= 1.0.0.1  && < 1.1
+    , bytestring            >= 0.10     && < 0.11
+    , exceptions            >= 0.8      && < 0.9
+    , http-api-data         >= 0.1      && < 0.3
+    , http-client           >= 0.4.18.1 && < 0.6
+    , http-client-tls       >= 0.2.2    && < 0.4
```
soenkehahn: This theoretically allows to have `http-client-0.5.*` and `http-client-tls-0.2.*` which is a combination that doesn't compile. Is that a problem in practice?

servant-client/src/Servant/Common/Req.hs
```diff
@@ -148,7 +148,11 @@ performRequest reqMethod req manager reqHost = do
   partialRequest <- liftIO $ reqToRequest req reqHost
 
   let request = partialRequest { Client.method = reqMethod
+#if MIN_VERSION_http_client(0,5,0)
+                               , checkResponse = \ _req _res -> return ()
+#else
                                , checkStatus = \ _status _headers _cookies -> Nothing
+#endif
```
soenkehahn: I think it would be nicer to read if the CPP were used to implement two different variants of a top-level declaration. Something like `disableStatusCheck :: Response -> Response`.

soenkehahn:
@phadej: Why did you reformat the stack files? Is that a side-effect of a tool you used?

The last commit (https://github.com/haskell-servant/servant/pull/538/commits/a1882f7a1264a1a7b714c467992afb995e4fea88) LGTM, apart from my comments.

Upper bounds are being discussed here: #534.

servant-client/servant-client.cabal
phadej: If `http-client-tls` of some version doesn't compile with some version of its dependencies, it's problem of `http-client-tls`. I.e. downstream shouldn't worry about that. If there is a failure, I.e. Incorrect bounds, package maintainers or Hackage Trustees will correct that.

servant-client/src/Servant/Common/Req.hs
phadej: Commented on your PR

phadej:
About reformating: I prefer having `resolver` at the first entry. And everything else alphabetically sorted. Seems that I have a bug in a tool I use for that (plan to make PR to `stack` itself)

soenkehahn:
Fixed through #545 and #546.

PR #542, opened 2016-07-08
Replace () with NoContent

### LINKED ISSUE #531: A lot of places still use () instead of NoContent
Changing `()` to `NoContent` in ComprhensiveAPI breakes pretty much everything =) 

We should be consistent and destroy the presence of `()` in my opinion. As docs are now unclear on which one to use.

arianvp:
Fixes #531 .

### DIFF
--- a/doc/tutorial/tutorial.cabal
+++ b/doc/tutorial/tutorial.cabal
@@ -54,6 +54,7 @@ test-suite spec
   default-language: Haskell2010
   hs-source-dirs: test
   main-is: Spec.hs
+  other-modules: JavascriptSpec
   build-depends: base == 4.*
                , tutorial
                , hspec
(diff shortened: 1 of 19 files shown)


arianvp:
Reasoning:

In servant-server we advice to use `NoContent` instead of `()` but all the other code was still using `()`. Basically. a 0.7 API could not use any other library but `servant-server`.  This PR should fix this.

@haskell-servant/maintainers could someone take a good look at this if I changed this correctly?

soenkehahn:
LGTM. It's of course hard to tell whether you managed to replace every occurrence of `()` as a return type. Did you do some grepping?

And many thanks for fixing this!

arianvp:
I've grepped and I think we're Okay.

arianvp:
Also. this should probably be added released with 0.8 as it's a breaking API change.

soenkehahn:
Again LGTM. Do you want to merge, @arianvp?

arianvp:
LGTM

arianvp:
@soenkehahn I squashed the commits

jkarni:
LGTM

PR #543, opened 2016-07-09
servant-client: update stack-ghcjs.yaml file for testsuite

soenkehahn:
Unfortunately I couldn't reproduce a working `ghcjs` setup with the existing stack file. With this one it works again. But I suspect that `ghcjs` projects with stack files are not really reproducible yet. :(


soenkehahn:
Build failures are due to broken `master`.

PR #545, opened 2016-07-09
support http-client-0.5

soenkehahn:
This is an alternative PR to #538. I'd like to make a release to hackage that works with `http-client-0.5` soon. Since #538 contains other, more controversial changes I've created this PR in the hope in can be merged and released faster.

@haskell-servant/maintainers

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -35,24 +35,24 @@ library
     Servant.Common.BasicAuth
     Servant.Common.Req
   build-depends:
-      base >=4.7 && <5
-    , aeson
-    , attoparsec
-    , base64-bytestring
-    , bytestring
-    , exceptions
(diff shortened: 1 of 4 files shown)


jkarni:
LGTM (maybe with @phadej as author of the commit).

phadej:
I'd leave a note that if we drop http-client-0.4 support, we should use `parseRequest`, thus disabling status check isn't necessary anymore. 

Or rather I'd directly shim `parseRequest`.

soenkehahn:
@phadej: Disabling the status check causes `http-client` to not throw exceptions in case of non-`2xx` status codes. This allows us to report e.g. redirects without throwing exceptions. The newer `parseRequest` is about how exceptions are thrown that are caused by parsing invalid URLs. Unless I'm misunderstanding this, these two issues are unrelated.

Issue #191, opened 2015-08-13
autogenerated javascript can have invalid identifier


berdario:
If you have a route like

```
"preferred-hours" :> Get '[JSON] Int
```

servant-js will generate a function with name `getPreferred-hours`.

alpmestan:
With the code currently in *master* (i.e in the upcoming 0.5 release), you can give a function that "modifies" names. Take a look [here](https://github.com/haskell-servant/servant/blob/master/servant-js/src/Servant/JS/Internal.hs#L32). So you can just specify a function that watches for "preferred-hours" and gives a custom, valid name, and just leaves all other names as they are. Is that a satisfactory solution for you?

freezeboy:
What if we split url portions when we generate the segments ? this would then generate:

```haskell
usingConcat: preferredhours
usingCamel: preferredHours
usingSnake: preferred_hours
```

alpmestan:
What should we split on, aside from `-`? Sounds a bit arbitrary but if you have a suggestion that works well enough in practice, I'm all ears.

berdario:
I think we should just split using something like this regexp when matching over (example in Python):

    >>> re.split('[^\w]', 'a_-a\U00010330b\u180ec')
    ['a_', 'a𐌰b', 'c']

I can write a patch for this, as soon as I'll have time (which means that someone else might want to do it earlier... )

jkarni:
[Here](https://mathiasbynens.be/notes/javascript-identifiers) is a description of valid identifiers in JS. The simplest thing that might work is to just strip invalid ones, and prefix an underscore if the result is a reserved keyword. This doesn't allow camel/snake casing, but the advantage of keeping it simple is that it'd also be predictable.

freezeboy:
@jkarni why do you say this doesn't allow camel/snake case ? instead of stripping invalid chars we just split into pieces before joining them using the camel/snake function ?

jkarni:
@freezeboy I was wrong. 

Anyone know what's the status on this? It feels like a lot of changes that are relevant have happened since.

freezeboy:
I haven't touched it since that day, but I think we can insert some "sanitizing" on each part of the function name before giving the list to the camel/snake/transforming function if you still think bad things can happen

berdario:
I'm having a look at the code, I found some weird changes

https://github.com/haskell-servant/servant/commit/f6ee02eb917f231899de6a744e01026babc3472b

[...]

PR #556, opened 2016-07-24
Remove javascript sanitization from servant-foreign (was: Fix javascript function name generation)

berdario:
There's a lot of duplication in JS/Vanilla, JS/Jquery, etc.etc. but fixing that would be out of scope to fix this issue

Also, you might want to tweak the logic in `toValidFunctionName`. In particular, I think it could make more sense to have it as `Text -> [Text]`, rather than `Text -> Text`, so every invalid character (`-`, `.`, weird unicodes, spaces, etc) would eventually a CamelCase or snake_case in the following character


berdario:
Ok, I need to fix the tests

servant-foreign/test/Servant/ForeignSpec.hs
```diff
@@ -19,8 +19,6 @@ camelCaseSpec = describe "camelCase" $ do
   it "converts FunctionNames to camelCase" $ do
     camelCase (FunctionName ["post", "counter", "inc"])
       `shouldBe` "postCounterInc"
-    camelCase (FunctionName ["get", "hyphen-ated", "counter"])
```
phadej: why this is removed?

phadej: And you should add a test for the case PR fixes as well

berdario:
@phadej 
It's removed since the code it was testing does not exist anymore: that code does not belong in servant-foreign, but rather in servant-js

I want to add a few tests, but currently there's no infrastructure in place for some integration tests (which are the only reasonable ones to test that our generated code is parsable), so it'll take a bit of time to get done

my idea is to add `node` in travis, check the presence of the node binary and skip the tests otherwise, and then `node -c` to syntax check the generated code without executing it

I think it would be better/more manageable to have a few smaller PRs, rather than a huge code drop that will encompass everything

phadej:
I cannot comment on `node` (I am not a maintainer, and IIRC someone was against complicating CI setup), but you could at least add an example which broke your code.

berdario:
Sure, the example I used is the same one that I wrote in the [linked issue](https://github.com/haskell-servant/servant/issues/191#issuecomment-131757316)

You can write a route in servant accessible at `"foo-bar\66352baz\6158quux"`: making requests at it will work just fine, but the generated js function name will be invalid.

There might be other characters that might lead to generation of invalid code, but this PR at least is a starting point to go back from special-casing issues in a few different places in the code...

I'd also like to see what someone else thinks about changing the type of that function to `Text -> [Text]`... that would be one of the next few steps I had in mind

jkarni:
Thanks!
I'm in favor of the changes, but like @phadej I'd like for the original issue to be put in a test, even if integration tests would eventually make more sense.

berdario:
Note for myself:
use `Language.ECMAScript3.Parser.parse` and `Language.ECMAScript3.Lexer.identifier`... pass the generated identifier through it, and the output should be the same as the input

phadej:
Migrated to https://github.com/haskell-servant/servant-js/pull/1

berdario:
I realized that this should still be merged (I fixed the merge conflicts, now) since the changes to servant-foreign cannot be migrated to servant-js

(they are just removal changes, since sanitization of hyphens is something language-specific that doesn't really belong in servant-foreign)

phadej:
@berdario True, sorry, i mixed up the pr's. Could you change the title of this one?

phadej:
Merged in #668

Issue #565, opened 2016-08-10
[discussion] Switch travis CI to stack?


soenkehahn:
I wonder whether we should switch our CI over to `stack`. We went with `tinc` so far, mainly because it helped us catch hackage breakage. But now we have lots of upper bounds (since #546), at least in `servant`, `servant-server` and `servant-client`. So for most of our dependencies our CI doesn't even detect whether there are new versions on hackage, that are excluded by the bounds. Additionally it'd be nice to have CI for the `stack` files.

Any opinions?

fizruk:
Personally, I like Stack for CI. For every new Stackage snapshot build we usually get a notification on GitHub from Stackage team if any bounds are too low or anything breaks. Stack is also what I believe most of us use for our own projects and it is how we build servant locally.

phadej:
Breakdown:
- `tinc` does test against *newer versions inside the bounds*
- `tinc` allows each package to pick  different version of dependencies
    - e.q. in `QuickCheck-2.9` PR, `servant` tests pick `2.8` as there was some transitive dependency restricting it, yet `servant-mock` pick `2.9.1` as its tests weren't restricted by `servant`'s tests.
    - with `stack` we'll be forced to pin dependencies across all packages, which may or may not be restrictful.
- leaving bounds for the sake that CI would detect when they are too loose is just *bad*
    - proper way is to have a job with `--allow-newer`, dunno if `tinc` can do it (`stack` can btw).
    - yet the upper-bounds job should be triggered periodically, not on push

OTOH:
- with `stack` one could use solver to pick an install plan within specified bounds, in *addition* to using pinned version builds in `stack.yaml`s.

I'm +/- 0 on this, as far as implications of stack are understood. IMHO it's great development tool, but not so great CI tool.

soenkehahn:
@phadej: Could you elaborate how to use `stack` with `--allow-newer`, please? I can't find any command that has that flag.

I would also prefer if we could have a periodic build for checking hackage breakage and running `packdeps`, independent of pushes. Is there a CI service that allows to do something like this? I don't really want to start to maintain our own CI server.

I've tried `stack init --force --resolver ghc-7.10.3` in an attempt to let `stack` invoke `cabal-install` and write an install-plan into a stack file. `cabal-install` didn't terminate (for like 5 minutes or so). @phadej: Is there another way to let `stack` come up with a non-snapshot install plan?

soenkehahn:
[...]

PR #568, opened 2016-08-13
switch ci to stack and add ci cron job to test hackage breakage

soenkehahn:
This uses the experimental cron job feature by travis: https://docs.travis-ci.com/user/cron-jobs/

It uses `stack` for the normal push-triggered CI.

The cron job script still uses `tinc`. We could switch this to `stack` once this is fixed: https://github.com/commercialhaskell/stack/issues/2475

### DIFF
--- a/scripts/ci-cron.sh
+++ b/scripts/ci-cron.sh
@@ -1,8 +1,15 @@
 #!/usr/bin/env bash
 
+set -o nounset
 set -o errexit
+set -o verbose
 
-for package in $(cat sources.txt) doc/tutorial ; do
+export PATH=$(stack path --bin-path):$PATH
+
+stack install cabal-install
+cabal update
+
+for package in $(cat sources.txt) ; do
   echo testing $package
   pushd $package
   tinc
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,33 +3,27 @@ sudo: false
 language: c
 
 env:
-  - GHCVER=7.8.4 CABALVER=1.22
-  - GHCVER=7.10.3 CABALVER=1.22
-  - GHCVER=8.0.1 CABALVER=1.24
+  - STACK_YAML=stack-ghc-7.8.4.yaml
+  - STACK_YAML=stack.yaml
+  - STACK_YAML=stack-ghc-8.0.1.yaml
 
 addons:
   apt:
-    sources:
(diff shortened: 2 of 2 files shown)


jkarni:
Relates to #565

scripts/ci-cron.sh
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env bash
 
+set -o nounset
 set -o errexit
 
-for package in $(cat sources.txt) doc/tutorial ; do
+export PATH=$(stack path --bin-path):$PATH
+
+stack install cabal-install
+cabal update
```
jkarni: Why install `cabal-install` via `stack`? Why even install `stack` in the first place? So that we can install ghc?

jkarni:
I'm okay with or without this change. My opinion has been that `stackage` itself has pretty good infrastructure for testing breakage of their package sets, and they are very good about notifying us about issues, so I don't tend to worry too much about stackage breakage. That said, more testing isn't a bad thing.

scripts/ci-cron.sh
soenkehahn: The `install` section of the `.travis.yml` also gets executed for cron jobs. (That's where `stack` is installed: https://github.com/haskell-servant/servant/pull/568/files#diff-354f30a63fb0907d4ad57269548329e3R18 .) So when executing `ci-cron.sh` on travis we already have `stack`. But no `ghc` and no `cabal-install`. So I thought it would be easiest to just use `stack`'s `ghc` and use it to install `cabal-install`.

Alternatively we could install `ghc` and `cabal-install` from hvr's repos, like before. That would have the disadvantage of slowing down the push-triggered builds unnecessarily.

Generally the `ci-cron.sh` script is not really how I want it to be. But I think it should be worked on in follow-up PRs. So that we can get the stack CI in place. (And I think it's already better than what we have now.)

soenkehahn:
What we didn't have (before this PR) is CI that checks that the state on `master` works with `stack`. This was also broken more than once on `master`. `stackage` doesn't help with that.

I still have to make sure that the cron job succeeds, then I'll merge.

jkarni:
@soenkehahn fair point. I'm in favor of the change, then. 

But CI is broken :) . Looks like an issue with `wai-app-static`?

PR #570, opened 2016-08-14
servant-client: add missing class constraint when building with http-…

soenkehahn:
…client < 0.4.30

@phadej: This is necessary, at least if built with `stack`.


phadej:
:+1: but i'd rather use `liftM` than introduce redundant constraint (with newer `bases`)

soenkehahn:
Good call. I'll merge with `liftM`.

Issue #527, opened 2016-06-17
Shouldn't  `MkLink (Verb m s ct a) = Link`  ?


vwwv:
Sorry if I'm missing something obvious, but shouldn't it ?

The type `Link` is opaque making it safe, but none of the base cases produce it (the 2 base cases, yield `URI` instead). Again sorry if I'm missing something obvious, but wouldn't this make the `Link` type unusable ?

Is there a reason it is not implemented like: 

``` haskell
instance HasLink (Verb m s ct a) where
    type MkLink (Verb m s ct a) = Link
    toLink = id
```

?

jkarni:
@vwwv yes, nice catch.

vwwv:
Ok, then, (unless somebody do it first) once I have some minutes for it, I'll write a PR for this change, I might to change some code elsewhere to pass the tests as this change the type (Haven't read the test code yet).

PR #572, opened 2016-08-14
Minor bug fix for HasLink

vwwv:
As discuss on the issue https://github.com/haskell-servant/servant/issues/527 .

### DIFF
--- a/servant/src/Servant/Utils/Links.hs
+++ b/servant/src/Servant/Utils/Links.hs
@@ -307,8 +307,8 @@ instance HasLink sub => HasLink (BasicAuth realm a :> sub) where
 
 -- Verb (terminal) instances
 instance HasLink (Verb m s ct a) where
-    type MkLink (Verb m s ct a) = URI
-    toLink _ = linkURI
+    type MkLink (Verb m s ct a) = Link
+    toLink _ = id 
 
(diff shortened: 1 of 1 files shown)


phadej:
As you can see this is a *breaking change*. Would force us to bump versions to `0.9` (and need to fix test suites as well).

vwwv:
Thanks for your quick feedback! yes, it is a breaking change (btw, my bad I did not see it had to modify the test suites).

Could it be an option to patch to this PR a modification to the test suite so the change could be added later on a version update? (I mean, this tiny change is probably not worth a version change, but maybe could be added in a future when there's a version change anyway).

soenkehahn:
> I mean, this tiny change is probably not worth a version change [...]

@vwwv: Don't worry about this being a breaking change. We'd much rather release more often with major version bumps than letting people wait for features to land on hackage.

It'd be nice if you:

- mention in one of the commit messages that this is a breaking change,
- add something to the `CHANGELOG` and
- of course, fix the travis build.

vwwv:
@soenkehahn: thanks for the advise, I'll add those suggestions to the PR in a couple days.

phadej:
Merged in #666. Thanks again for spotting this issue.

Issue #348, opened 2016-01-21
Move From/ToFormUrlEncoded


jkarni:
To `http-api-data`. The issue there is [here](https://github.com/fizruk/http-api-data/issues/23). 

This would be more consistent with other mime-related classes, and would allow people to more easily write instances for it, without depending on `servant`.

jkarni:
`http-api-data` side of this is done (thanks @fizruk and @cdepillabout !)

jkarni:
Done in #582

PR #580, opened 2016-08-31
Support http-api-data-0.3

phadej:
Btw: @fizruk Why you shuffled the module names, from `Web.HttpApiData.Internal ` to `Web.Internal.HttpApiData`?

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.3
+    , http-api-data         >= 0.1      && < 0.4
     , http-client           >= 0.4.18.1 && < 0.6
     , http-client-tls       >= 0.2.2    && < 0.4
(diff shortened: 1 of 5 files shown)


jkarni:
Wow that was quick! Thanks!

The new version contains the FormUrlEncoded classes that are also still here (see #348). We should remove them (conditionally on http-api-data version?). But I'm +1 on merging first.

phadej:
I'm :-1: on conditional API. If `FormUrlEncoded` are provided by `http-api-data-0.3` then we should bump lower bound on it.

I'd rather make a PRs for [other packages using `http-api-data`](http://packdeps.haskellers.com/reverse/http-api-data), there aren't that many. `persistent` is upgraded already it seems.

jkarni:
@phadej I made changes to replace our `FormUrlEncoded` versions with the new `http-api-data` ones on top of this PR in #582.

fizruk:
@phadej I shuffled internal module so that both internal modules are grouped under `Web.Internal`. No major reason. But you shouldn't import `Internal` modules, those functions are available at `Web.HttpApiData` already (and have been for a few past releases).

jkarni:
Can I merge?

phadej:
Superseded by https://github.com/haskell-servant/servant/pull/582

PR #582, opened 2016-09-02
Use From/ToForm from http-api-data

### LINKED ISSUE #348: Move From/ToFormUrlEncoded

jkarni:
And remove our version. The `http-api-data` version is a big improvement, and that is anyhow the place for it.

Closes #348.
Supersedes #580.

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.3
+    , http-api-data         >= 0.1      && < 0.4
     , http-client           >= 0.4.18.1 && < 0.6
     , http-client-tls       >= 0.2.2    && < 0.4
     , http-media            >= 0.6.2    && < 0.7
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -77,6 +77,7 @@ test-suite spec
     , bytestring
     , deepseq
     , hspec == 2.*
+    , http-api-data
     , http-client
     , http-media
     , http-types
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.4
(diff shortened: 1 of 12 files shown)


jkarni:
@fizruk, @cdepillabout and @phadej might be interested.

servant-client/servant-client.cabal
```diff
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.3
+    , http-api-data         >= 0.1      && < 0.4
```
fizruk: You should bump the lower bound to `>= 0.3`. Or use CPP around `import Web.FormUrlEncoded`.
I'd prefer the former.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -43,10 +43,17 @@ import           Network.Wai                (Application, Request, Response,
 import           Prelude                    ()
 import           Prelude.Compat
 import           Web.HttpApiData            (FromHttpApiData)
+#if MIN_VERSION_http_api_data(0,3,0)
+import           Web.Internal.HttpApiData   (parseHeaderMaybe,
```
fizruk: Don't import `Internal` module, those functions are available at `Web.HttpApiData`.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -348,11 +346,11 @@ eitherDecodeLenient input =
...
     mimeUnrender _ = eitherDecodeLenient
 
--- | @decodeFormUrlEncoded >=> fromFormUrlEncoded@
+-- | @urlDecodeAsForm@
 -- Note that the @mimeUnrender p (mimeRender p x) == Right x@ law only
 -- holds if every element of x is non-null (i.e., not @("", "")@)
-instance FromFormUrlEncoded a => MimeUnrender FormUrlEncoded a where
-    mimeUnrender _ = decodeFormUrlEncoded >=> fromFormUrlEncoded
+instance FromForm a => MimeUnrender FormUrlEncoded a where
+    mimeUnrender _ = left TextS.unpack . urlDecodeAsForm
```
fizruk: Hm, should we replace `Either String` with `Either Text` to avoid such unnecessary conversions?

jkarni: Yes. But in a separate PR, in my opinion.

soenkehahn:
Replaced by #592.

PR #583, opened 2016-09-02
Servant dates (again)

axman6:
See #281 for previous discussion. I believe all the issues in that conversation have been resolves, I've added some more docs and all tests pass. The only thing I'm not sure about is the module name - `Servant.API.Times` or `.Time`.


alpmestan:
I think I like `Servant.API.Time` better. Other than that, LGTM!

axman6:
Agreed, and changed.

jkarni:
Sorry, I still think this should be in a separate package! @haskell-servant/maintainers thoughts?

soenkehahn:
Nice idea with the format string at the type-level.

If I understand correctly this is a new datatype for times, with instances for `{To,From}HTTPApiData`. Is that correct?

If yes, I definitely agree with @jkarni: This should be in a separate package.

christian-marie:
I feel like this is pretty debatable, could all data representation/marshalling things like this go into an officially supported servant-datatypes (help with naming please) package?

So long the list of supported datatypes is kept sane (dates and times certainly are), the burden of maintaining it shouldn't be too high. This kind of thing *really* adds to the immediate usefulness of servant for rapid prototyping.

codedmart:
While I originally was thinking the same thing as @jkarni I think @christian-marie brings up a good point.

soenkehahn:
IIUC `Servant.API.Time` doesn't even depend on anything from `servant`. If you want to have this somewhere upstream, it should be in `http-api-data`. (And unlike `servant` that package already depends on `time` directly.)

cc: @fizruk

jkarni:
> This kind of thing really adds to the immediate usefulness of servant for rapid prototyping.

I agree. But to my mind the usefulness is added by the *existence* of these things, not their being in `servant` itself. And conceptually, this isn't really related to servant, even if it happens to be useful with it - witness the fact that @soenkehahn pointed out.

alpmestan:
It's not conceptually related to servant but the discoverability is increased by a very large factor if it lives in servant and is documented like the rest of servant.

jkarni:
>  discoverability (for servant users, that is) is increased by a very large factor if it lives in servant and is documented like the rest of servant.

As it would be if we created a page in the tutorial, "The Wider Servant Ecosystem", linking to it, (as well as to all the other useful packages out there), and specifying its purpose and relationship to servant in one paragraph.

alpmestan:
Well, unless there's an immediate plan for such a page, I'd be in favor of merging this in servant. =P

soenkehahn:
I remain somewhat strongly against merging this into `servant`. Being useful in conjunction with `servant` is not a good enough reason to get something merged. We have to get used to the idea, that there's a `servant` ecosystem and people need to pick and choose the packages they want to use.

Also, think about people who want to use this (and `http-api-data`), but don't want to use `servant`. (Wasn't this scenario the reason for splitting out `http-api-data`?)

I'd be more or less neutral towards putting this into `http-api-data` (vs. putting it in a separate package). That would also make it automatically available for `servant` users. Albeit slightly less discoverable, since the module won't be in `servant`. This is, of course, up to @fizruk.

3noch:
I agree that the primary issue is "discoverability." For such a common data element as dates and times users will expect servant to have an obvious and well-documented story. Whether or not the code is in one package or another seems like an implementation detail at that point. But in the absence of an "obvious and well-documented story" I think including it in the package is second best. Still, maybe this case is enough impetus to get such documentation underway as I expect this is not the only "common-ish" example.

christian-marie:
I'm most concerned about the inertia to overcome  keeping whole ecosystem cohesive when not under a common testing and release process.

A servant-contrib meta-package/examples repo for a nightly build may solve this issue? This at least gives us a warning of dying code.

Then we could at least poke authors and remove dead code from the verified working contrib code. This would double as a list for discoverability and working examples.

Otherwise, perhaps stack already covers stability sufficiently? I suspect not, though.

3noch:
@christian-marie That's a good point. But if something *really is* common, then I don't expect it will die. If it does die, then it wasn't common enough.

christian-marie:
Either way, this all seems to point towards a new repo for this code.

@axman6, input?

axman6:
I mostly agree with Christian's thoughts on this, and my preference would be to initially have this in Servant until a) there is some external user asking for the functionality and b) we figure out how to generalise this further - I believe this idea of having the formatting of certain types (times, integral numbers [hex, octal, decimal], who knows what else) represented in the type system might be a big win in terms of precisely defining APIs going forward; particularly useful for swagger docs etc.

I also think that times are so fundamental for so many APIs it's worth making sure they're available in the key servant infrastructure such as the servant and servant-server packages (what's special about ints, text, doubles, JSON that isn't special about time?). Perhaps this is a good argument for having it somewhere like http-api-data. 

I don't personally have the time or resources to maintain such a library on its own, particularly for such a small package. If someone wants to take the code and make a new package (whether it be elsewhere or part of the servant repo) I'm ok with that.

So summing up, I'd prefer to see this begin life in Servant, and then get split off once we better understand where we can take it and there's demand for it elsewhere.

phadej:
Could I propose this functionality to live in a separate package, though in `haskell-servant/servant` repo? We are good at splitting them out, and for sake of experimentation and development I see it beneficial to be "close to core".

Also there is http://hackage.haskell.org/package/time-locale-compat, so we don't need to introduce flags (to be removed later).

axman6:
@phadej yeah that would be my preferred option for splitting this into its own package, mainly so it doesn't get orphaned by my lack of time to maintain it.

fizruk:
Unless I miss something, `FTime` seems like a generally useful datatype, not just for `servant`.

Maybe we should open a PR for [time](https://github.com/haskell/time) to add `FTime`? Then we could add appropriate instances to `aeson`, `http-api-data`, `swagger2`, etc. And then we could just use it as normal data type in `servant` and just document an example for this (for discoverability).

Creating a separate package for `FTime` to me seems like an overkill, and I would rather add it directly to `http-api-data` if it's impossible to add it to `time` (although again, I don't think `FTime` is useful only for web).

dredozubov:
I strongly agree with @fizruk on this matter, it should be addressed beyond the scope of servant. The main reason behind it: it's applicable extremely wide in a lot of serialization settings, such as reads/writes to databases. At work we use dates parametrized by a type-level strings since forever. Sadly, we never got to making it into a package.

soenkehahn:
Alright, since this is not going to be in `servant`, I'll close this. (Feel free to continue the discussion, of course.)

bitemyapp:
@axman6 did this code ever get upstreamed or land anywhere else?

axman6:
@bitemyapp no idea if this went anywhere, I didn't have time or a need for this any more, so if no one else took the work and added it to something else, it probably didn't happen.

PR #585, opened 2016-09-02
Replace use of ToByteString with HttpApiData for GetHeaders

axman6:
Fixes servant/#581

* Version bump because this changes the API for GetHeaders

### DIFF
--- a/servant-mock/test/Servant/MockSpec.hs
+++ b/servant-mock/test/Servant/MockSpec.hs
@@ -40,8 +40,8 @@ data TestHeader
 
 instance ToHttpApiData TestHeader where
   toHeader = toHeader . show
-  toUrlPiece _ = error "ToHttpApiData.toUrlPiece not implemented for TestHeader"
-  toQueryParam _ = error "ToHttpApiData.toQueryParam not implemented for TestHeader"
+  toUrlPiece = toUrlPiece . show
+  toQueryParam = toQueryParam . show
 
 
 instance Arbitrary TestHeader where
--- a/servant-mock/test/Servant/MockSpec.hs
+++ b/servant-mock/test/Servant/MockSpec.hs
@@ -8,9 +8,7 @@
 module Servant.MockSpec where
 
 import           Data.Aeson as Aeson
-import           Data.ByteString.Conversion.To
 import           Data.Proxy
-import           Data.String
 import           GHC.Generics
 import           Network.Wai
 import           Servant.API
@@ -40,8 +38,11 @@ data TestHeader
   | ArbitraryHeader
   deriving (Show)
 
-instance ToByteString TestHeader where
-  builder = fromString . show
+instance ToHttpApiData TestHeader where
+  toHeader = toHeader . show
+  toUrlPiece _ = error "ToHttpApiData.toUrlPiece not implemented for TestHeader"
+  toQueryParam _ = error "ToHttpApiData.toQueryParam not implemented for TestHeader"
+
 
 instance Arbitrary TestHeader where
   arbitrary = return ArbitraryHeader
--- a/servant/CHANGELOG.md
+++ b/servant/CHANGELOG.md
@@ -2,7 +2,7 @@ next
 ----
 
 * Add `CaptureAll` combinator. Captures all of the remaining segments in a URL.
-* replace use of `ToByteString` with `HttpApiData` for `GetHeaders`
+* BACKWARDS INCOMPATIBLE replace use of `ToFromByteString` with `To/FromHttpApiData` for `GetHeaders/BuildHeadersTo`
 
 0.8
 ---
--- a/servant/CHANGELOG.md
+++ b/servant/CHANGELOG.md
@@ -2,6 +2,7 @@ next
 ----
 
 * Add `CaptureAll` combinator. Captures all of the remaining segments in a URL.
+* replace use of `ToByteString` with `HttpApiData` for `GetHeaders`
 
 0.8
 ---
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -1,5 +1,5 @@
 name:                servant
-version:             0.8.1
+version:             0.8
 synopsis:            A family of combinators for defining webservices APIs
 description:
   A family of combinators for defining webservices APIs and serving them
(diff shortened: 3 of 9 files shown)


servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -31,8 +31,9 @@ module Servant.API.ResponseHeaders
     ) where
 
 import           Data.ByteString.Char8       as BS (pack, unlines, init)
-import           Data.ByteString.Conversion  (ToByteString, toByteString',
+import           Data.ByteString.Conversion  (--ToByteString, toByteString',
```
fizruk: Just remove old code, there's no need to keep it commented

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -31,8 +31,9 @@ module Servant.API.ResponseHeaders
...
                                               FromByteString, fromByteString)
+import Web.HttpApiData (ToHttpApiData,toHeader)
```
fizruk: Please, keep the indentation homogenous, e.g.:

```haskell
import           Web.HttpApiData             (ToHttpApiData(..))

```

jkarni: We also need to remove `FromByteString` (and replace it with `FromHttpApiData`)

servant-mock/test/Servant/MockSpec.hs
```diff
@@ -40,8 +38,11 @@ data TestHeader
   | ArbitraryHeader
   deriving (Show)
 
-instance ToByteString TestHeader where
-  builder = fromString . show
+instance ToHttpApiData TestHeader where
+  toHeader = toHeader . show
+  toUrlPiece _ = error "ToHttpApiData.toUrlPiece not implemented for TestHeader"
```
jkarni: Any reason to not just implement one of these?

jkarni:
`servant-docs`, still uses `ToByteString`, which should also be changed. And I think `MockSpec` and `ServerSpec` can get rid of their `Bytestring.Conversion` imports. I *think* in the end we should be able to remove `bytestring-conversion` from all cabal files.

jkarni:
(And thanks!)

servant-mock/test/Servant/MockSpec.hs
axman6: I assumed the intent of TestHeader was only to test headers - happy to add the trivial instances if you'd prefer, but it didn't make much sense to me.

axman6:
I'll have a look at the other packages and see how easy it'll be to remove `bytestring-conversion`.

servant/servant.cabal
```diff
@@ -1,5 +1,5 @@
 name:                servant
-version:             0.8
+version:             0.8.1
```
jkarni: It'll actually have to be a major version bump, but don't do that in this PR. Leave it '0.8' and we'll bump the versions when releasing.

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -31,8 +31,8 @@ module Servant.API.ResponseHeaders
     ) where
 
 import           Data.ByteString.Char8       as BS (pack, unlines, init)
-import           Data.ByteString.Conversion  (ToByteString, toByteString',
-                                              FromByteString, fromByteString)
+import           Web.HttpApiData             (ToHttpApiData,toHeader
```
jkarni: Spaces after commas

servant/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ next
 ----
 
 * Add `CaptureAll` combinator. Captures all of the remaining segments in a URL.
+* replace use of `ToByteString` with `HttpApiData` for `GetHeaders`
```
jkarni: Preface with "BACKWARDS INCOMPATIBLE" (new thing, so isn't in any of the other ones yet.
And now it is "Replace `To/FromByteString` with `To/FromHttpApiData` for `GetHeaders/BuildHeadersTo`"

servant/servant.cabal
axman6: Yeah I realised that after pushing this, I'll put it back.

servant-mock/test/Servant/MockSpec.hs
jkarni: I prefer implementing because (a) it's shorter and less attention grabbing, and (b) in my opinion tests should sort of set an example for how to do things.

jkarni: Oh, and (c) the fewer 'errors' in the code the better.

jkarni:
Other than maybe implementing the other methods, LGTM.

servant-mock/test/Servant/MockSpec.hs
axman6: Fair point, I'll do that now.

jkarni:
@haskell-servant/maintainers ping (need another +1)

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -31,8 +31,8 @@ module Servant.API.ResponseHeaders
     ) where
 
 import           Data.ByteString.Char8       as BS (pack, unlines, init)
-import           Data.ByteString.Conversion  (ToByteString, toByteString',
-                                              FromByteString, fromByteString)
+import           Web.HttpApiData             (ToHttpApiData, toHeader,
+                                             FromHttpApiData, parseHeader)
 import qualified Data.CaseInsensitive        as CI
```
axman6: There was a comment earlier about just using `(ToHttpApiData(..)...` here, would that be preferable? Or `(ToHttpApiData(toHeader)...` to make it clear it's coming from that class?

alpmestan: I personally wouldn't mind `ToHttpApiData(..)`, this module is small enough that we do not have to spend a lot of time chasing where some function comes from.

alpmestan:
LGTM :+1:

jkarni:
Thanks!

axman6:
Thanks team! Glad to help

PR #590, opened 2016-09-03
Make the tutorial compile

diekmann:
As just discussed in person on MuniHac

### DIFF
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -28,7 +28,6 @@ import Prelude.Compat
 
 import Control.Monad.Except
 import Control.Monad.Reader
-import Data.Aeson.Compat
 import Data.Aeson.Types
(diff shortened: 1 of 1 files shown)


doc/tutorial/Server.lhs
```diff
@@ -37,7 +36,6 @@ import Data.Maybe
 import Data.String.Conversions
 import Data.Time.Calendar
 import GHC.Generics
-import Lucid
```
arianvp: We actually do need the Lucid import I think.  As we're using all sorts of lucid combinators later on.

The https://github.com/haskell-servant/servant/tree/master/doc/tutorial  directory is just a cabal project for all the literate haskell files.    So that way you can check if the code works.

diekmann:
Thanks, I'll cancel my pull request and make a new one later :+1:

PR #591, opened 2016-09-03
Make tutorial/example code compile.

diekmann:
Document where to find the dependencies to make the tutorial compile.

### DIFF
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -58,7 +58,7 @@ fully-fledged webserver. This means that in your applications, you can just add
 else.
 
 The dependencies to compile this example can be found in the 
-(cabal file)[https://github.com/haskell-servant/servant/blob/master/doc/tutorial/tutorial.cabal] 
+(cabal file)[https://github.com/haskell-servant/servant/blob/v0.8/doc/tutorial/tutorial.cabal] 
 of the tutorial.
 
 We will write a server that will serve the following API.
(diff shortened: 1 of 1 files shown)


jkarni:
LGTM

jkarni:
Well - slight problem. Now *every* version of the tutorial will point to `master` of the cabal file. Ideally, version 0.8 (from git tag 0.8) would point to the cabal file for that tag, and so on.

diekmann:
Right, ...
I just couldn't find where the github links are defined.
By the way, the "Edit on GitHub" link (top right of the tutorial page) also points to a 404 github page.

jkarni:
@diekmann thanks for the heads up about edit on github. Looks like it's because of [this](https://github.com/rtfd/readthedocs.org/issues/2130).

jkarni:
There's an [issue](https://github.com/rtfd/readthedocs.org/issues/2195) with old branches, so the 'stable' doc points to something that no longer exists (ugh), but for everything else `Edit on Github` should be working.

diekmann:
Updated link to cabal file. Cannot fix "Edit on GitHub"-issue.

soenkehahn:
I think pointing to the `v0.8` branch is a bad idea, since we would have to update that manually every time we cut a new release. (It's already outdated, since we have `v0.8.1` now.) On the other hand I think pointing to the version of `master` is not *that* bad.

jkarni:
RTD allows us to link to the correct version: http://read-the-docs.readthedocs.io/en/latest/vcs.html

diekmann:
Maybe someone with RTD access should finish this pull request ;-)

alpmestan:
The tutorial & cookbook are now CI checked, can we close this?

diekmann:
Sure, thanks!

alpmestan:
Thank you, and sorry for dropping the ball on this one, but at least it happened :-)

PR #592, opened 2016-09-03
Use http api forms

jkarni:
Replaces #582 . I'm doing this because having branches with numbers as names confuses read-the-docs, and I want to delete that one.

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.3
+    , http-api-data         >= 0.1      && < 0.4
     , http-client           >= 0.4.18.1 && < 0.6
     , http-client-tls       >= 0.2.2    && < 0.4
     , http-media            >= 0.6.2    && < 0.7
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.4
+    , http-api-data         >= 0.3      && < 0.4
     , http-client           >= 0.4.18.1 && < 0.6
     , http-client-tls       >= 0.2.2    && < 0.4
     , http-media            >= 0.6.2    && < 0.7
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -78,6 +78,7 @@ test-suite spec
     , bytestring
     , deepseq
     , hspec == 2.*
+    , http-api-data
     , http-client
     , http-media
     , http-types
--- a/servant-client/test/Servant/ClientSpec.hs
+++ b/servant-client/test/Servant/ClientSpec.hs
@@ -30,7 +30,7 @@ module Servant.ClientSpec where
 import           Control.Applicative        ((<$>))
 #endif
 import           Control.Arrow              (left)
-import           Control.Concurrent         (ThreadId, forkIO, killThread)
+import           Control.Concurrent         (forkIO, killThread, ThreadId)
 import           Control.Exception          (bracket)
 import           Control.Monad.Trans.Except (throwE )
 import           Data.Aeson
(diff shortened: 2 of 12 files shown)


servant-client/servant-client.cabal
```diff
@@ -41,7 +41,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , http-api-data         >= 0.1      && < 0.3
+    , http-api-data         >= 0.3      && < 0.4
```
soenkehahn: Yes. I think this is better than supporting older versions with CPP.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -22,43 +22,40 @@ module Servant.Server.Internal
...
                                               QueryParam, QueryParams, Raw,
-                                              RemoteHost, ReqBody, Vault,
+                                              ReflectMethod (reflectMethod),
+                                              RemoteHost, ReqBody, Vault, Verb,
                                               WithNamedContext)
 import           Servant.API.ContentTypes    (AcceptHeader (..),
                                               AllCTRender (..),
-                                              AllCTUnrender (..),
-                                              AllMime,
+                                              AllCTUnrender (..), AllMime,
```
soenkehahn: Are all these formatting changes intended? I usually try hard to avoid these. (Was there anything else changed, apart from formatting?)

jkarni: Removed the `Web.HttpApiData.Internal` import (in favor of `Web.HttpApiData`). Yeah, this is the result of `stylish-haskell`. Not quite sure how to deal with them... Should I undo them?

soenkehahn: Well, personally I try very hard not to make formatting changes in normal PRs. Just to keep the noise level down and to avoid (possibly tool-supported) edit wars. Otherwise I would also change formatting all the time. So from my perspective: yes, please revert.

jkarni:
@soenkehahn fixed the imports.

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -30,7 +30,7 @@ module Servant.ClientSpec where
 import           Control.Applicative        ((<$>))
 #endif
 import           Control.Arrow              (left)
-import           Control.Concurrent         (forkIO, killThread, ThreadId)
+import           Control.Concurrent         (ThreadId, forkIO, killThread)
```
soenkehahn: There's still some formatting changes to imports like this one that I would remove. In some other places, too.

jkarni: What's the underlying principle, though? That it be a separate PR? Or a separate commit?

soenkehahn: I wouldn't make these changes at all. If you're not touching code, don't change formatting.

jkarni: But it's a pain to be consistent about formatting without using automated tools, and automated tools will touch parts that are inconsistent. My opinion is that we should aim to follow some stylish-haskell configuration, and *then* running the tool wouldn't change anything besides what you touched.  But if we still haven't reached that stage, we'll have issues like this.

soenkehahn: My guess would be that it'd be hard to come to a consensus, which style and tool to use for code-formatting. For example, I personally use https://github.com/zalora/normalize-imports. And I also don't think it's worth the effort, at least not at this point.

I *do* think that we shouldn't have formatting changes scattered around in unrelated PRs and commits like this. In my opinion, if you touch a piece of code, please, go nuts and reformat at will. If you don't, then don't.

What I usually do is to run my code-formatting tools during development and they might change formatting where I don't want them to. Then I commit changes selectively with `git add -p`, while not staging the formatting changes in parts of the code that I didn't really touch. Afterwards I run `git checkout -p`, to revert the unwanted changes the tools introduced.

I think another good workflow that I heard about is this: You configure your editor to allow you to format only regions of the code. Like the selected code block, or the enclosing expression.

jkarni:
@soenkehahn I fixed the imports - properly this time, I think.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -290,12 +287,12 @@ instance OVERLAPPABLE_
          ToJSON a => MimeRender JSON a where
     mimeRender _ = encode
 
--- | @encodeFormUrlEncoded . toFormUrlEncoded@
+-- | @urlEncodeAsForm"
```
soenkehahn: What's with the weird quote?

soenkehahn:
@jkarni: Apart from the one comment, LGTM. (Sorry about being so pedantic about the import formatting.)

jkarni:
> (Sorry about being so pedantic about the import formatting.) 

I was just thinking how great it is that we have you to be pedantic! This would otherwise turn into a zoo ; )

(Still not sure I agree with the general policy about import formatting, though!)

Issue #563, opened 2016-08-09
Servant.Client.client variants with baked in Manager and/or BaseUrl


runeksvendsen:
I have an API that exports 7 endpoints/resources. So, `Servant.Client.client` returns 7 functions that I can use to query the server, all of which require both a Manager and BaseUrl as the last two parameters.

It would be extremely helpful for me to be able to "bake in" either just a BaseUrl or both a Manager and a BaseUrl into the returned functions, such that I would call `client` like so:

```
api = Proxy :: Proxy HugeApi
manager <- newManager defaultManagerSettings
let baseUrl = BaseUrl Http "localhost" 8000 ""
firstFunc :<|> secondFund :<|> [..] = client api $ manager baseUrl
```

And then have all the client/query functions (`firstFunc`, `secondFunc`, ...) only take the API-specified argument(s) (if any). So that we get universally reusable functions that can be evaluated from anywhere by themselves (without reading the right BaseUrl from a config and applying this plus a Manager to all the functions returned by `client`).

soenkehahn:
Thanks for the report.

Note that we don't want to change how `client` works, because lots of the time we want to be able to bind the client functions at the top-level. Then we need to be able to pass in the `Manager` and often-times also the `BaseUrl` to the individual client functions.

But this is definitely something that would be nice to have *in addition* to how `client` currently works. @amarpotghan and I spent some time trying to get this working, but in the end didn't come up with a very nice solution. This is the current state of our efforts: https://github.com/haskell-servant/servant/pull/484.

I actually like the sketched solution in that discussion: Let `ClientM` be some kind of `Reader` over `BaseUrl` and `Manager` and then use `Enter` to implement `staticClient`. (And probably rename `staticClient` to something better.)

chkl:
Apparently it was discussed once to change ClientM to be Reader, but it got discarded. See this thread:
https://github.com/haskell-servant/servant/pull/432#issuecomment-202701482

jkarni:
Anyone want to take this up?

chkl:
So basically we would change ClientM from

```haskell
 ClientM = ExceptT ServantError IO                                                              
```
to
```haskell
ClientM = ReaderT Manager (ReaderT BaseUrl (ExceptT ServantError IO))                      
```
or
```haskell
ClientM = ReaderT (Manager, BaseUrl) (ExceptT ServantError IO)
```

I would be interested in doing that, although I might need some help as I'm a bloody beginner.

[...]

PR #594, opened 2016-09-04
make ClientM  a Reader

chkl:
As discussed in https://github.com/haskell-servant/servant/issues/563

This would change the definition of clientM from
```haskell
type ClientM = ExceptT ServantError IO
```

to
```haskell
newtype ClientM a = ClientM { runClientM' :: ReaderT ClientEnv (ExceptT ServantError IO) a }
```

where ClientEnv is just a record of `Manager` and `BaseUrl`

I added `mtl` as a dependency,  but I am not sure about the version boundaries.

### DIFF
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -15,11 +15,10 @@ need to have some language extensions and imports:
 
 module Client where
 
-import Control.Monad.Trans.Except (ExceptT, runExceptT)
 import Data.Aeson
 import Data.Proxy
 import GHC.Generics
-import Network.HTTP.Client (Manager, newManager, defaultManagerSettings)
+import Network.HTTP.Client (newManager, defaultManagerSettings)
 import Servant.API
 import Servant.Client
 ```
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -71,19 +71,13 @@ What we are going to get with **servant-client** here is 3 functions, one to que
 ``` haskell
 position :: Int -- ^ value for "x"
          -> Int -- ^ value for "y"
-         -> Manager -- ^ the HTTP client to use
-         -> BaseUrl -- ^ the URL at which the API can be found
-         -> ExceptT ServantError IO Position
+         -> ClientM Position
 
 hello :: Maybe String -- ^ an optional value for "name"
-      -> Manager -- ^ the HTTP client to use
-      -> BaseUrl -- ^ the URL at which the API can be found
-      -> ExceptT ServantError IO HelloMessage
+      -> ClientM HelloMessage
 
 marketing :: ClientInfo -- ^ value for the request body
-          -> Manager -- ^ the HTTP client to use
-          -> BaseUrl -- ^ the URL at which the API can be found
-          -> ExceptT ServantError IO Email
+          -> ClientM Email
 ```
 
 Each function makes available as an argument any value that the response may
@@ -120,17 +114,17 @@ data BaseUrl = BaseUrl
 That's it. Let's now write some code that uses our client functions.
 
 ``` haskell
-queries :: Manager -> BaseUrl -> ExceptT ServantError IO (Position, HelloMessage, Email)
-queries manager baseurl = do
-  pos <- position 10 10 manager baseurl
-  message <- hello (Just "servant") manager baseurl
-  em  <- marketing (ClientInfo "Alp" "alp@foo.com" 26 ["haskell", "mathematics"]) manager baseurl
+queries :: ClientM (Position, HelloMessage, Email)
+queries = do
+  pos <- position 10 10 
+  message <- hello (Just "servant") 
+  em  <- marketing (ClientInfo "Alp" "alp@foo.com" 26 ["haskell", "mathematics"])
   return (pos, message, em)
 
 run :: IO ()
 run = do
   manager <- newManager defaultManagerSettings
-  res <- runExceptT (queries manager (BaseUrl Http "localhost" 8081 ""))
+  res <- runClientM queries (ClientEnv manager (BaseUrl Http "localhost" 8081 ""))
   case res of
     Left err -> putStrLn $ "Error: " ++ show err
     Right (pos, message, em) -> do
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -53,7 +53,6 @@ library
     , text                  >= 1.2      && < 1.3
     , transformers          >= 0.3      && < 0.6
     , transformers-compat   >= 0.4      && < 0.6
-    , http-conduit
     , mtl
   hs-source-dirs: src
   default-language: Haskell2010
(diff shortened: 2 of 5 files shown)


jkarni:
CI is failing for ghc 7.8.4 (run `stack test --stack-yaml=stack-ghc-7.8.4.yaml` to test it locally). I think because of the version of `mtl` picked.

doc/tutorial/Client.lhs
```diff
@@ -120,17 +113,17 @@ data BaseUrl = BaseUrl
 That's it. Let's now write some code that uses our client functions.
 
 ``` haskell
```
jkarni: This is great. All the signatures have become a lot cleaner!

servant-client/src/Servant/Common/Req.hs
```diff
@@ -149,20 +152,39 @@ parseRequest url = liftM disableStatusCheck (parseUrl url)
 displayHttpRequest :: Method -> String
 displayHttpRequest httpmethod = "HTTP " ++ cs httpmethod ++ " request"
 
-type ClientM = ExceptT ServantError IO
+-- previously: 
+--type ClientM = ExceptT ServantError IO
```
jkarni: You can remove this. The place for such notes is in the changelog.

servant-client/src/Servant/Common/Req.hs
```diff
@@ -149,20 +152,39 @@ parseRequest url = liftM disableStatusCheck (parseUrl url)
...
+
+data ClientEnv
+  = ClientEnv
+  { manager :: Manager
+  , baseUrl :: BaseUrl
+  }
+
+newtype ClientM a = ClientM { runClientM' :: ReaderT ClientEnv (ExceptT ServantError IO) a }
```
jkarni: A haddock here and above is important. Something like:

```
-- | @ClientM@ is the monad in which client functions run. Contains the 'Manager' and 'BaseUrl' used for requests in the reader environment.
```

servant-client/src/Servant/Common/Req.hs
```diff
@@ -149,20 +152,39 @@ parseRequest url = liftM disableStatusCheck (parseUrl url)
...
+                    deriving ( Functor, Applicative, Monad, MonadIO
+                             , MonadReader ClientEnv
```
jkarni: `deriving Generic` too

servant-client/src/Servant/Common/Req.hs
```diff
@@ -149,20 +152,39 @@ parseRequest url = liftM disableStatusCheck (parseUrl url)
...
 
-type ClientM = ExceptT ServantError IO
+-- previously: 
+--type ClientM = ExceptT ServantError IO
+
+data ClientEnv
+  = ClientEnv
+  { manager :: Manager
+  , baseUrl :: BaseUrl
+  }
```
jkarni: Instance [`HasHttpManager`](https://hackage.haskell.org/package/http-client-0.5.3.1/docs/Network-HTTP-Client.html#t:HasHttpManager).

jkarni: It may be interesting to provide convenience functions:

```
defaultLocalHostEnv :: Port -> IO ClientEnv
defaultLocalHostEnv port = do
   mgr <- newManager defautlManagerSettings
   return $ ClientEnv mgr (BaseUrl Http "localhost" port "")

defaultHttpsClient :: String -> IO ClientEnv
defaultHttpsClient host = ...
```

jkarni:
This is a great change even aside from the fact that it'll make using `enter` possible! Thanks!

arianvp:
Okay so the following problem occurs:

We need `mtl` for `MonadReader`.  But the version of `mtl` in  `lts-2.22` which we use to test `GHC-7.8.4`   does not yet have `Control.Monad.Except`, whilst the `transformers` package in `lts-2.22` already has `Control.Monad.Trans.Except`.  Sadly enough `lts-2.22` was the latest LTS supporting `7.8.4` so we need to consider building this in a different way I guess?

I'm sure there is a valid build plan for 7.8.4.


An option would be:

Manually add a version of `mtl` that does have `Control.Monad.Except` to the  `stack-7.8.4.yaml` file.

jkarni:
Why do we need `Control.Monad.Except`? Won't `Control.Monad.Trans.Except` do?

jkarni:
Also, see [mtl-compat](https://hackage.haskell.org/package/mtl-compat-0.2.1.2/docs/Control-Monad-Except.html)

chkl:
First of all thank you very much for your feedback; this is very valuable to me.
I'm having a little trouble to understand the relationship between `mtl` and `transfomers`. First of all just adding `mtl-compat` to the cabal-file makes it compile. But I guess having less dependencies is always a good thing so tried to work without `mtl`.

The reason that used `mtl` was that I wanted to give `ClientM` a `MonadError` instance and I think that that class is only defined in `mtl`.  @jkarni : You asked for an instance of MonadError, do we need really need one? Otherwise I would remove that class and revert the other change where I replaced `throwE` with `throwError`. This way we wouldn't need `mtl` nor `mtl-compat`.

jkarni:
> I'm having a little trouble to understand the relationship between mtl and transfomers. 

`mtl` provides a bunch of typeclasses, usually describing effects. `transformers` provides monad transformers, and these usually are the primary instances for the `mtl` typeclass. So while `mtl` might have a `MonadError`, it's `transformers` that has the concrete `ExceptT`. (But note that `mtl` re-exports things from `tranformers`!). 

Now, the problem isn't that `MonadError` didn't exist previously - only that it existed in `Control.Monad.Error.Class` (from `mtl`), which has been deprecated. So instead of using `mtl-compat`, you could do something like:

```
{-# LANGUAGE CPP #-}
...
#if MIN_VERSION_mtl(2,2,0)
import Control.Monad.Except (MonadError(..))
#else
import Control.Monad.Error.Class (MonadError(..))
#endif
import Control.Monad.Trans.Except

```

I *think* that should work.vBut if it doesn't, I'm also okay with having `mtl-compat`. And I'm definitely okay having `mtl` in the dependencies. I'm sure it transitively already is.

alpmestan:
@jkarni isn't what you describe the purpose of `mtl-compat` though? =)

I agree that depending explicitly on `mtl` isn't a problem.

servant-client/servant-client.cabal
```diff
@@ -53,6 +53,7 @@ library
     , text                  >= 1.2      && < 1.3
     , transformers          >= 0.3      && < 0.6
     , transformers-compat   >= 0.4      && < 0.6
+    , http-conduit
```
jkarni: Definitely not worth this dependency for the `HasManager` instance. Sorry! I didn't know it didn't exist previously!

chkl: Ok, I removed it.

jkarni:
@chkl great! Last two things - do you mind adding an entry about this change into the servant-client changelog (labeled `BACKWARDS INCOMPATIBLE`), and squashing the commits into one or two?

chkl:
I added an entry to the changelog. I squashed all commits on a new branch https://github.com/chkl/servant/tree/clientm. I didn't see a way to edit my pull-request. Do I open a new one?

jkarni:
@chkl either that, or force-pushing into this branch.

chkl:
I created a new pull request: https://github.com/haskell-servant/servant/pull/599

jkarni:
Superseded by #599

PR #598, opened 2016-09-07
Add instances for IsSecure

jkarni:


jkarni:
I don't think this needs review.

PR #602, opened 2016-09-17
Release

soenkehahn:
cc: @haskell-servant/maintainers

### DIFF
--- a/servant-client/CHANGELOG.md
+++ b/servant-client/CHANGELOG.md
@@ -1,10 +1,9 @@
-0.8.1
------
+0.9
+---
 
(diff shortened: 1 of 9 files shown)


servant-client/CHANGELOG.md
```diff
@@ -1,11 +1,10 @@
-0.8.1
------
+0.9
```
soenkehahn: @jkarni, @chkl: This was `0.8.1` before, but it should be `0.9` now, right?

jkarni: Right

PR #609, opened 2016-10-02
client-ghcjs: Rebase on master + Fix MVar blocking exception

FPtje:
As discussed in https://github.com/haskell-servant/servant/issues/51#issuecomment-250737727.

This PR contains a rebase onto master. It also contains the fix for the following exception, which is thrown after every request:

```
thread blocked indefinitely in an MVar operation
```


arianvp:
Hmm something fishy is going on here. there are still conflicts, and a lot of commits are already in master.   My git foo is not good enough to see directly what went wrong here...

arianvp:
Ahh I see... your branch is rebased (hence the history is rewritten) on master but of course this one isnt... so it will try an explicit merge.   Could you instead open up this pull request against the master branch instead of the ghcjs-client branch? that would fix the issue

arianvp:
Hmm If I change the base of the merge request to master. only one commit shows up... odd.

FPtje:
This pull request rebases client-ghcjs upon master, as per [your request](https://github.com/haskell-servant/servant/issues/51#issuecomment-250794503). Since the client-ghjs branch currently isn't rebased, Github is confused. To merge this branch you would first have to locally force pull it to then force push it to Github. After all, a rebase changes history.

If this is not desirable I still have [the branch with the merge commit](https://github.com/LumiGuide/servant/commits/client-ghcjs). This branch will merge without any force being required.

soenkehahn:
1) I think we should always merge released versions of `master` into `client-ghcjs`. Say `v0.9`. (Current `master` is `v0.9` so I think that works out. :))
2) We should avoid force pushing to `client-ghcjs`. If there's any people out there using that branch, they probably have it fixed down by git commit, and we should make sure, the commits don't disappear from the branch.

@FPtje: Nice to see someone putting some work into this!

FPtje:
Closing in favour of PR with merge commit:
https://github.com/haskell-servant/servant/pull/610

PR #610, opened 2016-10-03
client-ghcjs: Merged master + Fix MVar blocking exception

FPtje:
As discussed in https://github.com/haskell-servant/servant/issues/51#issuecomment-250737727 and https://github.com/haskell-servant/servant/pull/609. 

This pull request contains a **merge** commit rather than a rebase. It also contains the fix for the following exception, which is thrown after every request:

```
thread blocked indefinitely in an MVar operation
```


FPtje:
I pushed some commits removing redundant imports. Looks like your CI problem is [solved](https://travis-ci.org/haskell-servant/servant/builds/164583944) too, @jkarni.

alpmestan:
Terrific, thanks!

jkarni:
Hmm, looking through the code, it's not obvious to me where or how GHCJS is being tested. @soenkehahn any insight?

In other news, stack does have binary distributions for GHCJS. Also, a trick I've seen in some GHCJS package (I can't remember which now) to deal the build timing out with cold caches is have a conditional check on the CI script to check whether the cache is cold. If so, only try building part of all the things, and then error out (someone would manually have to restart the build in those cases, or trigger another one with a commit).

FPtje:
I don't think ghcjs is being tested at all.

jkarni:
Ah. Sorry, maybe I wasn't clear. The CI problem I was thinking of was that we still didn't have a good way of testing GHCJS on CI.

FPtje:
I can imagine that being difficult, since you'd want to compile servant-client with both ghc and ghcjs and then perform tests with both of them.

jkarni:
Yeah... On the other hand, releasing something without CI into the wild sounds like a recipe for disaster.

 We do have servers that we could use though. So we could have something that just runs a cronjob or handles a github hook, without starting from scratch every time. If you tell me how to run the tests locally, I could try to get that setup on my server.

alpmestan:
(Don't have time to check right now.) Wasn't there a library for facilitating the definition of github hooks?

jkarni:
@alpmestan yes, [this](https://hackage.haskell.org/package/servant-github-webhook). 

Of course, we'd be running our own CI service, in essence. That's an unfortunate complication to CI, though if we simply can't get travis or Circle to cut it, one I'm okay with.

FPtje:
If travis can deal with ghcjs, it'll probably doable when the ghcjs client is separated from servant-client.

jkarni:
What do you mean separated? I thought from previous discussion the idea was a separate package (rather than a flag), but I imagined the codebase being the same. 

@FPtje any chance you'll be in London for Haskell exchange? If so, we could try to figure out CI there together (@arianvp will also be there)

FPtje:
Yes, I mean having one `servant-client` package for the whole shebang with ghc, and a `servant-client-ghcjs` package purely for a ghcjs client.

FPtje:
Sadly, I won't be in London, by the way.

jkarni:
A few of us are in London, and are talking about this. On Saturday we're
going to sit down and try to get CI working, even if it's just a cron job
on my server.

Sorry this is all a bit slow and a bit of a drag, but thanks for picking it
up again!

jkarni:
@dmj pointed out https://docs.travis-ci.com/user/languages/nix - we could
try using that for a binary distribution of ghcjs...

arianvp:
That looks cool. I did not know Travis could build nix expressions!

We already build nix expressions for all our repo's right? We can just adapt those to use ghcjs instead

FPtje:
Hey, I haven't heard from this for a while. Have there been any updates?

arianvp:
@FPtje , I got some more time to work on Servant again  (scriptie was taking its toll ;P ), interested in grabbing a coffee soon?

FPtje:
@arianvp Sounds like a plan! I've sent you an email.

jkarni:
@FPtje #803 makes it easy to now implement `servant-client-ghcjs` as a separate library, which hopefully makes all of this much more pleasant. I'm closing this.

Issue #522, opened 2016-06-01
Docs for HasServer (Header sym a) show incorrect Server type


axman6:
The docs for the Header instance for HasServer show that for `Header sym a` the `Server` type will be `a -> ...` instead of `Maybe a -> ...`.

PR #614, opened 2016-10-11
Allow multiple content-types for single Accept

phadej:
WIP, not-tested

The situation is to be able to write:

``` hs
class Accept JPG where
    contentTypes = "image/jpg" :| [ "image/jpeg" ]
```
- [x] tests
- [x] `servant-client` changes

---

this is different problem then in #522. Though with (too) clever `MimeUnrender` the problem there could be solved with this approach too.

### DIFF
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -60,11 +60,15 @@ library
     , http-types            >= 0.8  && < 0.10
     , mtl                   >= 2.0  && < 2.3
     , mmorph                >= 1    && < 1.1
-    , semigroups            >= 0.16 && < 0.19
     , text                  >= 1    && < 1.3
     , string-conversions    >= 0.3  && < 0.5
     , network-uri           >= 2.6  && < 2.7
     , vault                 >= 0.3  && < 0.4
+
+  if !impl(ghc >= 8.0)
+    build-depends:
+      semigroups            >= 0.16 && < 0.19
+
   hs-source-dirs: src
   default-language: Haskell2010
   other-extensions: CPP
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -113,12 +113,17 @@ test-suite spec
     , attoparsec
     , bytestring
     , hspec == 2.*
+    , http-media
     , QuickCheck
     , quickcheck-instances
     , servant
     , string-conversions
     , text
     , url
+
+  if !impl(ghc >= 8.0)
+    build-depends:
+      semigroups            >= 0.16 && < 0.19
 
 test-suite doctests
  build-depends: base
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -60,6 +60,7 @@ library
     , http-types            >= 0.8  && < 0.10
     , mtl                   >= 2.0  && < 2.3
     , mmorph                >= 1    && < 1.1
+    , semigroups            >= 0.16 && < 0.19
     , text                  >= 1    && < 1.3
     , string-conversions    >= 0.3  && < 0.5
     , network-uri           >= 2.6  && < 2.7
--- a/servant/src/Servant/API/ContentTypes.hs
+++ b/servant/src/Servant/API/ContentTypes.hs
@@ -81,6 +81,7 @@ import qualified Data.ByteString                  as BS
 import           Data.ByteString.Lazy             (ByteString, fromStrict,
                                                    toStrict)
 import qualified Data.ByteString.Lazy.Char8       as BC
+import qualified Data.List.NonEmpty               as NE
 import           Data.Maybe                       (isJust)
 import           Data.String.Conversions          (cs)
 import qualified Data.Text                        as TextS
@@ -119,6 +120,9 @@ data OctetStream deriving Typeable
 --
 class Accept ctype where
     contentType   :: Proxy ctype -> M.MediaType
+
+    contentTypes  :: Proxy ctype -> NE.NonEmpty M.MediaType
+    contentTypes  =  (NE.:| []) . contentType
 
 -- | @application/json@
 instance Accept JSON where
@@ -219,9 +223,10 @@ instance AllMime '[] where
     allMime _ = []
 
 instance (Accept ctyp, AllMime ctyps) => AllMime (ctyp ': ctyps) where
-    allMime _ = (contentType pctyp):allMime pctyps
-      where pctyp  = Proxy :: Proxy ctyp
-            pctyps = Proxy :: Proxy ctyps
+    allMime _ = NE.toList (contentTypes pctyp) ++ allMime pctyps
+      where
+        pctyp  = Proxy :: Proxy ctyp
+        pctyps = Proxy :: Proxy ctyps
 
 canHandleAcceptH :: AllMime list => Proxy list -> AcceptHeader -> Bool
 canHandleAcceptH p (AcceptHeader h ) = isJust $ M.matchAccept (allMime p) h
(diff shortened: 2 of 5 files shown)


phadej:
I tried quickly in https://github.com/phadej/servant-yaml/pull/10, `servant-server` part *just works*

jkarni:
@phadej I'm confused about the #522 reference.

I'm :+1: on this. The alternative would be writing something like:

```
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
...
data JPG
....
newtype JPEG = JPEG JPG deriving ( ... , Accept , ..)

type API = ... Get '[JPG, JPEG] ...
```

But I think this is nicer since end users shouldn't really have to be aware of the fact that there are two possible mimetypes for JPEG.

phadej:
Sorry, I meant #552

phadej:
ALso my motivation is for `servant-yaml` which seems to have 4-5 mime-types, See http://stackoverflow.com/a/38000954/1308058 and https://www.ietf.org/mail-archive/web/media-types/current/msg00688.html

servant/servant.cabal
```diff
@@ -60,6 +60,7 @@ library
     , http-types            >= 0.8  && < 0.10
     , mtl                   >= 2.0  && < 2.3
     , mmorph                >= 1    && < 1.1
+    , semigroups            >= 0.16 && < 0.19
```
soenkehahn: I'm worried about adding dependencies to `servant`, since this is what everything depends on.

Can we do this without `semigroups`? How many new transitive dependencies is this introducing?

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -219,9 +226,10 @@ instance AllMime '[] where
     allMime _ = []
 
 instance (Accept ctyp, AllMime ctyps) => AllMime (ctyp ': ctyps) where
-    allMime _ = (contentType pctyp):allMime pctyps
-      where pctyp  = Proxy :: Proxy ctyp
-            pctyps = Proxy :: Proxy ctyps
+    allMime _ = NE.toList (contentTypes pctyp) ++ allMime pctyps
+      where
+        pctyp  = Proxy :: Proxy ctyp
+        pctyps = Proxy :: Proxy ctyps
```
soenkehahn: I wish we would have less cryptic variable names. `pctyps` is not a good name. (I'm aware that you didn't introduce these names here.) `contentTypesProxy` (or `contentTypesP`) is better IMO.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -235,25 +243,31 @@ class (AllMime list) => AllMimeRender (list :: [*]) a where
                   -> [(M.MediaType, ByteString)]    -- content-types/response pairs
 
 instance OVERLAPPABLE_ ( MimeRender ctyp a ) => AllMimeRender '[ctyp] a where
-    allMimeRender _ a = [(contentType pctyp, mimeRender pctyp a)]
-        where pctyp = Proxy :: Proxy ctyp
+    allMimeRender _ a = [ (ct, bs) | ct <- NE.toList $ contentTypes pctyp ]
```
soenkehahn: Wouldn't this be easier to read without list comprehensions? Maybe that's just me.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -235,25 +243,31 @@ class (AllMime list) => AllMimeRender (list :: [*]) a where
...
 instance OVERLAPPABLE_
          ( MimeRender ctyp a
          , AllMimeRender (ctyp' ': ctyps) a
          ) => AllMimeRender (ctyp ': ctyp' ': ctyps) a where
-    allMimeRender _ a = (contentType pctyp, mimeRender pctyp a)
-                       :(allMimeRender pctyps a)
-        where pctyp = Proxy :: Proxy ctyp
-              pctyps = Proxy :: Proxy (ctyp' ': ctyps)
+    allMimeRender _ a =
+        [ (ct, bs) | ct <- NE.toList $ contentTypes pctyp ]
```
soenkehahn: Again, wouldn't it be easier to have `map (, bs) $ ...`?

servant/servant.cabal
phadej: avoiding `semigroups` is essentially impossible. We could have it optional for GHC 8.0+ though.
[`attoparsec`](http://hackage.haskell.org/package/attoparsec) depends on `semigroups`, for example, so it's in the transitive closure already.

soenkehahn:
Since `MimeRender` now allows multiple content types, I wonder whether we should make the actual sent content-type available to `MimeUnrender`.

This would also allow to write a custom (non-standard compliant) content type as a solution for #552. @cdepillabout: What do you think?

(@phadej: This should be done in a separate PR.)

servant/src/Servant/API/ContentTypes.hs
phadej: Well, I find the comprehension quite clear, yet pick any from:

```hs
-- map:
= map (, bs) $ ...
-- fmap:
= (, bs) <$> ...
-- do:
= do
  ct <- ....
  return (ct, bs)
-- list comprehension:
[ (ct, bs) | ct <- ...]
```

phadej: IMHO, the `do` notation is the 2nd most clear.

soenkehahn: I'd prefer the first.

phadej:
@soenkehahn I had that idea too, giving the content type to `MimeUnrender`, I guess it can be introduced in not-so-breaking way too.

servant/servant.cabal
soenkehahn: Ah, I see. Never mind then.

servant/src/Servant/API/ContentTypes.hs
phadej: I guess there the name is choose to resemble type level variable:
```hs
pctyp ~ ctyp
pctyps ~ ctyps
```
would
```hs
ctypProxy = Proxy :: Proxy ctyp
```
be good middleground?

servant/src/Servant/API/ContentTypes.hs
phadej: done

phadej:
Didn't change the proxy variable name. Probably would be better to do as whole file refactor afterwards? Consistency is IMHO more important then local improvement.

phadej:
Added tests

phadej:
ping @jkarni @soenkehahn , is this good to go?

soenkehahn:
@phadej: Sorry for the delay. I'd like to give this another review, hope to do so today.

servant/src/Servant/API/ContentTypes.hs
soenkehahn: Well, tbh, I wish we had less cryptic type level variables, too. I would strongly prefer `contentTypes` over `ctyps`, both at the term and the type level.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -219,9 +227,10 @@ instance AllMime '[] where
     allMime _ = []
 
 instance (Accept ctyp, AllMime ctyps) => AllMime (ctyp ': ctyps) where
-    allMime _ = (contentType pctyp):allMime pctyps
-      where pctyp  = Proxy :: Proxy ctyp
-            pctyps = Proxy :: Proxy ctyps
+    allMime _ = NE.toList (contentTypes pctyp) ++ allMime pctyps
```
soenkehahn: It would be nice to have tests for this method.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -235,25 +244,31 @@ class (AllMime list) => AllMimeRender (list :: [*]) a where
...
 instance OVERLAPPABLE_
          ( MimeRender ctyp a
          , AllMimeRender (ctyp' ': ctyps) a
          ) => AllMimeRender (ctyp ': ctyp' ': ctyps) a where
-    allMimeRender _ a = (contentType pctyp, mimeRender pctyp a)
-                       :(allMimeRender pctyps a)
-        where pctyp = Proxy :: Proxy ctyp
-              pctyps = Proxy :: Proxy (ctyp' ': ctyps)
+    allMimeRender _ a =
+        (map (, bs) $ NE.toList $ contentTypes pctyp)
```
soenkehahn: We don't seem to have a test for this new behavior.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -235,25 +244,31 @@ class (AllMime list) => AllMimeRender (list :: [*]) a where
...
+        pctyps = Proxy :: Proxy (ctyp' ': ctyps)
 
 
 -- Ideally we would like to declare a 'MimeRender a NoContent' instance, and
 -- then this would be taken care of. However there is no more specific instance
 -- between that and 'MimeRender JSON a', so we do this instead
 instance OVERLAPPING_ ( Accept ctyp ) => AllMimeRender '[ctyp] NoContent where
-    allMimeRender _ _ = [(contentType pctyp, "")]
-      where pctyp = Proxy :: Proxy ctyp
+    allMimeRender _ _ = map (, "") $ NE.toList $ contentTypes pctyp
```
soenkehahn: Here also: we don't have a test for the new behavior.

servant/src/Servant/API/ContentTypes.hs
phadej: There is no tests for `canHandleAcceptH` or `allMime`.

phadej:
I'm sorry, but I'm not interested enough in writing tests for existing but not tested functionality (e.g. `allMimeRender` directly). I actually don't know what tests to add.

phadej:
Also rebased.

soenkehahn:
> I'm sorry, but I'm not interested enough in writing tests for existing but not tested functionality (e.g. allMimeRender directly). I actually don't know what tests to add.

Fair enough. I'll see whether I can add some tests. Thanks for the work so far.

phadej:
rebased

phadej:
ping @jkarni @soenkehahn

phadej:
Can this be merged (I don't have rights to do so myself, if you wonder).

soenkehahn:
@phadej: Sorry, I'm not working on servant right now. I'll withdraw my review. Pinging @haskell-servant/maintainers.

PR #615, opened 2016-10-12
Mime unrender multiple

phadej:
Follow-up to #614, makes possible to resolve problem in #552 without complicating general `servant-client` code: the problem solution will go into client application.

/cc @jkarni @soenkehahn @cdepillabout 

I added an `aeson-compat` dependency, as it makes test pass on ghc-7.8. With old enough `aeson`, vanilla `decode` doesn't accept top-level scalars (i.e. accepts only arrays and objects, which leads to wtf-situations). Alternatively one could bump aeson version for tests.

### DIFF
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -111,10 +111,10 @@ test-suite spec
       base == 4.*
     , base-compat
     , aeson
+    , aeson-compat >=0.3.3 && <0.4
     , attoparsec
     , bytestring
     , hspec == 2.*
-    , http-media
     , QuickCheck
     , quickcheck-instances
     , servant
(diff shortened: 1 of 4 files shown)


servant-client/src/Servant/Common/Req.hs
```diff
@@ -7,13 +7,17 @@
...
+import Prelude hiding (any)
+
 #if !MIN_VERSION_base(4,8,0)
 import Control.Applicative
 #endif
 import Control.Exception
 import Control.Monad
 import Control.Monad.Catch (MonadThrow)
 
+import Data.Foldable (toList, any)
```
jkarni: Sort of annoying comment, but can we remove the extra space between this and the rest of the imports (and put this one in alphabetical order)?

jkarni:
@haskell-servant/maintainers another review?

phadej:
rebased

alpmestan:
LGTM.

PR #620, opened 2016-10-15
servant-client: use base-compat in order to use less CPP

soenkehahn:


servant-client/servant-client.cabal
```diff
@@ -36,6 +36,7 @@ library
     Servant.Common.Req
   build-depends:
       base                  >= 4.7      && < 4.10
+    , base-compat           >= 0.9.1
```
phadej: I'd add an upper bound.

soenkehahn: Done.

PR #622, opened 2016-10-21
Add a field to Delayed for storing a clean up action

alpmestan:
I need something like this in [servant-multipart](https://github.com/haskell-servant/servant-multipart) in order to delete files created in a temporary dir once the handler has run.

This is not the definitive code, but at least we can start a discussion.

For instance, @jkarni pointed out that the clean up might not always run with the code as is because of exceptions.

I decided to make the clean up actions live in `IO` (and not `DelayedIO`) because most of the time, if not always, the code we want to run there lives in `IO` and doesn't really need access to the request, but rather to something created in the body.

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -22,6 +22,7 @@ module Servant.Server.Internal
   , module Servant.Server.Internal.ServantErr
   ) where
 
+import           Control.Exception          (finally)
 import           Control.Monad.Trans        (liftIO)
 import qualified Data.ByteString            as B
 import qualified Data.ByteString.Char8      as BC8
@@ -398,11 +399,13 @@ instance HasServer Raw context where
   type ServerT Raw m = Application
 
   route Proxy _ rawApplication = RawRouter $ \ env request respond -> do
-    (r, _) <- runDelayed rawApplication env request
-    case r of
-      Route app   -> app request (respond . Route)
-      Fail a      -> respond $ Fail a
-      FailFatal e -> respond $ FailFatal e
+    (r, cleanup) <- runDelayed rawApplication env request
+    go r request respond `finally` cleanup
+
+    where go r request respond = case r of
+            Route app   -> app request (respond . Route)
+            Fail a      -> respond $ Fail a
+            FailFatal e -> respond $ FailFatal e
 
 -- | If you use 'ReqBody' in one of the endpoints for your API,
 -- this automatically requires your server-side handler to be a function
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -398,7 +398,7 @@ instance HasServer Raw context where
   type ServerT Raw m = Application
 
   route Proxy _ rawApplication = RawRouter $ \ env request respond -> do
-    r <- runDelayed rawApplication env request
+    (r, _) <- runDelayed rawApplication env request
     case r of
       Route app   -> app request (respond . Route)
       Fail a      -> respond $ Fail a
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -399,8 +399,12 @@ instance HasServer Raw context where
   type ServerT Raw m = Application
 
   route Proxy _ rawApplication = RawRouter $ \ env request respond -> do
-    (r, cleanup) <- runDelayed rawApplication env request
-    go r request respond `finally` cleanup
+    -- note: a Raw application doesn't register any cleanup
+    -- but for the sake of consistency, we nonetheless run
+    -- the cleanup once its done
+    cleanupRef <- newCleanupRef
+    r <- runDelayed rawApplication env request cleanupRef
+    go r request respond `finally` runCleanup cleanupRef
 
     where go r request respond = case r of
             Route app   -> app request (respond . Route)
--- a/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
+++ b/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
@@ -9,6 +9,7 @@
 {-# LANGUAGE TupleSections              #-}
 module Servant.Server.Internal.RoutingApplication where
 
+import           Control.Exception                  (bracket)
 import           Control.Monad                      (ap, liftM)
 import           Control.Monad.Trans                (MonadIO(..))
 import           Control.Monad.Trans.Except         (runExceptT)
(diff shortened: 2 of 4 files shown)


servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -258,8 +271,11 @@ runAction :: Delayed env (Handler a)
           -> (RouteResult Response -> IO r)
           -> (a -> RouteResult Response)
           -> IO r
-runAction action env req respond k =
-  runDelayed action env req >>= go >>= respond
+runAction action env req respond k = do
+  (routeResult, cleanup) <- runDelayed action env req
+  resp <- go routeResult
+  cleanup
```
jkarni: I think this should be `bracket (runDelayed action env eq) snd (go . fst)`?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -103,6 +105,10 @@ data Delayed env c where
              , authD     :: DelayedIO auth
              , bodyD     :: DelayedIO body
              , serverD   :: captures -> auth -> body -> Request -> RouteResult c
+             , cleanupD  :: body -> IO ()
```
jkarni: As I mentioned elsewhere, I don't know exactly why you need the `body` arg. Or why *just* the `body`, for that matter. `DelayedIO (IO ())` seems like a more intuitive signature, to me.

alpmestan: 1/ Why is it more intuitive? Clean up will rarely (if ever) need to look at the raw request.

2/ In my case, I get filepaths of temp files and what not by looking at `body`. We can indeed add other arguments to cleanupD, but not make it a simple action, because I otherwise don't have any mean for my cleanup thing to get its hands on the temp files it is supposed to delete. More generally, this feature is useful anytime we want to clean up after body checks and other stuffs (anything, really). Except maybe for somewhat "global" stuffs, we need a way to for the clean up action to get its hands on whatever it is that needs cleaning up, so I use an argument for that.

I would not mind adding more arguments (results of the other checks). But making it a `DelayedIO` so that I have to parse everything from the request again? No.

alpmestan:
Yes, I'll write tests before it's approved for merging.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -196,8 +203,9 @@ addBodyCheck :: Delayed env (a -> b)
              -> Delayed env b
 addBodyCheck Delayed{..} new =
   Delayed
-    { bodyD   = (,) <$> bodyD <*> new
-    , serverD = \ c a (z, v) req -> ($ v) <$> serverD c a z req
+    { bodyD    = (,) <$> bodyD <*> new
+    , serverD  = \ c a (z, v) req -> ($ v) <$> serverD c a z req
+    , cleanupD = cleanupD . fst -- not sure that's right
```
soenkehahn: Seems right to me. :)

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -240,13 +248,19 @@ passToServer Delayed{..} x =
...
-  DelayedIO (\ req -> return $ serverD c a b req)
+           -> IO (RouteResult a, IO ())
+runDelayed Delayed{..} env req = do
+  cleanupRef <- newIORef (return ())
+  routeRes <- runDelayedIO
+    (do c <- capturesD env
+        methodD
+        a <- authD
+        b <- bodyD
+        liftIO (writeIORef cleanupRef $ cleanupD b)
```
soenkehahn: I would feel better if we were able to use something like `bracket` for this. Not sure that's easily doable.

alpmestan: For writing the `IORef`, or for the whole sequence of operations?

alpmestan: Note that this is going to change a bit. @jkarni pointed out that adding clean up actions can happen from within `bodyD` and that we should not wait for it to terminate in order to add a clean up action, because if an exception occurs the cleanup never gets run. Particularly annoying in the case of file uploads where files can pile up in /tmp.

alpmestan:
@soenkehahn Yes, I'll add tests that mimic my use case as far as Delayed/cleanup are concerned. In the meantime, you might want to take a look here: https://github.com/haskell-servant/servant-multipart/blob/master/src/Servant/Multipart.hs#L267

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -112,12 +115,33 @@ instance Functor (Delayed env) where
...
       } -- Note [Existential Record Update]
 
+-- | A mutable cleanup action
+newtype CleanupRef = CleanupRef (IORef (IO ()))
+
+newCleanupRef :: IO CleanupRef
+newCleanupRef = CleanupRef <$> newIORef (return ())
+
+-- | Add a clean up action to a 'CleanupRef'
+addCleanup' :: IO () -> CleanupRef -> IO ()
```
jkarni: I'd just have this is a `where` clause to `addCleanup` so it's obvious that it's not meant to be used elsewhere.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -398,11 +399,17 @@ instance HasServer Raw context where
   type ServerT Raw m = Application
 
   route Proxy _ rawApplication = RawRouter $ \ env request respond -> do
-    r <- runDelayed rawApplication env request
-    case r of
-      Route app   -> app request (respond . Route)
-      Fail a      -> respond $ Fail a
-      FailFatal e -> respond $ FailFatal e
+    -- note: a Raw application doesn't register any cleanup
```
jkarni: I *think* this is not just for consistency. E.g.:

```
type API = FileUpload :> Raw
```

Isn't in this case not having this going to cause the cleanup action to not be run? (It probably makes sense to have a test for (1) cleanup being run for `Raw`; (2) cleanup only being run *once*)

alpmestan:
Added some basic tests to make sure that clean up operations are run under various circumstances, among which the presence of exceptions in `bodyD` or `serverD`. Also rebased against master. Could anyone take a look and let me know what's missing?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
phadej: I agree

servant-server/test/Servant/Server/Internal/RoutingApplicationSpec.hs
```diff
@@ -0,0 +1,51 @@
...
+ok :: IO (RouteResult ())
+ok = return (Route ())
+
+delayed :: DelayedIO () -> RouteResult (Handler ()) -> Delayed () (Handler ())
+delayed body srv = Delayed
+  { capturesD = \() -> DelayedIO $ \_req _cl -> ok
+  , methodD = DelayedIO $ \_req_ _cl -> ok
+  , authD = DelayedIO $ \_req _cl -> ok
+  , bodyD = do
+      liftIO (writeFile "delayed.test" "hia")
```
phadej: I'd prefer `IORef`, so the tests don't write to disk.

jkarni:
@phadej @alpmestan is this superseded by more recent PRs? Can we close?

phadej:
AFAICS all comments here are addressed already.

Issue #608, opened 2016-10-01
Improve documentation of addHeader


thoughtpolice:
Recently in the Twitt-o-sphere, I described some frustration with Response Headers, and how to send multiple headers at once, while working on a new donation site for Haskell.org. This is because I am a huge whiner (and twitter is where I vent).

But I also found the documentation quite poor, honestly! The only example I ever found was in the test suite, of how to add two headers at once - and that's not great! And it's especially nasty because if you ask me, the way you do it isn't immediately obvious.

Here's an example of why. Let's say I want to add one header. Easy enough:

```haskell
type CorsHeader v = Headers '[ Header "Access-Control-Allow-Origin" String ] v

addCors :: String -> v -> CorsHeader v
addCors origin v = addHeader origin v
```

Now, you have to return `CorsHeader` as part of the type, something like:

```haskell
"foo" :> Get '[PlainText] (CorsHeader String)
```

for example. Then do `return (addCors str)` or something.

OK, so now I want to add multiple headers. My initial intuition was that, despite the fact `Headers` uses an HList, I should have threaded the request something like this:

```haskell
type CorsHeader2 v = Headers '[ Header "Access-Control-Allow-Origin" String ] (Headers [ ... ] v)
```

Why would I think this? Well, because in the documentation the type of `addHeader` isn't explained (I'll get to that in a moment) very much. So I got the impression "Well, clearly this is just some kind of data structure wrapping the final return result. So I should wrap multiple times to make it all work". It's like how if I had `Maybe a` and extended that to `Maybe (Maybe a)`.

OK, so under the assumption I keep "nesting" them, I would then do:

```haskell
addCors2 :: String -> String -> v -> CorsHeader2 v
addCors2 domain next v = addHeader domain $ addHeader next v
```

Except this is totally wrong! Because `addHeader` is built on a type class using an `HList`, this is *the correct definition*, but it absolutely *has the wrong type*.

The proper type of `addCors2` would be something like:

```haskell
type CorsHeader2 v = Headers '[ Header "Access-Control-Allow-Origin" String, Header "Access-Control-..." String ] v
```

That is, because the HList contains the list of headers, the type class instances for `addHeader` "folds" over the HList, adding header entries based on the list position.

This took me like +15 minutes to figure out, *most* of it searching for an example (by frantically searching for `addHeader` in every project using Servant). Why? Because the type of `addHeader` is totally impenetrable!

[...]

PR #623, opened 2016-10-21
Add noHeader function.

### LINKED ISSUE #621: Optional response headers
Currently response headers can be optionally empty, but the header itself is always present. Two different people today were asking about optional headers, and the present situation is a little embarrassing.

Fixing it is relatively easy, I think. We need an overlapping instance for `BuildHeaderTo` (I believe) for `Maybe`, which only write the header in case of `Just`. 

cc: @bitemyapp

jkarni:
Which allows not adding a header where a signature declares one, making
    response headers optional. 

   Fixes #621.

### DIFF
--- a/servant/src/Servant/API.hs
+++ b/servant/src/Servant/API.hs
@@ -75,7 +75,7 @@ import           Servant.API.QueryParam      (QueryFlag, QueryParam,
 import           Servant.API.Raw             (Raw)
 import           Servant.API.RemoteHost      (RemoteHost)
 import           Servant.API.ReqBody         (ReqBody)
-import           Servant.API.ResponseHeaders (AddHeader (addHeader),
+import           Servant.API.ResponseHeaders (AddHeader, addHeader, noHeader,
                                               BuildHeadersTo (buildHeadersTo),
                                               GetHeaders (getHeaders),
                                               HList (..), Headers (..),
--- a/servant/src/Servant/API/ResponseHeaders.hs
+++ b/servant/src/Servant/API/ResponseHeaders.hs
@@ -23,7 +23,9 @@
 -- example above).
 module Servant.API.ResponseHeaders
     ( Headers(..)
-    , AddHeader(addHeader)
+    , AddHeader
+    , addHeader
+    , noHeader
     , BuildHeadersTo(buildHeadersTo)
     , GetHeaders(getHeaders)
     , HeaderValMap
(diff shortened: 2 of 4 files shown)


jkarni:
I initially thought this would break the symmetry between client and server, since I thought the client only got all headers or an error, but that's not the case, and actually this is re-instituting symmetry.

Also helps a little with #608 .

alpmestan:
LGTM.

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -108,17 +110,48 @@ instance OVERLAPPABLE_ ( KnownSymbol h, GetHeaders (HList rest), ToHttpApiData v
 -- We need all these fundeps to save type inference
 class AddHeader h v orig new
     | h v orig -> new, new -> h, new -> v, new -> orig where
-  addHeader :: v -> orig -> new  -- ^ N.B.: The same header can't be added multiple times
+  addHeader' :: Header h v -> orig -> new  -- ^ N.B.: The same header can't be added multiple times
```
soenkehahn: Maybe we can find a better name than `addHeader'`. Maybe `addOptionalHeader`?

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -108,17 +110,48 @@ instance OVERLAPPABLE_ ( KnownSymbol h, GetHeaders (HList rest), ToHttpApiData v
...
+addHeader :: AddHeader h v orig new => v -> orig -> new
+addHeader = addHeader' . Header
+
+-- | Deliberately do not add a header to a value.
+--
+-- >>> let example1 = noHeader "hi" :: Headers '[Header "someheader" Int] String
+-- >>> getHeaders example1
+-- []
+noHeader :: AddHeader h v orig new => orig -> new
+noHeader = addHeader' MissingHeader
```
soenkehahn: To clarify: We had `MissingHeader` before, but it wasn't possible to *not* set a specified header, because users always had to go through `addHeader`. So this is not just a convenience function. Is that correct?

servant/src/Servant/API.hs
```diff
@@ -75,7 +75,7 @@ import           Servant.API.QueryParam      (QueryFlag, QueryParam,
 import           Servant.API.Raw             (Raw)
 import           Servant.API.RemoteHost      (RemoteHost)
 import           Servant.API.ReqBody         (ReqBody)
-import           Servant.API.ResponseHeaders (AddHeader (addHeader),
+import           Servant.API.ResponseHeaders (AddHeader, addHeader, noHeader,
```
soenkehahn: Do you want to add an entry to the changelog for this?

PR #625, opened 2016-10-23
add err418

soenkehahn:
What I've been working on during Hacφ.


alpmestan:
So _that_ is what servant has been missing all along =)

PR #626, opened 2016-10-24
Add more documentation about response headers to tutorial.

### LINKED ISSUE #608: Improve documentation of addHeader

jkarni:
Fixes #608 

cc: @thoughtpolice

### DIFF
--- a/doc/tutorial/Server.lhs
+++ b/doc/tutorial/Server.lhs
@@ -753,7 +753,29 @@ myHandler :: Server MyHandler
 myHandler = return $ addHeader 1797 albert
 ```
 
-Note that the type of `addHeader x` is different than the type of `x`!
+Note that the type of `addHeader header x` is different than the type of `x`!
+And if you add more headers, more headers will appear in the header list:
+
+``` haskell
+type MyHeadfulHandler = Get '[JSON] (Headers '[Header "X-A-Bool" Bool, Header "X-An-Int" Int] User)
+
+myHeadfulHandler :: Server MyHeadfulHandler
+myHeadfulHandler = return $ addHeader True $ addHeader 1797 albert
+```
+
+But what if your handler only *sometimes* adds a header? If you declare that
+your handler adds headers, and you don't add one, the return type of your
+handler will be different than expected. To solve this, you have to explicitly
+*not* add a header by using `noHeader`:
+
+``` haskell
+type MyMaybeHeaderHandler
+  = Capture "withHeader" Bool :> Get '[JSON] (Headers '[Header "X-An-Int" Int] User)
+
+myMaybeHeaderHandler :: Server MyMaybeHeaderHandler
+myMaybeHeaderHandler x = return $ if x then addHeader 1797 albert
+                                       else noHeader albert
+```
 
 ## Serving static files
 


thoughtpolice:
LGTM!

doc/tutorial/Server.lhs
```diff
@@ -753,7 +753,30 @@ myHandler :: Server MyHandler
 myHandler = return $ addHeader 1797 albert
 ```
 
-Note that the type of `addHeader x` is different than the type of `x`!
+Note that the type of `addHeader x` is different than the type of `x`! More
```
soenkehahn: Shouldn't this say that the type of `addHeader headerValue x` is different from `x`?

soenkehahn: Maybe `s/More/Added/`?

doc/tutorial/Server.lhs
```diff
@@ -753,7 +753,30 @@ myHandler :: Server MyHandler
...
+headers will appear in the header list:
+
+``` haskell
+type MyHeadfulHandler = Get '[JSON] (Headers '[Header "X-A-Bool" Bool, Header "X-An-Int" Int] User)
+
+myHeadfulHandler :: Server MyHeadfulHandler
+myHeadfulHandler = return $ addHeader True $ addHeader 1797 albert
+```
+
+But what if your handler adds a header *conditional* on some run-time data? If
```
soenkehahn: `s/conditional/optionally depending/`?

doc/tutorial/Server.lhs
```diff
@@ -753,7 +753,30 @@ myHandler :: Server MyHandler
...
+
+``` haskell
+type MyMaybeHeaderHandler
+  = Capture "withHeader" Bool :> Get '[JSON] (Headers '[Header "X-An-Int" Int] User)
+
+myMaybeHeaderHandler :: Server MyMaybeHeaderHandler
+myMaybeHeaderHandler x = return $ if x then addHeader 1797 albert
+                                       else noHeader albert
+```
+
```
soenkehahn: Maybe use one newline?

doc/tutorial/Server.lhs
jkarni: Fixed! Also rephrased something else.

doc/tutorial/Server.lhs
jkarni: Hm, that doesn't sound quite right either. Maybe:

> But what if your handler only *sometimes* add a header?

soenkehahn: Yes, better.

doc/tutorial/Server.lhs
jkarni: Done.

PR #630, opened 2016-10-31
[RFC] Add TypeRep of api end-point to Foreign.Req.

fisx:
Not ready for merge, just interested in your thoughts on this.  If you're short on time feel free to ignore or close.

If you like this change, I will rebase this to current master, run the tests (should i add any for this?), and consider any other changes you may require.

Thanks!


fisx:
argh.  just noticed this isn't working quite yet.  anyway, interested in hearing how you like the idea!  (-:

fisx:
[somewhat fixed it]
[updated description above]

open problem: api types `"only_end_point" :> Get '[JSON] Int` has the default type (because it doesn't hit the `:<|>` instance), and `AuthCombinator :> (route1 :<|> route2)` doesn't mention the `AuthCombinator` in the types (because it hits the `:<|>` instance before it gets to that).

one solution would be to introduce an explicit type that steers update of `reqApiType`, and that we need to wrap each end-point in.  this would have the benefit that the change to servant, servant-js would be minimal (only change the `Req` type), and the drawback that it would uglyfy the api types.

phadej:
@haskell-servant/maintainers can someone comment on this, `servant-foreign` code is quite foreign to me.

alpmestan:
If @fisx still wants this, I would not mind. Would this break packages depending on -foreign? If it would, we might want to involve the maintainers of those packages. @fisx would you mind leading that discussion, if you still want this patch to be merged?

fisx:
I don't need this right now, but I do not want it to go to waste.  I'll try to get it rebased by the end of the week.

tchoutri:
@fisx Hi, any plans for this PR? Otherwise can I close it?

PR #635, opened 2016-11-15
Use natural-transformation-0.4

phadej:


phadej:
One issue, is to change `Nat` suffix to `NT`. Personally I first think about natural numbers when I see `Nat`. But that would be bigger breaking change.

3noch:
Or `NatTrans`

jkarni:
@haskell-servant/maintainers another review please?

@phadej can you add this to the changelog?

Issue #637, opened 2016-11-17
Add Lift


dmjio:
Sometimes it's beneficial to have only concrete types expressed inside your servant DSL. 

Example:
```haskell
handlers :: String :<|> String
handlers = "foo" :<|> "bar"

type API = "foo" :> Get '[JSON] String :<|> "bar" :> Get '[JSON] String
```

Serving this DSL will obviously fail since it's not operating inside of `ExceptT ServantErr IO`.

```haskell
main :: IO ()
main = run 3000 $ serve (Proxy :: Proxy API) handlers

{--
Main.hs:125:46-51: error: …                                                                                                                                                       
    • Couldn't match type ‘[Char]’ with ‘ExceptT ServantErr IO [Char]’                                                                                                            
      Expected type: Server API                                                                                                                                                   
        Actual type: String :<|> String                                                                                                                                           
    • In the second argument of ‘serve’, namely ‘things’                                                                                                                          
      In the second argument of ‘($)’, namely                                                                                                                                     
        ‘serve (Proxy :: Proxy API) things’                                                                                                                                       
      In the expression: run 3000 $ serve (Proxy :: Proxy API) things                                                                                                             
Compilation failed.    
--}
```

An initial attempt to transform this DSL into `ExceptT ServantErr IO` using `Nat` and `Enter`'s instance for `(:~>)` will also fail since `String` is a concrete type of kind `*`. `Nat` requires the type it's transforming to be at least `* -> *`.

For reference, `Nat` and its `Enter` instance are defined as: 
```haskell
newtype m :~> n = Nat { unNat :: forall a. m a -> n a }
instance Enter (m a) (m :~> n) (n a) where enter (Nat f) = f    
```

```haskell
main :: IO ()
main = run 3000 $ serve (Proxy :: Proxy API) (enter (Nat return) handlers)

{--
Main.hs:125:58-63: error: …                                                                                                                                                       
    • Couldn't match type ‘a’ with ‘m0 a’                                                                                                                                         
      ‘a’ is a rigid type variable bound by                                                                                                                                       
        a type expected by the context:                                                                                                                                           
          forall a. m0 a -> n0 a                                                                                                                                                  
        at /Users/dmj/Desktop/falad/Main.hs:125:54                                                                                                                                
      Expected type: m0 a -> n0 a                                                                                                                                                 
        Actual type: a -> n0 a                                                                                                                                                    
    • In the first argument of ‘Nat’, namely ‘return’                                                                                                                             
[...]

PR #638, opened 2016-11-17
Initial commit of Lift datatype

dmjio:
#637

### DIFF
--- a/servant/src/Servant/Utils/Enter.hs
+++ b/servant/src/Servant/Utils/Enter.hs
@@ -49,6 +49,13 @@ instance C.Category (:~>) where
 instance Enter (m a) (m :~> n) (n a) where
     enter (Nat f) = f
 
+-- | A lifting function from @a@ to @n a@. Used to `enter` particular
+-- datatypes.
+newtype x :^ n = Lift { unLift :: forall a. a -> n a} deriving Typeable
+
+instance Enter a (a :^ n) (n a) where
+    enter (Lift f) = f
+
 -- | Like `lift`.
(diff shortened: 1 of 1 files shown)


dmjio:
@haskell-servant/core would appreciate a review / critique from someone at some point.

phadej:
I wonder whether we can use `Identity :~> n`, and provide combinators which makes dealing with `Identity` transparent?

dmjio:
@phadej, yea that'd be interesting. If we used `Identity` it would require all other types in the DSL to be `Identity` as well, the more polymorphic we make it the harder it is for `Enter` to infer anything. `Lift` just says, I know there's already a `Monad m => m` there, so just put this `a` into it with all the rest. Otherwise, we could make 2 APIs, one full of `Identity` that gets transformed into the same type as `ExceptT ServantErr IO`, but I don't think this is necessary. I don't see `Lift` as hostile at all, and I don't think making it an instance of `Enter` introduces ambiguity. The argument seems to be whether or not it should be in `servant` proper. Regardless, I can keep it in the package I'm working on and add the `Enter` instance there. Just think others might find this useful.

servant/src/Servant/Utils/Enter.hs
```diff
@@ -49,6 +49,13 @@ instance C.Category (:~>) where
 instance Enter (m a) (m :~> n) (n a) where
     enter (Nat f) = f
 
+-- | A lifting function from @a@ to @n a@. Used to `enter` particular
+-- datatypes.
+newtype x :^ n = Lift { unLift :: forall a. a -> n a} deriving Typeable
```
phadej: Why it's `newtype x :^ n`, is phantom `x` useful for something?

Issue #344, opened 2016-01-20
Generate client types and functions with TH


fizruk:
As brought up by @jtdaugherty in https://github.com/haskell-servant/servant/issues/174#issuecomment-172602367 it would be nice to have TH generation for Haskell client types and functions.

The idea is to translate this API:

``` haskell
type Bar = Value
type Foo = Value
type FooId = Int

type API
    = "bar" :> Get '[JSON] Bar
 :<|> "foo" :> Capture "foo_id" FooId :>
         ( "a" :> Get '[JSON] Foo
      :<|> "b" :> Get '[JSON] Foo )
```

into this client structures and functions:

``` haskell
data ApiClient = ApiClient
  { getBar :: ExceptT ServantError IO Bar
  , mkFooClient :: FooId -> FooClient
  }

data FooClient = FooClient
  { getFooA :: ExceptT ServantError IO Foo
  , getFooB :: ExceptT ServantError IO Foo
  }

mkClient :: BaseUrl -> ApiClient
mkClient baseurl = ApiClient{..}
  where
    apiClient = client (Proxy :: Proxy API) baseurl 
    getBar :<|> fooClient = client baseurl

    mkFooClient fooId = FooClient{..}
      where
        getFooA :<|> getFooB = fooClient fooId
```

The naming for the functions/types can be derived from URL pieces (e.g. `"foo"`, `"a"`, `"b"`).

tdietert:
`apiClient = client (Proxy :: Proxy API) baseurl`
`getBar :<|> fooClient = client baseurl`

Is this a typo? How does client have type 
`client :: Proxy a -> BaseUrl -> ..
and
`client :: BaseUrl -> ..

And why is it being used twice. I feel like I understand what is happening here until I reach those lines. Please clarify?

fizruk:
@tdietert I meant `getBar :<|> fooClient = apiClient`, of course!

tdietert:
On that note, Servant 0.7 `client` function has type `HasClient layout => Proxy layout -> Client layout`, whereas the client function you use here takes a `BaseUrl`. I'm currently trying to use this approach to nest my API call that ALWAYS requires a `QueryParam "client_id" Text` combinator, no matter the endpoint. However, with the recent changes to servant-client, a `BaseUrl` and `Manager` must be passed to each client functions that queries and endpoint, instead of the baseurl being passed to the client function. Does this change the approach outlined above in any way? Thank you so much for your help :)

fizruk:
[...]

Issue #558, opened 2016-07-27
BasicAuth does not distribute over :<|>


tommyengstrom:
Given a simple API and client
``` Haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeOperators #-}

import Servant
import Servant.Client
import Control.Monad.Except
import Network.HTTP.Client (Manager)

type API = "a" :> Get '[JSON] String
      :<|> "b" :> Get '[JSON] String

api :: Proxy API
api = Proxy

getA :: Manager -> BaseUrl -> ExceptT ServantError IO String
getB :: Manager -> BaseUrl -> ExceptT ServantError IO String
getA :<|> getB = client api
```
I was expecting to be able to add authorization like this:
``` Haskell
type BA = BasicAuth "role" Bool

authApi1 :: Proxy AuthAPI1
authApi1 = Proxy

type AuthAPI1 = BA :> API

auth1GetA :: BasicAuthData -> Manager -> BaseUrl -> ExceptT ServantError IO String
auth1GetB :: BasicAuthData -> Manager -> BaseUrl -> ExceptT ServantError IO String
auth1GetA :<|> auth1GetB = client authApi1

```
But the client fails to compile with error

```
 Couldn't match type ‘BasicAuthData
                         -> (Manager -> BaseUrl -> ClientM [Char])
                            :<|> (Manager -> BaseUrl -> ClientM [Char])’
                   with ‘t :<|> t1’
    Expected type: t :<|> t1
      Actual type: Client AuthAPI1
    Relevant bindings include
      auth1GetA :: t (bound at Bug.hs:29:1)
      auth1GetB :: t1 (bound at Bug.hs:29:16)
    In the expression: client authApi1
    In a pattern binding: auth1GetA :<|> auth1GetB = client authApi1
```
But the following works fine
```Haskell 
type AuthAPI2 = BA :> "a" :> Get '[JSON] String
       :<|> BA :> "b" :> Get '[JSON] String

authApi2 :: Proxy AuthAPI2
authApi2 = Proxy


auth2GetA :: BasicAuthData -> Manager -> BaseUrl -> ExceptT ServantError IO String
auth2GetB :: BasicAuthData -> Manager -> BaseUrl -> ExceptT ServantError IO String
auth2GetA :<|> auth2GetB = client authApi2
```

After talking to @arianvp it is my understanding that authAPI1 and authAPI2 is intended to be equivalent.

arianvp:
[...]

PR #640, opened 2016-11-21
Generic client

fierce-katie:
This PR is related to #344 and adds the class that matches client functions generated using `client` with data structure (its representation as sum of products).
The structure itself has to be written manually (no TH used here), but we avoid pattern-matching, which is replaced with just one line of code.

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -43,7 +43,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
-    , generics-sop
+    , generics-sop          >= 0.1.0.0  && < 0.3
     , http-api-data         >= 0.3      && < 0.4
     , http-client           >= 0.4.18.1 && < 0.6
     , http-client-tls       >= 0.2.2    && < 0.4
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -30,6 +30,7 @@ source-repository head
 library
   exposed-modules:
     Servant.Client
+    Servant.Client.Generic
     Servant.Client.Experimental.Auth
     Servant.Common.BaseUrl
     Servant.Common.BasicAuth
@@ -42,6 +43,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
+    , generics-sop
     , http-api-data         >= 0.3      && < 0.4
     , http-client           >= 0.4.18.1 && < 0.6
     , http-client-tls       >= 0.2.2    && < 0.4
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -98,3 +98,4 @@ test-suite spec
     , transformers-compat
     , wai
     , warp
+    , generics-sop
--- a/servant-client/src/Servant/Client/Generic.hs
+++ b/servant-client/src/Servant/Client/Generic.hs
@@ -5,12 +5,14 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 module Servant.Client.Generic
   ( ClientLike(..)
-  , genericMkClient
+  , genericMkClientL
+  , genericMkClientP
   ) where
 
-import Generics.SOP   (Generic, I(..), NP(..), NS(Z), Rep, SOP(..), to)
+import Generics.SOP   (Code, Generic, I(..), NP(..), NS(Z), Rep, SOP(..), to)
 import Servant.API    ((:<|>)(..))
 import Servant.Client (ClientM)
 
@@ -40,20 +42,20 @@ import Servant.Client (ClientM)
 -- > instance (Client API ~ client) => ClientLike client APIClient
 -- >
 -- > data NestedClient = NestedClient
--- >  { getString ::  ClientM String
+-- >  { getString :: ClientM String
 -- >  , postBaz   :: Maybe Char -> ClientM ()
 -- >  } deriving GHC.Generic
 -- >
--- > instance Generic.SOP.Generic
+-- > instance Generic.SOP.Generic NestedClient
 -- > instance (Client NestedAPI ~ client) => ClientLike client NestedClient
 -- >
 -- > mkAPIClient :: APIClient
 -- > mkAPIClient = mkClient (client (Proxy :: Proxy API))
 class ClientLike client custom where
   mkClient :: client -> custom
-  default mkClient :: (Generic custom, GClientLikeP client xs, SOP I '[xs] ~ Rep custom)
+  default mkClient :: (Generic custom, Code custom ~ '[xs], GClientList client '[], GClientLikeL (ClientList client '[]) xs)
     => client -> custom
-  mkClient = genericMkClient
+  mkClient = genericMkClientL
 
 instance ClientLike client custom
       => ClientLike (a -> client) (a -> custom) where
@@ -62,9 +64,7 @@ instance ClientLike client custom
 instance ClientLike (ClientM a) (ClientM a) where
   mkClient = id
 
--- | This class is used to match client functions to the
--- representation of client structure type as sum of products
--- and basically does all the internal job to build this structure.
+-- GClientLikeP
 class GClientLikeP client xs where
   gMkClientP :: client -> NP I xs
 
@@ -75,8 +75,37 @@ instance (GClientLikeP b (y ': xs), ClientLike a x)
 instance ClientLike a x => GClientLikeP a '[x] where
   gMkClientP a = I (mkClient a) :* Nil
 
+-- GClientLikeL
+class GClientLikeL (xs :: [*]) (ys :: [*]) where
+  gMkClientL :: NP I xs -> NP I ys
+
+instance GClientLikeL '[] '[] where
+  gMkClientL Nil = Nil
+
+instance (ClientLike x y, GClientLikeL xs ys) => GClientLikeL (x ': xs) (y ': ys) where
+  gMkClientL (I x :* xs) = I (mkClient x) :* gMkClientL xs
+
+type family ClientList (client :: *) (acc :: [*]) :: [*] where
+  ClientList (a :<|> b) acc = ClientList a (ClientList b acc)
+  ClientList a acc = a ': acc
+
+class GClientList client (acc :: [*]) where
+  gClientList :: client -> NP I acc -> NP I (ClientList client acc)
+
+instance (GClientList b acc, GClientList a (ClientList b acc))
+  => GClientList (a :<|> b) acc where
+  gClientList (a :<|> b) acc = gClientList a (gClientList b acc)
+
+instance {-# OVERLAPPABLE #-} (ClientList client acc ~ (client ': acc))
+  => GClientList client acc where
+  gClientList c acc = I c :* acc
+
 -- | Generate client structure from client type.
-genericMkClient :: (Generic custom, GClientLikeP client xs, SOP I '[xs] ~ Rep custom)
+genericMkClientL :: (Generic custom, Code custom ~ '[xs], GClientList client '[], GClientLikeL (ClientList client '[]) xs)
   => client -> custom
-genericMkClient = to . SOP . Z . gMkClientP
+genericMkClientL = to . SOP . Z . gMkClientL . flip gClientList Nil
+
+genericMkClientP :: (Generic custom, Code custom ~ '[xs], GClientLikeP client xs)
+  => client -> custom
+genericMkClientP = to . SOP . Z . gMkClientP
 
--- a/servant-client/src/Servant/Client/Generic.hs
+++ b/servant-client/src/Servant/Client/Generic.hs
@@ -12,7 +12,7 @@ module Servant.Client.Generic
   , genericMkClientP
   ) where
 
-import Generics.SOP   (Code, Generic, I(..), NP(..), NS(Z), Rep, SOP(..), to)
+import Generics.SOP   (Code, Generic, I(..), NP(..), NS(Z), SOP(..), to)
 import Servant.API    ((:<|>)(..))
 import Servant.Client (ClientM)
 
@@ -38,7 +38,7 @@ import Servant.Client (ClientM)
 -- >   , mkNestedClient :: Int -> NestedClient
 -- >   } deriving GHC.Generic
 -- >
--- > instance Generic.SOP.Generic APIClient
+-- > instance Generics.SOP.Generic APIClient
 -- > instance (Client API ~ client) => ClientLike client APIClient
 -- >
 -- > data NestedClient = NestedClient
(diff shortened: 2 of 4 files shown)


servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+import Servant.Client (ClientM)
+
+-- | This class allows us to match client structure with client functions
+-- produced with 'client' without explicit pattern-matching.
+--
+-- The client structure needs a 'Generics.SOP.Generic' instance.
+--
+-- Example:
+--
+-- type API
```
jkarni: Can you make this a code section (start each line with `>`)?

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+--     = "foo" :> Capture "x" Int :> Get '[JSON] Int
+--  :<|> "bar" :> QueryParam "a" Char :> QueryParam "b" String :> Post '[JSON] [Int]
+--  :<|> Captre "nested" Int :> NestedAPI
```
jkarni: `Captre` -> `Capture`.

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+-- instance Generic.SOP.Generic APIClient
+-- instance (Client API ~ client) => ClientLike client APIClient
+--
+-- data NestedClient = NestedClient
+--  { getString ::  ClientM String
+--  , postBaz   :: Maybe Char -> ClientM ()
+--  } deriving GHC.Generic
+--
+-- instance Generic.SOP.Generic
+-- instance (Client NestedAPI ~ client) => ClientLike client NestedAPI
```
jkarni: I think the second `NestedAPI` should be `NestedClient`

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+--
+-- mkAPIClient :: APIClient
+-- mkAPIClient = mkClient (client (Proxy :: Proxy API))
+class ClientLike client custom where
+  mkClient :: client -> custom
```
jkarni: It make be that I'm missing something, but if we had:

```haskell
class (HasClient api) => ClientLike api custom where
   mkClient :: Client client -> custom
```

Then wouldn't we be able to say:

```haskell
instance ClientLike NestedAPI NestedClient
```
Instead of:
```haskell
instance (Client NestedAPI ~ client) => ClientLike client NestedClient
```

And maybe even get away with:

```haskell
data NestedClient = ... deriving (ClientLike NestedAPI)
```

fizruk:
A minor warning: for some APIs this might cause GHC panic (@fierce-katie can provide the error message). We've noticed this when tried to refactor our API a bit and haven't yet come up with a reason/solution (might be unrelated).

fierce-katie:
The error is:
```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.0.1 for x86_64-unknown-linux):
	pprIfaceCo

Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
```

It may appear if you have a more complicated client structure.

jkarni:
Yeah, that's a [bug](https://ghc.haskell.org/trac/ghc/ticket/12242) I've been hitting in a different project too. Presumably you're using `stack`? I don't think it's triggered without `ddump-hi` (which `stack` turns out by default).

GHC 8.0.2 has the fix, but it hasn't been released yet. If it's only triggered in client code that uses `mkClient`, in my opinion that's okay.

servant-client/src/Servant/Client/Generic.hs
fizruk: Well, GHC simply does not allow type family synonym instances (even with `TypeSynonymInstances` enabled). So we work around that with type equality constraint. Since in practice client structure (e.g. `NestedClient`) will only have one instance, I don't see this as a problem (only as an inconvenience).
I am not sure why that's not allowed, but I didn't investigate it.

We have also explored the possibility of having `ClientLike` parametrized with an API type rather than client type. But that leads to many more problems (e.g. you now have to know API type for every recursing `mkClient` call and also have separate instances for API combinators).

jkarni: Fair enough - thanks for the explanation!

alpmestan:
That's terrific! In my opinion this is even blog-post worthy (either on servant's or on your own).

fierce-katie:
Also, I played with TH trying to generate client structures (or at least separate functions) declarations, as proposed in the mentioned issue, but encountered a problem getting TH types representations. TH does not expand type synonyms, so `runQ [t| Client API|]` results in an obvious `AppT (ConT Client) (ConT API)`. There's [th-expand-syns](https://hackage.haskell.org/package/th-expand-syns-0.4.1.0), but it doesn't support type families now.
My current code is very awkward and error-prone, but if I come up with something better, this could be in a separate PR and most probably be an `Experimental` module. Any suggestions are highly appreciated :)

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+-- >  { getString ::  ClientM String
+-- >  , postBaz   :: Maybe Char -> ClientM ()
+-- >  } deriving GHC.Generic
+-- >
+-- > instance Generic.SOP.Generic
+-- > instance (Client NestedAPI ~ client) => ClientLike client NestedClient
+-- >
+-- > mkAPIClient :: APIClient
+-- > mkAPIClient = mkClient (client (Proxy :: Proxy API))
+class ClientLike client custom where
```
fizruk: > I'd add a note why there isn't `| client -> xs` functional dependency.

Sure. I would say that there is more that one way to turn `Client api` into another data type.

One example we're working on is for APIs like `(a :<|> b) :<|> c` (note the parentheses!). Currently this code will only match a client for this API with a structure like this:

```haskell
data TopClient = TopClient
  { mkNestedClient :: NestedClient
  , getC :: ... }

data NestedClient = NestedClient
  { getA :: ...
  , getB :: ... }
```

In practice sometimes it is reasonable to embed `NestedClient` in `TopClient`:

```haskell
data TopClient = TopClient
  { getA :: ...
  , getB :: ...
  , getC :: ... }
```

But I'm not sure how to allow both structures.

> Haven't noticed that this PR doesn't flatten the client. That might be useful too.

Sure, that can be handy too. But that could be a different transformation that can be combined with an approach shown here.

jkarni: I had also been thinking about flattening clients, but I agree that we can add that later - it seems to be just a type family away.

jkarni: Why? Shouldn't 

```haskell
type family Flatten client where ...
mkAPIClient = mkClient (client (Proxy :: Proxy (Flatten API)))
```

Just work?

fizruk: No, because `GClientLike` goes from `Generic` representation of custom client, and currently it demands an N-tuple (not nested tuples) for N-alternative API. Note that `Flatten API` would be represented by nested 2-tuples, not an N-tuple.

jkarni: I see - but there's no reason why in principle we can't come to support @phadej 's idea by incorporating a few more instances, no?

kosmikus:
Looks nice. I'll try to take a more detailed look tonight. But if you want to merge already, don't let that stop you.

servant-client/servant-client.cabal
```diff
@@ -42,6 +43,7 @@ library
     , base64-bytestring     >= 1.0.0.1  && < 1.1
     , bytestring            >= 0.10     && < 0.11
     , exceptions            >= 0.8      && < 0.9
+    , generics-sop
```
kosmikus: Dependency should have bounds.

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+-- >
+-- > instance Generic.SOP.Generic APIClient
+-- > instance (Client API ~ client) => ClientLike client APIClient
+-- >
+-- > data NestedClient = NestedClient
+-- >  { getString ::  ClientM String
+-- >  , postBaz   :: Maybe Char -> ClientM ()
+-- >  } deriving GHC.Generic
+-- >
+-- > instance Generic.SOP.Generic
```
kosmikus: This looks incomplete.

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,82 @@
...
+-- > instance (Client NestedAPI ~ client) => ClientLike client NestedClient
+-- >
+-- > mkAPIClient :: APIClient
+-- > mkAPIClient = mkClient (client (Proxy :: Proxy API))
+class ClientLike client custom where
+  mkClient :: client -> custom
+  default mkClient :: (Generic custom, GClientLikeP client xs, SOP I '[xs] ~ Rep custom)
```
kosmikus: I prefer `Code custom ~ '[ xs ]` over `SOP I '[ xs ] ~ Rep custom`, but both conditions are equivalent.

servant-client/src/Servant/Client/Generic.hs
kosmikus: Regarding the associativity:

I think I'd intuitively prefer accepting the non-nested definition for left-nested occurrences of `:<|>` instead of requiring the nested one. This would be easy to do, and seems a bit more symmetric to me.

Regarding the flattening:

I don't understand the argument about N-tuples here. I think @jkarni is right. Flattening is an orthogonal operation. If you have a type-level computation `Flatten` that turns your API into flattened form, then you can do the entire construction on `Flatten API`, and then successfully map to a flat datatype.

alpmestan: `Flatten` has been referred to as `Canonicalize` in the past, for what it's worth, if I understand what `Flatten` would be correctly. There's an implementation for it in some ticket.

fizruk: > I think I'd intuitively prefer accepting the non-nested definition for left-nested occurrences of `:<|>` instead of requiring the nested one. This would be easy to do, and seems a bit more symmetric to me.

That's actually not that easy to do. Now we take only the head of the type level list, but if we go left and right, we'd have to use `xs ++ ys` which can't be used in an instance head. So we'd need to define a separate class with associated type family to split `zs` into `xs` and `ys` based on `a :<|> b` client. This is doable, of course, just messy.

This approach also makes `a :<|> b` decide how the custom client should be split... which is probably not a good idea, although I am not 100% sure. One concern is that `a :<|> b :<|> c` client can be represented by either 3 fields in a client record, or 1 field (client sub-record). I suppose that means that if we flatten the nested `:<|>` on the left, sometimes GHC wouldn't be able to resolve the types. To eliminate this issue we could add `instance TypeError ... => ClientLike (a :<|> b) '[x]` so that `:<|>` would have to be flattened always...

So I guess, we can do it, but I am not sure yet that we should.

> I don't understand the argument about N-tuples here. I think @jkarni is right. Flattening is an orthogonal operation. If you have a type-level computation Flatten that turns your API into flattened form, then you can do the entire construction on Flatten API, and then successfully map to a flat datatype.

Here's a simple example of a client and flattened client. Which instances should I add to `ClientLike` to allow `mkClient :: Client -> FlattenedClient`? I doesn't work out of the box right now and I don't see an immediate solution here (accept the one that is essentially its own type class for flattening, [such as one here](http://hackage.haskell.org/package/smsaero-0.4/docs/SMSAero-Utils.html))

```haskell
type Client = Maybe Int -> Maybe String -> (ClientM Int :<|> ClientM String)

type FlattenedClient
    = (Maybe Int -> Maybe String -> ClientM Int)
 :<|> (Maybe Int -> Maybe String -> ClientM String)
```

kosmikus: My version is here: https://gist.github.com/kosmikus/108b4bb11a17a9955495fc3db8b40661

And I still also think that a `Flatten` type family suffices. Why would you want to convert at the term level?

jkarni:
Here's the `Canonicalize` that @alpmestan mentioned. Note that there was a bug in this! I don't remember what it was, but @alpmestan might - he worked on it, if I recall correctly:

```haskell
-type family Canonicalize api :: * where
 -- requires UndecidableInstances
 Canonicalize (a :> (b :<|> c)) = a :> Canonicalize b :<|> a :> Canonicalize c
 Canonicalize ((a :<|> b) :> c) = a :> Canonicalize c :<|> b :> Canonicalize c
 Canonicalize (a :> b)          = Redex b (Canonicalize b) a
 Canonicalize (a :<|> b)        = Canonicalize a :<|> Canonicalize b
 Canonicalize a                 = a

type family Redex a b c :: * where
  Redex a a first = Canonicalize first :> a
  Redex a b first = Canonicalize (first :> b)
```

servant-client/src/Servant/Client/Generic.hs
fizruk: Ok, I've finally understood what I've been missing about flattening: one doesn't have to flatten the client on the term-level since it can generate already flattened client for a flattened API! Sorry everyone! :)

alpmestan:
@jkarni Are you thinking of https://github.com/haskell-servant/servant/issues/558#issuecomment-244818924 ?

alpmestan:
By the way, this all makes me think we really ought to put `Canonicalize` (with another name) somewhere in one of the core packages. Duplicating it (with possibly different bugs in different implementations, that haven't been as battle-tested) is not the best option.

kosmikus:
@alpmestan Agreed.

jkarni:
@fierce-katie could you address @kosmikus' comments?

fierce-katie:
I've implemented what @kosmikus suggests, but I think it's reasonable to have both implementations, use one of them as default and manually declare `instance ClientLike` if you want to use the other one. We've tried this on some real APIs, and it turned out to be rather convenient to have separate clients for logically separated nested APIs. So the idea is, depending on how `instance ClientLike` is declared for client data structure, its nested clients are either expanded or not.

If there are no objections, I'll add it to this PR.

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,162 @@
...
+  mkClient = genericMkClientL
+
+instance ClientLike client custom
+      => ClientLike (a -> client) (a -> custom) where
+  mkClient c = mkClient . c
+
+instance ClientLike (ClientM a) (ClientM a) where
+  mkClient = id
+
+-- GClientLikeP
```
fizruk: This comment is not helpful.

servant-client/src/Servant/Client/Generic.hs
```diff
@@ -0,0 +1,162 @@
...
+  gMkClientP :: client -> NP I xs
+
+instance (GClientLikeP b (y ': xs), ClientLike a x)
+      => GClientLikeP (a :<|> b) (x ': y ': xs) where
+  gMkClientP (a :<|> b) = I (mkClient a) :* gMkClientP b
+
+instance ClientLike a x => GClientLikeP a '[x] where
+  gMkClientP a = I (mkClient a) :* Nil
+
+-- GClientLikeL
```
fizruk: And this one too.

phadej:
I see that everyone is happy with this, can @fierce-katie rebase this so we can merge?

fierce-katie:
@phadej rebased

phadej:
Thanks, merged!

PR #644, opened 2016-11-28
Make mimeRender failable

benweitzman:
I have some serializers that can possibly fail depending on the value given, having `mimeRender` return a `Maybe` makes everything much neater for those cases. Maybe this is useful for others?

### DIFF
--- a/servant/src/Servant/API/ContentTypes.hs
+++ b/servant/src/Servant/API/ContentTypes.hs
@@ -157,7 +157,12 @@ newtype AcceptHeader = AcceptHeader BS.ByteString
 -- > type MyAPI = "path" :> Get '[MyContentType] Int
 --
 class Accept ctype => MimeRender ctype a where
-    mimeRender  :: Proxy ctype -> a -> ByteString
+    {-# MINIMAL mimeRenderMaybe | mimeRender #-}
+    mimeRenderMaybe  :: Proxy ctype -> a -> Maybe ByteString
+    mimeRenderMaybe p val = Just $ mimeRender p val
+
+    mimeRender :: Proxy ctype -> a -> ByteString
+    mimeRender _ _ = error "no mimeRender defn"
 
 class (AllMime list) => AllCTRender (list :: [*]) a where
     -- If the Accept header can be matched, returns (Just) a tuple of the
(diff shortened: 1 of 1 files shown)


fizruk:
@benweitzman can you give an example of such a serializer?

I did not encounter such problem, but I think if some data cannot always be serialized then maybe you can add a new data type (and conversion functions) for that? Why is making `mimeRender` failable a better approach?

It appears to me that failure to render is an application error and should be handled in the application layer.

benweitzman:
I've been working on a versioning layer on top of aeson (https://github.com/vfiles/aeson-versioned/tree/vfiles-version). One of the design goals of this project is to lay things out in such a way that the application does not care about the version of the data type until it needs to be serialized. This is desirable because having to maintain multiple code paths for different versions of the same request is tedious. 

This has the consequence that by the time the application gets to serialization, it may be unable to fulfill the request for the particular version. Consider the case of adding a new constructor to a union type. This will be a breaking change for any clients. Any endpoint that returns a value using the new constructor will be unable to serialize the data in a way that clients who are using the old version can understand. I think in this case a 406 error is the most appropriate thing to respond with. 

I experimented with writing instances of `AllMimeRender` and `AllCTRender` directly (instead of `MimeRender`) since those are both failable already, but it is quite annoying. It requires any failable content types to be at the front of the list (which is a confusing requirement) and it duplicates content type checking code.

jkarni:
The use case is pretty compelling, but I am not sure this is the right thing to do. Consider some state-changing endpoint that returns data. If a client requests v1, and it turns out the data can't be serialized as v1, as far as I can tell the handler will run, the side-effect will take place, but the response will be 406. That's not good, since the client can't know whether the side-effect occurred (and normally 406 means it didn't).

Generally we try to follow the [HTTP decision diagram](https://github.com/for-GET/http-decision-diagram), where the 406 check happens *before* the handler is run.

benweitzman:
Hmm that's a good point, perhaps there is a way to restrict possibly failing serializers to GET requests.

phadej:
I mark this blocked. Let's find a way to make only `GET` request failable; otherwise close this PR (after say few month of inactivity).

This overlaps with https://github.com/haskell-servant/servant/issues/732 a bit

jkarni:
Closing - it seems like there are still unsolved problems, and solving them might complicate `servant` unnecessarily. Please comment if you disagree.

PR #646, opened 2016-12-03
add HTTP status code 422

genya0407:
Implemented ServantErr of HTTP status code 422(Unprocessable Entity).

Definition of Unprocessable Entity: https://tools.ietf.org/html/rfc2518#section-10.3


jkarni:
Thanks!

PR #649, opened 2016-12-12
Throw error on query parameter parse failure

Philonous:
`QueryParam` and `QueryParams` pass `Nothing` to the handler when parsing the parameters fails. This is counter-intuitive and IMO makes the parsing functionality almost useless. 

This pull-request changes the HasServer instance to throw error400 when parsing fails. In case where parse failures are to be dealt with gracefully the user can define a new type:

```haskell
data MayFail a = Success a
               | Failed Text Text -- Original Text and error message

instance FromHttpApiData a => FromHttpApiData (MayFail a) where
  parseQueryParam txt =
    Right $ case parseQueryParam txt of
              Left e -> Failed txt e
              Right r -> Success r
```
or similar or just take `Text` parameters directly. 

(Maybe servant should provide this for convenience?)

Related to #256

#### Technical Details:

* I added a new `serverD` block to the `Delayed` type because none of the existing blocks seemed to be appropriate. The block is executed last, even after bodyD because this is the only way to guarantee that 400 has the least priority (bodyD can also return 406 and 415)

#### RFCs:
* When returning errors in the 400 range, the HTTP spec recommends :
> The 4xx class of status code is intended for cases in which the client seems to have erred. Except when responding to a HEAD request, the server SHOULD include an entity containing an explanation of the error situation, and whether it is a temporary or permanent condition. These status codes are applicable to any request method. User agents SHOULD display any included entity to the user.

And of course this would greatly aid debugging of requests, but I left this as a TODO because I wasn't sure what format to put the error messages in and whether to include the parse errors. 

* Parse error propagation for `Header`s could use the same `Delayed` block, but this isn't implemented in this PR (yet?). In this case it might be better to rename it to something different than `paramD`. Suggestions?

### DIFF
--- a/servant-server/CHANGELOG.md
+++ b/servant-server/CHANGELOG.md
@@ -1,3 +1,10 @@
+0.10
+------
+
+* Changed `HasServer` instances for `QueryParam` and `QueryParam` to throw 400
+  when parsing fails
+* Added `paramD` block to `Delayed`
+
 0.7.1
 ------
 
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -26,7 +26,8 @@ import           Control.Monad.Trans        (liftIO)
 import qualified Data.ByteString            as B
 import qualified Data.ByteString.Char8      as BC8
 import qualified Data.ByteString.Lazy       as BL
-import           Data.Maybe                 (fromMaybe, mapMaybe)
+import           Data.Either                (partitionEithers)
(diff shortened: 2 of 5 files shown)


Philonous:
Travis seams to fail because of -Wall + -Werror on warnings I didn't introduce here. Can I work around this?

phadej:
I'm :+1: to this change. Others, @haskell-servant/maintainers ?

fizruk:
I'm 👍, but TODOs have to be figured out before merge.

phadej:
Superseded by #670

PR #653, opened 2017-01-01
Support aeson-1.1

phadej:


phadej:
travis breakage seems due some test-deps, will check next week

phadej:
green!

alpmestan:
LGTM. I assume the extra packages that got removed went into newer stackage snapshots?

Any objection, anyone?

phadej:
Yes packages removed have same or newer versions in new stackage snapshot. season-1.1 breaking change is related to how json is parsed, so doesn't affect servant as such. hspec-wai change is the same as in servant-mock

Issue #656, opened 2017-01-07
Memory leak in servant-client


edsko:
Although the `Req` type uses a lazy bytestring for the `reqBody`, this lazy bytestring is fully evaluated in memory. Here's a test case against servant 0.9.1.1:

``` haskell
{-# LANGUAGE DataKinds             #-}
{-# LANGUAGE FlexibleInstances     #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings     #-}
{-# LANGUAGE ScopedTypeVariables   #-}
{-# LANGUAGE TypeFamilies          #-}
{-# LANGUAGE TypeOperators         #-}

module Main (main) where

import Control.Monad.Except
import Network.HTTP.Media
import Network.Wai
import System.Environment
import qualified Data.ByteString          as Strict
import qualified Data.ByteString.Lazy     as Lazy
import qualified Network.Wai.Handler.Warp as Warp
import qualified Network.HTTP.Client      as Client

import Servant
import Servant.Client
import Servant.Common.Req (setRQBody)
import Servant.Server.Internal.RoutingApplication

{-------------------------------------------------------------------------------
  Large request bodies

  This allows clients to explicitly provide a lazy bytestring
-------------------------------------------------------------------------------}

data LargeBody

instance HasClient sub => HasClient (LargeBody :> sub) where
  type Client (LargeBody :> sub) = Lazy.ByteString -> Client sub

  clientWithRoute Proxy req bs =
      clientWithRoute
        (Proxy :: Proxy sub)
        (setRQBody bs genericMediaType req)
    where
      genericMediaType :: MediaType
      genericMediaType = "application"//"octet-stream"

instance HasServer sub ctxt => HasServer (LargeBody :> sub) ctxt where
  type ServerT (LargeBody :> sub) m = IO Strict.ByteString -> ServerT sub m

  route Proxy ctxt sub =
      route
        (Proxy :: Proxy sub)
        ctxt
        (addBodyCheck sub (withRequest (return . requestBody)))

{-------------------------------------------------------------------------------
  Test API
-------------------------------------------------------------------------------}

type API = LargeBody :> Post '[JSON] NoContent

testServer :: Server API
testServer getChunk = liftIO dumpChunks
  where
    dumpChunks :: IO NoContent
    dumpChunks = do
      bs <- getChunk
      if Strict.null bs
        then return NoContent
        else Strict.putStr bs >> dumpChunks

testClient :: Client API
[...]

PR #661, opened 2017-01-12
servant-client: generalize the function to set the request body

kosmikus:
NOTE: This is most probably not ready to be merged yet. It's primarily to aid the discussion in #656.

Rather than hard-coding the `RequestBodyLBS` constructor and be
limited to lazy bytestrings, the new function `setReqBody` just
takes any value of type `RequestBody`.

This has the advantage the we can use streaming request bodies
such as for example constructed by the `streamFile` function in
http-client.

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -1,5 +1,5 @@
 name:                servant-client
-version:             0.9.1.1
+version:             0.9.2.0
 synopsis: automatical derivation of querying functions for servant webservices
 description:
   This library lets you derive automatically Haskell functions that
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -406,7 +406,7 @@ instance (MimeRender ct a, HasClient api)
   clientWithRoute Proxy req body =
     clientWithRoute (Proxy :: Proxy api)
                     (let ctProxy = Proxy :: Proxy ct
-                     in setRQBody (mimeRender ctProxy body)
+                     in setReqBodyLBS (mimeRender ctProxy body)
                                   -- We use first contentType from the Accept list
                                   (contentType ctProxy)
                                   req
--- a/servant-client/src/Servant/Common/Req.hs
+++ b/servant-client/src/Servant/Common/Req.hs
@@ -87,7 +87,7 @@ instance Exception ServantError
 data Req = Req
   { reqPath   :: String
   , qs        :: QueryText
-  , reqBody   :: Maybe (ByteString, MediaType)
+  , reqBody   :: Maybe (RequestBody, MediaType)
   , reqAccept :: [MediaType]
   , headers   :: [(String, Text)]
   }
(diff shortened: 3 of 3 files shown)


alpmestan:
I like that this doesn't change the `ReqBody` instance. I'm wondering however if we could allow users to pick between LBS, streaming etc in an easy way, without having to define a new combinator for request bodies that should be sent using some particular constructor. Otherwise everyone will just end up writing similar combinators just to pick the streaming constructor for example.

Also @phadej is actively thinking about ways to make -server and -client less about wai/http-client, he might want to chime in.

Note: we should obviously find a better name for the new request body setting function, but we're not going to merge this as is anyway and this is a detail, so we can keep that for the end :)

kosmikus:
@alpmestan Yes, this all makes sense. However, I think it would be better to disentangle the bigger picture from a small local fix that could even be fully backwards-compatible. This is because AFAIK, the problem @edsko was mentioning is currently a blocker in a project he's working on. So it might be good to have a quick fix and release if there's a small change we can all agree on for now.

phadej:
AS said in the issue, I'm ok with throwing `http-client` specific stuff in atm. Elegant but broken solution is worse than not-so-elegant but working.

And this is elegant enough.

alpmestan:
OK, I think it indeed makes sense to merge this in master and ship it in the next release. My hope for a way to be able to easily switch from LBS to streaming for the request body is a user-facing API problem and should not block this from being merged.

alpmestan:
I'm all for merging this. Unless anyone objects, let's merge this and include it in the upcoming release?

phadej:
I don't have rights to merge myself, but go ahead!

kosmikus:
@alpmestan Ok. I think we should wait for @edsko to confirm that this does indeed fix his situation, too, and not just the simplified example he created. Also, we should then agree on the actual name(s) being exposed. I chose setReqBody for the new function because it was free and because setRQBody is a naming scheme I cannot completely explain anyway. But having both function, especially without comments, is not very nice. Any better suggestions?

phadej:
I'd rename `setRQBody` to `setReqBodyLBS`, and maybe keep the `setRQBody` name for sometime, but deprecated.

kosmikus:
Ok, this is close to what I had in mind. So if I gather correctly, the suggestion is to:

- deprecate `setRQBody` and say it has been renamed to `setReqBodyLBS`
- add `setReqBodyLBS` with the same functionality as `setRQBody`
- keep the new `setReqBody` under this name
- add short documentation for both `setReqBody` and `setReqBodyLBS`.

I'll try to amend the PR.

phadej:
Feel free to merge after @edsko is verified it actually fixes the problem.

kosmikus:
I've implemented the name changes as discussed and also bumped the version and added since-annotations for the new functions (none of the other functions have any, but we can at least start somewhere).

codedmart:
LGTM, seems straight forward enough.

alpmestan:
Thanks @kosmikus :)

PR #664, opened 2017-01-16
Make servant buildable with cabal new-build

phadej:
The downside is that packages with doctests has to be `build-type: Custom`, but this approach is far more resilient.


phadej:
Note: there are some doctests in testsuites, and these scripts lose them. I'll probably just hardcode them, before we get proper doctest support in `cabal`. https://github.com/haskell/cabal/issues/2327

sol:
No comment.

PR #667, opened 2017-01-16
Make Handler a newtype

### LINKED ISSUE #641: Abstract Handler
@kosmikus suggested this a while back, but: is everyone okay with making `Handler` it's own abstract monad (rather than a type synonym)? This will make it easier to add more instances without breaking backwards compatibility.

Currently I'm thinking of making it a `MonadResource` (as part of the work on file uploads, I'd be using `MonadResource` anyhow, and it seems reasonable to allow handlers to also register cleanup actions). I've also thought about adding functions:

```haskell
etagIs :: Etag -> Handler ()
lastModifiedAt :: DateTime -> Handler ()
```

With the functionality that if there is a `If-Match`/`If-None-Match` (or `If-Modified-Since`/`If-Unmodified-Since`) header in the request that doesn't/does match the mentioned `Etag`/`DateTime`, and appropriate `throwError` is thrown. (This seems like a nice API for servers, but I still haven't figured out how it would work for clients.)

phadej:
Resolve #641

### DIFF
--- a/servant-client/servant-client.cabal
+++ b/servant-client/servant-client.cabal
@@ -75,10 +75,8 @@ test-suite spec
     , Servant.Common.BaseUrlSpec
   build-depends:
       base == 4.*
-    , base-compat
-    , transformers
-    , transformers-compat
     , aeson
+    , base-compat
     , bytestring
     , deepseq
     , hspec == 2.*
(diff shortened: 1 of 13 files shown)


codedmart:
:+1: LGTM

PR #669, opened 2017-01-16
Change JSON content type to add the charset

### LINKED ISSUE #263: Response with Text type does not contain unicode characters
As in title, when connecting to the /test endpoint using web browser, the unicode characters are not displayed as they should.

``` haskell
{-# LANGUAGE DataKinds, TypeOperators, OverloadedStrings #-}
module Main where

import Control.Monad.Trans.Either
import Servant
import Network.Wai
import Data.Text

type API = "test" :> Get '[JSON] Text

test :: EitherT ServantErr IO Text
test = return "ąółżźć sdfasdf"

proxyAPI :: Proxy API
proxyAPI = Proxy

app :: Application
app = serve proxyAPI test

main :: IO ()
main = run 8080 app
```

Tested with Firefox 42.0, Chromium 46.0:

```
"Ä…Ã³Å‚Å¼ÅºÄ‡ sdfasdf"
```

phadej:
This resolves #263.

But, it's the revert of https://github.com/haskell-servant/servant/commit/99374c5868b7e20f526532d7bcbe917c55d11ad1 by @jkarni. Can you comment why it was introduced, so we don't go back and forth?

### DIFF
--- a/servant-server/test/Servant/ServerSpec.hs
+++ b/servant-server/test/Servant/ServerSpec.hs
@@ -173,7 +173,7 @@ verbSpec = describe "Servant.API.Verb" $ do
           it "sets the Content-Type header" $ do
             response <- THW.request method "" [] ""
             liftIO $ simpleHeaders response `shouldContain`
-              [("Content-Type", "application/json")]
+              [("Content-Type", "application/json;charset=utf-8")]
 
   test "GET 200" get200 methodGet 200
   test "POST 210" post210 methodPost 210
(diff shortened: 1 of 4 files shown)


phadej:
FWIW, I'm personally against this change. I'd rather close #263 as *wontfix*.

jkarni:
I believe I changed it because I noticed `charset` is only meant for `text` content-types. It's a little unfortunate that browsers fail to follow this policy. I don't mind if the `charset` is added, though.

phadej:
I just wonder whether we'll break more than fix. @haskell-servant/maintainers comments on this?

fizruk:
AFAICT adding `charset=utf-8` won't break anything. If it'll fix things for Chromium and Firefox, then let's have it!

arianvp:
AFAIK,  `application/json` does not imply `utf-8`  the RFC that defines `json` mentions that `json` is "Unicode"  but it does not further comment on what encoding should be used

PR #670, opened 2017-01-16
Throw error on param parse failure.

phadej:
Supersedes #649

### DIFF
--- a/servant-server/CHANGELOG.md
+++ b/servant-server/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 * Add `err422` Unprocessable Entity
   ([#646](https://github.com/haskell-servant/servant/pull/646))
+* Changed `HasServer` instances for `QueryParam` and `QueryParam` to throw 400
+  when parsing fails
+  ([#649](6e77453b67dc164e5381fb867e5e6475302619a3))
+* Added `paramD` block to `Delayed`
 
 * `Handler` is now an abstract datatype. Migration hint: change `throwE` to `throwError`.
   ([#641](https://github.com/haskell-servant/servant/issues/641))
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -28,7 +28,8 @@ import           Control.Monad.Trans.Resource (runResourceT)
 import qualified Data.ByteString            as B
 import qualified Data.ByteString.Char8      as BC8
 import qualified Data.ByteString.Lazy       as BL
-import           Data.Maybe                 (fromMaybe, mapMaybe)
+import           Data.Either                (partitionEithers)
+import           Data.Maybe                 (fromMaybe)
 import           Data.String                (fromString)
 import           Data.String.Conversions    (cs, (<>))
 import           Data.Typeable
@@ -45,7 +46,7 @@ import           Network.Wai                (Application, Request, Response,
 import           Prelude                    ()
 import           Prelude.Compat
 import           Web.HttpApiData            (FromHttpApiData, parseHeaderMaybe,
-                                             parseQueryParamMaybe,
+                                             parseQueryParam,
                                              parseUrlPieceMaybe,
                                              parseUrlPieces)
 import           Servant.API                 ((:<|>) (..), (:>), BasicAuth, Capture,
@@ -311,14 +312,23 @@ instance (KnownSymbol sym, FromHttpApiData a, HasServer api context)
     Maybe a -> ServerT api m
 
   route Proxy context subserver =
-    let querytext r = parseQueryText $ rawQueryString r
-        param r =
-          case lookup paramname (querytext r) of
-            Nothing       -> Nothing -- param absent from the query string
-            Just Nothing  -> Nothing -- param present with no value -> Nothing
-            Just (Just v) -> parseQueryParamMaybe v -- if present, we try to convert to
-                                        -- the right type
-    in route (Proxy :: Proxy api) context (passToServer subserver param)
+    let querytext req = parseQueryText $ rawQueryString req
+        parseParam req =
+          case lookup paramname (querytext req) of
+            Nothing       -> return Nothing -- param absent from the query string
+            Just Nothing  -> return Nothing -- param present with no value -> Nothing
+            Just (Just v) ->
+              case parseQueryParam v of
+                  -- TODO: This should set an error description (including
+                  -- paramname)
+                  Left _e -> delayedFailFatal err400 -- parsing the request
+                                                     -- paramter failed
+
+                  Right param -> return $ Just param
+        delayed = addParameterCheck subserver . withRequest $ \req ->
+                    parseParam req
+
+    in route (Proxy :: Proxy api) context delayed
     where paramname = cs $ symbolVal (Proxy :: Proxy sym)
 
 -- | If you use @'QueryParams' "authors" Text@ in one of the endpoints for your API,
@@ -352,12 +362,20 @@ instance (KnownSymbol sym, FromHttpApiData a, HasServer api context)
         -- named "foo" or "foo[]" and call parseQueryParam on the
         -- corresponding values
         parameters r = filter looksLikeParam (querytext r)
-        values r = mapMaybe (convert . snd) (parameters r)
-    in  route (Proxy :: Proxy api) context (passToServer subserver values)
+        parseParam (paramName, paramTxt) =
+          case parseQueryParam (fromMaybe "" paramTxt) of
+              Left _e -> Left paramName -- On error, remember name of parameter
+              Right paramVal -> Right paramVal
+        parseParams req =
+          case partitionEithers $ parseParam <$> parameters req of
+              ([], params) -> return params -- No errors
+              -- TODO: This should set an error description
+              (_errors, _) -> delayedFailFatal err400
+        delayed = addParameterCheck subserver . withRequest $ \req ->
+                    parseParams req
+    in  route (Proxy :: Proxy api) context delayed
     where paramname = cs $ symbolVal (Proxy :: Proxy sym)
           looksLikeParam (name, _) = name == paramname || name == (paramname <> "[]")
-          convert Nothing = Nothing
-          convert (Just v) = parseQueryParamMaybe v
 
 -- | If you use @'QueryFlag' "published"@ in one of the endpoints for your API,
 -- this automatically requires your server-side handler to be a function
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -28,10 +28,11 @@ import           Control.Monad.Trans.Resource (runResourceT)
 import qualified Data.ByteString            as B
 import qualified Data.ByteString.Char8      as BC8
 import qualified Data.ByteString.Lazy       as BL
+import           Data.Maybe                 (fromMaybe, mapMaybe)
 import           Data.Either                (partitionEithers)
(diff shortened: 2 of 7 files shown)


phadej:
FWIW, we need `paramsD` check to make mandatory query parameters: #256

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -246,7 +266,8 @@ runDelayed Delayed{..} env = runDelayedIO $ do
   methodD
   a <- authD
   b <- bodyD
-  DelayedIO (\ req -> return $ serverD c a b req)
+  p <- paramsD -- Has to be after body to respect the relative error order
```
jkarni: Here it seems like params are being checked after the body. But in the documentation on line 91 it states that params come first. Which one is it?

servant-server/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 * Add `err422` Unprocessable Entity
   ([#646](https://github.com/haskell-servant/servant/pull/646))
+* Changed `HasServer` instances for `QueryParam` and `QueryParam` to throw 400
+  when parsing fails
```
jkarni: I would add some more alarmist note (maybe `IMPORTANT` or just `NOTICE`), since the change may break expectations of existing applications without a compilation error.

phadej: @jkarni I'll edit the changelogs to contain "BREAKING CHANGES" explicitly at the top. But until it, it's just easier to have append only log (also asking others "could you add a changelog entry" is quite vague request")

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
phadej: I'll make a test, good catch.

phadej:
@jkarni, added a test, and found a problem with previous "let's bundle everything into `bodyD` check".
Exploded it into more granular checks. 

One question: now contentType list is traversed twice, where once could be enough. I'll see if `contentTypeCheck` could remember which deserialiser we picked. But I hope it's ok to leave it as is, if it get's hairy.

I'll update haddocks after the try out.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -310,14 +312,22 @@ instance (KnownSymbol sym, FromHttpApiData a, HasServer api context)
...
-    in route (Proxy :: Proxy api) context (passToServer subserver param)
+    let querytext req = parseQueryText $ rawQueryString req
+        parseParam req =
+          case lookup paramname (querytext req) of
+            Nothing       -> return Nothing -- param absent from the query string
+            Just Nothing  -> return Nothing -- param present with no value -> Nothing
+            Just (Just v) ->
+              case parseQueryParam v of
+                  Left e -> delayedFailFatal err400
+                      { errBody = cs $ "Parsing param name " <> paramname <> " failed: " <> e
```
fizruk: How about `"Error parsing query parameter " <> paramname <> ": " <> e`?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -351,12 +361,23 @@ instance (KnownSymbol sym, FromHttpApiData a, HasServer api context)
         -- named "foo" or "foo[]" and call parseQueryParam on the
         -- corresponding values
         parameters r = filter looksLikeParam (querytext r)
-        values r = mapMaybe (convert . snd) (parameters r)
-    in  route (Proxy :: Proxy api) context (passToServer subserver values)
+        parseParam (paramName, paramTxt) =
+          case parseQueryParam (fromMaybe "" paramTxt) of
```
fizruk: Why are we using `fromMaybe ""` here? Shouldn't we ignore empty values (like it was before with `mapMaybe`)?

servant-server/src/Servant/Server/Internal.hs
fizruk: Or at least `"Parsing param "` instead of `"Parsing param name "`.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -245,8 +279,11 @@ runDelayed Delayed{..} env = runDelayedIO $ do
   c <- capturesD env
   methodD
   a <- authD
+  acceptD
+  contentD
+  p <- paramsD       -- Has to be after "body", but after "body content" checks
```
fizruk: Before body?

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -319,6 +326,54 @@ queryParamSpec = do
...
+
+      it "generates an error on parse failures of multiple parameters" $
+        (flip runSession) (serve queryParamApi qpServer) $ do
+        let params = "?ages=2&ages=foo"
+        response <- Network.Wai.Test.request defaultRequest{
+            rawQueryString = params,
+            queryString = parseQuery params,
+            pathInfo = ["multiparam"]
+           }
+        liftIO $ statusCode (simpleStatus response) `shouldBe` 400
```
fizruk: Would be nice to test for error message too.

phadej:
One way is to change to

```
addBodyCheck
    :: Delayed env (a -> b)
    -> DelayedIO c         -- content type check 'ByteString -> Either String a'
    -> (c -> DelayedIO a)  -- actual body check, which tries to read and parse body
    -> Delayed env b
```

servant-server/src/Servant/Server/Internal.hs
phadej: Hmm, I guess we have to decide what to when

```hs
type API = QueryParams "foo" MyType :> ...
```

and querystring is

```
foo&foo&foo[]=1&foo[]=2
```

should we pass an empty string, or strip them?

based on

```diff
-          convert Nothing = Nothing
 -          convert (Just v) = parseQueryParamMaybe v
```

we did strip empty params

fizruk: I would at least not change previously existing behaviour here (which was to ignore params without values). If this is a concern we can solve it in a different issue/PR.

servant-server/test/Servant/ServerSpec.hs
phadej: We don't test error messages for body parsing failures either. Won't do in this PR. See https://github.com/haskell-servant/servant/issues/673

phadej:
Ready for re-review. As I had to untrivially touch Delayed, if @kosmikus have time to review, that would be great as well.

kosmikus:
@phadej Could you please summarize the changes in behaviour and the reasons why they seem necessary? I've looked at the discussions and the other PR, but I don't see this argument being made anywhere, at least not explicitly.

phadej:
@kosmikus 
- previously `paramsD` check couldn't fail, but we want it be able to. That's the main motivator.
- also previously `Accept` and `Content-Type` checks were done n `bodyD`, i.e. last, but `addAcceptCheck` was cleverly prepending (not appending) them to the `bodyD`, so they were done before body parsing.
- Now `Accept` and `Content-Type` checks are own fields in `Delayed`, so we can perform them before `paramsD`
- and `paramsD` is before `bodyD`.
- `contentD` returns a value passed to `bodyD`, so we can pass it over `paramsD` check, this is done to not lookup content-type based parser twice.

kosmikus:
@phadej Thanks for the explanation. Superficially, this looks ok to me.

phadej:
@kosmikus, I'll take that as approval :). Let's see what @jkarni says in his re-review

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -345,18 +354,27 @@ instance (KnownSymbol sym, FromHttpApiData a, HasServer api context)
...
-    where paramname = cs $ symbolVal (Proxy :: Proxy sym)
-          looksLikeParam (name, _) = name == paramname || name == (paramname <> "[]")
-          convert Nothing = Nothing
-          convert (Just v) = parseQueryParamMaybe v
+  route Proxy context subserver = route (Proxy :: Proxy api) context $
+      subserver `addParameterCheck` withRequest paramsCheck
+    where
+      paramname = cs $ symbolVal (Proxy :: Proxy sym)
+      paramsCheck req =
+          -- We could collect all parse errors, have to use 'Validate' Applicative then.
```
jkarni: What about (to have all errors):

```haskell
case partitionEithers $ parseQueryParam <$> params of
   ([], parsed) -> return parsed
   (failed, _)  -> ...
```

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -345,18 +354,27 @@ instance (KnownSymbol sym, FromHttpApiData a, HasServer api context)
...
+          case traverse parseQueryParam params of
+              Right parsed -> return parsed
+              Left e       -> delayedFailFatal err400
+                  { errBody = cs $ "Error parsing one of query parameter " <> paramname <> " failed: " <> e
```
jkarni: Maybe `"Error parsing query parameter " <> paramname <> ": " <> e`:

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -186,24 +210,33 @@ addAuthCheck :: Delayed env (a -> b)
...
+             -> (c -> DelayedIO a)  -- ^ body check
              -> Delayed env b
-addBodyCheck Delayed{..} new =
+addBodyCheck Delayed{..} newContentD newBodyD =
   Delayed
-    { bodyD   = (,) <$> bodyD <*> new
-    , serverD = \ c a (z, v) req -> ($ v) <$> serverD c a z req
+    { contentD = (,) <$> contentD <*> newContentD
+    , bodyD    = \(content, c) -> (,) <$> bodyD content <*> newBodyD c
+      -- bodyD    :: content -> DelayedIO body
```
jkarni: I'm not sure if these comments help!

phadej: They did when I tried to make types align, but I can remove them

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -92,23 +92,36 @@ toApplication ra request respond = ra request routingRespond
...
 data Delayed env c where
   Delayed :: { capturesD :: env -> DelayedIO captures
+             , paramsD   :: DelayedIO params
              , methodD   :: DelayedIO ()
              , authD     :: DelayedIO auth
-             , bodyD     :: DelayedIO body
-             , serverD   :: captures -> auth -> body -> Request -> RouteResult c
+             , acceptD   :: DelayedIO ()
+             , contentD  :: DelayedIO content
+             , bodyD     :: content -> DelayedIO body
```
phadej: will do

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -92,23 +92,36 @@ toApplication ra request respond = ra request routingRespond
...
+-- can cause 415 and 406 errors.
+--
+-- 5. Query parameter checks. They require parsing and can cause 400 if the
+-- parsing fails. Query parameter checks provide inputs to the handler
+--
+-- 6. Body check. The request body check can cause 400.
 --
 data Delayed env c where
   Delayed :: { capturesD :: env -> DelayedIO captures
+             , paramsD   :: DelayedIO params
```
phadej: Yes, good idea.

phadej:
@jkarni, fixed based on your comments

Issue #576, opened 2016-08-22
Add custom type errors


fizruk:
We all know `servant`'s type errors are awfully long and hard to read (especially for newbies).
Let's change that with new feature of GHC 8 — [custom type errors](https://ghc.haskell.org/trac/ghc/wiki/Proposal/CustomTypeErrors)!

I don't have anything concrete yet, rather just clearing my head of this.
I don't think this is a very hard task, but probably not newcomer-friendly.

jkarni:
It would be great. The only thing we have to figure out is how to test this: the only way simple I can think of doing it is with doctest, but we'd need to branch on GHC version. I guess we can CPP documentation?

fizruk:
IIRC when you CPP documentation, you have to CPP the whole block.
You can see this in [`http-api-data`'s sources](https://github.com/fizruk/http-api-data/blob/master/src/Web/HttpApiData/Internal.hs#L147-L190).

However, we could add a separate named haddock chunk with doctests and description of nice errors.

soenkehahn:
Since `doctest` is meant to be read as a `ghci` session I think it's generally acceptable to stick to one version of `ghc`. Otherwise we would have to provide different documentation for different `ghc` versions, which sounds unpractical. So I would consider just disabling `doctest` for all but one `ghc` version.

In this special case, we should prominently point out that the type error messages reflect what `ghc-8` outputs. And that previous versions might yield much different (and much worse) type errors.

jkarni:
+1. I think a positive feedback loop with regards to better error messages is
us showing how much better things are in GHC 8, so that more of our users
switch to GHC 8, so that the impact of writing better custom errors becomes
larger, &c.

On Thu, Aug 25, 2016 at 06:31:02AM -0700, Sönke Hahn wrote:
> Since `doctest` is meant to be read as a `ghci` session I think it's generally acceptable to stick to one version of `ghc`. Otherwise we would have to provide different documentation for different `ghc` versions, which sounds unpractical. So I would consider just disabling `doctest` for all but one `ghc` version.
>
> In this special case, we should prominently point out that the type error messages reflect what `ghc-8` outputs. And that previous versions might yield much different (and much worse) type errors.
>
> --
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/haskell-servant/servant/issues/576#issuecomment-242386414

--
Julian K. Arni
Haskell Consultant, Turing Jump
https://turingjump.com

[...]

PR #671, opened 2017-01-16
TypeError annotations for AllCTRender

reactormonk:
Just a proof-of-concept. It compiles, but I'm not sure how to check if it works / write a test for it.

I had to up the stackage resolver, otherwise base would be < 4.9.0

Relevant issue: https://github.com/haskell-servant/servant/issues/576

### DIFF
--- a/servant/src/Servant/API/ContentTypes.hs
+++ b/servant/src/Servant/API/ContentTypes.hs
@@ -185,7 +185,7 @@ instance OVERLAPPABLE_
             lkup = fmap (\(a,b) -> (a, (fromStrict $ M.renderHeader a, b))) amrs
 
 #if MIN_VERSION_base(4,9,0)
-instance TL.TypeError (TL.Text "No instance for (), use NoContent instead.")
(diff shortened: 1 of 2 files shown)


jkarni:
What I've been doing for testing `TypeError`s is using `doctest`.

phadej:
@reactormonk See https://github.com/haskell-servant/servant/blob/555b60ea06dcdbb69f8823d982d2639746803ccd/servant/test/Servant/Utils/LinksSpec.hs#L70 and https://github.com/haskell-servant/servant/blob/555b60ea06dcdbb69f8823d982d2639746803ccd/servant/Setup.lhs#L109

phadej:
We have `stack-ghc-8.0.2.yaml`, so no need to bump `stack.yaml` (which we won't; because we'll loose one GHC from travis matrix).

@reactormonk can you fix tests? (you can run them for different compilers with e.g.`STACK_YAML=stack-ghc-8.0.2 stack test`)

PR #674, opened 2017-01-17
Delayed cleanup ioref tests

phadej:
Rebase of #622, with tests changed to use `IORef`

### DIFF
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -109,6 +109,7 @@ test-suite spec
       Servant.ArbitraryMonadServerSpec
       Servant.Server.ErrorSpec
       Servant.Server.Internal.ContextSpec
+      Servant.Server.Internal.RoutingApplicationSpec
       Servant.Server.RouterSpec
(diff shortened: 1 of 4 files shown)


phadej:
I'm ok with merging this, and then rewriting `DelayedIO` to use

```hs
newtype DelayedIO a = DelayedIO { runDelayedIO' :: ReaderT Request (ResourceT IO) (RouteResult a) }
```

opinions on this, @alpmestan, @jkarni?

alpmestan:
There's some build failure to address (https://travis-ci.org/haskell-servant/servant/jobs/192826872#L393) -- for ghc 7.8. Other than that, LGTM.

jkarni:
Why not just use `ResourceT` (and remove `CleanupRef`) from the start? It seems like we're just mimicking `ResourceT`s functionality, which, given how subtle these things can be (e.g., what happens in the presence of forks), is probably going to bite us.

phadej:
@jkarni I'll make a refactor PR today.

PR #675, opened 2017-01-18
Use resourcet for resource managment

phadej:


servant-server/src/Servant/Server/Internal.hs
```diff
@@ -404,9 +403,8 @@ instance HasServer Raw context where
     -- note: a Raw application doesn't register any cleanup
     -- but for the sake of consistency, we nonetheless run
     -- the cleanup once its done
-    cleanupRef <- newCleanupRef
-    r <- runDelayed rawApplication env request cleanupRef
-    go r request respond `finally` runCleanup cleanupRef
+    r <- runDelayed rawApplication env request
+    go r request respond
```
phadej: wait, here I guess I'll run cleanups too early. Wouldn't it?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -264,18 +303,17 @@ passToServer Delayed{..} x =
 runDelayed :: Delayed env a
            -> env
            -> Request
-           -> CleanupRef
            -> IO (RouteResult a)
```
phadej: so should this return `ResourceT IO (RouteResult a)`?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -32,6 +40,58 @@ data RouteResult a =
...
+instance Monad RouteResult where
+    return = Route
+    Route a     >>= f = f a
+    Fail e      >>= _ = Fail e
+    FailFatal e >>= _ = FailFatal e
+
+newtype RouteResultT m a = RouteResultT { runRouteResultT :: m (RouteResult a) }
+  deriving (Functor)
+
+-- As we write these instances, we get instances for `DelayedIO` with GND.
```
jkarni: I don't think the comment is necessary. Instances are always good in my opinion.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -115,53 +175,30 @@ instance Functor (Delayed env) where
...
+    , MonadBase IO
+    , MonadThrow
+    , MonadResource
+    )
 
-instance Applicative DelayedIO where
-  pure = return
-  (<*>) = ap
+returnRouteResult :: RouteResult a -> DelayedIO a
+returnRouteResult x = DelayedIO $ lift . lift $ RouteResultT . return $ x
```
jkarni: Not sure about the name, but can't think of anything better (maybe `delayRouteResult`?)

servant-server/test/Servant/Server/Internal/RoutingApplicationSpec.hs
```diff
@@ -59,3 +100,10 @@ spec = do
       _ <- simpleRun $ delayed (liftIO $ throwIO DivideByZero) (Route $ return ())
       cleanUpDone <- isJust <$> readIORef delayedTestRef
       cleanUpDone `shouldBe` False
+  describe "ResApi" $
+    with (return $ serve resApi resServer) $ do
+      it "writes and cleanups resources" $ do
+        request "GET" "foobar" [] "" `shouldRespondWith` "foobar"
+        liftIO $ do
+          cleanUpDone <- isJust <$> readIORef delayedTestRef
```
jkarni: This test can in theory pass if `check` in line 67 were never run. Can we maybe write something other than `Nothing` (the initial ref state) to be sure the value is coming from deallocation?

servant-server/test/Servant/Server/Internal/RoutingApplicationSpec.hs
```diff
@@ -40,6 +51,36 @@ simpleRun d = fmap (either ignoreE id) . try $
...
+
+resApi :: Proxy ResApi
+resApi = Proxy
+
+resServer :: Server ResApi
+resServer ref = liftIO $ fmap (maybe "<empty>" T.pack) $ readIORef ref
+
+-------------------------------------------------------------------------------
+-- Spec
+-------------------------------------------------------------------------------
```
jkarni: I definitely don't insist on it, but it perhaps would be nice to have a test where the combinator allocating the resource forks a thread, and that thread waits on an MVar (with `readMVar`). Then the test checks (a) that before it `putMVar`s, the resource is not yet deallocated, and (b) after it `putMVar`s, the resource is deallocated.

What I have in mind is forking jobs that are longer running, which I imagine would be commonly used in conjunction with resources, especially if we make `Handler` a `MonadResource`.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -264,18 +303,17 @@ passToServer Delayed{..} x =
...
+           -> ResourceT IO (RouteResult a)
+runDelayed Delayed{..} env req  =
   runDelayedIO
     (do c <- capturesD env
         methodD
         a <- authD
         b <- bodyD
-        DelayedIO $ \ r _cleanup -> return (serverD c a b r)
+        r <- ask
+        returnRouteResult (serverD c a b r)
```
jkarni: Now that `Handler` is abstract, and that this is being added, I feel like it makes sense to make `Handler` a `MonadResource` so that handlers can also register, allocate, and release. That said, that can just as well be a separate PR.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
phadej: What's about `liftRouteResult` ?

servant-server/test/Servant/Server/Internal/RoutingApplicationSpec.hs
phadej: I'll make a `IORef` three way, `None | InUse a | Cleared`.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
phadej: `Handler` cannot, it's just `ExceptT ServerErr IO a`. Then we'll need to add `ResourceT` there too. I'm neutral about that.

.stylish-haskell.yaml
```diff
@@ -65,7 +63,7 @@ steps:
 
 # A common setting is the number of columns (parts of) code will be wrapped
 # to. Different steps take this into account. Default: 80.
-columns: 80
+columns: 120
```
phadej: not really, except with current style the import lists are almost always spanning multiple lines, which I find annoying. There's a bit of settings in newest `stylish-haskell` so if the line is too long, imports could be aligned as
```hs
import           Data.List (foldl')
import qualified VeryLongNamespace.Data.Foo.Bar
                 (foo, bar, quux, longMethodName, anotherLongMethodName, Foobar,
                 MonadEveryhing, zyxxy)
```

or at least I remember making some such configuration possible.

I'll revert this change and investigate more (subjectively) pleasant (yet diff friendly) formatting style.

servant-server/test/Servant/Server/Internal/RoutingApplicationSpec.hs
phadej: I'm not understand, could you write an example, assuming that `Handler` is `MonadResource`?

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
phadej: (There's no technical problem adding it, the `runHandler` is inside `runResourceT` anyway).

PR #677, opened 2017-01-19
Alternative instance for ClientM

sordina:
Supposedly there used to be an Alternative instance for ClientM before it became a real data-type. I think that would be useful. Although my implementation is of questionable quality, the utility of such an instance seems obvious and this might serve as a starting point for adding one back in.

Happy to change this PR if anyone has feedback!

### DIFF
--- a/servant-client/src/Servant/Common/Req.hs
+++ b/servant-client/src/Servant/Common/Req.hs
@@ -21,6 +21,8 @@ import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Trans.Except
 
 import GHC.Generics
+import Data.Monoid
+import Control.Applicative
 import Control.Monad.Base (MonadBase (..))
 import Control.Monad.IO.Class ()
 import Control.Monad.Reader
@@ -82,6 +84,10 @@ instance Eq ServantError where
   _ == _ = False
 
 instance Exception ServantError
+
+instance Monoid ServantError where -- Outlaw Implementation
+  mempty = ConnectionError (SomeException (AssertionFailed "Empty"))
+  mappend a b = b
 
 data Req = Req
   { reqPath   :: String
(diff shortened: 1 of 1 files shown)


phadej:
What's the motivation?

More right approach would be to define e.g. `Semigroup` instance for `Client` (or/and [`Alt`](https://hackage.haskell.org/package/semigroupoids-5.1/docs/Data-Functor-Alt.html)

`Semigroup` on `ServantError` makes sense, but should it be `First` or `Last`, or something smarter?

sordina:
@phadej The motivation is to be able to stay within the ClientM monad, but use Alternative for API fetching, such as:

        updateUser updates = do
           details <- getDetails <|> createBlankUser
           putUser (merge details updates)

alpmestan:
I'm in favor of having such an instance, that little code snippet definitely looks like a nice way to do what it's doing.

sordina:
@phadej The ideas about Alt and First/Last make sense to me. Unless Servant error had some kind of CompoundError case added... but I'm not sure that's very restful!

phadej:
I'm slightly hesistant to add dependency on `semigroupoids`, so if you can get away only with `Semigroup`'s `<>`, that would be great.

I'm slightly against `Semigroup` instance for `ServantError` as it would be quite arbitrary. "semigroupiness" of errors can be backed into `Semigroup (ClientM a)`.

jkarni:
Was there really an `Alternative` instance before? I guess that means we used to have a `Monoid` instance for `ServantError`?

sordina:
I don't know from personal experience, but I was asking about it on freenode#haskell:

     2017-01-17 14:25:09	orzo	0.8.1
     2017-01-17 14:25:16	lyndon	I'm using servant-client 0.9.1.1
     2017-01-17 14:25:24	lyndon	(& servant)
     2017-01-17 14:26:05	lyndon	Unless I need to import something else...
     2017-01-17 14:26:55	orzo	looks like they made a newtype in later versions
     2017-01-17 14:27:08	orzo	in 0.8.1 ClientM was just a type alias for a transformers type ExceptT
     2017-01-17 14:27:20	orzo	so they lost the Alternative instance along the way

alpmestan:
It definitely makes sense to recover it.

phadej:
There is `Monoid e, ... => Alternative (ExcepT e m(` instance. But `ServantError` was never a `Monoid`. So "Was there a `Alternative ClientM` instance?" is a trick question.

But now there is `Alt ClientM`, #679. hopefully it's good enough.

sordina:
@phadej Fantastic!

PR #679, opened 2017-01-19
Semigroup

phadej:
Related to #677

### DIFF
--- a/servant-client/src/Servant/Common/Req.hs
+++ b/servant-client/src/Servant/Common/Req.hs
@@ -16,7 +16,8 @@ import Control.Exception
 import Control.Monad
 import Control.Monad.Catch (MonadThrow, MonadCatch)
 import Data.Foldable (toList)
-import Data.Semigroup (Semigroup (..))
+import Data.Functor.Alt (Alt (..))
+import Data.Semigroup ((<>))
 
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Trans.Except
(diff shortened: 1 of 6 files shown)


servant-client/src/Servant/Common/Req.hs
```diff
@@ -214,6 +215,10 @@ instance MonadBaseControl IO ClientM where
   -- restoreM :: StM ClientM a -> ClientM a
   restoreM st = ClientM (restoreM st)
 
+-- | Try clients in order, last error is preserved.
+instance Semigroup (ClientM a) where
+  a <> b = a `catchError` \_ -> b
```
fizruk: I think using `<>` for `<!>` is morally wrong. The appropriate instance here would be:

```haskell
-- this requires semigroupoids
instance Alt ClientM where
  a <!> b = a `catchError` \_ -> b

-- this is probably useless in practice, so I wouldn't even add this instance
instance Semigroup a => Semigroup (ClientM a) where
  (<>) = liftA2 (<>)
```

Instead of making a morally wrong `Semigroup` instance I would rather add `semigroupoids` dep.
I can't tell right now, but I suspect that it is already a transitive dependency in more than one way.

fizruk: An alternative to using `Alt` would be to add a custom combinator. E.g.:

```haskell
-- | Try clients in order, last error is preserved.
orClient :: ClientM a -> ClientM a -> ClientM a
orClient a b = a `catchError` \_ -> b
```

I would prefer this over a wrong `Semigroup` instance.

phadej: I actually don't have strong opinion on this.
- I'd prefer to add `semigroupoids` dependency (it's in `lens` closure), and not define `Semigroup` instance at all, because people have strong opinions for both
    - `(<>) = (<|>)`
    -  `(<>) = liftA2 (<>)`

fizruk:
@haskell-servant/maintainers is anyone against adding `semigroupoids` as a dependency?

jkarni:
I'm happy with either of the proposed options (semigroupoids or `orElse`)

alpmestan:
Same as @jkarni.

PR #680, opened 2017-01-19
Update .stylish-haskell

phadej:
With one file formatted as an example.

AFAICS, this is only configuration where indentation/alignment won't change globally if you add/remove long module name or qualified imports.

One option would be to have short lists right after the module name, but there isn't such configuration option in `stylish-haskell` atm.

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -15,6 +15,14 @@ steps:
   #     # true.
   #     add_language_pragma: true
 
+  # Align the right hand side of some elements.  This is quite conservative
+  # and only applies to statements where each element occupies a single
+  # line.
+  - simple_align:
+      cases: false
+      top_level_patterns: false
+      records: false
+
   # Import cleanup
   - imports:
       # There are different ways we can align names and lists.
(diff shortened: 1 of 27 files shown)


fizruk:
I'm not a big fan of aligned imports anyway, but I'm all for simpler diffs. Looks good to me! 👍

phadej:
I'd like others @haskell-servant/maintainers to comment, as I plan to reformat everything, and try to remember to do that regularly in the future

alpmestan:
Do we really want to stick so hard to stylish haskell? I won't fight it, but I'll never encourage it either =)

phadej:
@alpmestan looking across `servant` codebase, the import styles vary a lot, what worse even inside single file. As a contributor I'm always confused "now I have to add new module import, where and how should I put it". From that POV uniform style will help.

I'm not the huge fan of the style proposed, but it's one of diff friendly ones.

Another variant is to separate explicit import lists and qualified imports into two groups:

```hs
import Data.Maybe
       (mapMaybe)
import Data.List
       (foldl')
import GHC.Generics
       (Generic)

import qualified Data.ByteString as BS
import qualified Data.Text as T
```

or third:

```hs
import Data.Maybe (mapMaybe, fromJust,
                  catMaybe)
import Data.List (foldl')
import GHC.Generics (Generic)

import qualified Data.ByteString as BS
import qualified Data.Text as T
```

(third isn’t supported by stylish-haskell, but it’s an easy addition)

jkarni:
I'm in favor of a unified style, but don't much care for which, so I'm happy either way here.

phadej:
I'm merging this. I'm so used to `%!styli<tab>` to reformat imports, and even this format is vertically verbose, it's somewhat diff-friendliest

PR #681, opened 2017-01-19
Add test illustrating that query params don't affect router grouping

phadej:
Relating to the email discussion.

We do merge segments, even when there are `QueryParam`. See `type PermuteRef`.


fizruk:
Nice! I somehow missed that moment when `QueryParam` stopped using `WithRequest` (which is now called `RawRouter`) and moved completely to `Delayed`.

This changes things, thanks for the explicit test proving me wrong! 👍

PR #682, opened 2017-01-19
Add ComprehensiveAPI test for servant-foreign

phadej:


servant-foreign/test/Servant/ForeignSpec.hs
```diff
@@ -6,9 +6,11 @@ module Servant.ForeignSpec where
 import Data.Monoid ((<>))
 import Data.Proxy
 import Servant.Foreign
+import Servant.API.Internal.Test.ComprehensiveAPI
```
arianvp: Good! We should perhaps add the Auth combinators to the ComprehensiveAPI as well such that the basic auth bug would have been a compiler error.

arianvp: I could also add them in my basicAuth PR.  the relevant issue is #465   whatever you want.  You can merge this without that change

phadej: Please add in your PR.

PR #686, opened 2017-01-24
Fixes to tutorial (Client.lhs file)

Tarrasch:
I add two minor commits. Each with commit messages describing the changes.

Also: This library looks amazing! Good job all contributors! :)

### DIFF
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -107,6 +107,7 @@ data BaseUrl = BaseUrl
   { baseUrlScheme :: Scheme -- ^ URI scheme to use
   , baseUrlHost :: String   -- ^ host (eg "haskell.org")
   , baseUrlPort :: Int      -- ^ port (eg 80)
+  , baseUrlPath :: String   -- ^ path (eg "/a/b/c")
   }
 ```
 
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -27,8 +27,8 @@ Also, we need examples for some domain specific data types:
 
 ``` haskell
 data Position = Position
-  { x :: Int
-  , y :: Int
+  { xCoord :: Int
+  , yCoord :: Int
   } deriving (Show, Generic)
 
 instance FromJSON Position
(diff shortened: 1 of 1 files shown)


arianvp:
Looks good to me!

@haskell-servant/core do we currently check the documentation in the CI pipeline?

doc/tutorial/Client.lhs
```diff
@@ -27,8 +27,8 @@ Also, we need examples for some domain specific data types:
 
 ``` haskell
 data Position = Position
-  { x :: Int
-  , y :: Int
+  { xCoord :: Int
```
phadej: Down in the file we print `Position`, and the `x` and `y` field names are in the example output

doc/tutorial/Client.lhs
```diff
@@ -107,6 +107,7 @@ data BaseUrl = BaseUrl
   { baseUrlScheme :: Scheme -- ^ URI scheme to use
   , baseUrlHost :: String   -- ^ host (eg "haskell.org")
   , baseUrlPort :: Int      -- ^ port (eg 80)
+  , baseUrlPath :: String   -- ^ path (eg "/a/b/c")
```
phadej: good catch. @arianvp this is in ignored block, thus not caught by CI.

phadej: ... as well as isn't previous inconsistency. We compile tutorial, but don't run it.

doc/tutorial/Client.lhs
Tarrasch: Nice catch. I'll fix this tomorrow. :)

Tarrasch: Fixed now. I also notice there's an educational point in calling the field name `xCoord` and the `Capture` argument `x`, to show that they don't need to be equal.

doc/tutorial/Client.lhs
Tarrasch: But I suppose you can't turn this into an not ignored block. Because then the subsequent code couldn't disambiguate this `BaseUrl` from the (correct) imported `BaseUrl`. No?

phadej: Yes. One /could/ write tests that type `a` is isomorphic to `b`, but then we'd need to have a qualified annotation somewhere.

phadej:
Thanks!

PR #687, opened 2017-01-25
Bump version to 0.10, preparation for the release

phadej:


alpmestan:
There seems to be a problem with bounds with ghc 7.8.

Other than that, we'll have to:
- update changelogs
- update haddocks if/when appropriate
- update the tutorial

phadej:
the stackage is making lts-8 in less than two weeks, so I'll check the dependencies of servant today (so there isn't obvious problems in changes we introduced), and then go thru changelogs/haddocks/tutorial.

It would be great to make a release on Monday

PR #688, opened 2017-01-28
Make type consistent with whats written on line 20

christiantakle:
https://github.com/haskell-servant/servant/blob/master/doc/tutorial/ApiType.lhs#L20


fizruk:
I think you have to `import Data.Time (UTCTime)`.

Issue #689, opened 2017-01-29
Out-of-the-box configurable server


alpmestan:
This is mostly to act as a summary of a discussion that I had with @dcoutts this week, though if we consider that this proposal makes sense, we can certainly go ahead and implement it in servant-server.

Note that this is in the same alley (but generalizes a fair bit) as @jkarni's https://github.com/haskell-servant/servant/issues/685. Duncan and I thought of the same thing sometimes this week independently to address the same problem.

# Problem

Currently, the behavior of the routing machinery as well as that of all existing combinators is fixed and we have no way whatsoever to tweak any of it. What if we'd like to add some "hooks" to be called here and there instead?

Examples:
- what if we're not happy with the standard error responses and instead would like JSON errors all the time, for all the HTTP errors _thrown by the router_? (see #685)
- what if we'd like to see some debugging output to figure out which endpoint a given request is routed to (see https://github.com/haskell-servant/servant/issues/449)?
- what if we'd like to allow the server to be more lenient with respect to content types, e.g make checks on the request's `Content-Type` header optional, because of some existing clients that don't follow the HTTP RFCs?
- etc

# Solution

Augment the default context with either:
- a bunch of entries for each configurable bit at the toplevel, or
- a named subcontext with a bunch of entries in it, one for each configurable bit, or
- a single context entry of some `ServerCfg` data type, with a field for each configurable bit.

(I'd go for 2, 3, 1, in that order, I think)

By "augment", here, I mean that the `serve*` functions would now require that the context contains at least our server configuration (whatever its final form is), and `serve` would just use some default behavior (e.g the current one).

We could alternatively do lookups in the context and do some standard thing if no "hook" is provided by the user, allowing us to still be able to work with an empty `Context`. I don't like this idea though because the default config would be "implicit", spread accross all the `HasServer` instances, instead of in one place.

It's important to notice that, while this would be a pervasive change, it would provide us with a definitive solution to "I'd like to make combinator X do Y instead of what it currently does", most of the time forcing people to duplicate an existing combinator, only to override the behavior of one bit of the `HasServer` instance. They would only have to find the right config entry to tweak some particular behavior, override it in their code, done.

[...]

PR #690, opened 2017-01-30
Use classes to access context

phadej:
As an example what I mean in #689: We don't need to force which construct is used as `context`. Users should be free to pick their own "extensible-records" library (if they want to use one).

### DIFF
--- a/servant-server/servant-server.cabal
+++ b/servant-server/servant-server.cabal
@@ -59,6 +59,7 @@ library
       , exceptions         >= 0.8  && < 0.9
       , http-api-data      >= 0.3  && < 0.4
       , http-types         >= 0.8  && < 0.10
+      , lens
       , network-uri        >= 2.6  && < 2.7
       , monad-control      >= 1.0.0.4 && < 1.1
       , mtl                >= 2    && < 2.3
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -131,11 +131,11 @@ import           Servant.Utils.Enter
 -- > main :: IO ()
 -- > main = Network.Wai.Handler.Warp.run 8080 app
 --
-serve :: (HasServer api '[]) => Proxy api -> Server api -> Application
-serve p = serveWithContext p EmptyContext
+serve :: (HasServer api ()) => Proxy api -> Server api -> Application
+serve p = serveWithContext p ()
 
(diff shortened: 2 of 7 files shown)


jkarni:
I don't mind the proposal, but I don't also see any huge gain. This is hardly the sort of access pattern that really requires sophisticated record libs (you just construct the datatype once, and never again modify it)

(Presumably we don't want to actually depend on `lens` just for this?)

phadej:
`lens:` well we don't need to. There wasn't setters at all previously. The only lost feature is `Inject`, but I have hard time to come up with generally useful combinator requiring it.

The gain: it's much more approachable (especially if we don't use lens, but only getter functions) to people who aren't comfortable with `HList`s. And for them who are, we don't force them to use our implementation. In our codebase I construct `context` I partially apply to handlers (maybe I should use `Reader` `enter`, but anyway), so to get e.g auth working, i'd just write an instance `HasAuthCheck Ctx` and pass it to `serveWithContext`...

The `enter runReaderTNat` and `serverWithContext context` solve about the same problem, but with very differently. I'm not sure if `MonadReader context (Handler context)` (with `context :: *` as here) is a good idea, but maybe it is.

jkarni:
Hmm, interesting. On the one hand I think configuring handlers and combinators are distinct enough to be kept as separate notions, but on the other I do see how it'd be useful if the same datatype could be used for both so that one ends up with a single configuration datatype. 

I'm overall coming over to this idea, but still have a question. Let's say we have a default config - the one one `serve` uses. We have a datatype for it, and instances for all the classes our base combinator instances need. If there's one more combinator that needs something else in the context, it seems like doing that all of a sudden gets a lot harder. Previously you'd just prepend the new data. Now you need to declare a new datatype, and a whole lot of new instances. This is solved decently well by `makeClassy`, but now we're back to using `lens` (or `microlens`).

phadej:
@jkarni: we could still provide hlist-`Context`, at least for some time. And/or write own TH method for generating boilerplate: if it's as simple "derive servant built-in classes, except ones in this list", it's very trivial. 

So it would be better to have only getters, as then `()` could be default.

phadej:
I'll improve this after 0.10 is out. Showing code will probably help.

alpmestan:
I have a few ideas in that space I'd like us to experiment with, so I agree that we should postpone this to after 0.10.

PR #699, opened 2017-02-12
Issue 698

phadej:


phadej:
Let's see if https://github.com/sol/doctest/pull/151 will be merged, than we could improve the tests (and visible haddock error doctests greatly!)

Issue #715, opened 2017-03-19
linkURI adds unwanted `[]` to specified parameter names for QueryParams


mpilgrem:
`linkURI`, in generating a `URI` from a `Link`, adds `[]` to the end of the specified parameter names for `QueryParams`, which consequently generates URIs that are inconsistent with those generated by the use of `client`.

~~~~
linkURI :: Link -> URI
...
    makeQuery (ArrayElemParam k v) = escape k <> "[]=" <> escape (Text.unpack v)
~~~~

This has been problematic for me, because I am seeking to query an API that encourages (and sometimes requires) the initial request to be digitally signed using HMAC-SHA1 in producing the final request (through the addition of a `signature=<CALCULATED_SIGNATURE>` parameter). I was expecting to be able to use `linkURI` to generate a copy of the relevant part of the URI that I could then sign, not a different URI.

If the behaviour of `linkURI` is fixed, perhaps a function `linkURI' :: Bool -> Link -> URI` could also be provided, with the `Bool` argument providing a switch between adding the `[]` or not.

phadej:
So do I understand right, there is inconsitency between `servant-server`, `servant-client` and `servant` (`Links`) in how they handle `QueryParams`?

AFAICS `Links` and `HasServer` are consistent, aren't they?

phadej:
I see, `HasServer`:

> Lookup the values associated to the sym query string parameter and try to extract it as a value of type [a]. This is typically meant to support query string parameters of the form param[]=val1&param[]=val2 and so on. Note that servant doesn't actually require the []s and will fetch the values just fine with param=val1&param=val2, too.

So is `servant-server` not accepting either one?

mpilgrem:
I've not used the `server` side of `servant`, but, as far as I can see, the 'problem' (from my perspective) is limited to function `linkURI` and the `safeLink` side of things in `Servant.Utils.Links`:

* as you identify, the `server` side is tolerant of incoming requests with `[]` added on to the parameter names specified in the API (`looksLikeParam name = name == paramname || name == (paramname <> "[]")` in the instance for `HasServer`);

* the `client` side of things does not add `[]` on to specified parameter names in making requests (if you want a parameter name to end in `[]`, you have to specify that expressly) (I am looking here at `reqToRequest` in module `Servant.Common.Req`); but

* `linkURI :: Link -> URI` adds `[]` on to specified parameter names. I think `linkURI` is the only way to get an `URI` from an API specification.

[...]

PR #717, opened 2017-03-22
Link accessors

qrilka:
Currenly it's only possible to create `URI` from `network-uri` which e.g. serializes both path and query parameters. With this change it will be possible to construct `URI.ByteString.URIRef` for instance.

### DIFF
--- a/servant/CHANGELOG.md
+++ b/servant/CHANGELOG.md
@@ -1,3 +1,8 @@
+next
+----
+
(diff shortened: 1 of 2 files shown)


alpmestan:
LGTM.

3noch:
This will need to interop with this discussion somehow, but likely not before being merged: https://github.com/haskell-servant/servant/issues/715

phadej:
Superseded by #745

PR #720, opened 2017-03-27
Add HasLink instance for AuthProtect.

mschristiansen:
I needed this instance for routes authenticated with `servant-auth-cookie` and thought I might as well submit it here. Let me know if there is anything I've overlooked and I'll try to fix it.


alpmestan:
LGTM

aaronlevin:
LGTM

PR #721, opened 2017-03-28
Add typeclass default for HasLink.

mschristiansen:
There is quite a few of the servant combinators that don't influence the url generated by `toLink` and it therefore seems reasonable to include a default typeclass implementation that ignores the combinator.

### DIFF
--- a/servant/src/Servant/Utils/Links.hs
+++ b/servant/src/Servant/Utils/Links.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables    #-}
 {-# LANGUAGE TypeFamilies           #-}
 {-# LANGUAGE TypeOperators          #-}
+{-# LANGUAGE UndecidableInstances   #-}
 {-# OPTIONS_HADDOCK not-home        #-}
 
 -- | Type safe generation of internal links.
@@ -176,11 +177,15 @@ safeLink
 safeLink _ endpoint = toLink endpoint (Link mempty mempty)
 
 -- | Construct a toLink for an endpoint.
+--
+-- Default implementation for combinators not used in link e.g. headers.
 class HasLink endpoint where
     type MkLink endpoint
+    type MkLink endpoint = MkLink endpoint
     toLink :: Proxy endpoint -- ^ The API endpoint you would like to point to
          -> Link
          -> MkLink endpoint
+    toLink _ = toLink (Proxy :: Proxy endpoint)
 
 -- Naked symbol instance
 instance (KnownSymbol sym, HasLink sub) => HasLink (sym :> sub) where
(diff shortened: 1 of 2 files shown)


mschristiansen:
Fails only on GHC 7.8.4. Looks like Travis CI error.

mschristiansen:
Restart Travis please @haskell-servant/maintainers

mschristiansen:
@phadej Team mentions possibly not working and Travis CI needing restart as far as I can tell.

arianvp:
I tried restarting is but Travis is acting up

arianvp:
https://www.traviscistatus.com/incidents/93wfcv0ksqrz

servant/src/Servant/Utils/Links.hs
```diff
@@ -176,11 +177,15 @@ safeLink
...
 -- | Construct a toLink for an endpoint.
+--
+-- Default implementation for combinators not used in link e.g. headers.
 class HasLink endpoint where
     type MkLink endpoint
+    type MkLink endpoint = MkLink endpoint
     toLink :: Proxy endpoint -- ^ The API endpoint you would like to point to
          -> Link
          -> MkLink endpoint
+    toLink _ = toLink (Proxy :: Proxy endpoint)
```
phadej: this loops

mschristiansen:
@phadej is right about the loop and that might be what is causing the Travis failure. I spend some time trying to resolve the loop, but can't get the types to align (Any advice?). Closing the PR.

phadej:
Something like
```haskell
type family NextEndpoint (endpoint :: *) :: * where
    NextEndpoint (combinator :> endpoint) = endpoint

class HasLink endpoint where
    type MkLink endpoint
    type MkLink endpoint = MkLink (NextEndpoint endpoint)

    toLink :: Proxy endpoint -- ^ The API endpoint you would like to point to
           -> Link
           -> MkLink endpoint
	default toLink :: HasLink (NextEndpoint endpoint)
	               => Proxy endpoint
				   -> Link
				   -> MkLink (NextEndpoint endpoint)
	toLink _ = toLink (Proxy :: (NextEndpoint endpoint))
```

But I'm still :-1: of adding this functionality. @haskell-servant/maintainers  opinions (FWIW, @mschristiansen , only members of the group (or organisation) can ping it)

alpmestan:
I don't feel strongly about having or not having a default implementation for `HasLink`. Presumably, most (if not everything) of the things that appear in URIs already have a combinator in servant, which means most `HasLink` instances for new combinators will just be dummy ones like the default implementation  above. So maybe I'll just say I'm +0.5 on this.

fizruk:
I am 👎 for the same reasons @phadej has mentioned.

Additionally, I don't think that most of the things that appear in URIs are already in servant. We only have basic tools for that, but it is easy to have a custom type that behaves like, say, `Capture`, but has more structure (restrictions) built into it. What I mean is something like type-level `newtype`ing or other kinds of wrapping for basic combinators.

kosmikus:
I'm also against. Type errors as a reminder for potentially missing cases are valuable.

mschristiansen:
You would still get a "no instance" type error until you explicitly specify using the default.

E.g. "MyCombinator is not instance of HasLink" until adding:

```
instance HasLink sub => HasLink (MyCombinator :> sub)
```

Out of 13 current HasLink instances 5 can use the default implementation and I would guess *all* future combinators (all new auth combinators, multipart file, etc.).

PR #722, opened 2017-04-01
Make servant-foreign build on GHC 8.2

RyanGlScott:
`servant-foreign` currently fails to build with GHC 8.2. This is because the validity checking for associated type family instances [tightened up](https://ghc.haskell.org/trac/ghc/wiki/Migration/8.2?version=19#Associatedtypefamilyinstancesarepickier) a bit, and now require using the same type variables in the associated family instance as in the instead head itself.

Luckily, this amounts to a single-character change.


phadej:
Thanks!

PR #724, opened 2017-04-06
Throw 400 on header parse failure

Philonous:
Similar to #649 and #670, header handling ignores parse errors and just passes Nothing as though the header isn't present. 
This pull requests adds a headersD block to Delayed and throws 400 when header parsing fails.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
+++ b/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
@@ -160,7 +160,9 @@ toApplication ra request respond = ra request routingRespond
 -- 5. Query parameter checks. They require parsing and can cause 400 if the
 -- parsing fails. Query parameter checks provide inputs to the handler
 --
--- 6. Body check. The request body check can cause 400.
+-- 6. Header Checks. They also require parsing and can cause 400 if parsing fails.
+--
+-- 7. Body check. The request body check can cause 400.
 --
 data Delayed env c where
   Delayed :: { capturesD :: env -> DelayedIO captures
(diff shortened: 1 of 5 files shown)


servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -160,6 +160,8 @@ toApplication ra request respond = ra request routingRespond
 -- 5. Query parameter checks. They require parsing and can cause 400 if the
 -- parsing fails. Query parameter checks provide inputs to the handler
 --
+-- 6. Header Checks. They also require parsing and can cause 400 if parsing fails.
+--
 -- 6. Body check. The request body check can cause 400.
```
phadej: -> 7.

Philonous:
Fixed the numbering

Philonous:
The PR has been open for almost a month. Can I do something to help it get merged?

phadej:
It's a breaking change. We could have `0.11` released this/next week (there's at least one breaking change pending merge as well).

Opinions; @haskell-servant/maintainers ?

fizruk:
@phadej I suppose you can always get back to lenient parsing with [`LenientData`](http://hackage.haskell.org/package/http-api-data-0.3.7/docs/Web-HttpApiData.html#g:8).

 👍 from me

alpmestan:
:+1: from me as well.

christian-marie:
I thought merging to master was ok & code freezes / curation & testing are the releaser's responsibility? Our users should have CI too.

Do we want to set up a next major release branch so that maintainers can take the burden off committers for merging incompatible ideas and breaking type changes? I personally appreciate quick upstream merges and transfer of responsibility.

phadej:
We have a policy
> We require two +1 from the maintainers of the repo. If you feel like there has not been a timely response to a PR, you can ping the Maintainers group (with @haskell-servant/maintainers).

And I want to stick to that, IMHO it's very good principle.  E.g. https://github.com/haskell-servant/servant/pull/740 has only my comments. So even it looks good to me, I don't want to merge it.

christian-marie:
Ok I'm talking more about parallelism. I guess this works fine.

alpmestan:
Welcome back @christian-marie :) I don't think we're really crawling under PRs so the current modus operandi should be fine for a while, as the sugestion you made above make the workflow a little more complicated.

PR #727, opened 2017-04-07
Improve Client docs

### LINKED ISSUE #606: URL formation documentation is lacking
I am trying to build some code to access an api for Nomad. I need to send a delete request to 
"/v1/job/<<JOBID>>", the documentation is rather lacking. I think I should be using one of "Capture, ReqBody, QueryParam" or something, but in this page http://haskell-servant.readthedocs.io/en/stable/tutorial/Client.html it is not explained exactly what capture does, or or what "And friends" means.

If there is some more basic documentation I was not able to find it. 

In additon the magic generation of functions `position :<|> hello :<|> marketing = client api` really needs a better explanation 

As a developer who is new to haskell (Though has experience with in Erlang and Elm) I found reading the servant tutorial left me rather confused.

declension:
* Use suggestion in #606 to document pattern matching client behaviour better
 * Helpful suggestion of reading in order in the index
 * A couple of extra internal links
 * Fix some typos / punctuation.

Fixes #606

### DIFF
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -1,12 +1,11 @@
 # Querying an API
 
-While defining handlers that serve an API has a lot to it, querying an API is simpler: we do not care about what happens inside the webserver, we just need to know how to talk to it and get a response back. Except that we usually have to write the querying functions by hand because the structure of the API isn't a first class citizen and can't be inspected to generate a bunch of client-side functions.
+While defining handlers that [serve an API](Server.lhs) has a lot to it, querying an API is simpler: we do not care about what happens inside the webserver, we just need to know how to talk to it and get a response back. Except that we usually have to write the querying functions by hand because the structure of the API isn't a first class citizen and can't be inspected to generate a bunch of client-side functions.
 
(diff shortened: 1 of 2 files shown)


alpmestan:
LGTM

doc/tutorial/Client.lhs
```diff
@@ -1,12 +1,11 @@
 # Querying an API
 
-While defining handlers that serve an API has a lot to it, querying an API is simpler: we do not care about what happens inside the webserver, we just need to know how to talk to it and get a response back. Except that we usually have to write the querying functions by hand because the structure of the API isn't a first class citizen and can't be inspected to generate a bunch of client-side functions.
+While defining handlers that [serve an API](Server.lhs) has a lot to it, querying an API is simpler: we do not care about what happens inside the webserver, we just need to know how to talk to it and get a response back. Except that we usually have to write the querying functions by hand because the structure of the API isn't a first class citizen and can't be inspected to generate a bunch of client-side functions.
```
3noch: Not quite correct grammar: Replace "Except that..." with "The only exception is that..." and it will be correct.

declension: Hmm yes. To me it's not so much wrong but colloquial / chatty in style. Happy to change it though.

How about _That said, we usually have to write.._ (it might not be the _only_ exception), and perhaps changing _a bunch of_ to just _the_?

3noch: Great!

3noch: The sentence had no main subject and verb, hence the reason I said it was not correct grammar (you must have both of those to have a valid sentence).

declension:
Thanks!

Issue #728, opened 2017-04-09
Add QueryParamAll


isairz:
CaptureAll is exist, but QueryParamAll is not.

I suggest *QueryParamAll* to get all list of query params.
```hs
-- /books?title=eq.servant&category=eq.haskell
data QueryParamAll a b
type MyApi = "books" :> QueryParamAll ByteString ByteString :> Get '[JSON] [Book]
books :: [(ByteString, Maybe ByteString)] -> AppHandler ByteString
books filter = ...
```

I think it is not hard to implement this because [QueryParam](https://github.com/haskell-servant/servant/blob/master/servant-server/src/Servant/Server/Internal.hs#L316) alreay do [parseQueryText](https://hackage.haskell.org/package/http-types-0.9.1/docs/Network-HTTP-Types.html#v:parseQueryText)

fizruk:
I think it would be nice if we instead use `ToForm`/`FromForm` from `http-api-data` and parse have a combinator to parse all query parameters using those:

```haskell
type MyApi = "book" :> QueryParamForm BookSearchParams :> Get '[JSON] [Book]
```

Your version of `QueryParamAll` would then just be a special case with

```haskell
type BookSearchParams = HashMap Text Text
```

isairz:
It looks cool!

I think I can create a pull request tomorrow.

PR #729, opened 2017-04-10
QueryParamForm Combinator

isairz:
#728 
```hs
type MyApi = "book" :> QueryParamForm BookSearchParams :> Get '[JSON] [Book]
```

This is my first pull request in servant.
Please feel free to review my commits 😃 

- [x] Add API
- [x] instance and test for servant-server
- [ ] instance and test for servant-client
- [ ] instance and test for servant-doc
- [ ] instance and test for servant-foreign


servant-server/src/Servant/Server/Internal.hs
```diff
@@ -405,6 +406,48 @@ instance (KnownSymbol sym, HasServer api context)
           examine v | v == "true" || v == "1" || v == "" = True
                     | otherwise = False
 
+-- | If you use @'QueryParamForm' BookSearchParams@ in one of the endpoints for your API,
+-- this automatically requires your server-side handler to be a function
+-- that takes an argument of type @['BookSearchParams']@.
```
fizruk: Of type `@'BookSearchParams'@`, right?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -405,6 +406,48 @@ instance (KnownSymbol sym, HasServer api context)
...
+-- to your type by simply providing an instance of 'FromForm' for your type.
+--
+-- Example:
+--
+-- > data BookSearchParams = BookSearchParams
+-- >   { title :: Text
+-- >   { authors :: [Text]
+-- >   , page :: Maybe Int
+-- >   } deriving (Generic)
+-- > instance FromForm BookSearchParams
```
fizruk: We should probably mention that we default implementation uses `FromHttpApiData` for each field.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -405,6 +406,48 @@ instance (KnownSymbol sym, HasServer api context)
...
+      => HasServer (QueryParamForm a :> api) context where
+
+  type ServerT (QueryParamForm a :> api) m =
+    a -> ServerT api m
+
+  route Proxy context subserver = route (Proxy :: Proxy api) context $
+      subserver `addParameterCheck` withRequest paramsCheck
+    where
+      paramsCheck req =
+          case urlDecodeAsForm (BL.drop 1 . BL.fromStrict $ rawQueryString req) of
```
fizruk: I guess `BL.drop 1` is about dropping leading `?`?
Can `rawQueryString` contain extra info (e.g. HTML anchors?), would it ruin parsing?

fizruk:
Also, I don't see any changes for `servant-client`, for that you should symmetrically use `ToForm`.

alpmestan:
@fizruk the `servant-client` interpretation is still not ticked in the first comment in this issue :) I agree with fizruk's feedback, lacking the time to do a proper review but this is a good start! I'll review when it's all ready.

fizruk:
@isairz are you going to work on this soon or should we take over?

isairz:
@fizruk Sorry, I was so busy and did not care.
I will work on this soon.

Issue #698, opened 2017-02-12
Doctest failure


DanBurton:
As seen on the nightly stackage build.

```
> /tmp/stackage-build14/servant-0.10$ dist/build/doctests/doctests
### Failure in /tmp/stackage-build14/servant-0.10/test/Servant/Utils/LinksSpec.hs:84: expression `apiLink (Proxy :: Proxy WrongContentType)'
expected: ...
          ...Could not deduce...
          ...
 but got:
          <interactive>:27:1: error:
              <E2><80><A2> OctetStream expected in list '[JSON]
              <E2><80><A2> In the expression: apiLink (Proxy :: Proxy WrongContentType)
                In an equation for <E2><80><98>it<E2><80><99>:
                    it = apiLink (Proxy :: Proxy WrongContentType)
Examples: 141  Tried: 138  Errors: 0  Failures: 1
```

DanBurton:
Since this looks benign I've simply marked servant as "expected test failure" on stackage until this is resolved.

phadej:
This should be fixed, but let's see after `servant-0.11` reaches the nightly and we enable tests, and they pass.

bergmark:
Passed in the latest nightly build so this can be closed.

PR #730, opened 2017-04-17
Allow aeson-1.2

phadej:


alpmestan:
It seems the CI fails on 8.0.2 because of a different error message GHC gives.

phadej:
yeah, it's related to #698...

Issue #734, opened 2017-04-28
Instances for Enter are secretly overlapping


fizruk:
Turns out these two instances secretly overlap:

```haskell
instance (Enter b arg ret) => Enter (a -> b) arg (a -> ret)
instance Enter (m a) (m :~> n) (n a)
```

It happens only when `m ~ n` and I suspect that's why no one noticed that so far.

In the "real world" this results in non-descriptive type errors like this:

```
    • Couldn't match type ‘App’ with ‘(->) (Maybe Param)’
        arising from a use of ‘enter’
    • In the expression: enter (Nat id :: App :~> App) server
      In an equation for ...
```

This prevents me from applying `App :~> App` transformation to handlers directly.
Currently I work around this by using a pair of transformation:

```haskell
Nat IdentityT :: App :~> IdentityT App

Nat (f . runIdentityT) :: IdentityT App :~> App
```

instead of just

```haskell
Nat f :: App :~> App
```

Unfortunately I don't have a minimal failing test case at this moment. For instance, this works just fine:

```
>>> :set -XDataKinds
>>> :set -XTypeOperators
>>> import Servant
>>> type Endpoint1 = QueryParam "param" Int :> Get '[JSON] NoContent
>>> type Endpoint2 = QueryParam "param" String :> Get '[JSON] NoContent
>>> type API = Endpoint1 :<|> Endpoint2
>>> data App a
>>> type S layout = ServerT layout App
>>> server :: S API; server = undefined
>>> :t enter (Nat id :: App :~> App) server
enter (Nat id :: App :~> App) server
  :: (Maybe Int -> App NoContent)
     :<|> (Maybe [Char] -> App NoContent)
```

Perhaps there's something extra in my project's API that prevents GHC to infer types correctly.

phadej:
Hmm the fact that pair of transformations works kind of rules out that you have non `App` endpoint impl. Weird.

fizruk:
Apparently it works differently in REPL.
Here's a minimal example:

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeOperators #-}

import Servant

data App a

f :: App :~> App
f = Nat id

server :: App Int :<|> (String -> App Bool)
server = undefined

server' :: App Int :<|> (String -> App Bool)
server' = enter f server
```

Here's what happens if you try load it in GHCi:

```
>>> :l ~/temp/servant-enter.hs
[...]

PR #736, opened 2017-04-28
Redo Enter

phadej:
"Fixes" #734

### DIFF
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -207,7 +207,7 @@ layoutWithContext p context =
 -- monad. Or have your types ensure that your handlers don't do any IO. Enter
 -- `enter`.
 --
--- With `enter`, you can provide a function, wrapped in the `(:~>)` / `Nat`
+-- With `enter`, you can provide a function, wrapped in the `(:~>)` / `NT`
 -- newtype, to convert any number of endpoints from one type constructor to
 -- another. For example
 --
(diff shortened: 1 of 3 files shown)


fizruk:
Nice!

Can we also add some custom `TypeError`s here?
Otherwise I except there to be some unreadable type errors.

Can we interpret `Enter` as a closed function (type-level to term-level) and disallow any other instances (filling the holes with custom type errors)?

phadej:
I guess we can drop the `ret` from `Enter` definition; and have a `TypeError` in `Entered`.
Via constraints the GHC should be able to reason "back" `ret -> typ` anyway.

alpmestan:
I still think `Nat` is the wrong thing to use there. Enter should really just be about walking down `:<|>`-separated stuffs, idealy. But +1 as this improves on the status quo.

fizruk:
@alpmestan where's the prior discussion about `:~>` and why we generalised over it?
Can you link it here?

phadej:
I guess it may work with generic `transform`, not `m :~> n`, but then we have to specify that it can be inverted (that what `enter`'s functional dependencies force us to do). Not sure how the lack of injectivity (if we don't' want to specify the inverse) would work.

alpmestan:
> @alpmestan where's the prior discussion about :~> and why we generalised over it?
Can you link it here?

@fizruk I'm not sure there ever was a discussion about generalizing the kind of transformation we can apply, as the original motivation for `Enter` was to go from a monad to another. It's just been on my personal wishlist for years. I think `Enter` should allow us to apply any transformation "that can be applied on the given chain of `:<|>`-separated stuffs", whatever those `:<|>`-separated stuffs are. I ideally would like a solution such that "if the types align, we should be able to do it". But I don't want to block this PR on that, just using the occasion to maybe pull you guys down that rabbit hole with me =)

phadej:
@alpmestan in this case we have to fuse old `Enter` with `m :~> n` so we can more specifically write
"if you have `typ` api, you'll get `ret` api". We could use defunctionalization to make `Enter` generic, but that is far from elegant (IMHO).

phadej:
@fizruk how urgent this is for you; I'd like to release 0.11 only after ZuriHac + few weeks; as then there
might be other major stuff in (e.g. `Lenient` params)

fizruk:
@phadej not urgent, I have a workaround in place for now.

Issue #733, opened 2017-04-26
http library independent servant-client


phadej:
See #731 for explanation of one approach (thru `MonadHTTP`).

Successful outcome would give us:
- servant-client
- servant-http-client
- servant-http-streams

Having two client "backends" is essential, to verify that generalisation actually works (and not only adds one indirection layer).

gmalecha:
Either version of the proposed `HasClient` changes seems reasonable to me. The one @phadej proposed is a bit simpler so it be preferred. It might also mirror servant-server a bit more.

What is the plan for implementing? Should I work on a PR?

phadej:
**ZuriHac note:** This is expert task, but if someone wants to start tacking this, it would be great!

abailly:
Some work towards this in #800

jkarni:
Done in #803

PR #740, opened 2017-05-05
Include the request that failed in FailureResponse.

stevana:
Sometimes I find it hard to figure out what API call caused the error that servant-client throws, this PR fixes that by including the request in the error.

### DIFF
--- a/servant-client/src/Servant/Common/Req.hs
+++ b/servant-client/src/Servant/Common/Req.hs
@@ -48,7 +48,8 @@ import Web.HttpApiData
 
 data ServantError
   = FailureResponse
-    { responseStatus            :: Status
+    { failingRequest            :: Request
+    , responseStatus            :: Status
     , responseContentType       :: MediaType
     , responseBody              :: ByteString
     }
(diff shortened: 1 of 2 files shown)


phadej:
I'd rather include the servant-client's `Req`, not `http-client` `Req`. That's important for #733

servant-client/src/Servant/Common/Req.hs
```diff
@@ -71,7 +72,7 @@ data ServantError
   deriving (Show, Typeable)
 
 instance Eq ServantError where
-  FailureResponse a b c == FailureResponse x y z =
+  FailureResponse _ a b c == FailureResponse _ x y z =
```
phadej: `Req` can be made an instance of `Eq`.

stevana: Hmm, `Req` contains a `RequestBody` which doesn't have an `Eq` instance...

Also, I'm not sure one wants to include the request in the failure response equality check -- two responses can be equal even if they didn't originate from the same request.

stevana:
I made a change to use `BaseUrl` and `Req` instead of `Request`.

phadej:
@haskell-servant/maintainers any objections?

alpmestan:
LGTM, if users find this to be more helpful than how it was before I'm all for it.

PR #748, opened 2017-05-15
ServerT Raw m = Tagged m Application

phadej:
Supersedes #282

ping @avieth

By having `ServerT Raw m = Tagged m Application` the solution is quite generic. I'll see how if it can be applied to `servant-client` too.

### DIFF
--- a/doc/tutorial/Docs.lhs
+++ b/doc/tutorial/Docs.lhs
@@ -220,12 +220,10 @@ api :: Proxy DocsAPI
 api = Proxy
 
 server :: Server DocsAPI
-server = Server.server3 :<|> serveDocs
-
-  where serveDocs _ respond =
-          respond $ responseLBS ok200 [plain] docsBS
-
-        plain = ("Content-Type", "text/plain")
+server = Server.server3 :<|> Tagged serveDocs where
+    serveDocs _ respond =
+        respond $ responseLBS ok200 [plain] docsBS
+    plain = ("Content-Type", "text/plain")
 
 app :: Application
 app = serve api server
--- a/servant/src/Servant/Utils/Enter.hs
+++ b/servant/src/Servant/Utils/Enter.hs
@@ -22,9 +22,9 @@ import qualified Control.Monad.State.Lazy    as LState
 import qualified Control.Monad.State.Strict  as SState
 import qualified Control.Monad.Writer.Lazy   as LWriter
 import qualified Control.Monad.Writer.Strict as SWriter
+import           Data.Tagged                 (Tagged, retag)
 import           Prelude                     ()
 import           Prelude.Compat
-
 import           Servant.API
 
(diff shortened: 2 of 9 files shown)


avieth:
> By having ServerT Raw m = Tagged m Application the solution is quite generic.

That's the same as what's in #282 so LGTM. Thanks for saving me from the rebase :)

jkarni:
:+1: from me too. Does this require a type annotation when used with `enter` (when `m` *should* be deducible from the other handlers)?

phadej:
@jkarni I removed the deducibility from `res` in ``Enter` previously while fixing https://github.com/haskell-servant/servant/issues/734

The overlap happens when `m` = `(->) x`. So now we have the `Entered` type-family and constraint instances allowing to use `(->) a` in `Enter`.

doc/tutorial/Docs.lhs
```diff
@@ -220,12 +220,10 @@ api :: Proxy DocsAPI
 api = Proxy
 
 server :: Server DocsAPI
-server = Server.server3 :<|> serveDocs
-
-  where serveDocs _ respond =
-          respond $ responseLBS ok200 [plain] docsBS
-
-        plain = ("Content-Type", "text/plain")
+server = Server.server3 :<|> Tagged serveDocs where
```
alpmestan: It's a bit ugly to have `Tagged` appear here, but I don't have a better idea. If you've got one, I'm all ears.

phadej: I guess it's price to pay; the `type family Tag m a where Tag m a = a` won't work, as it will reduce too early, so `Enter` couldn't match it (and we cannot have type family application in the LHS of `Entered` or in `instance` declaration).

Now as I think, the `Entered` is probably not required; yet it might help generate better error messages, I should test a bit.

servant/src/Servant/Utils/Enter.hs
```diff
@@ -22,20 +22,21 @@ import qualified Control.Monad.State.Lazy    as LState
...
     Entered m n (m a)            = n a
     Entered m n (a -> api)       = a -> Entered m n api
     Entered m n (api1 :<|> api2) = Entered m n api1 :<|> Entered m n api2
+    Entered m n (Tagged m a)     = Tagged n a
 
 class ( Entered m n typ ~ ret
       , Entered n m ret ~ typ
       )
-    => Enter typ m n ret
+    => Enter typ m n ret | typ m n -> ret, ret m n -> typ
```
jkarni: Don't *any* three determine the fourth?

phadej: Seems to be, good catch!

phadej:
@jkarni about type synonym: Do you have any in mind? `RawServer` would make sense in `servant-server` context; but I'm not sure if it really helps.

PR #753, opened 2017-05-16
Add a type representing an empty API

DaveCTurner:
This is for where you want to compose an API out of bits that come from various sources. For example, we have a Servant API, `CommonAPI`, and a function that has a type rather like this:

    defaultMain :: HasServer specialAPI => Proxy specialAPI -> Server specialAPI -> ... -> IO a

the purpose of which is to serve `CommonAPI :<|> specialAPI`, plus a bunch of other common setup/teardown code. Most of the time there's interesting stuff in `specialAPI`, but in one or two cases there's nothing much to put there.

Morally speaking, this would be a unit for the `:<|>` operator (up to isomorphism).

This issue follows a [discussion on the mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/haskell-servant/J76EcFgr8m8/Rahw0sriBAAJ) concluding with @alpmestan saying:

> OK I think I'm convinced enough to give my +1 for a suitable PR that implements this all properly. Thank you guys for taking the time to make the case. Bonus points if the PR augments the tutorial with a section about this that documents the workflow of using this for "not implemented yet" situations.

### DIFF
--- a/doc/tutorial/ApiType.lhs
+++ b/doc/tutorial/ApiType.lhs
@@ -326,6 +326,16 @@ type ProtectedAPI11
  :<|> BasicAuth "my-realm" User :> UserAPI2 -- this is protected by auth
 ```
 
+### Empty APIs
+
+TODO motivation...
+
+``` haskell ignore
+type UserAPI12
+     = UserAPI
+ :<|> EmptyAPI -- this adds nothing to the API
+```
+
 ### Interoperability with `wai`: `Raw`
 
 Finally, we also include a combinator named `Raw` that provides an escape hatch
--- a/doc/tutorial/ApiType.lhs
+++ b/doc/tutorial/ApiType.lhs
@@ -321,7 +321,7 @@ data BasicAuth (realm :: Symbol) (userData :: *)
 Which is used like so:
 
 ``` haskell
-type ProtectedAPI12
+type ProtectedAPI11
      = UserAPI                              -- this is public
  :<|> BasicAuth "my-realm" User :> UserAPI2 -- this is protected by auth
 ```
@@ -334,7 +334,7 @@ you want to plug a [wai `Application`](http://hackage.haskell.org/package/wai)
 into your webservice:
 
 ``` haskell
-type UserAPI11 = "users" :> Get '[JSON] [User]
+type UserAPI12 = "users" :> Get '[JSON] [User]
                  -- a /users endpoint
 
             :<|> Raw
--- a/doc/tutorial/ApiType.lhs
+++ b/doc/tutorial/ApiType.lhs
@@ -334,7 +334,7 @@ you want to plug a [wai `Application`](http://hackage.haskell.org/package/wai)
 into your webservice:
 
 ``` haskell
-type UserAPI12 = "users" :> Get '[JSON] [User]
+type UserAPI13 = "users" :> Get '[JSON] [User]
                  -- a /users endpoint
 
             :<|> Raw
--- a/doc/tutorial/ApiType.lhs
+++ b/doc/tutorial/ApiType.lhs
@@ -345,7 +345,8 @@ type UserAPI12Alone = UserAPI12 EmptyAPI
 ```
 
 This also works well as a placeholder for unfinished parts of an API while it
-is under development.
+is under development, for when you know that there should be _something_ there
+but you don't yet know what. Think of it as similar to the unit type `()`.
 
 ### Interoperability with `wai`: `Raw`
 
--- a/doc/tutorial/ApiType.lhs
+++ b/doc/tutorial/ApiType.lhs
@@ -328,13 +328,24 @@ type ProtectedAPI11
 
 ### Empty APIs
 
-TODO motivation...
+Sometimes it is useful to be able to generalise an API over the type of some
+part of it:
 
-``` haskell ignore
-type UserAPI12
-     = UserAPI
- :<|> EmptyAPI -- this adds nothing to the API
+``` haskell
+type UserAPI12 innerAPI
+     = UserAPI             -- this is the fixed bit of the API
+ :<|> "inner" :> innerAPI  -- this lets us put various other APIs under /inner
 ```
+
+If there is a case where you do not have anything extra to serve, you can use
+the `EmptyAPI` combinator to indicate this:
+
+``` haskell
+type UserAPI12Alone = UserAPI12 EmptyAPI
+```
+
+This also works well as a placeholder for unfinished parts of an API while it
+is under development.
 
 ### Interoperability with `wai`: `Raw`
 
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -77,8 +77,6 @@ hello :: Maybe String -- ^ an optional value for "name"
 
 marketing :: ClientInfo -- ^ value for the request body
           -> ClientM Email
-
-emptyClient :: EmptyAPIClient
 ```
 
 Each function makes available as an argument any value that the response may
@@ -91,7 +89,7 @@ the function `client`. It takes one argument:
 api :: Proxy API
 api = Proxy
 
-position :<|> hello :<|> marketing :<|> emptyClient = client api
+position :<|> hello :<|> marketing :<|> EmptyClient = client api
 ```
 
 `client api` returns client functions for our _entire_ API, combined with `:<|>`, which we can pattern match on as above. You could say `client` "calculates" the correct type and number of client functions for the API type it is given (via a `Proxy`), as well as their implementations.
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -62,6 +62,7 @@ Enough chitchat, let's see an example. Consider the following API type from the 
 type API = "position" :> Capture "x" Int :> Capture "y" Int :> Get '[JSON] Position
       :<|> "hello" :> QueryParam "name" String :> Get '[JSON] HelloMessage
       :<|> "marketing" :> ReqBody '[JSON] ClientInfo :> Post '[JSON] Email
+      :<|> EmptyAPI
 ```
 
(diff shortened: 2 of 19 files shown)


DaveCTurner:
#fa3f186 is a very rough first draft to gather feedback on naming/style/etc. I've coarsely shoehorned some examples into the tutorial; my intention is to do something more elegant once the bikeshedding has died down.

DaveCTurner:
Updated the test suites to match

servant-server/src/Servant/Server.hs
```diff
@@ -17,6 +17,8 @@ module Servant.Server
   , -- * Handlers for all standard combinators
     HasServer(..)
   , Server
+  , EmptyAPIServer
```
phadej: I'd name these just
```haskell
EmptyServer
emptyServer
emptyClient
EmptyClient
```

Not sure whether `EmptyAPI` should also be just `Empty`, we already have clash-y names in api-type vocabulary, so I'm :+1: to dropping `API` from the names.

servant/src/Servant/API/Empty.hs
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# OPTIONS_HADDOCK not-home    #-}
+module Servant.API.Empty(EmptyAPI(..)) where
+
+import           Data.Typeable    (Typeable)
+import           Prelude ()
+import           Prelude.Compat
+
+-- | An empty API: one which serves nothing.
```
phadej: could mention that `EmptyAPI` (or `Empty`) is a unit of `:<|>`.

servant-server/src/Servant/Server.hs
phadej: @haskell-servant/maintainers opinions?

phadej:
Also have to remember to add a test for `Enter`.

servant-server/src/Servant/Server.hs
kosmikus: Dropping API from the other names sounds good to me; but I think I'd keep `EmptyAPI` over `Empty`, as it is more explicit.

servant/src/Servant/API/Empty.hs
kosmikus: It's only "morally" the unit, though, right? It's still very much observable whether `EmptyAPI` is used or not.

phadej: @kosmikus somwhere, e.g in `docs` and `servant-swagger` it will be actually a unit.

I want something to guide the implementors of other interpretations of API types to treat `EmptyAPI` as unit as possible.

phadej: .. which reminds me, that `safeLink` should probably fail if pointed into `EmptyAPI`.

kosmikus: Yes, I'm happy to say that interpretations should treat it as the unit of `:<|>` as much as possible.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -532,6 +532,16 @@ instance HasServer api context => HasServer (HttpVersion :> api) context where
   route Proxy context subserver =
     route (Proxy :: Proxy api) context (passToServer subserver httpVersion)
 
+data EmptyAPIServer = EmptyAPIServer
```
phadej: we probably want `Eq` and `Show` instances (at least), `:<|>` have them too.

Same classes as in `EmptyAPI`?

servant/src/Servant/API/Empty.hs
```diff
@@ -0,0 +1,10 @@
...
+data EmptyAPI = EmptyAPI deriving (Typeable, Eq, Show, Bounded)
```
phadej: I'd derive `Enum` too.

doc/tutorial/Client.lhs
```diff
@@ -88,7 +91,7 @@ the function `client`. It takes one argument:
 api :: Proxy API
 api = Proxy
 
-position :<|> hello :<|> marketing = client api
+position :<|> hello :<|> marketing :<|> emptyClient = client api
```
phadej: I'd rather write `EmptyAPIClient` (or `EmptyClient` after rename) directly here and elaborate that you can do so (and it's safe) because it's a always succeeding irrefutable pattern match like:

```haskell
let (x, ()) = (42, ())
```

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -142,7 +145,8 @@ getGet
   :<|> getRawFailure
   :<|> getMultiple
   :<|> getRespHeaders
-  :<|> getDeleteContentType = client api
+  :<|> getDeleteContentType
+  :<|> EmptyAPIClient = client api
```
phadej: :+1: would use this idiom in docs too.

DaveCTurner:
Great, thanks for all the feedback. All seem like good points. I should be able to get on with them tonight.

phadej:
LGTM. Happy to include into `0.11`. Does other @haskell-servant/maintainers have any comments?

DaveCTurner:
Hang on - two more things:

1. I haven't done anything with `Enter` because, frankly, I don't understand it. Can someone else help?
2. The changes to the tutorial needs a lot of polish.

phadej:
@DaveCTurner there is a doctest in servant-server/src/Servant/Server.hs, probably if you add 

```diff
--- >>> type ReaderAPI = "ep1" :> Get '[JSON] Int :<|> "ep2" :> Get '[JSON] String :<|> Raw
--- >>> let readerServer = return 1797 :<|> ask :<|> Tagged (error "raw server") :: ServerT ReaderAPI (Reader String)
+-- >>> type ReaderAPI = "ep1" :> Get '[JSON] Int :<|> "ep2" :> Get '[JSON] String :<|> Raw :<|> EmptyAPI
+-- >>> let readerServer = return 1797 :<|> ask :<|> Tagged (error "raw server") :<|> emptyServer :: ServerT ReaderAPI (Reader String)
```

the tests should pass.

And good point. Let's not forget the tutorial.

servant-server/src/Servant/Server.hs
alpmestan: I agree with Andres.

servant/src/Servant/API/Empty.hs
alpmestan: This has always been "a problem" with servant: terms in our DSL are just symbolic, so whenever we talk about their semantics, we're really talking about the semantics of their various interpretations.

alpmestan:
When do you plan on releasing 0.11 @phadej ? I'd be happy to ship this PR once polished (especially once the changes to the tutorial are done, it's important to document this well), not sure whether it's worth holding the next release over though, depending on how urgent the said release is and on when David will have time to polish everything.

phadej:
I hope that we'll get contributions in ZuriHac which will force us to do another major bump anyway. So I could release `0.11` this week, and `0.12` after the things are settled after ZuriHac. 

I guess this is better than having `0.11` after `ZuriHac`.

DaveCTurner:
Ok, I've added the `Enter` test as described (and fixed the bug it found - the type of `emptyServer` wasn't general enough) and put something a bit more helpful into the tutorial.

Probably wants squashing down to one commit before merging, but apart from that I think I'm done.

doc/tutorial/ApiType.lhs
```diff
@@ -321,11 +321,32 @@ data BasicAuth (realm :: Symbol) (userData :: *)
...
+
+If there is a case where you do not have anything extra to serve, you can use
+the `EmptyAPI` combinator to indicate this:
+
+``` haskell
+type UserAPI12Alone = UserAPI12 EmptyAPI
+```
+
+This also works well as a placeholder for unfinished parts of an API while it
+is under development.
```
alpmestan: To give a concrete idea, it might be nice to mention that this is somewhat equivalent to how we use `undefined` (yeah, bad bad bad, I know =P) or `()` or what have you when we need to fill in an expression but don't want to worry about the actual value of it until later.

doc/tutorial/Server.lhs
```diff
@@ -1020,6 +1020,16 @@ serverFor = error "..."
 -- or the mailing list if you get stuck!
 ```
 
+If the API contains the `EmptyAPI` combinator, the corresponding server is
+called `emptyServer`:
```
alpmestan: Could you describe the behavior of `emptyServer` in a few words? Maybe a better wording overall would be:

> When your API contains the `EmptyAPI` combinator, you'll want to use `emptyServer` in the corresponding slot for your server, which will simply fail with 404 whenever a request reaches it.

?

doc/tutorial/Client.lhs
```diff
@@ -93,6 +93,18 @@ position :<|> hello :<|> marketing = client api
 
 `client api` returns client functions for our _entire_ API, combined with `:<|>`, which we can pattern match on as above. You could say `client` "calculates" the correct type and number of client functions for the API type it is given (via a `Proxy`), as well as their implementations.
 
+If there is an `EmptyAPI` within your API, this matches the `EmptyClient`
```
alpmestan: I think it might be nice to clarify things a bit more here as well. Something like:

> If you have an `EmptyAPI` in your API, servant-client will hand you a value of type `EmptyClient` in the corresponding slot, where `data EmptyClient = EmptyClient`, as a way to indicate that you can't do anything useful with it.

?

DaveCTurner:
Nitpicking is more than welcome!

I've made the changes requested. The mention of `EmptyAPI` in the pages about documentation and Javascript functions feel a little awkward - that's why I left them out before. If you've a better idea then let's hear it.

I also avoided mentioning `undefined` - it's more like `()` so I said that.

alpmestan:
I'm happy with the current state of this PR. @phadej feel free to merge it when you're happy with it as well.

phadej:
This will make it into `0.11`, great!

Issue #754, opened 2017-05-16
servant-docs implicitly drops ReqBody samples


osa1:
`docs` in `servant-docs` drops samples other than the first one for `ReqBody`. As far as I can see, this is not documented, and implementation ignores `DocOptions`.

We could do something like this:

```diff
diff --git a/servant-docs/src/Servant/Docs/Internal.hs b/servant-docs/src/Servant/Docs/Internal.hs
index 2884473..83f8c2b 100644
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -801,11 +801,13 @@ instance HasDocs Raw where
 instance (ToSample a, AllMimeRender (ct ': cts) a, HasDocs api)
       => HasDocs (ReqBody (ct ': cts) a :> api) where
 
-  docsFor Proxy (endpoint, action) =
-    docsFor subApiP (endpoint, action')
+  docsFor Proxy (endpoint, action) opts@DocOptions{..} =
+    docsFor subApiP (endpoint, action') opts
 
     where subApiP = Proxy :: Proxy api
-          action' = action & rqbody .~ sampleByteString t p
+          action' :: Action
+          action' = action & rqbody .~ take _maxSamples
+                               (map (\(_,mt,bs) -> (mt,bs)) (sampleByteStrings t p))
                            & rqtypes .~ allMime t
           t = Proxy :: Proxy (ct ': cts)
           p = Proxy :: Proxy a
```

But this still ignores sample names (which is important in my case).

If anyone has better ideas for this I can implement this myself. I'd really like to see this solved.

alpmestan:
Please feel free to put any change in place that will fix this admittedly buggy behavior.

osa1:
OK, I'll open a pull request sometime later today.

alpmestan:
PR merged.

PR #755, opened 2017-05-16
Don't drop samples in `HasDocs ReqBody` instance

osa1:
As reported in #754, `HasDocs` instance of `ReqBody` was dropping samples other
than the first one. With this patch we show at most `_maxSamples` samples for
`ReqBody`, and also include the sample title in the docs.


alpmestan:
LGTM. @haskell-servant/maintainers ?

ivan-m:
If this is for 0.11, why was it merged and released as 0.10.0.1? (cf #784)

PR #757, opened 2017-05-20
Replace `User` with `Account` in Generalized Authentication section 

justanotherdot:
I may be mistaken here but when reading the Generalized Authentication section of the docs I noticed some discrepancies, mainly that `User` was referenced in the enclosing text but `Account` is in the actual code. 

I also may be wrong about the `Request -> Handler req` with using `req` but I assumed that `req` was a type variable referencing whatever datatype is expected. If needed, I can update that to reflect `Account`, as well.

And thanks for the awesome library!


alpmestan:
@aaronlevin looks good to you?

doc/tutorial/Authentication.lhs
```diff
@@ -246,7 +246,7 @@ your feedback!
 ### What is Generalized Authentication?
 
 **TL;DR**: you throw a tagged `AuthProtect` combinator in front of the endpoints
-you want protected and then supply a function `Request -> Handler user`
```
jkarni: It seems like `user` is *less* misleading, no? `req` to me means "request"

justanotherdot: Ah right, my bad, as it's a `Handler` of some 'application-specific datatype'. Would it be better to say something like `Request -> Handler a` or use lowercase `account` to keep in line with the subsequent example?

jkarni: `a` sounds best to me, but maybe all of this indicates that we could use another sentence of explanation (something like "where `a` is a type of your choice representing the data returned by successful authentication - e.g., a `User` or, in our example below, `Account`")

justanotherdot: I was thinking the same (for the latter sentence) otherwise people would have to read on to connect the two. I think your copy is quite apropos, so I'll tack that on.

justanotherdot:
Forgot to merge in master so I think that's why Travis died. Cool if I push another commit merging in master to fix?

justanotherdot:
@jkarni any idea why the 8.2.1 CI build would fail on doc changes? Perhaps it's temporary, but happy to change anything necessary. :slightly_smiling_face:

justanotherdot:
Seems to have been a glitch from the point in time I was working from. Merged in a squashed copy of master to fix.

justanotherdot:
Seems like it was a temp glitch. I've merged in a squashed version of master to fix it.

justanotherdot:
Would someone mind taking a look at this? :slightly_smiling_face:  @phadej @alpmestan @jkarni ?

alpmestan:
Thanks! Sorry it took so long.

justanotherdot:
No worries, glad I can contribute back somehow!

PR #762, opened 2017-06-02
Lower infix precedence for (:>) and (:<|>)

### LINKED ISSUE #761: Lower infix precedence for Sub (:>)
There's a discussion about parameter descriptions in https://github.com/haskell-servant/servant-swagger/issues/62. It's becoming obvious that there needs to be a support for them in servant. However, it can take a while before we solve design problems for it.

In any case, it is highly probable that we want to use infix operator to attach descriptions:

```haskell
-- style I like
type FooEndpoint
  = "foo"
 :> Capture "foo-id" FooId      --^ "The identifier"
 :> QueryParam "canonical" Bool --^ "Normalise the representation"
 :> Get '[JSON] Foo             --^ "Why not annotate the result too"
```

```haskell
-- style @phadej proposed
type FooEndpoint =
   "foo" :>
   Capture "foo-id" FooId      :? "The identifier" :>
   QueryParam "canonical" Bool :? "Normalise the representation" :>
   Get '[JSON] Foo             :? "Why not annotate the result too"
```

Unfortunately, none of these is currently possible to implement (even locally), because

```haskell
infixr 9 :>
```

9 is the highest available priority and I guess it should never be used in a library.
So I propose we decrease it to at least 8.

fierce-katie:
Closes #761.

### DIFF
--- a/servant-server/test/Servant/Server/Internal/ContextSpec.hs
+++ b/servant-server/test/Servant/Server/Internal/ContextSpec.hs
@@ -37,7 +37,7 @@ spec = do
 
         it "works with operators" $ do
           let cxt = ((1 :: Integer) :. 'a' :. EmptyContext) :<|> ('b' :. True :. EmptyContext)
-          show cxt `shouldBe` "(1 :. 'a' :. EmptyContext) :<|> ('b' :. True :. EmptyContext)"
+          show cxt `shouldBe` "1 :. 'a' :. EmptyContext :<|> 'b' :. True :. EmptyContext"
 
   describe "descendIntoNamedContext" $ do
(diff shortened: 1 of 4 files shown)


fizruk:
👍

alpmestan:
LGTM.

servant-server/test/Servant/Server/Internal/ContextSpec.hs
```diff
@@ -37,7 +37,7 @@ spec = do
 
         it "works with operators" $ do
           let cxt = ((1 :: Integer) :. 'a' :. EmptyContext) :<|> ('b' :. True :. EmptyContext)
```
phadej: I'd remove the parentheses there as well.

Issue #763, opened 2017-06-06
servant-foreign: comments for client functions


fierce-katie:
There's a discussion on `Description` combinator (https://github.com/haskell-servant/servant-swagger/issues/60). When adding `HasForeign` instance for it, we thought it would be reasonable to use this description as a comment for generated client function instead of just ignoring it. AFAIK currently there's no support for generating comments in foreign clients.

PR #767, opened 2017-06-08
Description and Summary combinators

fierce-katie:
This PR adds combinators discussed in https://github.com/haskell-servant/servant-swagger/issues/60. They allow you to add documentation for API endpoints right in the type declaration.

Both `Description` and `Summary` are needed for different `HasSwagger` instances:

```haskell
instance (KnownSymbol desc, HasSwagger api) => HasSwagger (Description desc :> api) where
  toSwagger _ = toSwagger (Proxy @api)
    & allOperations.description %~ (Just (Text.pack (symbolVal (Proxy @desc))) <>)

instance (KnownSymbol desc, HasSwagger api) => HasSwagger (Summary desc :> api) where
  toSwagger _ = toSwagger (Proxy @api)
    & allOperations.summary %~ (Just (Text.pack (symbolVal (Proxy @desc))) <>)

```

### DIFF
--- a/servant-foreign/src/Servant/Foreign/Internal.hs
+++ b/servant-foreign/src/Servant/Foreign/Internal.hs
@@ -350,6 +350,20 @@ instance HasForeign lang ftype api
   foreignFor lang ftype Proxy req =
     foreignFor lang ftype (Proxy :: Proxy api) req
 
+instance HasForeign lang ftype api
+  => HasForeign lang ftype (Summary desc :> api) where
+  type Foreign ftype (Summary desc :> api) = Foreign ftype api
+
+  foreignFor lang ftype Proxy req =
+    foreignFor lang ftype (Proxy :: Proxy api) req
+
+instance HasForeign lang ftype api
+  => HasForeign lang ftype (Description desc :> api) where
+  type Foreign ftype (Description desc :> api) = Foreign ftype api
+
+  foreignFor lang ftype Proxy req =
+    foreignFor lang ftype (Proxy :: Proxy api) req
+
 -- | Utility class used by 'listFromAPI' which computes
 --   the data needed to generate a function for each endpoint
 --   and hands it all back in a list.
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -58,7 +58,8 @@ import           Servant.API                 ((:<|>) (..), (:>), BasicAuth, Capt
                                               IsSecure(..), Header, QueryFlag,
                                               QueryParam, QueryParams, Raw,
                                               RemoteHost, ReqBody, Vault,
-                                              WithNamedContext)
+                                              WithNamedContext,
+                                              Description, Summary)
 import           Servant.API.ContentTypes    (AcceptHeader (..),
                                               AllCTRender (..),
                                               AllCTUnrender (..),
@@ -532,6 +533,18 @@ instance HasServer api context => HasServer (HttpVersion :> api) context where
 
   route Proxy context subserver =
     route (Proxy :: Proxy api) context (passToServer subserver httpVersion)
+
+-- | Ignore @'Summary'@ in server handlers.
+instance HasServer api ctx => HasServer (Summary desc :> api) ctx where
+  type ServerT (Summary desc :> api) m = ServerT api m
+
+  route _ = route (Proxy :: Proxy api)
+
+-- | Ignore @'Description'@ in server handlers.
+instance HasServer api ctx => HasServer (Description desc :> api) ctx where
+  type ServerT (Description desc :> api) m = ServerT api m
+
+  route _ = route (Proxy :: Proxy api)
 
 -- | Singleton type representing a server that serves an empty API.
 data EmptyServer = EmptyServer deriving (Typeable, Eq, Show, Bounded, Enum)
--- a/servant/src/Servant/API/Description.hs
+++ b/servant/src/Servant/API/Description.hs
@@ -18,7 +18,14 @@ data Summary (sym :: Symbol)
 --
 -- Example:
 --
--- >>> type MyApi = Description "Some longer implementation details here." :> "books" :> Capture "isbn" Text :> Get '[JSON] Book
+-- >>> :{
+--type MyApi = Description
+--  "This comment is visible in multiple Servant interpretations \
+--  \and can be really long if necessary. \
+--  \Haskell multiline support is not perfect \
+--  \but it's still very readable."
+-- :> Get '[JSON] Book
+-- :}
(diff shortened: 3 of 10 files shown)


fizruk:
Build fails for GHC 7.8 with

```
    /home/travis/build/haskell-servant/servant/servant-server/test/Servant/ServerSpec.hs:71:5:
        Context reduction stack overflow; size = 21
        Use -fcontext-stack=N to increase stack size to N
          Servant.Server.Internal.HasServer
            (Servant.API.Description.Description "foo" :> GET)
            '[NamedContext "foo" '[]]
        In the expression:
          serveWithContext comprehensiveAPI comprehensiveApiContext
        In a pattern binding:
          _ = serveWithContext comprehensiveAPI comprehensiveApiContext
```

@phadej should we increase stack size for the build?
I guess that error is because of `ComprehensiveAPI` being too comprehensive for GHC 7.8 :)

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -350,6 +350,20 @@ instance HasForeign lang ftype api
   foreignFor lang ftype Proxy req =
     foreignFor lang ftype (Proxy :: Proxy api) req
 
+instance HasForeign lang ftype api
```
fizruk: Here we ignore `Summary` and `Description`, but it would be super nice to adjust `servant-foreign` so that we could save descriptions to later translate them into comments.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -533,6 +534,18 @@ instance HasServer api context => HasServer (HttpVersion :> api) context where
   route Proxy context subserver =
     route (Proxy :: Proxy api) context (passToServer subserver httpVersion)
 
+-- | Ignore @'Summary'@ in server handlers.
```
fizruk: Can/should we use summary/description in error messages?

servant-foreign/src/Servant/Foreign/Internal.hs
fierce-katie: Opened #763 for it.

servant/src/Servant/API/Description.hs
```diff
@@ -0,0 +1,32 @@
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE PolyKinds          #-}
+{-# OPTIONS_HADDOCK not-home    #-}
+module Servant.API.Description (Description, Summary) where
+
+import           Data.Typeable (Typeable)
+import           GHC.TypeLits  (Symbol)
+-- | Add a short summary for (part of) API.
```
phadej: could we call it `Synopsis`? Would be more in line with Cabal terminology :P

servant/src/Servant/API/Description.hs
```diff
@@ -0,0 +1,32 @@
...
+-- >>> type MyApi = Summary "Get book by ISBN." :> "books" :> Capture "isbn" Text :> Get '[JSON] Book
+data Summary (sym :: Symbol)
+    deriving (Typeable)
+
+-- | Add more verbose description for (part of) API.
+--
+-- Example:
+--
+-- >>> type MyApi = Description "Some longer implementation details here." :> "books" :> Capture "isbn" Text :> Get '[JSON] Book
+data Description (sym :: Symbol)
```
phadej: I see the motivation, but is it really practical to write any >=80 char description in a type-level? @fizruk?

servant-server/src/Servant/Server/Internal.hs
phadej: I have no idea how the implementation would look like...

servant/src/Servant/API/Description.hs
fizruk: We call it `Summary` since it's what it's called in Swagger terminology, but I'm ok with `Synopsis` too :)

servant/src/Servant/API/Description.hs
fizruk: Why not? :)

```haskell
-- | This comment is only visible in Haskell sources and Haddocks.
-- It can be really long and is very easy to read.
-- But we can't extract it even with TH :(
type MyEndpoint = Get '[JSON] Resource

type MyEndpoint = Description
  "This comment is visible in multiple Servant interpretations \
  \and can be really long if necessary. \
  \Haskell multiline support is not perfect \
  \but it's still very readable."
 :> Get '[JSON] Resource

```

servant-server/src/Servant/Server/Internal.hs
fizruk: We could (optionally) inject description/summary for 4xx (client side) errors.

servant/src/Servant/API/Description.hs
phadej: @haskell-servant/maintainers bikeshed please!

servant-server/src/Servant/Server/Internal.hs
phadej: Yeah I understand. We'd need some kind of `mapError` post phase in `Delayed` to do that. Can you or @fierce-katie open a ticket about that. Let's not complicate this PR

phadej:
I'm :+1: with this. Having two very similar things: `Synopsis` / `Summary` and `Description` is somewhat arbitrary, but I see the motivation, so not against it.

phadej:
about `-fcontext-stack=N`, yes. please increase it in a spec file, I don't want to have it 0 because we won't catch looping stuff then (as we have `UndecidableInstances` etc.)

servant/src/Servant/API/Description.hs
alpmestan: I don't care =P

servant-server/src/Servant/Server/Internal.hs
fizruk: Created https://github.com/haskell-servant/servant/issues/768.

servant/src/Servant/API/Description.hs
fizruk: @fierce-katie this should get into examples/doctests/tutorial to demonstrate multiline descriptions.

3noch: Now...if only we had a quasiquoter that would produce `Symbol`s instead of `String`s.

phadej:
@fierce-katie is there something you want to do here, or is this done from your side? I'd like to merge this!

fierce-katie:
@phadej looks like I've made all the fixes discussed above, so feel free to merge :)

fierce-katie:
Ping @phadej

phadej:
Totally forgot, thanks for the ping.

PR #772, opened 2017-06-27
Allow http-media 0.7

felixonmars:


phadej:
Have you tested it compiles and test passes? Our CI doesn't test against newest versions, if not added to `stack.yaml` explicitly as `extra-deps`?

felixonmars:
Yes, and I have tried a few operations to make sure :P

phadej:
Thanks, but

https://github.com/haskell-servant/servant/commit/70aae4c5c0a3d20d6209d338931ea4a30da33214

:(

felixonmars:
Ah, I didn't package servant-client for Arch so I overlooked that part...

phadej:
It's ok, next time make a change to `stack.yaml`, then Travis will do the job ;)

felixonmars:
Got it, many thanks :P

PR #777, opened 2017-07-12
Set up Danger with danger-hlint

tmspzz:
This PR will set up hlint to run automatically on every new open PR. 

This is how it will look if merged: https://github.com/blender/servant/pull/1

It needs one more change that can only be done by the maintainers on the travis job. It requires a `DANGER_GITHUB_API_TOKEN` travis environment variable to be set. Token should be communicate via another channel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ addons:
     packages:
       - libgmp-dev
 
+before_install:
+  - rvm use 2.1 --install --binary --fuzzy
+
 install:
   - mkdir -p ~/.local/bin
   - export PATH=$HOME/.local/bin:$PATH
@@ -21,6 +24,12 @@ install:
   - stack --version
   - stack setup --no-terminal
   - (cd $HOME/.local/bin && wget https://zalora-public.s3.amazonaws.com/tinc && chmod +x tinc)
+  - rake hlint:install[2.0.9]
+  - bundle install --deployment
+
+
+before_script:
+  - bundle exec danger
 
 script:
   - if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then ./scripts/ci-cron.sh ; else stack test --ghc-options=-Werror --no-terminal ; fi
@@ -29,3 +38,4 @@ cache:
   directories:
     - $HOME/.tinc/cache
     - $HOME/.stack
+    - vendor/bundle
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,7 @@
+# Run hlint on modified files
+affected_files = git.added_files + git.modified_files
+
+haskell_files = affected_files.select { |file| file.end_with?('.hs') }
+
+hlint.lint(haskell_files, true, hint: "HLint.hs", extension: "hs")
+
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+source "https://rubygems.org"
+
+gem "danger"
+gem "danger-hlint"
+gem "pry"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,64 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    addressable (2.5.1)
+      public_suffix (~> 2.0, >= 2.0.2)
+    claide (1.0.2)
+    claide-plugins (0.9.2)
+      cork
+      nap
+      open4 (~> 1.3)
+    coderay (1.1.1)
+    colored2 (3.1.2)
+    cork (0.3.0)
+      colored2 (~> 3.1)
+    danger (5.3.3)
+      claide (~> 1.0)
+      claide-plugins (>= 0.9.2)
+      colored2 (~> 3.1)
+      cork (~> 0.1)
+      faraday (~> 0.9)
+      faraday-http-cache (~> 1.0)
+      git (~> 1)
+      kramdown (~> 1.5)
+      octokit (~> 4.7)
+      terminal-table (~> 1)
+    danger-hlint (0.0.3)
+      danger-plugin-api (~> 1.0)
+    danger-plugin-api (1.0.0)
+      danger (> 2.0)
+    faraday (0.12.1)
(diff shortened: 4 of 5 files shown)


phadej:
I'd rather prefer simpler approach like `aeson` uses:

https://github.com/bos/aeson/blob/2bc6f22f164d172d0d77d878422b67433a57a9d8/travis/script.sh#L22

blender:
You mean instead of the Rake file? Building hint every time is quite long.

phadej:
If `hlint` is installed with `stack` then it is easy to cache (`~/.stack-work/` which we do).

Also, the right thing to do *first*, is to make running `hlint` easy locally, and then fix or silent the hints. `aeson` does so by providing `Makefile` as glorified "oneline shell-script" dispatcher. That approach I like a lot.

blender:
@phadej From what I see in the `aeson` approach hlint is not fixing anything. It's just been build and then called to lint (not fix) via a make file.

I am happy to remove the Rakefile and have stack build hlint. As far as having hlint auto fix the lint problems in every PR I am not sure it's doable as it would need commit access to the PR itself.

Am I missing something?

phadej:
Auto-fix? No, I'm strongly against that :-1:. I don't want bot to touch my (or other contributors) code. Even commenting is IMHO noisy, as travis tells what's wrong (my workflow with e.g. `aeson`: looks like `hlint` job failed, run it locally, fix, push commit)

@haskell-servant/maintainers opinions, also in general about using `hlint`

blender:
I see. The point of commenting is to invite people to fix instead of failing the build for linting reasons.

If you and the other maintainers prefer failing the build and have contributors looks through travis logs then you don't need any of this. 

Let me know.

kosmikus:
FWIW, I don't see any need in running hlint as part of CI at all, and I'm strongly against auto-fixing anything.

blender:
@kosmikus there is no auto-fixing. It was my misunderstanding of @phadej comment's .

kosmikus:
Ok, if we're merely running hlint and making suggestions as github comments, then I don't mind. If we're auto-failing the build and essentially forcing to blindly apply all hlint suggestions, then I'm against it.

blender:
@kosmikus yes. The intention is to make comments. This is all it does, there is no failing involved. It I understand correctly failing is the approach that @phadej would like  (what aeson has).

kosmikus:
@blender Thanks for clarifying.

phadej:
@blender does it only comment about changed parts of the file?

Even in that case it may be irritating. ATM `hlint` generates a lot of "hints", and what would be frustrating is that copy&pasting the code (a lot of code has similar shape) and it would generate a nitpick

```
servant-server/test/Servant/ServerSpec.hs:305:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)

servant-server/test/Servant/ServerSpec.hs:317:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)

servant-server/test/Servant/ServerSpec.hs:330:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)

servant-server/test/Servant/ServerSpec.hs:343:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)

servant-server/test/Servant/ServerSpec.hs:354:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)

servant-server/test/Servant/ServerSpec.hs:367:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)

servant-server/test/Servant/ServerSpec.hs:379:9: Suggestion: Redundant bracket
Found:
  (flip runSession) (serve queryParamApi qpServer)
Why not:
  flip runSession (serve queryParamApi qpServer)
```

maybe those should be fixed, with *eta-reduction* its not so clear

```
servant-client/src/Servant/Client.hs:446:3: Warning: Eta reduce
Found:
  clientWithRoute Proxy req
    = clientWithRoute (Proxy :: Proxy api) req
Why not:
  clientWithRoute Proxy = clientWithRoute (Proxy :: Proxy api)

servant-client/src/Servant/Client.hs:452:3: Warning: Eta reduce
Found:
  clientWithRoute Proxy req
    = clientWithRoute (Proxy :: Proxy api) req
Why not:
  clientWithRoute Proxy = clientWithRoute (Proxy :: Proxy api)

servant-client/src/Servant/Client.hs:458:3: Warning: Eta reduce
Found:
  clientWithRoute Proxy req
    = clientWithRoute (Proxy :: Proxy api) req
Why not:
  clientWithRoute Proxy = clientWithRoute (Proxy :: Proxy api)
```

kosmikus:
That's my main problem with hlint. There are many suggestions which are usually good, but occasionally not what you want. Redundant brackets are a good example. In most cases, we don't need them, but sometimes, they are deliberately there to emphasize or clarify something. Eta-reduction is a similar case (aeson disables the eta reduction check, the servant hlint config currently doesn't).

Disabling all such controversial checks completely is one option. But really, I think hlint is better suited for providing guidance to developers locally rather than to actually enforce a particular style in a completely automated way.

blender:
@phadej it will lint the entire file affected by the PR not only the changed lines. I think that with some tweaking (that I can do) this can be avoided and only inline comments can be done.

To be clear, it does not lint the entire project just the files affected by the PR.

Also it is mentioned [here](http://danger.systems/guides/faq.html#i-want-to-ignore-a-fail-just-this-once) that inappropriate comments can be silenced by Danger.

alpmestan:
I 100% agree with @kosmikus here -- not a big fan of giving too much importance to hlint as part of the automated testing/reporting/etc of a project.

Issue #776, opened 2017-07-11
Auth handling tutorial - confusion over cookie/header


JoeMShanahan:
I think that the wording of the example could be improved for the tutorial, particularly for the provided auth handler:

```haskell
authHandler :: AuthHandler Request Account
authHandler =
  let handler req = case lookup "servant-auth-cookie" (requestHeaders req) of
        Nothing -> throwError (err401 { errBody = "Missing auth header" })
        Just authCookieKey -> lookupAccount authCookieKey
  in mkAuthHandler handler
```
The pattern match `Just authCookieKey` implies that we've retrieved a value from the cookie, where in reality all we've done is retrieved the _header value_ of `servant-auth-cookie`.

Using `Web.Cookie`, the following achieves what is implied:

```haskell
authHandler :: AuthHandler Request Account
authHandler =
  let handler req = case lookup "cookie" (requestHeaders req) of
        Nothing -> throwError (err401 { errBody = "Missing cookie header" })
        Just cookie -> case lookup "servant-auth-cookie" $ parseCookies cookie of
          Nothing -> throwError (err401 { errBody = "Missing token in cookie"})
          Just authCookieValue -> lookupAccount authCookieValue
  in mkAuthHandler handler
```
.. which I think is a little convoluted. I'd (personally) instead write it like:

```haskell
authHandler :: AuthHandler Request Account
authHandler = mkAuthHandler handler
  where
  maybeToEither e = maybe (Left e) Right
  throw401 msg = throwError $ err401 { errBody = msg }
  handler req = either throw401 lookupAccount $ do
    cookie <- maybeToEither "Missing cookie header" $ lookup "cookie" $ requestHeaders req
    maybeToEither "Missing token in cookie" $ lookup "servant-auth-cookie" $ parseCookies cookie
```

Of course it's much easier just to change the wording of the tutorial but I'd prefer to see a fully worked example that likely better represents a typical situation.

JoeMShanahan:
Happy to do the PR if needed once feedback has been received!

alpmestan:
I too think that implementing the actual cookie retrieval is nicer. I'd definitely be in favor of a PR that makes that change, if you feel like putting it together :)

JoeMShanahan:
Will add something hopefully later today!

JoeMShanahan:
PR in!

PR #780, opened 2017-07-17
Improve authentication docs

JoeMShanahan:
See #776 

Comments, welcome. Took a bit longer than I first envisioned to get around to this!

### DIFF
--- a/doc/tutorial/Authentication.lhs
+++ b/doc/tutorial/Authentication.lhs
@@ -281,7 +281,7 @@ lookupAccount key = case Map.lookup key database of
 For generalized authentication, servant exposes the `AuthHandler` type,
 which is used to wrap the `Request -> Handler user` logic. Let's
 create a value of type `AuthHandler Request Account` using the above `lookupAccount`
-method:
+method (note: we depend upon `cookie`'s `parseCookies` for this):
 
 ```haskell
 import           Web.Cookie             (parseCookies)
--- a/doc/tutorial/Authentication.lhs
+++ b/doc/tutorial/Authentication.lhs
@@ -286,9 +286,9 @@ method (note: we depend upon `cookie`'s `parseCookies` for this):
 ```haskell
 import           Web.Cookie             (parseCookies)
 
---- | The auth handler wraps a function from Request -> Handler Account
---- we look for a Cookie in the header and the perform a token look up 
---- on the cookie, passing it to `lookupAccount`.
+--- | The auth handler wraps a function from Request -> Handler Account.
+--- We look for a token in the request headers that we expect to be in the cookie.
+--- The token is then passed to our `lookupAccount` function.
 authHandler :: AuthHandler Request Account
 authHandler = mkAuthHandler handler
   where
--- a/doc/tutorial/Authentication.lhs
+++ b/doc/tutorial/Authentication.lhs
@@ -281,7 +281,8 @@ lookupAccount key = case Map.lookup key database of
 For generalized authentication, servant exposes the `AuthHandler` type,
 which is used to wrap the `Request -> Handler user` logic. Let's
 create a value of type `AuthHandler Request Account` using the above `lookupAccount`
-method (note: we depend upon `cookie`'s `parseCookies` for this):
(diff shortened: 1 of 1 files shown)


doc/tutorial/Authentication.lhs
```diff
@@ -281,17 +281,22 @@ lookupAccount key = case Map.lookup key database of
...
-method:
+method (note: we depend upon `cookie`'s `parseCookies` for this):
 
 ```haskell
--- | The auth handler wraps a function from Request -> Handler Account
--- we look for a Cookie and pass the value of the cookie to `lookupAccount`.
+import           Web.Cookie             (parseCookies)
+
+--- | The auth handler wraps a function from Request -> Handler Account
+--- we look for a Cookie in the header and the perform a token look up
```
phadej: is there a period missing? (in original as well?)

phadej: *and then lookup a token ...*

JoeMShanahan: Thanks, push incoming

doc/tutorial/Authentication.lhs
```diff
@@ -281,17 +281,22 @@ lookupAccount key = case Map.lookup key database of
 For generalized authentication, servant exposes the `AuthHandler` type,
 which is used to wrap the `Request -> Handler user` logic. Let's
 create a value of type `AuthHandler Request Account` using the above `lookupAccount`
-method:
+method (note: we depend upon `cookie`'s `parseCookies` for this):
```
alpmestan: I know this is going to sound "annoying", but could you please have "`cookie`" point to the package's page on hackage? To make it clear and unambiguous what "`cookies` refers to here.

JoeMShanahan: Not annoying, I agree it would be good.

Is there a way to link it easily? Some markdown magic I don't know? Perhaps it's just `[text](http://link)`?

alpmestan: I think it's the contrary, you have to do a link around `text`, like this: [`servant`](https://github.com/haskell-servant/servant)

JoeMShanahan: Pushed _something_ - is there a way to see what it actually looks like? Not a fan of the guessing game! Looks ok in general markdown formatters.

alpmestan: I don't know of a way to check before it's merged. Once it's merged, you can go to http://haskell-servant.readthedocs.io/ and click "v: stable" at the bottom and pick *master* instead.

JoeMShanahan:
What's the process now? Looking at these CI failures, doesn't appear as if this branch is the cause? Could be wrong.

doc/tutorial/Authentication.lhs
```diff
@@ -281,17 +281,23 @@ lookupAccount key = case Map.lookup key database of
...
 which is used to wrap the `Request -> Handler user` logic. Let's
 create a value of type `AuthHandler Request Account` using the above `lookupAccount`
-method:
+method (note: we depend upon [`cookie`](https://hackage.haskell.org/package/cookie)'s
+`parseCookies` for this):
 
 ```haskell
--- | The auth handler wraps a function from Request -> Handler Account
--- we look for a Cookie and pass the value of the cookie to `lookupAccount`.
+import           Web.Cookie             (parseCookies)
```
phadej: this makes Travis barf.

phadej: the tutorial are markdown-literated files. You should get the same errors as travis if you do `stack build` locally.

kakkun61:
I was confused about this bug, while I was learning Servant's authentication.
It makes beginners happy to merge this PR.

I have fixed compilation error. `stack build` succeeds.
@phadej Can I create new PR including this PR?
https://github.com/haskell-servant/servant/compare/master...kakkun61:improve-authentication-docs

alpmestan:
@kakkun61 Yes, please do if you feel like it :)

JoeMShanahan:
Yes please do, apologies for not getting around to this.

PR #788, opened 2017-07-28
Move location of header information

ivan-m:
Currently, it appears after the notes, authentication and capture
information... such that if any of those exist, then the list of
header sensitivity will appear to be an item of one of the previous
ones (as they provide Markdown headers).


ivan-m:
The alternative would be to just add a Markdown-header for HTTP-headers.

phadej:
I'd go with Markdown header

PR #790, opened 2017-07-28
Document the type of endpoint a parameter belongs to

### LINKED ISSUE #760: servant-docs and QueryParam
If I have a `QueryParam` in a path ending with a `Post`, and generate documentation for it with servant-docs, then it prints out that it is a `GET Parameter`... which looks out of place.

ivan-m:
Closes #760


ivan-m:
I'm not sure if this is the right approach or if they should just be referred to as query parameters.

phadej:
conflict

ivan-m:
Conflict resolved.

phadej:
@ivan-m, if you could write CHANGELOG.md entries for your recent changes to `servant-docs` it would be super great.  just make `next` section. (or if there are breaking changes, then bump directly to `0.12`)

PR #797, opened 2017-08-25
Remove `ToSample a => ToSample [a]` instance

osa1:
`ToSample [a]` was generating an infinite list of samples. This is far from
being useful; if you have two samples for `a` servant-docs generates 79kloc long
samples for `[a]`.

Because it's impossible to come up with a useful `ToSample [a]` that works for
all use cases, we leave that out.

### DIFF
--- a/servant-docs/CHANGELOG.md
+++ b/servant-docs/CHANGELOG.md
@@ -6,6 +6,8 @@
 * Add a markdown header for the Headers an endpoint is sensitive to.
 * Document the HTTP Method the parameters of an endpoint belong to
   (rather than assuming `GET` for all of them).
+* Remove `ToSample a => ToSample [a]` and `ToSample (ZipList a)` instances
+  (https://github.com/haskell-servant/servant/pull/797).
 
 0.11
 ----
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -895,11 +895,9 @@ instance (ToSample a, ToSample b, ToSample c, ToSample d, ToSample e, ToSample f
 
 instance ToSample a => ToSample (Maybe a)
 instance (ToSample a, ToSample b) => ToSample (Either a b)
-instance ToSample a => ToSample [a]
 
 -- ToSample instances for Control.Applicative types
 instance ToSample a => ToSample (Const a b)
-instance ToSample a => ToSample (ZipList a)
 
 -- ToSample instances for Data.Monoid newtypes
 instance ToSample All


alpmestan:
This is indeed better than what we have. I'm just wondering whether folks will be annoyed or not by the fact that there'll be just one sample, and therefore whether we should instead _not_ have the instance at all.

osa1:
Right, we can't come up with an instance that'll work for everyone. I think it's best that we remove the instance altogether. Do you want me to update the PR?

alpmestan:
@osa1 That would be great, thanks :)

osa1:
Done.

phadej:
IMHO removing instance isn't the solution. Rather `take 5` or so, when generating the docs document.

When I used `servant-docs`, need to add (orphan) `ToSample` instances was one reason I moved away from using `servant-docs`.

alpmestan:
Ideally we would not be doing this through typeclass instances. If anyone's got a better idea, I'm all ears.

In the meantime @phadej, I suggested to remove it altogether because there are at least two users with different needs that can't be satisfied simultaneously, as far as I can tell. The current instance as well as the one suggested by @osa1 are too "arbitrary" in a sense. And so would be your `take 5`.

osa1:
It's a solution if you can't provide an instance that's useful for at least most of the cases. This problem actually exists in many other libraries too, e.g. IMHO `Arbitrary Int` should be just removed because it's completely useless.

arianvp:
Maybe take the approach that `hedgehog` takes? Moves away from the typeclass driven system, but provide example generators explicitly?

alpmestan:
@arianvp I would have gone for that right from the start if servant was not build around typeclass-based type traversal. When you say `docsFor apiProxy`, you don't have access to each individual endpoint and the types that it uses, so you have to get the samples externally, therefore all the crappy typeclasses that we have in servant-docs (ToCapture & friends). If you see another way to implement this that allows us to provide samples through functions/values instead of typeclass instances, I'd be delighted :)

jkarni:
:-1: 

 I'm a little bit confused by the reasoning for removing this. If the instance isn't useful in some cases, doesn't the end user have to newtype wrap one way or the other? But at least in many cases it *is* useful, at least when fixed.

I think something like what @phadej proposed is better, though rather than using a fixed size I'd propose using `sized`/`getSize`, or `arbitrarySizedBoundedIntegral` or some custom range.

osa1:
> If the instance isn't useful in some cases

It's completely useless, pretty much the worst instance you could come up with. I removed the original description so that part may not be clear now: it's basically generating an infinite list of samples. We have two samples for `a` and `ToSample [a]` generates 79kloc-long samples and makes docs completely useless. In fact it makes docs so big browsers have trouble rendering it and searching in it.

> I think something like what @phadej proposed is better, though rather than using a fixed size I'd propose using sized/getSize, or arbitrarySizedBoundedIntegral or some custom range.

I don't know what are those functions for, could you link us the haddocks for those? If those are functions, how do you propose passing parameters to them?

alpmestan:
@jkarni I really don't like the fact that we have to pick a number or rely on a function that does. What we can do is offer functions for writing those various types of instances that we've been discussing, and then let people define instances the way they want them, maybe?

jkarni:
@osa1 I agree what we currently have isn't good - I just disagree about what to do about it.

- [`arbitrarySizedBoundedIntegral`](https://hackage.haskell.org/package/QuickCheck-2.10.0.1/docs/Test-QuickCheck-Arbitrary.html#v:arbitrarySizedBoundedIntegral)
- [`sized`](https://hackage.haskell.org/package/QuickCheck-2.10.0.1/docs/Test-QuickCheck-Gen.html#v:sized)
- [`getSize`](https://hackage.haskell.org/package/QuickCheck-2.10.0.1/docs/Test-QuickCheck-Gen.html#v:getSize)

`QuickCheck` allows the use of a size parameter - we'd just use that. But even if there weren't that option, I'm still in favor of picking some value (like 10) as the bound on lists, and generate values up to that size, over removing the instance. 

@alpmestan I don't think making people write orphan instances helps. First, it'll often be more work for people. Second, they're orphan. Third, that doesn't even solve the problem - even if people are allowed define an instance for `[a]` in full knowledge of their use-cases, they can't pick a single one because there are multiple different contexts in a single application.

(If anything, this is making me realize that I think we should *add* more instances - `Int`, etc.)

jkarni:
Ugh, I for some reason though QC was used for the instances. Ignore the `sized` comments. It's been too long away from the computer. 

I no longer have an opinion here.

arianvp:
I agree with @phadej here. Lets just make the instance  `take 10 $ genSamples`

osa1:
I disagree -- `take 10 $ genSamples ...` would be slightly more useful (because you could at least render the docs in a browser in reasonable amount of time), but it'd still just repeat same example 10 times in the example I gave. I'd much rather define an orphan instance in my program.

Also, why `10` ?

alpmestan:
I'm with @osa1 on this. If your data type is _huge_, 10 samples might be too much for the docs to be reasonably readable. This number will be inappropriate most of the time. I'm more and more convinced the best solution is for users to define orphans using a few helper functions that servant-docs would provide. Along with a little mention of this whole thing in the haddocks. How does it sound?

phadej:
Orphan works only for applications. If someone wants to use [a] instance in a library, they are screwed. I propose
- `take n`, with either static n or dynamic, based on some serialized size metric,
- [a] instance which chunks underlying toSamples: `[a, b, c, d, e, f, g...] -> [[], [a], [b, c], [d, e, f], [g...],...`

phadej:
Single element instance is bad, if you need ([a], b) instance you'll get only single value too: we cannot have that either.

ToSample should generate a list of "high entropy" data, for sampleja to be interesting, but not too high to be still digestible. Thus `enumerable`/`universe` approach performs badly, as first elements in naive enumeration are boring

I don't think this is hard problem, someone just need to thing a bit about it as a whole. Local "fixes" don't solve overall problem. Now we argue which local fix is less bad :)

`swagger2` would benefit from proper example generation too.

phadej:
btw. We **had** arbitrary take: `take 1`: https://github.com/haskell-servant/servant/pull/755, pr by @osa1.

osa1:
I don't see any `take 1`s in that PR. Which line are you talking about?

phadej:
https://github.com/haskell-servant/servant/pull/755/files#diff-c63d2094c40a80cd335a525f83087ec6L808, and [`sampleByteString`](https://github.com/osa1/servant/blob/d8dd1cb90a887e2a235179a03069c9552f6ba8e0/servant-docs/src/Servant/Docs/Internal.hs#L475) uses [`toSample`](https://github.com/osa1/servant/blob/d8dd1cb90a887e2a235179a03069c9552f6ba8e0/servant-docs/src/Servant/Docs/Internal.hs#L475) which is `listToMaybe`, which is kind of `take 1` :)

Issue #798, opened 2017-08-28
Make servant-client more generic to support different ways of running a request


abailly:
Currently, when one extracts a client from an API definition, she gets a tree of functions running inside `ClientM` monad:

```
type MyAPI = "foo" :> Capture "x" Int :> Get '[JSON] Foo
        :<|> "bar" :> Capture "y" String :> Get '[JSON] Bar

myAPI :: Proxy MyAPI
myAPI = Proxy

foo :: Int -> ClientM Foo
bar :: String -> ClientM Bar
foo :<|> bar = client myAPI
```

This implies that one needs to use  `runClientM foo env` to execute a query which uses http-client under the hood to actually perform the request in `IO`. However there are various use cases where one would like to change how the query is run:

1. Testing: one would like to provide canned answers and mock the server without having to fire up an actual HTTP server
2. Retry: In some circumstances, it would be desirable to retry failed requests according to some policy
3. Specialized run: One could want to tweak the requests in some ways or use different execution engines

A possible solution might be to parameterize the terminal combinators, e.g. `Verb`s that actually call `performRequests` and returns a `ClientM` action:

```
class MonadClient m where
  performRequestCT :: Proxy ct -> method -> Req -> m a
  performRequest     :: Proxy ct -> method -> Req -> m a

instance MonadClient ClientM where
 .... 

instance OVERLAPPING_
  -- Note [Non-Empty Content Types]
  ( MonadClient m, MimeUnrender ct a, BuildHeadersTo ls, ReflectMethod method, cts' ~ (ct ': cts)
  ) => HasClient (Verb method status cts' (Headers ls a)) where
  type Client (Verb method status cts' (Headers ls a))
    = m (Headers ls a)
  clientWithRoute Proxy req = do
    let method = reflectMethod (Proxy :: Proxy method)
    (hdrs, resp) <- performRequestCT (Proxy :: Proxy ct) method req
    return $ Headers { getResponse = resp
                     , getHeadersHList = buildHeadersTo hdrs
                     }
```

Another possible route would be to abstract the building of a request and provide different interpretations, something like what's done in https://github.com/soenkehahn/servant-request/

abailly:
I have started working on a PR for this feature but got stuck while trying to add `m` parameter in `Client` type family: https://github.com/abailly/servant/blob/master/servant-client/src/Servant/Client.hs

jkarni:
Great - this is something many people have wanted!

I'm not exactly sure what problem you're running into, but:

```
[...]

PR #800, opened 2017-09-03
Make client runner a type class

abailly:
This PR is a first attempt at addressing #798 

The idea is to abstract away request execution and expose it as a method `runRequest` in `RunClient` type-class. Specialized instances of this type-class can then be provided to handle various execution contexts:

* Basic execution with `ClientM`
* Specialized execution like `RetryM` context that encapsulates logic for retrying requests depending on a `Policy` 
* Testing without having to launch a server
* ...

### DIFF
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -108,7 +108,7 @@ data EmptyClient = EmptyClient deriving (Eq, Show, Bounded, Enum)
 -- > (getAllBooks :<|> EmptyClient) = client myApi
 instance HasClient m EmptyAPI where
   type Client m EmptyAPI = EmptyClient
-  clientWithRoute pm Proxy _ = EmptyClient
+  clientWithRoute _pm Proxy _ = EmptyClient
 
 -- | If you use a 'Capture' in one of your endpoints in your API,
 -- the corresponding querying function will automatically take
@@ -178,7 +178,7 @@ instance OVERLAPPABLE_
   (RunClient m ct ([H.Header], a), MimeUnrender ct a, ReflectMethod method, cts' ~ (ct ': cts)
   ) => HasClient m (Verb method status cts' a) where
   type Client m (Verb method status cts' a) = m a
-  clientWithRoute pm Proxy req = do
+  clientWithRoute _pm Proxy req = do
     (_hdrs, a) :: ([H.Header], a) <- runRequest (Proxy :: Proxy ct) method req
     return a
       where method = reflectMethod (Proxy :: Proxy method)
@@ -188,7 +188,7 @@ instance OVERLAPPING_
   , ReflectMethod method) => HasClient m (Verb method status cts NoContent) where
   type Client m (Verb method status cts NoContent)
     = m NoContent
-  clientWithRoute pm Proxy req = do
+  clientWithRoute _pm Proxy req = do
     _hdrs :: [H.Header] <- runRequest (Proxy :: Proxy NoContent) method req
     return NoContent
       where method = reflectMethod (Proxy :: Proxy method)
@@ -200,7 +200,7 @@ instance OVERLAPPING_
   ) => HasClient m (Verb method status cts' (Headers ls a)) where
   type Client m (Verb method status cts' (Headers ls a))
     = m (Headers ls a)
-  clientWithRoute pm Proxy req = do
+  clientWithRoute _pm Proxy req = do
     let method = reflectMethod (Proxy :: Proxy method)
     (hdrs, resp) <- runRequest (Proxy :: Proxy ct) method req
     return $ Headers { getResponse = resp
@@ -213,7 +213,7 @@ instance OVERLAPPING_
   ) => HasClient m (Verb method status cts (Headers ls NoContent)) where
   type Client m (Verb method status cts (Headers ls NoContent))
     = m (Headers ls NoContent)
-  clientWithRoute pm Proxy req = do
+  clientWithRoute _pm Proxy req = do
     let method = reflectMethod (Proxy :: Proxy method)
     hdrs <- runRequest (Proxy :: Proxy NoContent) method req
     return $ Headers { getResponse = NoContent
@@ -416,7 +416,7 @@ instance (RunClient m NoContent (Int, ByteString, MediaType, [HTTP.Header], Resp
     = H.Method ->  m (Int, ByteString, MediaType, [HTTP.Header], Response ByteString)
 
   clientWithRoute :: Proxy m -> Proxy Raw -> Req -> Client m Raw
-  clientWithRoute pm Proxy req httpMethod = do
+  clientWithRoute _pm Proxy req httpMethod = do
     runRequest (Proxy :: Proxy NoContent) httpMethod req
 
 -- | If you use a 'ReqBody' in one of your endpoints in your API,
--- a/servant-client/src/Servant/Client.hs
+++ b/servant-client/src/Servant/Client.hs
@@ -22,7 +22,7 @@ module Servant.Client
   , client
   , HasClient(..)
   , ClientM
-  , runClientM
+  , runClientM, inClientM, clientM
   , ClientEnv (ClientEnv)
   , mkAuthenticateReq
   , ServantError(..)
@@ -52,7 +52,8 @@ import           Servant.Common.Req
 
 -- * Accessing APIs as a Client
 
--- | 'client' allows you to produce operations to query an API from a client.
+-- | 'client' allows you to produce operations to query an API from a client within
+-- a given monadic context `m`
 --
 -- > type MyApi = "books" :> Get '[JSON] [Book] -- GET /books
 -- >         :<|> "books" :> ReqBody '[JSON] Book :> Post '[JSON] Book -- POST /books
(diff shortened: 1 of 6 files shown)


phadej:
LGTM so far! :+1: Great job.

abailly:
Thanks. Just noticed build is failing, will try to fix that. I am not 100% satisfied with the solution I have so far: find it rather cumbersome to carry around _two_ `Proxy`. Plus, a good example of a _second_ instance, e.g. for testing would probably be handy.

servant-client/src/Servant/Client/HttpClient.hs
```diff
@@ -0,0 +1,142 @@
...
+-- | Try clients in order, last error is preserved.
+instance Alt ClientM where
+  a <!> b = a `catchError` \_ -> b
+
+runClientM :: ClientM a -> ClientEnv -> IO (Either ServantError a)
+runClientM cm env = runExceptT $ (flip runReaderT env) $ runClientM' cm
+
+
+performRequest :: Method -> Req
+               -> ClientM ( Int, ByteString, MediaType
```
phadej: I'd make a record for this 5-tuple.

Also `Response` ties the lib to http-client still, does one need anything else than `requestBody`?

servant-client/src/Servant/Client.hs
```diff
@@ -1,14 +1,16 @@
...
-{-# LANGUAGE FlexibleContexts     #-}
-{-# LANGUAGE FlexibleInstances    #-}
-{-# LANGUAGE InstanceSigs         #-}
-{-# LANGUAGE OverloadedStrings    #-}
-{-# LANGUAGE PolyKinds            #-}
-{-# LANGUAGE ScopedTypeVariables  #-}
-{-# LANGUAGE TypeFamilies         #-}
-{-# LANGUAGE TypeOperators        #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE AllowAmbiguousTypes   #-}
```
jkarni: Is this still necessary? Why exactly?

servant-client/src/Servant/Client.hs
```diff
@@ -42,6 +44,8 @@ import           Prelude ()
 import           Prelude.Compat
 import           Servant.API
 import           Servant.Client.Experimental.Auth
+import           Servant.Client.HttpClient
```
jkarni: It seems to me like ultimately we'd want:

1) Everything except `Servant.Client.HttpClient` in a new library, `servant-client-core`. 
2) `Servant.Client.HttpClient` in `servant-client`, possibly along with `Servant.Client` as a re-export module.

servant-client/src/Servant/Client/HttpClient.hs
```diff
@@ -0,0 +1,142 @@
...
+                             , MonadThrow, MonadCatch
+                             )
+
+instance MonadBase IO ClientM where
+  liftBase = ClientM . liftBase
+
+instance MonadBaseControl IO ClientM where
+  type StM ClientM a = Either ServantError a
+
+  -- liftBaseWith :: (RunInBase ClientM IO -> IO a) -> ClientM a
```
jkarni: I'd remove these comments.

servant-client/src/Servant/Client/HttpClient.hs
```diff
@@ -0,0 +1,142 @@
...
+data ClientEnv
+  = ClientEnv
+  { manager :: Manager
+  , baseUrl :: BaseUrl
+  }
+
+
+-- | @ClientM@ is the monad in which client functions run. Contains the
+-- 'Manager' and 'BaseUrl' used for requests in the reader environment.
+
```
jkarni: Extra newline.

servant-client/src/Servant/Client/HttpClient.hs
jkarni: Agreed on both counts here.

servant-client/src/Servant/Client.hs
phadej: FWIW, the amount done in the PR is good. I can do module & package rearrangements later.

phadej:
@jkarni in

```haskell
client :: forall m api .HasClient m api => Proxy api -> Client m api
```

`m` is ambiguous. `Client` is non-injective type-family bla bla.

servant-client/test/Servant/ClientSpec.hs
```diff
@@ -358,7 +359,7 @@ wrappedApiSpec = describe "error status codes" $ do
         test (WrappedApi api, desc) =
           it desc $ bracket (startWaiApp $ serveW api) endWaiApp $ \(_, baseUrl) -> do
             let getResponse :: SCR.ClientM ()
-                getResponse = client api
+                getResponse = client (Proxy :: Proxy ClientM) api
```
phadej: I'd define `proxyClientM :: Proxy ClientM` on top-level and use it across the spec module.

servant-client/src/Servant/Client/Class.hs
```diff
@@ -0,0 +1,16 @@
...
+  (RunClient(..))
+where
+
+import Data.Proxy
+import Network.HTTP.Types
+import Servant.API
+import Servant.Common.Req
+
+class (Monad m) => RunClient m ct result where
+  runRequest :: Proxy ct
```
phadej: @jkarni, I don't think `Free` is good here. From what I look, it would be enough to have `Functor` constraint here, not a `Monad`. Could @abailly try if it's enough,

```diff
-  clientWithRoute pm Proxy req = do
-    (_hdrs, a) :: ([H.Header], a) <- runRequest (Proxy :: Proxy ct) method req
-    return a
+  clientWithRoute pm Proxy req =
+    fmap snd (runRequest (Proxy :: Proxy ct) method req :: m ([H.Header], a))
```

etc.

jkarni:
> @jkarni in client :: forall m api. HasClient m api => Proxy api -> Client m api m is ambiguous. Client is non-injective type-family bla bla.

@phadej right - my bad. `forall m x api. (HasClient m api, ClientMonadOf x ~ m) => Proxy api -> x ` works though, no?

phadej:
@jkarni, probably, I'm not 100% convinced it buys anything.  Having second proxy argument is not so bad IMHO. That's the code `servant-client` users don't see anyway, only combinator writers.

servant-client/src/Servant/Client/HttpClient.hs
abailly: I left mots of the existing code untouched. This should probably be part of another PR

servant-client/src/Servant/Client.hs
phadej: Now I note. This module shouldn't depend on `HttpClient`. That's the point of this PR, isn't it?

servant-client/src/Servant/Client/HttpClient.hs
phadej: I commented otherwise, but the curren t `Servant.Client` module shouldn't depend on `HttpClient`, thus this change is mandatory.

servant-client/src/Servant/Client.hs
abailly: Yes. I have added this import to reexport the symbols and provide a more or less tranparent change

servant-client/src/Servant/Client/Class.hs
abailly: Tried it but it does not seem to work: I need to _sequence_ evaluation before return value in the `NoContent` case

abailly:
Pushed new commits trying to address comments. CI is still failing for different reasons on different targets. Shall I put `CPP` directives based on GHC version? Seems this will become cumbersome very quickly...

phadej:
We already depend on [`base-compat`](http://hackage.haskell.org/package/base-compat). Leverage it.

servant-client/src/Servant/Client/HttpClient.hs
abailly: OK. I was expecting to provide this feature in such a way that existing code would stay compatible without any modification. If you want to sever dependency between `Client` and `HttpClient`, which indeed is the ultimate goal of #800 then I think it would make sense to also implement the decoupling of packages.

abailly:
@phadej Never used `base-compat`, will do that, thanks for pointing me at it.

servant-client/src/Servant/Client/HttpClient.hs
```diff
@@ -0,0 +1,139 @@
...
+import Servant.Common.BaseUrl
+import Servant.Common.Req
+
+import qualified Network.HTTP.Client as Client
+
+instance RunClient ClientM NoContent ( Int, ByteString, MediaType
+                                     , [HTTP.Header], Response ByteString) where
+  runRequest _ meth req = performRequest meth req
+
+instance (MimeUnrender ct a) =>
```
jkarni: Looking at this again, it seems a little obscure how the types determine or correspond with behavior. How would a person know that these are the instances they have to write, and with this behavior? Since `RunClient` will be part of the external API, this question matters.

Is it not sufficient to have `runRequest` have type:

```
  runRequest :: Proxy ct
             -> Method
             -> Req
             -> m ( Int, ByteString, MediaType, [HTTP.Header], Response ByteString)
```
And then define `runRequestCT` and `runRequestNoBody` in terms of these? Or if that doesn't work, to have a class with three methods (with maybe a `runRequest` `MINIMAL`)?

jkarni:
@abailly I have a little time to work on this (and also need it) - mind if I go ahead and work on this branch to make the code review changes?

abailly:
Of course not ! Please go ahead

jkarni:
@phadej why do you say the `Free` approach isn't good?

phadej:
@jkarni, I don't think we need it. I'm on the `mtl` side of `mtl` vs extensible-effects discussion. My gut feeling that the glorified `httpLbs`, 
```haskell
class Monad m => RunClient m where
    runRequest :: Req -> m Res
```
should be enough for most things in servant.

jkarni:
Closing in favor of #803 (which is nearly ready for review)

PR #803, opened 2017-09-07
Factor out servant-client-core

jkarni:
Since the change is quite substantial, I'll point out some that are significant, or might be controversial:

- `servant-client` has been split into `servant-client-core` and `servant-client`. 
- The module naming scheme has been changed. We no longer have the `Common` modules, and now instead have `Internal` ones. This is more consistent with our other packages (and I never really understood the original `servant-client` organization).
- `Req` has been renamed to `Request`; fields have been added, and field names have been changed for consistency
- Internal modules now export everything
- Some things that were `String` have been changed to `Text`/`ByteString`
- Some things that were lists have been changed to `Seq`
- Function names mentioned `Req` have been renamed to `Request`
- Imports have been qualified or made explicit
- Don't use `string-conversions` (so much inefficiency was sneaking in with `cs`!)
- Most of the tests stayed in `servant-client`, since it's easier to test with a backend. 
- Some of the things to change `Requests` are only exported from `Internal` modules now. On the one hand that sort of stuff is ripe for deprecation; on the other, I think maybe people are using it for authentication clients?

Thanks to @abailly for the initial work.


servant-client-core/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for servant-client-core
+
+## 0.11  -- YYYY-mm-dd
```
phadej: it will be 0.12

servant-client-core/src/Servant/Client/Core.hs
```diff
@@ -1,26 +1,28 @@
...
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE UndecidableInstances  #-}
 
 #include "overlapping-compat.h"
 -- | This module provides 'client' which can automatically generate
 -- querying functions for each endpoint just from the type representing your
 -- API.
-module Servant.Client
+module Servant.Client.Core
```
phadej: this module doesn't export everything. I don't like that we need to import `Internal` modules to use library.

I.e. either:
- export whole stable API from `Servant.Client.Core`
- or drop `Internal` from the other modules (we probably still do major bumps when we change something there).

jkarni:
Sorry - this isn't ready for review yet!

phadej:
@jkarni, I just commented my thought about `Internal` namespace. IMHO it's unnecessary infix.

abailly:
@jkarni Is there something I can do to help?

servant-server/servant-server.cabal
```diff
@@ -164,3 +164,5 @@ test-suite doctests
  if impl(ghc >= 8.2)
    x-doctest-options: -fdiagnostics-color=never
  include-dirs: include
+ other-modules:
```
phadej: don't do this. this breaks `sdist. I fixed the CI already. Will make it a bit cleaner soon-ish.

jkarni: OK, though if we need to use `Wno-missing-home-modules`, can we put that in the cabal or stack file rather than just `travis.yaml`? Otherwise everyone is going to run into compilation errors/warnings locally.

(Also, I'd love to understand what's going on better!)

jkarni:
@abailly want to review? Hopefully the invite I gave you should suffice for that.

servant-server/servant-server.cabal
phadej: Yes, I'll do that (putting warning suppression in a proper place) PR separately.

- Cabal 2.0 passes `-Wmissing-home-modules`, because it's good idea.
- So `Build_foo`, or `Paths_exe` or `Lexer` are reported missing.
- So you add them as `other-modules`
- But it breaks `cabal sdist`, so you declare `autogen-modules` (feature in `Cabal-2.0`)
   - this is *ok-ish*, because old-Cabal's would just omit the field, and you don't use them for `sdist`
- Except when `build-type: Custom` the situation is trick-ier. Then `sdist` command uses system `Cabal`, and not the one `cabal-install` or `stack` is compiled with.
    - Yet, we don't do `sdist` check in CI, so it's not a problem now
    - but I think we should, because that will catch forgotten test files etc.
    - `-Wmissing-home-modules` is not enough by itself, if we edit `extra-source-files` for example

There is no elegant solution atm. Proper ones are:
- `cabal new-doctest` (and `stack` support as a follow-up)
- ability to configure `ghc-options` per component

abailly:
@jkarni Thanks, I am on it !

soenkehahn:
@jkarni: Nice work. Could you add a link to the `servant-client-core` version that this works with, please? I guess that should be reviewed at the same time.

jkarni:
@soenkehahn not sure I follow - `servant-client-core` is being created in this PR, and will live in this repo.

servant-client-core/README.md
```diff
@@ -0,0 +1,30 @@
...
+3. Re-export the module Servant.Client.Core.Reexport so that your end-users
+   can be blissfully unaware of 'servant-client-core', and so each
+   backend-package comes closer to the warm hearth of the drop-in-replacement
+   equivalence class.
+
+## For combinator-writers
+
+You'l need to define a new `HasClient` instance for your combinator. There are
+plenty of examples to guide you in the
+[HasClient](src/Servant/Client/Core/Internal/HasClient.hs) module.
```
phadej: this should be absolute link (README is shown on Stackage and Hackage).

But we'll know the address only after merge :)

servant-client-core/src/Servant/Client/Core.hs
```diff
@@ -0,0 +1,597 @@
+-- | This module provides backend-agnostic functionality for generating clients
+-- from @servant@ APIs. By "backend," we mean something that concretely
+-- executes the request, such as:
+--
+--  * The 'http-client' library
```
phadej: `@http-client@`, quotes are for identifiers!

servant-client-core/src/Servant/Client/Core.hs
```diff
@@ -0,0 +1,597 @@
...
+--  * The 'haxl' library
+--  * GHCJS via FFI
+--
+-- etc.
+--
+-- Each backend is encapsulated in a monad that is an instance of the
+-- 'RunClient' class.
+--
+-- This library is primarily of interest to backend-writers, who are encouraged
+-- to re-export the parts of the
```
phadej: truncated?

servant-client-core/src/Servant/Client/Core.hs
```diff
@@ -0,0 +1,597 @@
...
+                                                         parseBaseUrl,
+                                                         showBaseUrl)
+import           Servant.Client.Core.Internal.BasicAuth
+import           Servant.Client.Core.Internal.HasClient
+import           Servant.Client.Core.Internal.Generic
+import           Servant.Client.Core.Internal.Request
+import           Servant.Client.Core.Internal.RunClient
+
+{-
+import           Control.Monad.Error.Class              (throwError)
```
phadej: what's this big block of commented code?

servant-client-core/src/Servant/Client/Core/Reexport.hs
```diff
@@ -0,0 +1,30 @@
+-- | This module is a utility for @servant-client-core@ backend writers. It
+-- contains all the functionality fron @servant-client-core@ that should be
```
soenkehahn: `s/fron/from/`

servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
```diff
@@ -0,0 +1,47 @@
...
+import qualified Data.Text                            as T
+import           Network.HTTP.Media                   (MediaType, matches,
+                                                       parseAccept, (//))
+import           Servant.API                          (MimeUnrender,
+                                                       contentTypes,
+                                                       mimeUnrender)
+import           Servant.Client.Core.Internal.Request (Request, Response (..),
+                                                       ServantError (..))
+
+class (MonadError ServantError m) => RunClient m where
```
phadej: I'd make a `class AsServantError e where buildServantError :: ServantError -> e`, or actual `Prism`

servant-client-core/src/Servant/Client/Core/Reexport.hs
```diff
@@ -0,0 +1,30 @@
...
+-- re-exported.
+module Servant.Client.Core.Reexport
```
phadej: Great idea :+1:

servant-client/servant-client.cabal
```diff
@@ -12,7 +12,7 @@ license:             BSD3
 license-file:        LICENSE
 author:              Servant Contributors
 maintainer:          haskell-servant-maintainers@googlegroups.com
-copyright:           2014-2016 Zalora South East Asia Pte Ltd, Servant Contributors
+copyright:           2014-2017 Zalora South East Asia Pte Ltd, Servant Contributors
```
phadej: does zalora people have copyright over 2017?

stack-ghc-8.2.1.yaml
```diff
@@ -1,6 +1,7 @@
-resolver: nightly-2017-09-01
+resolver: nightly-2017-09-13
```
phadej: sorry, my rebase comment was triggered by this.

I personally don't like bumping resolvers often (and tend to prefer beginning of the months. But that's just me, we don't need to have any policy on this).

jkarni: Fair enough. I'll switch back.

servant-client/servant-client.cabal
jkarni: I can't say I understand this. Happy either way.

servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
jkarni: And then `(MonadError e m, AsServantError e)`?

servant-client-core/src/Servant/Client/Core/Internal/Request.hs
```diff
@@ -0,0 +1,114 @@
...
+import           Network.HTTP.Media      (MediaType)
+import           Network.HTTP.Types      (Header, HeaderName, HttpVersion,
+                                          Method, QueryItem, Status, http11,
+                                          methodGet)
+import           Web.HttpApiData         (ToHttpApiData, toEncodedUrlPiece,
+                                          toHeader)
+
+-- | A type representing possible errors in a request
+--
+-- Note that this type substially change in 0.12
```
abailly: `substially change` -> `substantially changed`

phadej:
`ClientLike` is `Generic` stuff. http://hackage.haskell.org/package/servant-client-0.11/docs/Servant-Client-Generic.html

servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
phadej: Ah, you need to catch errors too. Then one would need `Prism`.

It would be nice to have `(MonadError e m, AsServantError e)` as then one doesn't really need `enter` in client side, as one could make its `MegaMonad` an instance of `RunClient`. But to be composable we need `AsServantError` (as `MegaMonad` might want to throw and catch other errors).

I'm completely ok to depend on `lens` (or only `profunctors` to get `Prism`s). But we can do that change later.

servant-client/servant-client.cabal
phadej: 2014-2016 Zalora, 2016-2017 Servant Contributors imho is closer to truth (not sure about the transition year)

servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
jkarni: The other option is to require `MonadCatch e` instead of `(MonadError e m, AsServantError e)`.

jkarni:
(Note that the last commit also parametrized `Request` by the path type to make it easier to work with, and to write instances for)

servant-server/servant-server.cabal
jkarni: :+1: on making `sdist` part of CI. From the sound of it `-Wno-missing-home-modules` is the way to go for now.

jkarni:
I think the only thing missing (from review so far) is deciding on `MonadError` vs `MonadCatch`. I'm leaning towards `MonadCatch`, since it makes it possible for many existing monads (including `IO`!) to be made instances of `RunClient` in a pretty sensible way.

phadej:
third option is to have own `throw` and `catch` (with default implementation in terms of either one)

jkarni:
@phadej as in, a new `MonadCatch`-like class? Why?

alpmestan:
I'm not a big fan of `MonadThrow`/`MonadCatch` in this case as a `MonadThrow m`/`MonadCatch m` constraint doesn't tell you what kind of things are thrown/catched. Especially since we know the error type here.

servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
phadej:
```haskell
class RunClient m where
  runRequest :: Request -> m Response
  throwServantError :: ServantError -> m ()
  catchServantError :: m a -> (ServantError -> m a) -> m a
```

`IO` can be instance of this, but we still tell what we `throw`/`catch`

jkarni: :+1: - done!

jkarni:
Can I merge? I've addressed all issues raised during review.

abailly:
+1

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -88,6 +88,8 @@ instance Alt ClientM where
 
 instance RunClient ClientM where
   runRequest = performRequest
+  throwServantError = throwError
+  catchServantError = catchError
```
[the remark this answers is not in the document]
jkarni: I'm not using it right now, but think I will soon, and also think it's generally useful. E.g. retrying on connection error in a backend-agnostic way.

jkarni:
Thanks @phadej @abailly !

PR #804, opened 2017-09-08
Add hoistServer to HasServer

phadej:
This is a refinement of an idea by @paf31.

I'll do documentation pass and add tests, if the idea is OK.


phadej:
I'd only deprecate `enter`, and keep it around for few major releases

phadej:
@alpmestan, should I rewrite whole `Enter` section in the tutorial (i.e. only mention it existed?)

alpmestan:
@phadej Well, we definitely want to explain the new stuffs. And maybe indeed add a little sentence or two that says "it replaces `Enter`, just do this to port your `enter`-based code over to `hoistServer`" ?

phadej:
@alpmestan I'll merge this now and update tutorial in separate PR (as I  guess there will be much more to review).

alpmestan:
@phadej very good idea :)

PR #809, opened 2017-09-20
servant-foreign: Derive Data for all types

ocharles:
I would like to be able to find all "foreign" types under the list of requests generated by `listFromApi`. With a `Data` instance, it's trivial to repeatedly walk the `[Req MyLangType]`, and use `cast` to see if I'm looking at a `MyLangType` or not.

I also moved some `StandaloneDeriving` clauses to just a normal `deriving` clause. If there was a reason you want these to remain standalone I can change it back.

### DIFF
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -41,20 +41,21 @@ library
   if impl(ghc >= 8.0)
     ghc-options: -Wno-redundant-constraints
   include-dirs: include
-  default-extensions:  CPP
-                     , ConstraintKinds
+  default-extensions:  ConstraintKinds
+                     , CPP
                      , DataKinds
+                     , DeriveDataTypeable
                      , FlexibleContexts
                      , FlexibleInstances
                      , GeneralizedNewtypeDeriving
                      , MultiParamTypeClasses
+                     , OverloadedStrings
+                     , PolyKinds
                      , ScopedTypeVariables
                      , TemplateHaskell
                      , TypeFamilies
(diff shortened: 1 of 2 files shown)


alpmestan:
That should be absolutely fine, LGTM. @phadej ?

servant-foreign/servant-foreign.cabal
```diff
@@ -49,7 +49,6 @@ library
                      , GeneralizedNewtypeDeriving
                      , MultiParamTypeClasses
                      , ScopedTypeVariables
-                     , StandaloneDeriving
                      , TemplateHaskell
```
phadej: Add `DeriveDataTypeable` here?

ocharles:
Done.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -9,6 +9,7 @@ module Servant.Foreign.Internal where
 
 import           Control.Lens (makePrisms, makeLenses, Getter, (&), (<>~), (%~),
                                (.~))
+import           Data.Data
```
phadej: seems the unrestricted import breaks `-Werror` as `Data.Data` exports `Proxy` too. Could you use explicit import lists, maybe then travis will be happy?

jkarni:
I'm not sure I understand the 7.8.4 CI failure, but a guess would be adding `Typeable` as well?

phadej:
I'll take ownership on this. Will correct & merge soon.

phadej:
superseded by #846

PR #814, opened 2017-09-22
Fix up .stylish-haskell.yaml

ivan-m:
The `records` attribute is no longer valid.

Not sure if this is quite the config you want to use (as it doesn't
match all the code present).

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -177,7 +177,7 @@ columns: 80
 # - crlf: Convert to CRLF ("\r\n").
 #
 # Default: native.
-newline: native
+newline: lf
 
 # Sometimes, language extensions are specified in a cabal file or from the
 # command line instead of using language pragmas in the file. stylish-haskell
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -15,6 +15,14 @@ steps:
   #     # true.
   #     add_language_pragma: true
 
+  # Align the right hand side of some elements.  This is quite conservative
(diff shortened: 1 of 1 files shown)


alpmestan:
Someone else might want to chime in, as I personally am not a huge fan (nor user) of stylish-haskell. If nobody does, let's just merge it.

.stylish-haskell.yaml
```diff
@@ -67,6 +167,18 @@ steps:
...
+# preferred format here.
+#
+# - native: Native newline format. CRLF on Windows, LF on other OSes.
+#
+# - lf: Convert to LF ("\n").
+#
+# - crlf: Convert to CRLF ("\r\n").
+#
+# Default: native.
+newline: native
```
phadej: `lf`, we don't want windows users screwing up things.

3noch:
If not everyone uses this then contributors will either,

  a) ignore it and get the styling wrong
  b) use it, but make big diffs because not everyone uses it.

I personally use stylish Haskell and like it. But I've also been bitten by it while contributing to other projects.

ivan-m:
Either the config needs to be updated or removed; currently it means automated application keeps breaking.

3noch:
Oh right. I neglected to notice that this was an update, not a new file. Right on.

ivan-m:
I missed #680 when I wrote this, so one of these should be closed (on phone so can't easily inspect differences).

jkarni:
Closing after #680 got merged (let us know if there are things you think are still missing). Sorry for the slow turnaround.

PR #818, opened 2017-09-24
servant-client-ghcjs

FPtje:
UPDATE: Now works, but tests have to be integrated into CI.

It's been silent for a bit, but let's get this show back on the road. This PR serves mainly as a starting point to get ghcjs support in servant. The [last PR](https://github.com/haskell-servant/servant/pull/610) crashed on three things:

- The desire for the ghcjs bit to be separated from the regular `servant-client`
- The lack of tests
- Inactivity (sorry for that)

With @jkarni implementing [#803](https://github.com/haskell-servant/servant/pull/803), the first one got quite a bit easier. Hence this PR.

Also pinging @arianvp

Here are some of the things that need to be discussed or done:
- [x] Tests, and I really mean tests. I honestly don't know if *any* of the code works yet.
- [x] ~~Continuous integration~~
- [x] Naming. Is e.g. `GhcjsClientM` a good name?
- [x] Documentation
- [x] General style, there *probably* is a lot to improve here, since it's largely a copy paste job from the previous PR.
- [x] The stuff with `tinc`, I'm not sure how that works with travis.
- [ ] Other feedback. I don't hack on servant very often. You might want to do things differently.

A final note: I just pulled from master and started porting things over from the previous PR. That means the commits of that PR are lost. Is that OK?


servant-client-ghcjs/src/Servant/Client/Internal/XhrClient.hs
```diff
@@ -0,0 +1,241 @@
...
+  -- branching between node (for testing) and browsers
+  "(function () {if (typeof require !== 'undefined') { return require('xhr2'); } else { return XMLHttpRequest; };})()"
+  requireXMLHttpRequestClass :: IO JSXMLHttpRequestClass
+
+foreign import javascript unsafe "new $1()"
+  newXMLHttpRequest :: JSXMLHttpRequestClass -> IO JSXMLHttpRequest
+
+-- * performing requests
+-- Performs the xhr and blocks until the response was received
+performXhr :: JSXMLHttpRequest -> BaseUrl -> Request -> IO ()
```
jkarni: A lot of the FFI calls and `xhr` functionality seems to be present in [ghcjs-base](https://github.com/ghcjs/ghcjs-base/blob/master/JavaScript/Web/XMLHttpRequest.hs). Is there a reason not to use that?

servant-client-ghcjs/servant-client-ghcjs.cabal
```diff
@@ -0,0 +1,54 @@
...
+    Servant.Client.Ghcjs
+    Servant.Client.Internal.XhrClient
+  build-depends:
+      base                  >= 4.7      && < 4.11
+    , bytestring            >= 0.10     && < 0.11
+    , case-insensitive      >= 1.2.0.0  && < 1.3.0.0
+    , containers            >= 0.5      && < 0.6
+    , exceptions            >= 0.8      && < 0.9
+    , ghcjs-base
+    , ghcjs-prim
```
jkarni: Version bounds here too.

jkarni:
I'm okay with commits having been lost if you are. The style looks ok. 

I think I'm in favor of the monad being called `ClientM` as well, so that this lib becomes very nearly a drop-in replacement for `servant-client` (they're never going to be used together), though I have no strong preferences.

I'll think a little more about tests...

alpmestan:
+1 from me for naming the monad `ClientM`, just like in the http-client backed variant.

servant-client-ghcjs/src/Servant/Client/Internal/XhrClient.hs
FPtje: It's possible, but status text would be lost, as its API doesn't support getting the status text. Besides there would have to be some conversion between servant and ghcjs' types (Request, Response, the exceptions). Using ffi is somewhat more direct.

FPtje:
- Name changed to `ClientM`
- Added version bounds for `ghcjs-base` and `ghcjs-prim`

phadej:
Before this get merged, I'd like to see CI testing GHCJS-stuff (not it's green, but that doesn't tell anything). If @FPtje doesn't know how to, please say about that early (btw, we don't use `tinc` in CI, so it's irrelevant).

FPtje:
I was planning on looking at what tests servant-client has. Maybe some of them apply to servant-client-ghcjs.

arianvp:
One problem I had last time with getting GHCJS working is that generating the GHCJS files (it used to be called booting or something?) takes longer than the amount of time Travis gives you. It kills the process before even finishing.

saurabhnanda:
> One problem I had last time with getting GHCJS working is that generating the GHCJS files (it used to be called booting or something?) takes longer than the amount of time Travis gives you. It kills the process before even finishing.

Random comment -- please see if you can ping @alexfmpe He had struggled quite a bit to get GHCJS working with Travis for the JS benchmarks. He *might* be able to help.

phadej:
@arianvp looks so :( https://github.com/haskell-servant/servant/pull/825

fizruk:
@phadej can we switch to Circle CI then?
I think I have managed to build GHCJS there a year ago or so.

phadej:
@fizruk, I'll investigate whether we can get binary GHCJS, so we don't need to build it on CI. For the greater good :)

arianvp:
I think `stack` nowadays can also ship GHCJS precompiled...

phadej:
@arianvp At least their documentation doesn't advocate that.

arianvp:
https://docs.haskellstack.org/en/stable/ghcjs/

phadej:
@arianvp yes, that what I used in #825. It still bootstraps dependencies (as you can see in 825 travis CI), which takes a long time.

fisx:
> https://docs.haskellstack.org/en/stable/ghcjs/

https://github.com/matchwood/ghcjs-stack-dist

this worked better for us for a while, now we're using https://github.com/liqula/nixpkgs, which gives you a more recent, less buggy ghcjs than the upstream nixpkgs repo.

3noch:
I believe `reflex-platform` also pulls in a custom version of ghcjs via nix.

alexfmpe:
> Random comment -- please see if you can ping @alexfmpe He had struggled quite a bit to get GHCJS working with Travis for the JS benchmarks. He might be able to help.

Yeah, Travis can be a major pain on several fronts and we eventually switched to CircleCI on the benchmarks repo.
I'd only recommend using Travis if you need macOS builds. 

There's some very annoying workarounds for Travis timeouts if you *really* need to. 
- [Build stages](https://docs.travis-ci.com/user/build-stages/) (each stage has its own timeout, [no implicit persistence](https://docs.travis-ci.com/user/build-stages/), unlike CircleCI's [workflows](https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs))
- [Incrementally caching the build](https://chriscummins.cc/2016/incremental-travis-builds/)

CircleCI only timeouts if there's no output for a while. And you can [set the timeout](https://circleci.com/docs/1.0/configuration/#modifiers)! 
Also, I had little trouble caching `/nix` (sample [config](https://github.com/alexfmpe/reflex-platform/blob/035af6065d18f3835de1f8c0e5e60dd27d94a623/.circleci/config.yml) & [build](https://circleci.com/gh/alexteves/reflex-platform/63#config/containers/0)) yet never managed to with Travis.

> I believe reflex-platform also pulls in a custom version of ghcjs via nix.

Yop, and even there Travis' timeouts are an everyday thing, unless pretty much everything is cached. Might also switch to CircleCI soon.
Pinging @ryantrinkle, since I didn't mess with ghcjs directly in CI, always used reflex-platform's scripts.

imalsogreg:
Is there a reason not to use `jsaddle` and `jsaddle-dom`'s xhr machinery? This may give users free ghc compatibility.

alexfmpe:
I tried adapting #825 to circleci ([config](https://github.com/alexfmpe/servant/blob/92eda542d0e6aeac2a4a8ea694f183438682116f/.circleci/config.yml) / [build](https://circleci.com/workflow-run/0f3acdb2-c3ff-4353-bea5-43e9f3be2e79)).

This is very much a WIP, but I could use feedback:
- still needs caching
- don't know what to make of ghcjs's build error
- stack-ghc-8.2.1 errs on Wmissing-home-modules; I suspect something in the `if []` is escaped/evaluated wrong

ryantrinkle:
@imalsogreg Not sure if this is relevant, but reflex-dom originally had its own XHR support due to some limitations in jsaddle.  My guess is that those limitations are now gone, and so we can probably eventually get rid of all traces of the original implementation and go with jsaddle completely.

FPtje:
Wow, awesome that so much is being discussed about it. I just pushed a commit fixing some headers not being set correctly.

With CI I see that I'm a bit in over my head. I thought the biggest challenge was to write the tests, not getting ghcjs to compile the tests...

FPtje:
I just wrote a really dirty test: https://github.com/FPtje/servant-client-ghcjs-test

It's run manually, you run the command to run the server, go to `localhost:8000` in the browser and open up the console to view the test results.

It's mostly a copy paste job from the tests in `servant-client`, but with the server and client separated and some other changes. Obviously we'd like to have it integrated with Travis or Circle CI, but I see there's still some discussion going on there. The test repo could act as a great starting point, though, as it shows what's needed to test `servant-client-ghcjs`: a server and a browser.

With ce35514 the test completely passes too. That means `servant-client-ghcjs` works! Woohoo!

FPtje:
By the way, we've noticed that we always use the following function for ghcjs servant calls:

```haskell
-- | Performs blocking AJAX request on the location of the browser window
callServant
    :: String
       -- ^ Path prefixed to HTTP requests.
    -> ClientM a
    -> IO (Either ServantError a)
callServant path m = do
    curLoc <- getWindowLocation

    jsStr_protocol <- getProtocol curLoc
    jsStr_port     <- getPort     curLoc
    jsStr_hostname <- getHostname curLoc

    let protocol
          | jsStr_protocol == "https:" = Https
          | otherwise                  = Http

        portStr :: String
        portStr = JSS.unpack jsStr_port

        port :: Int
        port | null portStr = case protocol of
                 Http  ->  80
                 Https -> 443
             | otherwise = read portStr

        hostname :: String
        hostname = JSS.unpack jsStr_hostname

    runClientM m (ClientEnv (BaseUrl protocol hostname port path))
```

With cross domain restrictions in browsers, all requests *must* necessarily go to the host that served the page. I reckon this should be built in `servant-client-ghcjs` somehow. Some options:

- Build into `runClientM`, making it impossible to run requests on other hosts (which should be disallowed anyway)
- Make the above function available, leaving the regular `runClientM` as it is

What do you think?

3noch:
> all requests must necessarily go to the host that served the page

This is not true with CORS. You can make requests to other hosts that have CORS enabled and you can even set `withCredentials` to `True` for cross-domain requests with authorization headers and cookies.

jkarni:
We're considering switching CI services, or substantially reworking our tests, for this. 

That is by no means necessarily bad - possibly the changes are themselves good. But I want to point out that it's absolutely an option, and I think a good one, to just have `servant-client-ghcjs` in a separate repo.

arianvp:
I agree with @jkarni here.

I think the fastest way to get this released is to put it in a separate repo under the `haskell-servant` Group.

This way, we don't slow down on bikeshedding about CI.  we can already release it, and then iterate on the CI choice. Even have a different CI choice than the main Servant repo.  

Would be sad to see this linger around for too long, as I think it would be a useful addition for many people.

phadej:
Ok: let's not do anything related to CI. Let's merge as soon as there are documentation, and @jkarni is happy. After that I'll separate the package into separate repository

FPtje:
Separate repo sounds like a good idea. For documentation I had the following ideas:

- Highlighting the (minor) differences between `servant-client` and `servant-client-ghcjs`
- How to create client functions that work for ghc and ghcjs. This is specifically useful for bigger projects.

alexfmpe:
> Even have a different CI choice than the main Servant repo.

I think you can have config files for two CI services in the same repo (e.g. keep the travis as it is and just build `servant-client-ghcjs` on the circleci one).

Not implying that's the only reason to separate things.

FPtje:
Where can I add documentation? Should I create a `README.md` or integrate it in `servant-docs`

arianvp:
I would just add a README for now with an example or two? We could later add examples to https://github.com/haskell-servant/servant/tree/master/doc  which is the website documentation.

alpmestan:
Yes, you can start adding docs wherever appropriate and when you/we're ready to make this a first class citizen in the ecosystem ("core package", so to speak), then we will want a dedicated section in the tutorial, indeed.

FPtje:
I've renamed `runClientM` to `runClientMOrigin` (not confident about that name), and added a new `runClientM`  which doesn't have the `ClientEnv` parameter. For browser requests this seems like the sensible default.

I've also started work on the README. I'll continue working on that tomorrow.

FPtje:
Rebased on latest master, updated and re-ran the [test suite](https://github.com/FPtje/servant-client-ghcjs-test) and finished the concept of the [README](https://github.com/LumiGuide/servant/blob/feat-client-ghcjs/servant-client-ghcjs/README.md).

Some open questions:

- Is the `runClientM` `runClientMOrigin` change the right idea? Different name for `runClientMOrigin` perhaps?
- Does the README explain the concepts well enough?

I've made some specific choices in the README, like using `PackageImports` to indicate where the imports are coming from. Imho they are beneficial to the understandability of the documentation, but they may not fall in line with servant documentation standards. I'm open for discussion there.

servant-client-ghcjs/README.md
```diff
@@ -0,0 +1,176 @@
+# servant-client-ghcjs
+
+Type safe querying of servant APIs from the browser.
+
+`servant-client-ghcjs` is much like `servant-client`, as both packages allow you to generate functions that query the endpoints of your servant API. Both packages should feel the same in usage. The big difference lies in how they perform the actual requests. `servant-client` (indirectly) uses your operating system's socket mechanisms, whereas `servant-client-ghcjs` uses your browser's [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest) mechanisms to send requests.
+
+This guide assumes knowledge of servant. Reading its [documentation](haskell-servant.readthedocs.io) is recommended if you're new to the subject.
```
alpmestan: Please add http[s]:// before the link here as e.g if viewed on github, it just sends the reader to some file named "haskell-servant.readthedocs.io" in the LumiGuide repo =)

servant-client-ghcjs/README.md
```diff
@@ -0,0 +1,176 @@
...
+{-# LANGUAGE TypeOperators  #-}
+
+module Main where
+
+import "aeson" Data.Aeson
+import "base" Data.Proxy
+import "base" GHC.Generics
+import "servant" Servant.API                       -- To define the API itself
+import "servant-client-ghcjs" Servant.Client.Ghcjs -- To generate client functions
+
```
alpmestan: Do we really need package imports here? This might confuse readers more than anything.

servant-client-ghcjs/README.md
```diff
@@ -0,0 +1,176 @@
...
+In the example below, the client functions are put in a data type called `APIClient`, which has `m` as a type parameter. The lowercase `apiClient` constructs this data type, demanding that `m` is indeed a monad that can run requests.
+
+```haskell
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+...
+import "servant-client-core" Servant.Client.Core
```
alpmestan: ditto for package imports

alpmestan:
Oh, didn't notice your comment about package imports before bringing that up, sorry. I would definitely suggest not using the. It should be enough to explicitly say where things come from whenever there's a chance for the reader to be confused. Even on the import line itself: `import Servant.API -- from the 'servant' package`.

Other than than that, I think runClientMOrigin is okay, even though I'm sure we can find a better name. I just don't have one to suggest. =) The README looks good. We could end up basically "replicating" it in the tutorial, as an addendum to the -client page of the tutorial.

FPtje:
Thanks for the feedback! I'll fix stuff as soon as I get the chance!

FPtje:
Done

alpmestan:
Great. How about merging this, @haskell-servant/maintainers ?

alpmestan:
Or maybe there's some more work necessary, for the CI or something?

alpmestan:
@FPtje This PR doesn't include anything that checks that the package builds (not even mentionning tests) so far right?

FPtje:
@alpmestan Yes, see the discussion above. Since ghcjs seems to pose major problems for CI, the decision has been made *not* to include it for now. I've tested the package manually with [this repo](https://github.com/FPtje/servant-client-ghcjs-test).

arianvp:
Lets merge, and then split up in a separate package afterwards.

phadej:
I'll merge this right after I cut 0.12 branch.

alpmestan:
Good to merge now? Any objection?

Issue #821, opened 2017-09-28
Secure HTTP not Working in servant-client


limaner2002:
When experimenting with the new functionality of PR #803 I found that the `secure` field is not being set even though the `Scheme` is `Https`. I simply modified the `requestToClientRequest` function in `Servant.Client.Internal.HttpClient` and it works now. Below is the new version of `requestToClientRequest` that works for me now.

```
requestToClientRequest :: BaseUrl -> Request -> Client.Request
requestToClientRequest burl r = Client.defaultRequest
  { Client.method = requestMethod r
  , Client.host = fromString $ baseUrlHost burl
  , Client.port = baseUrlPort burl
  , Client.path = BSL.toStrict
                $ fromString (baseUrlPath burl)
               <> toLazyByteString (requestPath r)
  , Client.queryString = renderQuery True . toList $ requestQueryString r
  , Client.requestHeaders =
      let orig = toList $ requestHeaders r
      in maybe orig (: orig) contentTypeHdr
  , Client.requestBody = body
  , Client.secure = isSecure
  }
  where
    (body, contentTypeHdr) = case requestBody r of
      Nothing -> (Client.RequestBodyLBS "", Nothing)
      Just (RequestBodyLBS body', typ)
        -> (Client.RequestBodyLBS body', Just (hContentType, renderHeader typ))
    isSecure = case baseUrlScheme burl of
      Http -> False
      Https -> True
```

jkarni:
Ugh, my bad! Mind sending a PR?

limaner2002:
Just submitted a PR for this issue. By the way thanks for the work on #803. I think I will be able to leverage it in one of my projects.

jkarni:
Fixed in #822

PR #822, opened 2017-09-28
Fix issue #821

limaner2002:
Added the secure fields to the http-client request in the function `requestToClientRequest` as per issue #821


alpmestan:
@jkarni happy with this patch?

jkarni:
@limaner2002 thank you!

PR #825, opened 2017-10-02
Try if Travis can handle GHCJS

phadej:


phadej:
We have GHCJS job now.

PR #829, opened 2017-10-05
Multiple content types

ivan-m:
If an instance of `Accept` has multiple `contentTypes`, then previously each one was listed individually, even though the contents were identical.

This attempts to fix that, but isn't perfect: it groups identical rendered output together (which actually may be sufficient from a documentation point of view in indicating that there is no conceptual difference in how the values are represented).

Furthermore, to prevent generated documentation from becoming long, it is possible to specify whether all or just one mime-type of each example is shown.

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -298,17 +298,29 @@ defAction =
 single :: Endpoint -> Action -> API
 single e a = API mempty (HM.singleton e a)
 
--- | How many examples should be shown?
-data ShowContentTypes = AllContentTypes | FirstContentType
+-- | How many content-types for each example should be shown?
+--
+--   @since 0.11.1
+data ShowContentTypes = AllContentTypes  -- ^ For each example, show each content type.
+                      | FirstContentType -- ^ For each example, show only one content type.
   deriving (Eq, Ord, Show, Read, Bounded, Enum)
 
 -- | Customise how an 'API' is converted into documentation.
+--
+--   @since 0.11.1
 data RenderingOptions = RenderingOptions
   { _requestExamples  :: !ShowContentTypes
+    -- ^ How many content types to display for request body examples?
   , _responseExamples :: !ShowContentTypes
+    -- ^ How many content types to display for response body examples?
   } deriving (Show)
 
 -- | Default API generation options.
+--
+--   All content types are shown for both 'requestExamples' and
+--   'responseExamples'.
+--
+--   @since 0.11.1
 defRenderingOptions :: RenderingOptions
 defRenderingOptions = RenderingOptions
   { _requestExamples  = AllContentTypes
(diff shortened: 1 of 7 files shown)


alpmestan:
This appears to implement exactly what we have discussed previously, I'm quite happy with the overall patch.

For this to be merged, once CI passes, I'd just like to see:
- an update to the haddocks for `markdown` that says something like `markdown = markdownWith defOptions`
- haddocks for `markdownWith`, short description, maybe with a tiny teeny example?
- an update to the servant-docs section of the tutorial that mentions this new capability, and if you feel like it that illustrates it with some code?

Would you mind taking care of this or should I? (no guarantee of promptness here unfortunately)

ivan-m:
I can take care of it tomorrow (along with updating servant-pandoc to do the same).

Are you happy with the name of `ApiOptions`? Should it be `GenerationOptions` or `RenderingOptions`? (Bikeshed now before it's merged in!)

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -674,7 +722,10 @@ markdown api = unlines $
                   []        -> ["- No response body\n"]
                   [("", t, r)] -> "- Response body as below." : contentStr t r
                   xs        ->
-                    concatMap (\(ctx, t, r) -> ("- " <> T.unpack ctx <> " (`" <> cs (show t) <> "`)") : contentStr t r) xs
+                    formatBodies _responseExamples xs
+
+uncurry3 :: (a -> b -> c -> d) -> (a,b,c) -> d
```
ivan-m: This was not meant to be here; it was used in an earlier experiment then accidentally committed.

alpmestan:
Hmm, now that you mention it... `RenderingOptions` gets my vote.

ivan-m:
@alpmestan I think that covers everything.

I'd like to have `servant-pandoc` match the functionality of `markdown` wrt `RenderingOptions`; if this PR is OK, any idea as to timelines for a new release?

alpmestan:
@ivan-m Terrific, the haddocks changes look great, thanks! Re: release schedule, it pretty much depends on how urgently people need patches, and on the amount of work left to do to accomodate for big changes (if any). @phadej how soon do you think we could cut a new release?

phadej:
@alpmestan I actually planned to go thru changes and update changelogs (will do next week). If nothing "this must be fixed before proper release" is found, we can have release shortly after HaskellX

ivan-m:
Oh, right, it actually tries to run my example when there are no instances.

alpmestan:
@ivan-m Great, you got CI to pass. Can I merge or is there something else you wanted to commit before I do?

ivan-m:
I think it's ready.  I'm now just duplicating the functionality in servant-pandoc, but it appears ready to go at this time.

Anything else I'll do another PR.

alpmestan:
Thanks!

PR #830, opened 2017-10-06
Add circleci config

alexfmpe:
Might be useful as a base for what #825 is trying to do. 

Noteworthy:
- Took out [`travis_retry`](https://github.com/haskell-servant/servant/blob/master/.travis.yml#L20)
- Took out the [`$TRAVIS_EVENT_TYPE`](https://github.com/haskell-servant/servant/blob/master/.travis.yml#L27) bit.
- Kept the `tinc` download and caching, but it doesn't seem to be used (at least ~/.tinc/cache was empty).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,47 @@
+aliases: &build
+    machine: true
+    steps: 
+      - checkout
+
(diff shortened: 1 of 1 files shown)


alpmestan:
@haskell-servant/maintainers Someone else than me will have to look into this.

phadej:
If the only benefit is to support longer build-times, I'll procrastinate on this for some time.

Issue #551, opened 2016-07-13
servant-foreign does not support all basic Content-Types


fizruk:
`PlainText`, `FormUrlEncoded` and `OctetStream` are available with basic `servant` package.
Would it be difficult to add support for those?

I have `PlainText` and `(CSV', DefaultOpts)` (from `servant-cassava`) in my API, however, I don't need JS generation for those parts right now.

jkarni:
`servant-client` doesn't either. 

Should we have an extra argument to the functions (the content-type)? In `servant-client` we could do this in a type-safe way - i.e., you're not allowed to request content-types that are not available.

fizruk:
What I meant was `servant-client` works for `Get '[PlainText] Text` while `servant-foreign` does not.

jkarni:
Ah, I see.

phadej:
Doesn't we have `[PlainText]` in `comprehensiveAPI`, or doesn't we have test `HasForeign ComprehensiveAPI`?

phadej:
Related #509

phadej:
Also https://github.com/haskell-servant/servant/issues/290

NorfairKing:
+1

fosskers:
I'm currently hit by this in trying to generate Elm code for my servant API with `servant-elm`. My issue is coming from my use of a custom `Accept` instance, a la:

```haskell
-- | The Content-Type @text/html@.                                                                        
data HTML

instance Accept HTML where
  contentType _ = "text" M.// "html"

instance MimeRender HTML B.ByteString where
  mimeRender _ = id
```
I'm doing this to serve pre-built html files as `Content-Type: text/html`. I realize I could use `Raw` for this, but I'm trying to be clever and serve these prebuilt files from custom endpoints (not following the paths matching `Raw`). For instance, the above allows me to make the endpoint `/foo/bar/` and serve a `ByteString` held in RAM to be interpreted as `text/html`.

Unfortunately this means that `elm-servant` (via `elm-export` via `servant-foreign`) can't generate Elm code for my API. What's my recourse?

fosskers:
For what it's worth, the compile-time error is:
```
No instance for servant-foreign-0.9.1.1:Servant.Foreign.Internal.NotFound
	arising from a use of ‘generateElmForAPI’
```
And `NotFound` [doesn't seem to be exposed](https://hackage.haskell.org/package/servant-foreign-0.10/docs/doc-index-N.html) to try to hack together an instance.

PR #832, opened 2017-10-09
Adding instances for non-JSON content types to `servant-foreign`

### LINKED ISSUE #509: (servant-foreign / servant-js) missing instances for PlainText content-type
ghc gives
```
No instance for servant-foreign-0.6.1:Servant.Foreign.Internal.NotFound
  arising from a use of ‘jsForAPI’
In the second argument of ‘($)’, namely ‘jsForAPI api vanillaJS’
In the first argument of ‘(:<|>)’, namely
  ‘(return $ jsForAPI api vanillaJS)’
In the second argument of ‘(:<|>)’, namely
  ‘(return $ jsForAPI api vanillaJS)
   :<|> (return $ decodeUtf8 . toStrict . encode . toSwagger $ api)’
```

when I try to make js output for ("api" :> Get '[PlainText] Text)

tmbull:
I am trying to fix #509 (and probably #551). I have only done the work for `Verb` thus far, but I wanted some feedback before I proceed with `ReqBody`. 

In this first approach, I have just modified `Req` to include a list of `MediaType` that comes from the Accept class. From the perspective of someone wanting to use `servant-foreign`, I'm not sure if this is the correct approach. It might be better to create a type similar to `HeaderArg` or create a typeclass separate from `Accept` specifically for this purpose.  Thoughts?

### DIFF
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -31,6 +31,7 @@ library
                      , Servant.Foreign.Internal
                      , Servant.Foreign.Inflections
   build-depends:       base       == 4.*
+                     , http-media == 0.7.*
                      , lens       == 4.*
                      , servant    == 0.11.*
                      , text       >= 1.2  && < 1.3
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -31,10 +31,10 @@ library
                      , Servant.Foreign.Internal
                      , Servant.Foreign.Inflections
   build-depends:       base       == 4.*
-                     , http-media == 0.7.*
                      , lens       == 4.*
                      , servant    == 0.11.*
                      , text       >= 1.2  && < 1.3
+                     , http-media
                      , http-types
   hs-source-dirs:      src
   default-language:    Haskell2010
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -34,7 +34,7 @@ library
                      , lens       == 4.*
                      , servant    == 0.11.*
                      , text       >= 1.2  && < 1.3
-                     , http-media
+                     , http-media >= 0.6  && < 0.8
                      , http-types
   hs-source-dirs:      src
   default-language:    Haskell2010
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -70,6 +70,10 @@ test-suite spec
                    , hspec >= 2.1.8
                    , servant
                    , servant-foreign
+  if !impl(ghc >= 8.0)
+    build-depends:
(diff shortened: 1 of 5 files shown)


alpmestan:
This seems quite reasonable to me, even though I have not followed servant-foreign closely for the past year, so maybe someone from @haskell-servant/maintainers more familiar with it than me can take a look?

tmbull:
Per @alpmestan's feedback, I've gone ahead and made the same changes to the `Req` type for `ReqBody`. The advantage I see of this approach is that it works for any type that implements `Accept`. Let me know what you guys think.

alpmestan:
I'm pretty happy with what I see in the patch. At this point maybe the other thing we should discuss and consider is: have you thought already about "the other side", i.e how easy or hard it is to use this for doing codegen in other languages in a way that's handy for people to use?

tmbull:
[Here ](https://github.com/tmbull/servant-elm/commit/ce848d0a81c251de781715f5913b0ca35ad4ce64#diff-7e13462cc66d19b6a0d2a0ddcd7e8864R384) is a link to my corresponding commit to `servant-elm` and here's a snippet: 
```
...
  expect =
      if
        any
        (\m -> m == F.contentType (Proxy :: Proxy F.JSON))
        (request ^. F.reqReturnContentTypes)
      then
        case request ^. F.reqReturnType of
...
```

Basically, I can just check if the expected content type is in the results. Calling `contentType` on the `Proxy` type allowed me to avoid using string literals. I do not have experience with any other codegen libraries that use `servant-foreign`, but for my purpose this seems reasonable to me. What do you guys think?

alpmestan:
Looks reasonable to me. Let's give a little bit more time (a few days) for people to complain about this patch or more generally give feedback, and then let's merge. Feel free to ping maintainers of servant-foreign powered libraries if you know of any, as it might make their lives a little easier to know about this change beforehand.

servant-foreign/servant-foreign.cabal
```diff
@@ -34,6 +34,7 @@ library
                      , lens       == 4.*
                      , servant    == 0.11.*
                      , text       >= 1.2  && < 1.3
+                     , http-media
```
jkarni: Could you include upper and lower bounds? (It's a mistake that we haven't done this for `http-types`.)

tmbull:
Sorry for the delay. I was pretty busy last week and didn't have a chance to update this.

@jkarni, I wasn't sure what to do about the version number in the CHANGELOG so I just incremented the current minor version by 1. If that's not the correct version number, let me know and I'll fix it.

alpmestan:
@jkarni @phadej Any other feedback?

phadej:
Let's have this as part of 0.12. As @jkarni said it's LGTM, sans changelog (which is now there). I'll merge this very soon.

phadej:
Superseded by #848

PR #843, opened 2017-10-27
Temporary fix for Header.

phaazon:
This fix exposes routes containing `Header _ a :> …` as `Maybe a …`. A typical problem is when you want to make request to such routes without using a header (typical when you’re working with the `Authorization` header, for instance).

I’m also making a PR to the `servant-elm` project.

### DIFF
--- a/servant-foreign/src/Servant/Foreign/Internal.hs
+++ b/servant-foreign/src/Servant/Foreign/Internal.hs
@@ -225,7 +225,7 @@ instance (Elem JSON list, HasForeignType lang ftype a, ReflectMethod method)
       method   = reflectMethod (Proxy :: Proxy method)
       methodLC = toLower $ decodeUtf8 method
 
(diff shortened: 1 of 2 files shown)


alpmestan:
@haskell-servant/maintainers Anyone knows whether this could be a problem for client libraries?

Note that this changes makes sense IMO because it would be consistent with what we do in -server and -client (see e.g https://hackage.haskell.org/package/servant-client-0.11/docs/src/Servant-Client.html#line-240).

phadej:
Tests are failing. @alpmestan it might be problem, we have `Header` in comprehensive API, so e.g. `servant-js` test-suite might fail (and we don't run it during our CI)...

phadej:
Also: please rebase on top of current master

phaazon:
I’m rebasing then.

alpmestan:
CI passes.

PR #848, opened 2017-11-06
Adding instances for non-JSON content types to `servant-foreign`

phadej:
Rebase of #832

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ release of `servant` package.
 | servant-cassava     | 0.7      | ?        | ?        |
 | servant-client      | 0.10     | 0.11     | 0.12     |
 | servant-docs        | 0.10     | 0.11     | 0.11.1   |
-| servant-foreign     | 0.10     | 0.10.0.1 | 0.10.2   |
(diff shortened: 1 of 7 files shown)


phadej:
- Now as I review this, I see that we generate foreign stuff also for non-JSON endpoints
- thus e.g. `servant-js` in current form will generate functions for non-JSON endpoints
- That would be simple fix to generate functions only for JSON endpoints
- but then we won't get **compile-time error**, if we forget `JSON` from some endpoint.

For me this sounds like a blocker. ping @jkarni.

phadej:
I'd postpone this from `0.12` preparations for now.

alpmestan:
We could quite simply make the JS stuffs available only if the endpoints all support JSON requests/responses, and give a custom type error otherwise?

exarkun:
I am interested in this change.  How can I help?

phadej:
https://github.com/haskell-servant/servant/pull/848#issuecomment-342126375 needs a solution

cdupont:
Hello, I hit this problem. Is there a solution? Thanks!

PR #851, opened 2017-11-06
MkLink: add instance for (:<|>)

roberth:
This lets you generate all links at once, which is useful in conjunction with servant-generic.


roberth:
There you go. I have add a test and I have added the `allLinks` function.
Otherwise you'd have to use `toLink` directly, which isn't fun because the empty link you need to provide is hard to obtain. `allLinks` does that for you without exposing the internals of the module, or decreasing the perceived type safety of `safeLink`.

roberth:
@phadej anything else I can do?

phadej:
Merged in #854, I added doctests.

PR #852, opened 2017-11-06
docs for streaming

gbaz:


doc/tutorial/Client.lhs
```diff
@@ -153,4 +153,47 @@ HelloMessage {msg = "Hello, servant"}
...
+``` haskell
+streamAPI :: Proxy StreamAPI
+streamAPI = Proxy
+
+posStream :: ClientM (ResultStream Position)
+
+posStream = client streamAPI
+```
+
+And here's how to just print out all elements from a `ResultStream`, to give some idea how to work with them.
```
alpmestan: to give some idea **of** how to work with them* ?

doc/tutorial/Server.lhs
```diff
@@ -1164,6 +1164,25 @@ app5 :: Application
 app5 = serve readerAPI (hoistServer readerAPI funToHandler funServerT)
 ```
 
+## Streaming endpoints
+
+We can create endpoints that don't just give back a single result, but give back a *stream* of results, served one at a time. Stream endpoints only provide a single content type, and also specify what framing strategy is used to deliniate the results. To serve these results, we need to give back a stream producer. Adapters can be written to `Pipes`, `Conduit` and the like, or written directly as `StreamGenerator`s. StreamGenerators are IO-based continuations that are handed two functions -- the first to write the first result back, and the second to write all subsequent results back. (This is to allow handling of situations where the entire stream is prefixed by a header, or where a boundary is written between elements, but not prior to the first element). The API of a streaming endpoint is needs to explicitly specify which sort of generator it produces. Note that the generator itself is returned by a `Handler` action, so that additional IO may be done in the creation of one.
```
alpmestan: "is needs" => "needs" ?

doc/tutorial/Server.lhs
```diff
@@ -1164,6 +1164,25 @@ app5 :: Application
...
+streamUsers :: StreamGenerator User
+streamUsers = StreamGenerator $ \sendFirst sendRest -> do
+                       sendFirst isaac
+                       sendRest  albert
+                       sendRest  albert
+
+app6 :: Application
+app6 = serve streamAPI (return streamUsers)
+```
+
```
alpmestan: Could you possibly "wrap up" the streaming section by a short description of what this simple example does? Something along the lines of "This simple application returns a stream of `User` values encoded in JSON format, where the encoding for two successive users is separated by a newline. The said stream's values are `isaac`, `albert` and `albert`".

alpmestan:
Oh wait, maybe CI is failing because this branch simply doesn't have the streaming stuffs :)

gbaz:
Changes made. I ran a spellcheck and so along the way fixed a few other misspellings that were caught as well.

alpmestan:
Lovely. Unless any major objection comes up, I'll merge both of your PRs sometimes in the upcoming days and we'll have streaming responses in the next release :)

phadej:
Let's not merge until CI is green.

alpmestan:
@gbaz Could you rebase the PR? If CI passes we're good to go.

gbaz:
done, merge?

alpmestan:
Wonderful, thanks :)

PR #866, opened 2017-11-26
servant-foreign: Expose ordered list of request parts.

mattjbray:
Allows libraries using `servant-foreign` (e.g. `servant-elm`) to generate client functions with argument lists in a consistent order.

See https://github.com/mattjbray/servant-elm/issues/30#issuecomment-347035299 for the motivation.

### DIFF
--- a/servant-foreign/src/Servant/Foreign.hs
+++ b/servant-foreign/src/Servant/Foreign.hs
@@ -5,6 +5,7 @@ module Servant.Foreign
   , HeaderArg(..)
   , QueryArg(..)
   , Req(..)
+  , ReqPart(..)
   , Segment(..)
   , SegmentType(..)
   , Url(..)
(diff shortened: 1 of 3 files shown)


alpmestan:
Hello!

Could you add a little bit of haddocks around the new data type, and perhaps even give an indication of how it can be used, maybe with a somewhat minimal example or something? I'm afraid that if we don't do that, we might end up being confused about that ourselves in a year or two =)

tchoutri:
@mattjbray Hi! Are you still willing to work on this? If so, it can be present on the next major release of Servant. :)

mattjbray:
@tchoutri apologies for the long silence. Unfortunately I'm not using servant/elm in my current work so I don't think I'll get around to this.

Will close. Feel free to reopen if someone else wants to pick it up.

PR #869, opened 2017-12-04
Bump servant-client-core upperbound in servant-client-ghcjs

cocreature:
I haven’t done particularly extensive testing against `0.12` but it definitely does build and in my limited testing it also seems to work just fine.


alpmestan:
Would be nice if https://github.com/haskell-infra/hackage-trustees/issues/119 was handled so that our CI would actually be useful...

phadej:
Please bump the lower-bound as well, we won't test anything in the repository against old versions of packages in this same repository.

phadej:
Also `servant-client-core-0.11` doesn't exist anyway :)

phadej:
I fixed Travis in `master`, could you rebase?

cocreature:
Looks like Travis is now happy after I’ve rebased.

phadej:
thanks

PR #871, opened 2017-12-06
Cookbook recipe for JWT with basic authentication

defanor:
Perhaps it can use more editing, but here's what I've got for now.

### DIFF
--- a/doc/cookbook/jwt-and-basic-auth/JWTAndBasicAuth.lhs
+++ b/doc/cookbook/jwt-and-basic-auth/JWTAndBasicAuth.lhs
@@ -1,0 +1,249 @@
+# Combining JWT-based authentication with basic access authentication
+
+In this example we will make a service with
+[basic HTTP authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)
+for Haskell clients and other programs, as well as
+with [JWT](https://en.wikipedia.org/wiki/JSON_Web_Token)-based
+authentication for web browsers. Web browsers will still use basic
+HTTP authentication to retrieve JWTs though.
+
+**Warning**: this is insecure when done over plain HTTP,
+so [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security)
+should be used.
+See [warp-tls](https://hackage.haskell.org/package/warp-tls) for that.
+
+While basic authentication comes with Servant itself,
+[servant-auth](https://hackage.haskell.org/package/servant-auth) and
+[servant-auth-server](https://hackage.haskell.org/package/servant-auth-server)
+packages are needed for the JWT-based one.
+
+This recipe uses the following ingredients:
+
+```haskell
+{-# LANGUAGE OverloadedStrings, TypeFamilies, DataKinds,
+  DeriveGeneric, TypeOperators #-}
+import Data.Aeson
+import GHC.Generics
+import Data.Proxy
+import System.IO
+import Network.HTTP.Client (newManager, defaultManagerSettings)
+import Network.Wai.Handler.Warp
+import Servant as S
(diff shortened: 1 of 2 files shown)


doc/cookbook/jwt-and-basic-auth/JWTAndBasicAuth.lhs
```diff
@@ -0,0 +1,168 @@
...
+
+port :: Int
+port = 3001
+```
+
+
+## Authentication
+
+That's how we'll represent a user: usually user identifier is handy to
+keep around, along with their role if RBAC is used, and other commonly
```
alpmestan: I would suggest expanding the acronym, it might not be obvious to everyone what this acronym means.

doc/cookbook/jwt-and-basic-auth/JWTAndBasicAuth.lhs
```diff
@@ -0,0 +1,168 @@
...
+import Control.Monad.IO.Class (liftIO)
+
+port :: Int
+port = 3001
+```
+
+
+## Authentication
+
+That's how we'll represent a user: usually user identifier is handy to
```
alpmestan: "Below is how  [...]" might be more natural here for the reader, as the example only comes _after_ this sentence.

alpmestan:
This is great, thanks!

I'm wondering whether we might want to switch to even a trivial db (an ioref with a Map, an sqlite db) to make the example "more real" and therefore show people how they can use a database for performing user lookup etc. Thoughts?

defanor:
Database usage was actually a thing I thought of as well, but more or
less proper one may become rather specific and out of scope of the
recipe (and of the Servant documentation, I guess), while using
something insecure and basic leaves a risk of a user just using it
exactly in the same way for their project (unfortunately, things like
that seem to happen rather often). Though the same applies to TLS, which
is even trickier to use for a runnable example. I've added warnings for
now.

Still, not sure if it would be more useful for an example than plain
pattern matching, while most likely it would complicate things a bit.
But no strong opinion here.

Meantime, applied your suggestions and added more links. Going to squash
everything into a single commit when it'll come to merging.

alpmestan:
This is looking better and better! Other than the worry about "how to pass a database connection to the auth check" (which is fairly simple once you know how to do it, but a lot of people just don't "see it" initially), I'm very very happy with this. Would you rather merge this as is and rely on another recipe (by you, me or someone else, later) to show how a DB connection could be threaded to the auth check?

defanor:
Oh, indeed: I've missed that.

So yes, I agree that it would be nice to add an example. Not sure if
actual pooling and DB connections should be used, or dummy ones: they
are not too verbose, though may be distracting a bit, and it would
slightly duplicate the pooling and DB recipes. What do you think would
be more appropriate?

alpmestan:
What I would personally recommend is either something that mimics the basic auth recipe that I wrote, which uses a "dummy" DB (a map), or the sqlite one. At least that would make the auth check get a new argument (a static user db in a map, an ioref to a map, or an sqlite connection) and show how in `main` you can then partially apply the auth check to your map/connection/etc and then call `serveWithContext` using a context augmented with your partially applied auth check. This is the key bit that a lot of people have been missing, so it would be _great_ if we could hint at how this should be done in general, with however simple or fancy a DB you want. Most if not all people should be able to figure out how to do it for their preferred DB backend and library (and then submit a new recipe that demonstrates it! :)).

If you however don't feel like tweaking your example, that's fine too of course! Also feel free to add your name somewhere if you want to.

defanor:
Here is a version with `Map`, and squashed.

alpmestan:
That's perfect, thank you very much!

Issue #862, opened 2017-11-14
Explicit Parsing Error for Headers


KtorZ:
Hi folks!

I am running into a simple issue: having custom headers on a request lead to have some `Maybe header` parameter in the corresponding handler function.

There are 3 scenarios resulting of parsing a header

- The header hasn't been provided
- The header didn't parse successfully
- The header is present and successfully parsed

Therefore, from within the `Handler`, I didn't find any way to know whether a header was wrongly specified or simply not specified. In the former, I want to return a clear error to the user to avoid silently swallowing the mistake. 

Headers of the response aren't accessible either, so it's not possible to lookup whether the header is provided and compare against the value of the handler's parameter to deduce whether the header didn't parse or was simply not provided.

Provided that the `FromHttpApiData` method has the following signature:

```hs
parseHeader :: ByteString -> Either Text a
```

Is there any reasons why the [Header type](http://hackage.haskell.org/package/servant-0.12/docs/Servant-API-Header.html#t:Header) isn't leveraged here since it seems to already make an explicit distinction between the 3 possible outcomes?

Thanks for helping.

alpmestan:
I have the feeling this might be somewhat related to https://github.com/haskell-servant/servant/pull/256 and https://github.com/haskell-servant/servant/issues/241 but I'll let @phadej and @fizruk comment on this, as they have given a lot more thought to these issues.

phadej:
Will be resolved in #873

phadej:
solved by #873

Issue #856, opened 2017-11-10
Modifiers for QueryParam, Header etc.


phadej:
```
13:49  phadej alp: what you think about QueryParam "name" '[Required, Lenient] Text, with possible modifiers Required/Optional; Lenient/Strict, whatever is first in the list wins
13:49  phadej I have thought about it for few days, you can add own "modifiers", but they won't affect how HasServer is implemented
13:50  phadej e.g. We could have '[Description "name of your pet"] there too, for servant-docs/-swagger to pick
13:50  phadej but we can make it so that HasServer/HasClient doesn't need to worry about those
13:51  phadej IMHO this is cleaner approach than (re-)inventing extensible records
13:51  phadej I could make a PoC soon for that
```

ping @alpmestan @fizruk

phadej:
Alternative variant is to have
```haskell
QueryParam "name" (Required :<> Lenient) Text
```
and have `ServerQueryParamMod` type-class, etc.

That would be more boilerplate, and  I don't really think we need to be open in this way for `QueryParam`. that one would like to change how `HasServer (QueryParam ... :> api)` works, at that point it's simpler&easier to write new completely new combinator.

alpmestan:
servant-js doesn't traverse APIs, it just uses servant-foreign for that and therefore doesn't care about servant's combinators per se, just the data about them that -foreign exports.

Issue #793, opened 2017-08-01
Optional request body


zudov:
It's sometimes useful to have an endpoint, that can either take the content as a body, or as query parameter (e.g. in the form of id that can be used to resolve that content).

A couple of examples:
- A `search` endpoint that can either take a simple query through parameters or a more complex query as json (elasticsearch does it)
- An endpoint that allows to submit a file can read a file from the body or download it from the url specified in the `url` paramater

I've been using the following combinator for my needs:

```haskell
data OptionalReqBody cts a

instance
  ( AllCTUnrender cts a
  , HasServer api context
  , reqBodyApi ~ (ReqBody cts a :> api)
  )
  => HasServer (OptionalReqBody cts a :> api) context where
  type ServerT (OptionalReqBody cts a :> api) m =
    Maybe a -> ServerT api m
  route _ context subserver =
    route @api Proxy context
      $ addBodyCheck subserver
          (withRequest contentTypeCheck)
          bodyCheck
    where
      contentTypeCheck
        :: Wai.Request
        -> DelayedIO (Maybe (LByteString -> Either String a))
      contentTypeCheck request =
        | hasMessageBody && not isEmptyMessageBody
            Just <$> getContentHandler
        | otherwise =
            pure Nothing
        where
          headers           = Wai.requestHeaders request
          hContentType      = headers & lookup Header.hContentType
          hContentLength    = headers & lookup Header.hContentLength
          hTransferEncoding = headers & lookup Header.hTransferEncoding

          -- https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html
          -- 4.3 Message body
          -- > The presence of a message-body in a request is
          -- > signaled by the inclusion of a Content-Length
          -- > or Transfer-Encoding header field in the request's
          -- > message-headers.
          hasMessageBody =
            isJust (hContentLength <|> hTransferEncoding)

          isEmptyMessageBody = fromMaybe False $ do
            contentLength <- parseContentLength =<< hContentLength
            pure $ contentLength <= 0
            where
              parseContentLength =
                (readMaybe :: String -> Maybe Double)
                  . Text.unpack . Text.decodeUtf8

          -- when Content-Type is not provided we assume it's octet-stream
          -- http://tools.ietf.org/html/rfc7231#section-3.1.1.5 RFC7231
          orOctetStream = fromMaybe "application/octet-stream"

          getContentHandler
            :: DelayedIO (LByteString -> Either String a)
          getContentHandler =
[...]

PR #873, opened 2017-12-10
Servant.API.Modifiers

phadej:
Implements modifiers #856.

### Combinators

- [x] QueryParam
- [x] Header
- [x] ReqBody

`QueryParams`? `Required` -> `NonEmpty a`?

### Implementations

- [x] servant (Links)
- [x] servant-client-core
- [x] servant-server
- [x] servant-docs
- [x] servant-foreign
- [x] Amend ComprehensiveAPI

### QA

-  `Capture` can also be lenient, but I'll leave it for now, but by definition they are always required.

There's some info in the commit messages.

### Resolves

- https://github.com/haskell-servant/servant/issues/862
- When `ReqBody` is amended, will solve https://github.com/haskell-servant/servant/issues/793
- Supersedes https://github.com/haskell-servant/servant/pull/256

### DIFF
--- a/servant-foreign/test/Servant/ForeignSpec.hs
+++ b/servant-foreign/test/Servant/ForeignSpec.hs
@@ -7,9 +7,11 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+#if __GLASGOW__HASKELL < 709
+{-# OPTIONS_GHC -fcontext-stack=41 #-}
+#endif
 #include "overlapping-compat.h"
 
 module Servant.ForeignSpec where
@@ -99,7 +101,7 @@ listFromAPISpec = describe "listFromAPI" $ do
     shouldBe postReq $ defReq
       { _reqUrl        = Url
           [ Segment $ Static "test" ]
-          [ QueryArg (Arg "param" "intX") Normal ]
+          [ QueryArg (Arg "param" "maybe intX") Normal ]
       , _reqMethod     = "POST"
       , _reqHeaders    = []
       , _reqBody       = Just "listX of stringX"
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -49,12 +49,13 @@ library
     Servant.API.HttpVersion
     Servant.API.Internal.Test.ComprehensiveAPI
     Servant.API.IsSecure
+    Servant.API.Modifiers
     Servant.API.QueryParam
     Servant.API.Raw
-    Servant.API.Stream
     Servant.API.RemoteHost
     Servant.API.ReqBody
     Servant.API.ResponseHeaders
+    Servant.API.Stream
     Servant.API.Sub
     Servant.API.TypeLevel
     Servant.API.Vault
@@ -77,6 +78,7 @@ library
     , mmorph                >= 1    && < 1.2
     , tagged                >= 0.7.3 && < 0.9
     , text                  >= 1    && < 1.3
+    , singleton-bool        >= 0.1.2.0 && <0.2
     , string-conversions    >= 0.3  && < 0.5
     , network-uri           >= 2.6  && < 2.7
     , vault                 >= 0.3  && < 0.4
--- a/servant/src/Servant/API/Header.hs
+++ b/servant/src/Servant/API/Header.hs
@@ -4,13 +4,15 @@
 {-# LANGUAGE PolyKinds          #-}
 {-# OPTIONS_HADDOCK not-home    #-}
 module Servant.API.Header (
-  Header(..),
-) where
+    Header, Header',
+    ) where
 
-import           Data.ByteString (ByteString)
 import           Data.Typeable   (Typeable)
 import           GHC.TypeLits    (Symbol)
+import           Servant.API.Modifiers
+
 -- | Extract the given header's value as a value of type @a@.
+-- I.e. header sent by client, parsed by server.
 --
 -- Example:
 --
@@ -18,10 +20,10 @@ import           GHC.TypeLits    (Symbol)
 -- >>>
 -- >>>            -- GET /view-my-referer
 -- >>> type MyApi = "view-my-referer" :> Header "from" Referer :> Get '[JSON] Referer
-data Header (sym :: Symbol) a = Header a
-                              | MissingHeader
-                              | UndecodableHeader ByteString
-    deriving (Typeable, Eq, Show, Functor)
+type Header = Header' '[Optional, Strict]
+
+data Header' (mods :: [*]) (sym :: Symbol) a
+    deriving Typeable
 
 -- $setup
 -- >>> import Servant.API
--- a/servant/src/Servant/API/Modifiers.hs
+++ b/servant/src/Servant/API/Modifiers.hs
@@ -1,0 +1,150 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE TypeOperators       #-}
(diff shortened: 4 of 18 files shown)


servant/servant.cabal
```diff
@@ -49,12 +49,13 @@ library
...
     Servant.API.Internal.Test.ComprehensiveAPI
     Servant.API.IsSecure
+    Servant.API.Modifiers
     Servant.API.QueryParam
     Servant.API.Raw
-    Servant.API.Stream
     Servant.API.RemoteHost
     Servant.API.ReqBody
     Servant.API.ResponseHeaders
+    Servant.API.Stream
```
alpmestan: Oh, nice one. I'm feeling better already =P

servant/src/Servant/Utils/Links.hs
```diff
@@ -101,8 +103,10 @@ module Servant.Utils.Links (
 import           Data.List
 import           Data.Monoid.Compat    ( (<>) )
 import           Data.Proxy            ( Proxy(..) )
+import           Data.Singletons.Bool  ( SBool (..), SBoolI (..) )
 import qualified Data.Text             as Text
 import qualified Data.Text.Encoding    as TE
+import           Data.Type.Bool (If)
```
jkarni: Alignment

servant/src/Servant/Utils/Links.hs
```diff
@@ -317,8 +330,33 @@ instance (ToHttpApiData v, HasLink sub)
...
+
+instance HasLink sub => HasLink (Summary s :> sub) where
+    type MkLink (Summary s :> sub) = MkLink sub
+    toLink _ = toLink (Proxy :: Proxy sub)
+
+instance HasLink sub => HasLink (HttpVersion :> sub) where
+    type MkLink (HttpVersion:> sub) = MkLink sub
+    toLink _ = toLink (Proxy :: Proxy sub)
+
+-- | /Note:/ the 'IsSecure' isn't visible in the resulting 'Link'.
```
jkarni: I don't entirely understand this note.

servant/test/Servant/Utils/LinksSpec.hs
```diff
@@ -75,11 +82,12 @@ spec = describe "Servant.Utils.Links" $ do
...
     it "can generate all links for an API that has only linkable endpoints" $ do
         let (allNames :<|> simple) = allLinks (Proxy :: Proxy LinkableApi)
-        simple
-          `shouldBeLink` "get"
-        allNames ["Seneca", "Aurelius"]
-          `shouldBeLink` "all/Seneca/Aurelius"
+        simple `shouldBeLink` "get"
+        allNames ["Seneca", "Aurelius"] `shouldBeLink` "all/Seneca/Aurelius"
 
+    it "can generate all inks for ComprehensiveAPIWithoutRaw" $ do
```
jkarni: typo ("links")

servant/src/Servant/API/Modifiers.hs
```diff
@@ -0,0 +1,150 @@
...
+--
+-- >>> :kind! FoldRequired '[Required, Optional]
+-- FoldRequired '[Required, Optional] :: Bool
+-- = 'False
+--
+-- >>> :kind! FoldRequired '[]
+-- FoldRequired '[] :: Bool
+-- = 'False
+--
+type FoldRequired mods = FoldRequired' 'False mods
```
jkarni: Hmm, so here, another option is to have a map:

``` [IsRequired := 'True, Description := "something"] ```

(And optionally, `type Required = IsRequired := 'True ; type Optional = IsRequired := 'False`)

Which simplifies the implementation here, insofar as instead of having separate `FoldX`s, you can just have a generic "Lookup".

File: servant/src/Servant/API/Modifiers.hs [hunk 6]
```diff
@@ -0,0 +1,150 @@
...
+-- | Fold a 'RequiredAgument' into a value
+foldRequiredArgument
+    :: forall mods a r. (SBoolI (FoldRequired mods))
+    => Proxy mods
+    -> (a -> r)        -- ^ 'Required'
+    -> (Maybe a -> r)  -- ^ 'Optional'
+    -> RequiredArgument mods a
+    -> r
+foldRequiredArgument _ f g mx =
+    case (sbool :: SBool (FoldRequired mods), mx) of
```
jkarni: Have you thought of making `SBool` an instance of `Reifies` (from `reflection`)? It seems like we might want non-boolean as well (e.g. `Description`), and then using a unified interface would be nice (rather than the one-off `SBoolI` class)

servant/src/Servant/API/Header.hs
```diff
@@ -4,24 +4,25 @@
...
 --
 -- >>> newtype Referer = Referer Text deriving (Eq, Show)
 -- >>>
 -- >>>            -- GET /view-my-referer
 -- >>> type MyApi = "view-my-referer" :> Header "from" Referer :> Get '[JSON] Referer
-data Header (sym :: Symbol) a = Header a
-                              | MissingHeader
-                              | UndecodableHeader ByteString
-    deriving (Typeable, Eq, Show, Functor)
+type Header = Header' '[]
```
jkarni: For these, should we maybe be explicit about `Optional` etc? That might serve as a little bit of documentation.

alpmestan: +1 on that.

servant/src/Servant/API/Modifiers.hs
phadej: Using map-like is an option.

TL;DR, current approach:
- pros: there isn't ill-kinded situations like `IsRequired := "true"`
- cons: slight boilerplate in the implementation (not use!)

IMHO pros outweight cons here.

---

I personally dislike "dynamic" feeling of `(:=) :: * -> k -> *`. Nothing really prevents you from saying `'[IsRequired := "true"]`. OTOH, having *typed* labels is a great improvements over `(:=) :: Symbol -> k -> *`, proposed earlier, as then typos in label name would get unnoticed.

Think about term level, we could have two APIs:

```haskell
defaultOptions &
  & required .~ True
  & description ?~ "Foo"
-- or
required <> description "Foo"
```

The type-level list approach is like the `Monoid` one.
Having all "type-safety" of lens one (value type is dependent on the key!) is tricky.
My gut feeling that in GHC `>= 8.0` one could encode that, but IMHO it's not worth that.

servant/src/Servant/API/Modifiers.hs
phadej: `reflection` isn't powerful enough. We need something to pattern match on (`SingI` from `singletons`, which I'd like to avoid for now as a dependency) to force `RequiredArgument mods a` into either `a` or `Maybe a`.

Note: the instance would be
```haskell
instance SBoolI b => Reifies Bool Bool where
    reflect _ = case sbool :: ... of STrue -> True; SFalse -> False
```

servant/src/Servant/Utils/Links.hs
phadej: `linkURI :: Link -> URI` should reflect the scheme, we use `IsSecure` combinator for links too, `f IsSecure = "https"; f NotSecure = "http"`.

alpmestan:
@taktoa I believe you were asking about this on #nixos-dev yesterday. Didn't find you on IRC so I'm pinging you here: this is the upcoming solution to the problem you were asking about.

taktoa:
@alpmestan Yup, I actually linked this in there. I was mainly asking because I thought someone might have an out-of-tree version of the code that I could use more easily than this patch, but I realized that I don't have a near-term need for the `HasServer` functionality, so I can just use the version that only has `HasClient` until this gets merged and upstreamed to Hackage.

servant-foreign/test/Servant/ForeignSpec.hs
```diff
@@ -7,9 +7,11 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
+#if __GLASGOW__HASKELL < 709
+{-# OPTIONS_GHC -fcontext-stack=41 #-}
```
alpmestan: :+1:

phadej:
Green. I'm positively surprised that `servant-js` isn't broken. I'll reread what I wrote tomorrow, and then merge.

PR #878, opened 2017-12-18
Fix #835. Use Escaped to prevent double-escaping

### LINKED ISSUE #835: double escape when creating link
In 54c30533425d4e9f05851df2bfbf316de80c7c56, a call to escape was added to linkURI to escape bad characters in the URL string.  The problem is that this causes a double escape sometimes because escape is also called in the HasLink instance.  That is, the HasLink instance for Capture is

```
instance (ToHttpApiData v, HasLink sub)
    => HasLink (Capture sym v :> sub) where
    type MkLink (Capture sym v :> sub) = v -> MkLink sub
    toLink _ l v =
        toLink (Proxy :: Proxy sub) $
addSegment (escape . Text.unpack $ toUrlPiece v) l
```

note the call to escape.  This causes the segment in Link to be escaped.  Then when it goes though linkURI, it gets escaped again.

For example,

```
type SomeRoute = "abc" :> Capture "email" Text :> Put '[JSON] ()
```

then

```
safeLink SomeRoute SomeRoute "test@example.com" 
   ~ Link { _segments = ["abc", "test%40example.com"]}
```

due to the escape in the HasLink instance.  But then calling linkURI turns it into

```
URI { uriPath = "abc/test%2540example.com" }
```

because the link segments are escaped again.

phadej:
- Reverting (and re-introducing different bug): *bad*
- Fixing for all: *good*

`Param` is unfortunately exposed, so I don't change it. This way this commit is backportable to 0.12.

### DIFF
--- a/servant/src/Servant/Utils/Links.hs
+++ b/servant/src/Servant/Utils/Links.hs
@@ -127,12 +127,24 @@ import Servant.API.Experimental.Auth ( AuthProtect )
 -- The only way of constructing a 'Link' is using 'safeLink', which means any
 -- 'Link' is guaranteed to be part of the mentioned API.
 data Link = Link
-  { _segments :: [String]   -- ^ Segments of "foo/bar" would be ["foo", "bar"]
+  { _segments    :: [Escaped]
   , _queryParams :: [Param]
   } deriving Show
 
+newtype Escaped = Escaped String
+
+escaped :: String -> Escaped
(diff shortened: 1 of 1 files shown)


alpmestan:
Just to be sure: is the doctest addition checking precisely that we're safe from #835 ?

alpmestan:
I just checked, it is.

PR #883, opened 2017-12-31
Support http-client’s CookieJar in servant-client

michalrus:
Could we consider this simple addition? Or something similar? :pray:

(This diff is just a draft to, hopefully, start a discussion.)

It’s incredibly useful when testing “cookied” endpoints (e.g. from https://github.com/zohl/servant-auth-cookie), when we don’t care about cookie values and want to just let them happen correctly behind the scenes (like a browser would do). /cc @zohl 

Perhaps, from Servant’s POV, it would be ideal if `Client.Manager` allowed this (and kept its own `Client.CookieJar`), but… ¯\\\_(ツ)\_/¯

----

Right now, as a workaround (in place of this PR), in our test code, we’re using this hack, which is not at all pretty (reduced imports and lifts because of ClassyPrelude):

```hs
module Backend.Unsafe.Tests.CookiedClientM where

import           Control.Monad.Except
import           Data.Proxy
import qualified Network.HTTP.Client                as Client
import           Network.HTTP.Types                 (statusCode)
import           Servant.Client.Core
import           Servant.Client.Internal.HttpClient

data CookiedClientEnv = CookiedClientEnv
  { clientEnv :: ClientEnv
  , cookieJar :: TVar Client.CookieJar
  }

newtype CookiedClientM a = CookiedClientM
  { runCookiedClientM' :: ReaderT CookiedClientEnv (ExceptT ServantError IO) a
  } deriving ( Functor
             , Applicative
             , Monad
             , MonadIO
             , Generic
             , MonadReader CookiedClientEnv
             , MonadError ServantError
             , MonadThrow
             , MonadCatch
             )

instance RunClient CookiedClientM where
  runRequest = performCookiedRequest
  throwServantError = throwError
  catchServantError = catchError

performCookiedRequest :: Request -> CookiedClientM Response
performCookiedRequest req = do
  m <- asks (manager . clientEnv)
  burl <- asks (baseUrl . clientEnv)
  cookieJar' <- asks cookieJar
  now <- liftIO getCurrentTime
  request <-
    atomically $ do
      oldCookieJar <- readTVar cookieJar'
      let (newRequest, newCookieJar) =
            Client.insertCookiesIntoRequest
              (requestToClientRequest burl req)
              oldCookieJar
              now
      writeTVar cookieJar' newCookieJar
      pure newRequest
  eResponse <- liftIO $ catchConnectionError $ Client.httpLbs request m
  case eResponse of
    Left err -> throwError err
    Right response -> do
      now' <- liftIO getCurrentTime
      atomically $
        modifyTVar'
          cookieJar'
          (fst . Client.updateCookieJar response request now')
      let status = Client.responseStatus response
          status_code = statusCode status
          ourResponse = clientResponseToReponse response
        throwError $ FailureResponse ourResponse
      return ourResponse

instance ClientLike (CookiedClientM a) (CookiedClientM a) where
  mkClient = id

cookiedClient ::
     HasClient CookiedClientM api => Proxy api -> Client CookiedClientM api
cookiedClient api = api `clientIn` (Proxy :: Proxy CookiedClientM)

runCookiedClientM ::
     CookiedClientM a -> CookiedClientEnv -> IO (Either ServantError a)
runCookiedClientM cm env =
  runExceptT . flip runReaderT env $ runCookiedClientM' cm
```

### DIFF
--- a/servant-client/src/Servant/Client/Internal/HttpClient.hs
+++ b/servant-client/src/Servant/Client/Internal/HttpClient.hs
@@ -16,17 +16,19 @@ module Servant.Client.Internal.HttpClient where
 import           Prelude                     ()
 import           Prelude.Compat
 
+import           Control.Concurrent.STM.TVar
 import           Control.Exception
 import           Control.Monad
 import           Control.Monad.Base          (MonadBase (..))
 import           Control.Monad.Catch         (MonadCatch, MonadThrow)
 import           Control.Monad.Error.Class   (MonadError (..))
 import           Control.Monad.Reader
+import           Control.Monad.STM           (atomically)
 import           Control.Monad.Trans.Control (MonadBaseControl (..))
 import           Control.Monad.Trans.Except
 import           Data.ByteString.Builder     (toLazyByteString)
 import qualified Data.ByteString.Lazy        as BSL
-import           Data.Foldable               (toList)
+import           Data.Foldable               (toList, traverse_)
 import           Data.Functor.Alt            (Alt (..))
 import           Data.Maybe                  (maybeToList)
 import           Data.Monoid                 ((<>))
@@ -34,6 +36,7 @@ import           Data.Proxy                  (Proxy (..))
 import           Data.Sequence               (fromList)
 import           Data.String                 (fromString)
 import qualified Data.Text                   as T
+import           Data.Time.Clock             (getCurrentTime)
 import           GHC.Generics
 import           Network.HTTP.Media          (renderHeader)
 import           Network.HTTP.Types          (hContentType, renderQuery,
@@ -47,6 +50,7 @@ data ClientEnv
   = ClientEnv
   { manager :: Client.Manager
   , baseUrl :: BaseUrl
+  , cookieJar :: Maybe (TVar Client.CookieJar)
   }
 
 -- | Generates a set of client functions for an API.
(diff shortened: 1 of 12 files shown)


michalrus:
(I don’t really know why only the last build succeeds.)

alpmestan:
@phadej this looks reasonable to me, thoughts?

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -104,12 +108,30 @@ performRequest :: Request -> ClientM Response
...
+                now
+        writeTVar cj newCookieJar
+        pure newRequest
 
   eResponse <- liftIO $ catchConnectionError $ Client.httpLbs request m
   case eResponse of
     Left err -> throwError $ err
     Right response -> do
+      now' <- liftIO getCurrentTime
+      traverse_
```
phadej: split his xpression, would `for_` help?
```haskell
for_ cookieJar' $ \cj -> liftIO . atomically $ do
    modifyTVar (...) cj
```

michalrus:
Hey!

Sorry for the silence, but I wasn’t sure this was the right approach.

And I think this intuition was right: just a while ago a need arose in our tests to spoof the `X-Real-IP`/`X-Forwarded-For` for all requests. I think such a request modification (for all requests) doesn’t really belong to API types…

So *maybe* it would be better to have some *general* way to modify `Request` before sending it in `performCookiedRequest` and post-process `Response`?

I.e. let the user provide `requestMod :: Request -> ClientM Request` and `responseSomething :: Response -> ClientM ()`?

Using those, I would be able to implement cookies (we might add that to the cookbook later, for example).

And also, I would be able to implement our XFF header spoofing globally.

@phadej, @alpmestan, what do you think?

phadej:
This is already merged as part of https://github.com/haskell-servant/servant/pull/897

Maybe something like you propose is possible. I agree that current `ClientM` is too monolithic, and something like what you propose might work.

- Maybe we should have `ClientT m` (because you needed to add `TVar` to make cookies work), and ask for optional `Request -> m Request`, `Response -> m Response`? 
- Or alternatively ask for `requestToClientRequest`, `requestToClientRequest` etc?

experimentation needed, I don't see one obviously best option.

michalrus:
Thank you very much for the merge and the additional work this required! =)

Yes, yes, `Request -> m Request` and `Response -> m Response` would be very nice! :pray: 

If I had `Request -> ClientM Request` and `Response -> ClientM`, while `ClientM` is in `MonadIO`, then the users could relatively easily keep their cookie jar outside of `ClientEnv`, so there’d be no need for a `ClientT m`. But maybe that wouldn’t be so easy for them conceptually? I dunno…

michalrus:
But which request and response should those be? Servant’s or http-client’s? Hmmm.

alpmestan:
Well, we went through the trouble of making servant-client backend agnostic, so it'd would be ideal if this feature was made available to all backends. On the other hand this would probably mean extending servant-client-core's notion of a request to be more comprehensive, wouldn't it?

phadej:
We really need second GHC (not ghcjs) `servant-client-core` "backend", e.g. `http-streams` (If I read http://hackage.haskell.org/package/http-streams-0.8.5.5/docs/Network-Http-Client.html right, there isn't built in Cookie implementation, which is :( but OTOH, cookies aren't the thing in APIs for machines so understandable)

michalrus:
> but OTOH, cookies aren't the thing in APIs for machines so understandable)

THB, I’m not really sure why. ^^, What’s the difference, whether you send a token in an `Authorization:` or a `Cookie:`. It makes a difference for SPAs in browser, though (`Cookie:` being more secure).

Issue #888, opened 2018-01-17
Add travis testing of stack configurations


mgsloan:
I noticed that https://github.com/haskell-servant/servant/blob/master/stack.yaml is broken, and certainly has been for some time.  I find this surprising, it is broken in ways that cause the build to immediately fail, and this would always be the case.  Namely, the `aeson-compat` version specified by the snapshot is not compatible with the specified `aeson` version.

If you do not want to have CI do both stack and cabal builds, at least run `stack build --dry-run --system-ghc`.  This will usually run very quickly and will tell you if there are version constraint problems.

phadej:
Will do. Though I will leave only single `stack.yaml` in place

mgsloan:
I am in favor of removing old configs unless there is a strong reason to keep them :+1:  One `stack.yaml` should do.

PR #889, opened 2018-01-17
Remove usage of default exts, now "stack ghci" loads all packages together

mgsloan:
`stack ghci` can now load all the modules from the servant packages at once. (well, `stack --stack-yaml stack-ghc-8.2.1.yaml ghci` due to the stack.yaml currently being broken - https://github.com/haskell-servant/servant/issues/888)


alpmestan:
Could you expand on what the problem is with those `default-extensions`?

mgsloan:
Here is the output:

```
mgsloan@treetop:~/oss/haskell/servant$ stack ghci --stack-yaml stack-ghc-8.2.1.yaml
servant-docs-0.11.1: initial-build-steps (lib + exe)
servant-foreign-0.11: initial-build-steps (lib)
Log files have been written to: /home/mgsloan/oss/haskell/servant/.stack-work/logs/
Completed 2 action(s).

* * * * * * * *
Warning: There are cabal settings for this project which may prevent GHCi from loading your code properly.
In some cases it can also load some projects which would otherwise fail to build.

-XCPP will be used, but it can cause issues with multiline strings.
See https://downloads.haskell.org/~ghc/7.10.2/docs/html/users_guide/options-phases.html#cpp-string-gaps
It is specified for:
    servant-foreign:lib
But not for: 
    servant:lib servant-client:lib servant-client-core:lib servant-docs:lib servant-docs:exe:greet-docs servant-server:lib servant-server:exe:greet

-XTemplateHaskell will be used, but it may cause compilation issues due to different parsing of '$' when there's no space after it.
It is specified for:
    servant-foreign:lib
But not for: 
    servant:lib servant-client:lib servant-client-core:lib servant-docs:lib servant-docs:exe:greet-docs servant-server:lib servant-server:exe:greet

-XOverloadedStrings will be used, but it can cause type ambiguity in code not usually compiled with it.
It is specified for:
    servant-foreign:lib
But not for: 
    servant:lib servant-client:lib servant-client-core:lib servant-docs:lib servant-docs:exe:greet-docs servant-server:lib servant-server:exe:greet

-XTypeFamilies will be used, but it implies -XMonoLocalBinds, and so can cause type errors in code which expects generalized local bindings.
It is specified for:
    servant-foreign:lib
But not for: 
    servant:lib servant-client:lib servant-client-core:lib servant-docs:lib servant-docs:exe:greet-docs servant-server:lib servant-server:exe:greet

To resolve, remove the flag(s) from the cabal file(s) and instead put them at the top of the haskell files.

It isn't yet possible to load multiple packages into GHCi in all cases - see
https://ghc.haskell.org/trac/ghc/ticket/10827
* * * * * * * *

Configuring GHCi with the following packages: servant, servant-client, servant-client-core, servant-docs, servant-foreign, servant-server

* * * * * * * *
The main module to load is ambiguous. Candidates are: 
1. Package `servant-docs' component exe:greet-docs with main-is file: /home/mgsloan/oss/haskell/servant/servant-docs/example/greet.hs
2. Package `servant-server' component exe:greet with main-is file: /home/mgsloan/oss/haskell/servant/servant-server/example/greet.hs
You can specify which one to pick by: 
 * Specifying targets to stack ghci e.g. stack ghci servant-docs:exe:greet-docs
 * Specifying what the main is e.g. stack ghci --main-is servant-docs:exe:greet-docs
 * Choosing from the candidate above [1..2]
* * * * * * * *

Specify main module to use (press enter to load none): 2
Loading main module from candidate 2, --main-is /home/mgsloan/oss/haskell/servant/servant-server/example/greet.hs

GHCi, version 8.2.1: http://www.haskell.org/ghc/  :? for help

/home/mgsloan/oss/haskell/servant/servant/src/Servant/API/Capture.hs:24:0: error:
     error: unterminated comment
     -- >>>            -- GET /src/*
     
   |
24 | -- >>>            -- GET /src/*
   | ^
`gcc' failed in phase `C pre-processor'. (Exit code: 1)

<no location info>: error:
    Could not find module ‘Servant’
    It is not a module in the current program, or in any known package.
Loaded GHCi configuration from /tmp/ghci22811/ghci-script
Prelude>
```

So the main thing is CPP balking on `/*` used in the middle of a haskell comment.  The warnings from `stack ghci` tell you about other 3 other extensions that can cause inconsistencies.  I noticed none of the other cabal files use `default-extensions`, so I made `servant-foreign` consistent with the rest.

alpmestan:
OK I see, thanks. Note that I was not asking because I have a problem with the patch, but simply so that we have a trace of what problem this is solving.

This makes a test fail in CI: https://travis-ci.org/haskell-servant/servant/jobs/329822165#L4139

As soon as CI is green I don't mind merging this patch.

mgsloan:
Well that's a real mystery!  I've also removed the `default-extensions` from the test-suite now.  But I can't get the test-suite to build locally.  I haven't seen behavior before where moving default-extensions into language pragmas cause a difference ¯\_(ツ)_/¯

phadej:
I see, does `stack` enable extensions for all packages, when any of them have `default-extensions`? Seems that `stack` loads whole project into ghci, not a single package. Interesting choice.

/Note:/ The failure is in `servant`, when `default-extensions` is in `servant-foreign`.

Is there an issue tracking this issue, the output should tell that `default-extensions` from one package apply to all other packages in the project! I guess my co-worker have seen that, but didn't understood why that happened.

I wonder, what would happen if one uses `PackageImprorts` (we won't, I'm just curious)?

phadej:
@mgsloan The problem with test-suite is that some other file in the library needs `PolyKinds`, things often compile even without it, except when you want to use them with something else than `*`.

mgsloan:
Yes, `stack ghci` is very powerful in this regard, I am usually loading many packages into it at once.  It defaults to loading all local packages together.  If you do `stack ghci pkg-x` it will load just one.

> I wonder, what would happen if one uses PackageImprorts (we won't, I'm just curious)?

Yep that causes trouble. I discussed this with ezyang at ICFP a while back and he opened this ghc ticket about it https://ghc.haskell.org/trac/ghc/ticket/10827 - hopefully one day this can be done in a more principled way than concatenating all the options, loading all the modules, and crossing fingers.  What we have works pretty darn well in practice, it just sometimes requires tweaking cabal files particularly when it comes to default-extensions.

mgsloan:
Updated, should be fixed now, let's hope CI is satisfied.

Issue #887, opened 2018-01-17
Should have more kind signatures, particularly (a :: *)


mgsloan:
I noticed that the definition of `Capture` is `data Capture (sym :: Symbol) a`.  The problem with this is that it allows me to write nonsense like `Capture "foo" Maybe :> Post '[JSON] ()`, and only get a type error later when matching up the implementations with the type specs.  The error would be much nicer if the definition of `Capture` was instead `data Capture (sym :: Symbol) (a :: *)`.  Similar should be done for some other types, definitely `CaptureAll`, `ReqBody`, `QueryParam`, `QueryParams`.  `BasicAuth` already has it.  This is not by any means an exhaustive list.  I'm new to servant and I just opened up a few modules to see how widespread this was.

I think this will make some errors a bit kinder.

Also, I was quite confused that I was able to write `Capture "foo" :> Post '[JSON] ()`.  Note the missing type argument on `Capture`.  I suppose the reason this isn't done is that it's desired to have an open world of types for APIs, and we don't have anything like "kind constraints"?  I guess I'm spoiled by typechecking and expect similar at the kind level.

alpmestan:
Hello @mgsloan,

You are making some good points. When we know that we are going to get something of kind `*` and that anything else just doesn't make sense (that's the case for all the combinators that you point out, namely `Capture`, `CaptureAll`, etc. I think we consistently do this for type level lists or strings so there's no reason not to do it for the most common of all kinds, `*`. I don't know how nicer the errors get with those changes either, but I'm convinced this would be a strict improvement.

Regarding the last case that you mention, namely the fact that you can define an API type without supplying all the parameters of `Capture`, this is indeed something we cannot do much about. While the RHS of `:>` is always of kind `*`, its LHS in practice either has kind `Symbol` or `*` (because we want to make type level strings special, we don't want to be forced say `type API = Static "foo" :> Capture "fooid" FooId :> Get '[JSON] Foo`, writing the strings directly is just so handy a nicer to read), so we to need to allow other kinds than `*` in the LHS of `:>`. The definition we have therefore is: `data (path :: k) :> a`.

Would you be interested in giving a shot at a PR to add the missing kind signatures?

mgsloan:
Sure!

mgsloan:
[...]

PR #893, opened 2018-01-20
Add two HasServer instances which won't ever exist

phadej:
Resolves https://github.com/haskell-servant/servant/issues/887


servant-server/src/Servant/Server/Internal.hs
```diff
@@ -703,3 +707,54 @@ instance (HasContextEntry context (NamedContext name subContext), HasServer subA
...
+-- type applications on LHS of ':>'.
+--
+-- >>> serve (Proxy :: Proxy (Capture "foo" :> Get '[JSON] Int)) (error "...")
+-- ...
+-- ...Expected something of kind Symbol or *, got: k -> l on the LHS of ':>'.
+-- ...Maybe you haven't applied enough arguments to
+-- ...Capture "foo"
+-- ...
+instance TypeError
+    ('Text "Expected something of kind Symbol or *, got: k -> l on the LHS of ':>'."
```
fizruk: Don't you want `'ShowType k :<>: 'Text "->" :<>: 'ShowType l` instead of `k -> l`?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -703,3 +707,54 @@ instance (HasContextEntry context (NamedContext name subContext), HasServer subA
...
+-- >>> serve (Proxy :: Proxy (Capture "foo" Int -> Get '[JSON] Int)) (error "...")
+-- ...
+-- ...No instance HasServer (a -> b).
+-- ...Maybe you have used '->' instead of ':>' between
+-- ...Capture "foo" Int
+-- ...and
+-- ...Verb 'GET 200 '[JSON] Int
+-- ...
+instance TypeError
+    ('Text "No instance HasServer (a -> b)."
```
fizruk: Don't you want `'ShowType a :<>: 'Text "->" :<>: 'ShowType b` instead of `a -> b`?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -703,3 +707,54 @@ instance (HasContextEntry context (NamedContext name subContext), HasServer subA
...
+-- ...Maybe you haven't applied enough arguments to
+-- ...Capture "foo"
+-- ...
+instance TypeError
+    ('Text "Expected something of kind Symbol or *, got: k -> l on the LHS of ':>'."
+    ':$$: 'Text "Maybe you haven't applied enough arguments to"
+    ':$$: 'ShowType arr)
+    => HasServer ((arr :: k -> l) :> api) context
+  where
+    type ServerT (arr :> api) m = Void
```
fizruk: Maybe use the same `TypeError` here?

servant-server/src/Servant/Server/Internal.hs
phadej: No I don't want. The `a` and `b` are printed below separately (as they can be huge).

servant-server/src/Servant/Server/Internal.hs
phadej:
```
Illegal qualified type: (TypeError ...) => Void
```

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -703,3 +707,54 @@ instance (HasContextEntry context (NamedContext name subContext), HasServer subA
...
+-- ...
+instance TypeError
+    ('Text "No instance HasServer (a -> b)."
+    ':$$: 'Text "Maybe you have used '->' instead of ':>' between "
+    ':$$: 'ShowType a
+    ':$$: 'Text "and"
+    ':$$: 'ShowType b)
+    => HasServer (a -> b) context
+  where
+    type ServerT (a -> b) m = Void
```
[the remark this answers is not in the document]
phadej:
```
Illegal qualified type: (TypeError ...) => Void
```

servant-server/src/Servant/Server/Internal.hs
phadej: We could, but then we'd need `{-# LANGUAGE TypeInType #-}`, and then i'd rather use `Type`, and not `*` everywhere we have.

The error message doesn't get much better:
```diff
-Expected something of kind Symbol or *, got: k -> l on the LHS of ':>'.
+Expected something of kind Symbol or *, got: * -> * on the LHS of ':>'.
```

My plan is to have 0.13 with GHC-7.8 support, but hopefully for next release GHC-8.4 will be out, so we can drop support of GHC-7.8 and GHC-7.10, and clean up a lot of things. This too. Does this sound as a good plan?

servant-server/src/Servant/Server/Internal.hs
fizruk: I fear that `a` and `b` in `a -> b` might confuse users who might think they have some ambiguous type variables or maybe they'll actually have `a` or `b` in their program.

OTOH I agree that huge types are not better.

servant-server/src/Servant/Server/Internal.hs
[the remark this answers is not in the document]
fizruk: Did you try

```haskell
type ServerT (a -> b) m = (TypeError ...) => Void
```

or

```haskell
type ServerT (a -> b) m = TypeError ...
```

?

servant-server/src/Servant/Server/Internal.hs
fizruk: Yeah, I guess. The problem I have with `k -> l` is the same as with `a -> b` below: `k` and `l` are not any of the user kind variables, so it can confuse them actually more than `* -> *`.

Anyway, it's ok to clean this up later, and the error message is a lot better already :)

servant-server/src/Servant/Server/Internal.hs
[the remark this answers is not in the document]
phadej: Ah, `TypeError` may be of any kind. Let's try

PR #895, opened 2018-01-25
Bump lower bounds

phadej:
- `text` is already 1.2.3.0 as https://github.com/fpco/stackage/issues/3147 is resolved AFAICS, though not closed.

- `http-types` bound is surprisingly low due: https://github.com/fpco/stackage/issues/2976

---

@alp and other @haskell-servant/maintainers: I'd like to try this bounds schema. It's less aggressive than "lowest versions there is build-plan for", but lessens maintaining burden (we have e.g. `MIN_VERSION_http_types` and `MIN_VERSION_wai_app_static` now, which I'd like to get rid off, we really don't test "other branch" that well)

### DIFF
--- a/doc/tutorial/tutorial.cabal
+++ b/doc/tutorial/tutorial.cabal
@@ -22,47 +22,60 @@ extra-source-files:
   static/ui.js
 
 library
+  default-language:    Haskell2010
+  ghc-options:         -Wall -pgmL markdown-unlit
   exposed-modules:     ApiType
                      , Authentication
                      , Client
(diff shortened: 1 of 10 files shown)


fizruk:
Looks good! And now I learned that it's possible to have multiple `build-depends` sections! :)

alpmestan:
CI is not happy, but I'm otherwise not opposed to the new bounds.

phadej:
Good. I'll go through the rest packages, and then merge (when travis is happy)

PR #897, opened 2018-01-26
Support http-client’s CookieJar in servant-client

phadej:


phadej:
@alpmestan I'll toy with idea if we can have

```haskell
prepare :: ClientM a -> (BaseUrl -> Request, Response -> Either ServantError a)
```

function, so for similar issues people will be able to hook functionality. It can s till make for 0.13, as I don't see why it won't be possible.

alpmestan:
@phadej To remove `ClientEnv` from the API? (or from the code altogether)

phadej:
@alpmestan, I won't remove `ClientEnv`, the

```haskell
runClientM :: ClientM a -> ClientEnv -> IO (Either ServantError a)
```

is useful for 99% of use cases still.

My idea is to make `runClientM` less opaque, i.e. `runClientM cli env = let (toRes, fromReq) = perform cli ... in plumb toRes fromReq where ...`, so downstream may use own plumbing. Now they cannot, as `servant-client` goes from `servant-client-core` structures directly to `IO`, but IMHO we should have intermediate functions to give `http-client` structures too.

alpmestan:
OK, sounds good to me. And no rush for 0.13, we can wait for this to go in.

Rosuavio:
Newb question, why isn't the `Maybe` inside the `TVar` for cookieJar instead of the `TVar` in the `Maybe`. It seems like it would be more ergonomic. Not trying to critique or criticize more than just trying to learn.

PR #899, opened 2018-01-30
Refactor servant-client-core Response+StreamingResponse

phadej:


servant-client/src/Servant/Client.hs
```diff
@@ -1,3 +1,14 @@
+{-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DeriveDataTypeable         #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
```
jkarni: Is there a reason for moving this around? I generally prefer to not have too much code in non-`Internal` modules since then I can export everything for testing etc. without cluttering the haddocks and API.

servant-client-core/src/Servant/Client/Core/Internal/Request.hs
```diff
@@ -65,14 +66,15 @@ type Request = RequestF Builder.Builder
...
-  , responseBody        :: LBS.ByteString
   , responseHeaders     :: Seq.Seq Header
   , responseHttpVersion :: HttpVersion
-  } deriving (Eq, Show, Generic, Typeable)
+  , responseBody        :: a
+  } deriving (Eq, Show, Generic, Typeable, Functor, Foldable, Traversable)
 
-data StreamingResponse = StreamingResponse { runStreamingResponse :: forall a. ((Status, Seq.Seq Header, HttpVersion, IO BS.ByteString) -> IO a) -> IO a }
+type Response = GenResponse LBS.ByteString
+newtype StreamingResponse = StreamingResponse { runStreamingResponse :: forall a. (GenResponse (IO BS.ByteString) -> IO a) -> IO a }
```
jkarni: Shouldn't this, for symmetry, be something like 
```
newtype StreamingBody = StreamingBody (forall a. ((IO ByteString) -> a) -> a)
type StreamingResponse = GenResponse StreamingBody
```

(not sure if those details are right, but what I'm referring to is having the continuation *within* `GenResponse` rather than containing it)

phadej: It won't work. see `performStreamingRequest`

servant-client/src/Servant/Client.hs
phadej: Not really (I planned to make non-IO client, but it's difficult given we have streaming now. I'll revert this change.

For the record, I partially disagree, I want e.g. `requestToClientRequest`to be a part of *stable public API* (I'd rather export it from `Servant.Client`)

PR #904, opened 2018-02-09
Make empty request path = slash. Fixes #781

### LINKED ISSUE #781: `parseBaseUrl` strips trailing slashes
What is the rationale for this? We have API endpoints which must be called with trailing slash, and we parse the endpoint URL from a configuration file with `parseBaseUrl`.

(Not sure whether this is a dup of #645; this is about not stripping a trailing slash rather than adding one.)

phadej:
This is small hack.

@edsko do you remember why you changed path to be `Builder` in https://github.com/haskell-servant/servant/commit/507f0219191ba220bb8c1f5256f2322502f6a12d , is it only to conveniently use `toEncodedUrlPiece`?

If it so, maybe we should make "own" builder, with notion of emptyness to avoid evaluating builder.

Also, @jkarni do you remember why there is `RequestF` is there use for non `Builder` `Request` somewhere?

---

0.13 will however be out with these changes.

---

@alp, @jkarni, @fizruk do we want `Free ClientF` into public API? We can export it later, it's handy :)

### DIFF
--- a/servant-client-core/test/Servant/Client/Core/Internal/HasClientSpec.hs
+++ b/servant-client-core/test/Servant/Client/Core/Internal/HasClientSpec.hs
@@ -1,0 +1,73 @@
+{-# LANGUAGE FlexibleContexts, DeriveFunctor, GeneralizedNewtypeDeriving, DataKinds, TypeOperators, OverloadedStrings #-}
+{-# OPTIONS_GHC -Werror #-}
+module Servant.Client.Core.Internal.HasClientSpec (spec) where
+
+import Prelude ()
+import Prelude.Compat
+
+import Control.Monad.Free (Free (..), liftF)
+import Servant.API
+import Servant.Client.Core
+import Data.Text (Text)
+import Data.Proxy (Proxy (..))
+import Servant.API.Internal.Test.ComprehensiveAPI
+import Test.Hspec
+
+import qualified Data.ByteString.Builder as B
+
+-- This declaration simply checks that all instances are in place.
+_ = client comprehensiveAPI
+
+spec :: Spec
+spec = describe "Servant.Client" $ do
+    issue781trailingSlashSpec
+
+type Issue781 = Get '[JSON] Text
+    :<|> "foo" :> "bar" :> Get '[JSON] Text 
+
+issue781 :: Proxy Issue781
+issue781 = Proxy
+
+issue781trailingSlashSpec :: Spec
+issue781trailingSlashSpec = describe "issue 781: trailing slash in baseurl" $ do
+    it "Empty request has / as a path" $ do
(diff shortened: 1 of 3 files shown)


servant-client-core/test/Servant/Client/Core/Internal/HasClientSpec.hs
```diff
@@ -0,0 +1,73 @@
...
+    :<|> "foo" :> "bar" :> Get '[JSON] Text 
+
+issue781 :: Proxy Issue781
+issue781 = Proxy
+
+issue781trailingSlashSpec :: Spec
+issue781trailingSlashSpec = describe "issue 781: trailing slash in baseurl" $ do
+    it "Empty request has / as a path" $ do
+        B.toLazyByteString . requestPath <$> matchSingleRequest emptyCli
+            `shouldBe` Just "/"
```
jkarni: Should this really be "/"? There'd be no way to make a request to `github.com`, no? Whereas if this were `""` then to make a request to `github.com/` would just involve adding an empty path component (which admittedly is non-obvious)

phadej: With:
```
% curl 'localhost:8888'
```

you get

```
% nc -l 8888
GET / HTTP/1.1
Host: localhost:8888
User-Agent: curl/7.47.0
Accept: */*
```

i.e. there is *no empty path*.

phadej: And the rationale is for 

```haskell
baseUrl = "server.com/some/path"
```

with api

```haskell
type API = Get '[JSON] Value -- not path component
```

the requests are done to

```
http://server.com/some/path i.e. /home/path
```

but reporter of #781 wants

```
server.com/some/path/ <- trailing path
```

As far as I understand HTTP spec those *can* return different results, servers done with `servant` won't.

In fact we want to control (and differ) endpoints with and without trailing slash. But that's more work (it affects how relative urls are calculated, so we might want to redirect from slashless endpoint to slash-ful).

jkarni: I still don't get it. If you want to `Get '[JSON] Value` to go to `server.com/some/path/`, why not change the `baseUrl`, appending the slash? It still feels like we're solving one problem but creating another (no easy way to have the request go to `server.com/some/path` _without_ the slash)

phadej: @jkarni 

Then

```haskell
type API = "foo" :> Get '[JSON] Bar
```

would go to `server.com/some/path//foo`, not the double slash.

phadej:
Ok, I won't include the fix into `0.13`, I want to release today.

PR #905, opened 2018-02-09
Allow BaseUrl to specify whether root has slash.

jkarni:
Previously this would cause an error with double slashes.

### DIFF
--- a/servant-client/src/Servant/Client/Internal/HttpClient.hs
+++ b/servant-client/src/Servant/Client/Internal/HttpClient.hs
@@ -25,6 +25,7 @@ import           Control.Monad.Trans.Control (MonadBaseControl (..))
 import           Control.Monad.Trans.Except
 import           Data.ByteString.Builder     (toLazyByteString)
 import qualified Data.ByteString.Lazy        as BSL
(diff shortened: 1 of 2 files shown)


phadej:
@jkarni, not that now you cannot get `baseurl` **with** trailing slash:

```haskell
λ *Servant.Client> parseBaseUrl "http://server.com/foo/bar/"
BaseUrl {baseUrlScheme = Http, baseUrlHost = "server.com", baseUrlPort = 80, baseUrlPath = "/foo/bar"}
```

jkarni:
```
λ> u <- parseBaseUrl "http://server.com/foo/bar/"
λ> u { baseUrlHost = baseUrlHost u <> "/" }
BaseUrl {baseUrlScheme = Http, baseUrlHost = "server.com/", baseUrlPort = 80, baseUrlPath = "/foo/bar"}
```

Not as convenient, I agree, but we have to default to *something*, and I don't see why defaulting to root _having_ trailing slash is better (but I could be wrong)

phadej:
IMO if we make this change we have to preserve `baseUrlPath` as given.

PR #913, opened 2018-02-27
Streaming request body for servant-client-core

### LINKED ISSUE #886: Have a streaming request body for servant-client
Currently, the [request body](https://hackage.haskell.org/package/servant-client-core-0.12/docs/Servant-Client-Core.html#t:RequestBody) may only be a lazy bytestring. This is a problem when large files must be included in the body, as the file must either be loaded fully in memory, or one has to use lazy I/O. It would be nice to have something similar to `http-client`'s [request body](https://hackage.haskell.org/package/http-client-0.5.8/docs/Network-HTTP-Client.html#t:RequestBody). I understand that it's not doable for servant-client-core to depend on http-client; however there may be a way to extract some types from http-client, which would also make the multipart implementation ( #884 ) simpler.

CC @jkarni @alpmestan

jkarni:
Fixes #886.

I also added tests to check that `Stream` ran in constant memory while I was at it.

### DIFF
--- a/servant-client/test/Servant/StreamSpec.hs
+++ b/servant-client/test/Servant/StreamSpec.hs
@@ -26,25 +26,26 @@
 #include "overlapping-compat.h"
 module Servant.StreamSpec (spec) where
 
-import           Prelude                                    ()
-import           Prelude.Compat
+import           Control.Monad       (replicateM_, void)
+import qualified Data.ByteString     as BS
 import           Data.Proxy
-import qualified Network.HTTP.Client                        as C
-import           System.IO.Unsafe                           (unsafePerformIO)
+import           GHC.Stats           (currentBytesUsed, getGCStats)
+import qualified Network.HTTP.Client as C
+import           Prelude             ()
+import           Prelude.Compat
+import           System.IO           (IOMode (ReadMode), withFile)
+import           System.IO.Unsafe    (unsafePerformIO)
 import           Test.Hspec
+import           Test.QuickCheck
 
-import           Servant.API                                ((:<|>) ((:<|>)),
-                                                             (:>),
-                                                             EmptyAPI, JSON,
-                                                             StreamGet,
-                                                             NewlineFraming,
-                                                             NetstringFraming,
-                                                             ResultStream(..),
-                                                             StreamGenerator(..))
+import           Servant.API         ((:<|>) ((:<|>)), (:>), JSON,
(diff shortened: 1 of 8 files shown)


jkarni:
Tests are failing on 7.8.4 for reasons that don't have anything to do with the PR. I tried reproducing it locally but can't. @phadej any ideas?

jkarni:
@phadej @alpmestan can I merge?

phadej:
Oh, I haven't seen this. I'll take a look tomorrow and merge if only small stuff there's

phadej:
I took a quick glance and I'm not particularly fond of replicating http-client API in the servant-client-core. It feels like a step back. I'Il think a little harder tomorrow

jkarni:
> I took a quick glance and I'm not particularly fond of replicating http-client API in the servant-client-core

I thought that was what we had come up with over IRC. But if there are nicer alternatives I'm all for them. The only constructor in `RequestBody` that didn't seem necessary/logical was `RequestBodyIO` (but funnily I ended up using it in `servant-streaming`, and now am in favor of it).

jkarni:
Any updates?

servant-client/test/Servant/StreamSpec.hs
```diff
@@ -77,14 +75,24 @@ bob :: Person
...
-       :<|>
-       emptyServer)
+server = serve sapi
+  $    return (StreamGenerator (\f r -> f alice >> r bob >> r alice))
+  :<|> return (StreamGenerator (\f r -> f alice >> r bob >> r alice))
+  :<|> return (StreamGenerator lotsGenerator)
+  where
+    lotsGenerator f r = do
+      f ""
+      withFile "/dev/urandom" ReadMode $
```
phadej: can we use `random` / `tf-random` etc. and not drain the (somewhat slow) secure random device?

jkarni:
Good call on testing GHCJS. I'm trying to do that but keep running into problems. Will update here when it works.

servant-client/test/Servant/StreamSpec.hs
jkarni: Tried `random` and `bytestring` random, but they used a lot more CPU and failed (without clear error messages) on CI. 

I can remove this test entirely, though

domenkozar:
@jkarni how can I help get this finalized?

jkarni:
@domenkozar finalized

PR #915, opened 2018-03-05
Fix the changelog with hoistServerWithContext

ndmitchell:
I tried what was in the changelog and that failed, so I tried the obvious guess and that worked - update attached.

### DIFF
--- a/servant/CHANGELOG.md
+++ b/servant/CHANGELOG.md
@@ -130,7 +130,7 @@
 
   ```haskell
   type ServerT (MyCombinator :> api) m = MyValue -> ServerT api m
-  hoistServerWithContext _ pc nt s = hoistServerWithContext (Proxy :: Proxy api) pc nt . s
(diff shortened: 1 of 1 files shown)


phadej:
@ndmitchell are you sure, check your `ServerT` definition

with argument (e.g. `Capture`)

https://github.com/haskell-servant/servant/blob/c1414478c6cd5b4cbb7da4a6630091d3773189bc/servant-server/src/Servant/Server/Internal.hs#L169-L172

and no argument (e.g. `Description`)

https://github.com/haskell-servant/servant/blob/c1414478c6cd5b4cbb7da4a6630091d3773189bc/servant-server/src/Servant/Server/Internal.hs#L670-L673

ndmitchell:
Yep, sorry, I hadn't appreciated the differences. Thanks for the detailed changelog, it's been most useful!

phadej:
No worries.

PR #919, opened 2018-03-09
relax the aeson constraint, allow 1.3.0.0

emmanueltouzery:
fixes https://github.com/fpco/stackage/issues/3337
I didn't actually test an app with the patched servant, I only ran the tests, but since it all compiles and the tests pass, it should be OK I assume.


phadej:
Thanks. 

Note however, we probably won't have a revision/release before we get GHC-8.4.1 compatibility. "blocked" on https://github.com/lpsmith/blaze-builder/issues/12

PR #920, opened 2018-03-11
Add Servant.Client.Free

phadej:
this is useful in testing, it nothing else.

### DIFF
--- a/servant-client-core/src/Servant/Client/Core/Internal/ClientF.hs
+++ b/servant-client-core/src/Servant/Client/Core/Internal/ClientF.hs
@@ -1,0 +1,10 @@
+{-# LANGUAGE DeriveFunctor #-}
+module Servant.Client.Core.Internal.ClientF where
+
+import Servant.Client.Core.Internal.Request
+
+data ClientF a
+    = RunRequest Request (Response -> a)
+    | StreamingRequest Request (StreamingResponse -> a)
+    | Throw ServantError
+  deriving (Functor)
--- a/servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
@@ -25,12 +25,11 @@ import           Servant.Client.Core.Internal.Request (Request, Response, GenRes
                                                        ServantError (..))
 import           Servant.Client.Core.Internal.ClientF
 
-class (Monad m) => RunClient m where
+class Monad m => RunClient m where
   -- | How to make a request.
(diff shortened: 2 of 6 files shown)


servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
```diff
@@ -48,3 +51,12 @@ decodedAs response contentType = do
     Right val -> return val
   where
     accept = toList $ contentTypes contentType
+
+instance ClientF ~ f => RunClient (Free f) where
```
phadej: dunno if should rather have `RunClientF f => RunClient (Free f)`

servant-client-core/src/Servant/Client/Core/Internal/RunClient.hs
```diff
@@ -48,3 +51,12 @@ decodedAs response contentType = do
...
+    runRequest req  = liftF (RunRequest req id)
+    streamingRequest req = liftF (StreamingRequest req id)
+    throwServantError = liftF . Throw
+    catchServantError x h = go x where
```
phadej: this is not particularly correct, as `RunRequest` could throw things catchable by `catchServantError` too.

jkarni:
Is this missing a file (`ClientF`)?

phadej:
oops

servant-client-core/src/Servant/Client/Core/Internal/ClientF.hs
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE DeriveFunctor #-}
+module Servant.Client.Core.Internal.ClientF where
+
+import Servant.Client.Core.Internal.Request
+
+data ClientF a
+    = RunRequest Request (Response -> a)
+    | StreamingRequest Request (StreamingResponse -> a)
+    | Throw ServantError
```
jkarni: Maybe adding a `Catch` constructor would help fix the problem with `catchServantError` that you mention below?

phadej: `catch` isn't an algebraic effect. So it won't really work out well.

phadej:
@jkarni @alpmestan I'd rather remove `catchServantError`, we don't use it itself, and I doubt anyone else

jkarni:
+1 to removing it. On my phone so can't review more carefully, but I think
that was my only concern.

alpmestan:
LGTM, and I don't mind the removal of `catchServantError` as well.

PR #925, opened 2018-03-15
Support GHC-8.4.1 and newer deps

phadej:


jkarni:
@phadej maybe cherry-picking from the commits that are going into `master` would make more sense? Otherwise we have two entirely different sets of commits on this branch and on the release branch, but doing the same things.

phadej:
I was slightly against doing one-by-one updates in master anyway (especially as it was known that there are other outdated dependencies).

But I'll forward port this to master, when `postgresql-client`, `http-media` and `servant-auth` allows it.

phadej:
@jkarni I don't see a problem with diverging histories, there is no intention they would converge.

Sorry for not addressing your comment properly. I feel there is a lot of pressure to put stuff out for GHC-8.4.1 or in into `master`.

As I said previously, I'd rather not merge `exceptions` #924 (yesterday) neither `aeson` #919 (5 days ago) PR, but wait till there's time to do this. I also would like to wait till `http-media` maintainer has time to react in https://github.com/zmthy/http-media/pull/25

PR #936, opened 2018-04-03
Add hoistClient to servant-client

alpmestan:
It's defined in terms of this new typeclass method of `HasClient` in `servant-client-core`:

``` haskell
class RunClient m => HasClient m api where
  [...]
  hoistClientMonad
    :: Proxy m
    -> Proxy api
    -> (forall x. mon x -> mon' x)
    -> Client mon api
    -> Client mon' api
```

Note that we don't expect source clients to live in the http client backend's `m` monad, so that we can say start from `ClientM`, go to some custom monad `App`, and finally land back in `IO`, using two successive calls to `hoistClientMonad`, or one that composes the transformations. This just seems like the right choice, because it's the most flexible option.

Feedback welcome!

### DIFF
--- a/doc/tutorial/Client.lhs
+++ b/doc/tutorial/Client.lhs
@@ -155,6 +155,63 @@ Email {from = "great@company.com", to = "alp@foo.com", subject = "Hey Alp, we mi
 
 The types of the arguments for the functions are the same as for (server-side) request handlers.
 
+## Changing the monad the client functions live in
+
+Just like `hoistServer` allows us to change the monad in which request handlers
+of a web application live in, we also have `hoistClient` for changing the monad
+in which _client functions_ live. Consider the following trivial API:
+
+``` haskell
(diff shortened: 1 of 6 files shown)


jkarni:
I'm about to be away from the internet for ten days, and am happy with this. One question though: isn't it already possible to do what `hoist` does simply by defining a new `RunClient` instance for the monad?

alpmestan:
@jkarni Maybe, yeah, except that a typeclass instance has to be provided for each target monad, and I think you're pretty much stuck with just one transformation per monad (you'd define the instance for `TargetMonad` as transformations over the instance of `SourceMonad`) so it's static. While with `hoistClient` you get to pick the transformation, and it can be dynamic: partially applying some arguments from some context first and finally transforming the return type of clients.

Not sure whether you were suggesting to refactor the PR in terms of `RunClient` instead or not, but I'm failing to see the advantages so I'm thinking probably not and that you were just curious. =)

alpmestan:
No other feedback? Should we merge?

PR #938, opened 2018-04-05
servant-client-ghcjs: Support binary requests

FPtje:
Introduces support for both sending and receiving binary data in `servant-client-ghcjs`.

Previously, the conversion to String before going to Bytestring would fail with any requests that sent or received binary data. With ArrayBuffers in the backend, sending Binary encoded data now works.

Also added a test case to https://github.com/FPtje/servant-client-ghcjs-test to test both sending and receiving of Binary data. It works and the tests don't show any regression.

One important note: I see that some recent version of servant has added a `streamingRequest` to the `RunClient`class. It is't implemented in servant-client-ghcjs, and this PR doesn't add it.

### DIFF
--- a/servant-client-ghcjs/src/Servant/Client/Internal/XhrClient.hs
+++ b/servant-client-ghcjs/src/Servant/Client/Internal/XhrClient.hs
@@ -35,6 +35,7 @@ import           Data.IORef                  (modifyIORef, newIORef, readIORef)
 import           Data.Proxy                  (Proxy (..))
 import qualified Data.Sequence               as Seq
 import           Data.String.Conversions
+import           Data.Typeable           (Typeable)
 import           Foreign.StablePtr
 import           GHC.Generics
 import qualified GHCJS.Buffer as Buffer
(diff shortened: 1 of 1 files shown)


alpmestan:
@arianvp Thanks for offering to look into it. :-)

phadej:
`fail`, not `error`. Even better would be to `throwIO` with custom exception type.

servant-client-ghcjs/src/Servant/Client/Internal/XhrClient.hs
```diff
@@ -217,42 +224,47 @@ setHeaders xhr request = do
...
 
   where
-    go :: RequestBody -> IO (Maybe String)
+    go :: RequestBody -> IO ArrayBuffer
     go x = case x of
-      RequestBodyLBS x -> return $ mBody x
+      RequestBodyLBS x -> return $ mBody $ BL.toStrict x
       RequestBodyBS x -> return $ mBody x
-      RequestBodyBuilder _ x -> return $ mBody $ toLazyByteString x
+      RequestBodyBuilder _ x -> return $ mBody $ BL.toStrict $ toLazyByteString x
```
phadej: As we on streaming case, we probably should add a note that `servant-client-ghcjs` doesn't respect any "constant space request body streaming" undertakings. (haddocks of `ClientM` and `client`?)

FPtje:
Updated: Rebased on latest master, added haddocks to `client` and `clientM` mentioning that streaming isn't supported. `streamingRequest` does a `throwIO` with a custom exception.

arianvp:
LGTM @phadej ?

arianvp:
Bump. Is this okay to merge?

Issue #125, opened 2015-06-17
support pagination


ibotty:
Whenever I find time I will start to write it myself. That bug report is to explore the design space.
- It should work with `Link` headers (`rel` in `next`, `prev`, `last`, `first`) (see [RFC5988](https://tools.ietf.org/html/rfc5988)).
- It should be possible to specify how many items to return (e.g.: "?per_page=100").
- I have used two different schemes to generate these `Link` urls in the past. There might be more worth supporting.
  - use a "?start_id=xxx" (This is usually way more efficient),
  - use a "?page=x"

alpmestan:
Just putting down some thoughts I've had on the topic, in the past.

``` haskell
data Paginated a = Paginated
  { page :: Int
  , totalPages :: Int
  , entries :: [a]
  } deriving (Eq, Show, Generic)

instance ToJSON a => ToJSON (Paginated a)

type Paginate c cts a = QueryParam "page" Int :> c cts (Paginated a)

-- Example
type API = "users" :> Paginate Get '[JSON] User
```

And then some machinery to help write request handlers that return paginated results.

christian-marie:
It'd be nice/crucial to design this so that the provider of these paginated results can load only the page it needs.

The whole point of pagination in many cases is that the data set is larger than memory,  or at least unwieldy, right? We should probably answer that question.

alpmestan:
Yes, but it's hard to be more helpful/specific without tying ourselves to a particular data store or something, isn't it? With some `Link` magic we could probably add a field with the URL to the next page, but what do you think this should look like? Feel free to drop a few ideas or lines of code.

jkarni:
The auth ideas in #70 show a way of being data-store agnostic that to my mind should be our general tack. (I.e., just a type "tag" that is associated with a `Range -> IO a` function that can be used as a parameter to `Paginate`).

I do wonder about using `Paginate Get ...` rather than `Paginate ... :> Get ...`. (Redirects, in #117 are relevant.) What's the general story about when you'd expect an overlapping instance in the last argument, when a type synonym over the method (as per @alpmestan suggestion), and when a `:>` combinator?

alpmestan:
@christian-marie by the way, just to clarify, the `entries` list isn't supposed to contain all the values, just the ones for the current page. But it's just a sketch anyway.

@jkarni `Paginate` above is a type synonym because it acts on the 2nd param of `Get` **and** adds a query string parameter.

As to the general story... My thinking:

[...]

PR #946, opened 2018-04-16
add recipe introducing servant-pagination

KtorZ:
As proposed in #125, here's a short introduction to [servant-pagination](https://hackage.haskell.org/package/servant-pagination-2.1.0). 

Feedback is welcome :)

### DIFF
--- a/doc/cookbook/pagination/Pagination.lhs
+++ b/doc/cookbook/pagination/Pagination.lhs
@@ -1,0 +1,250 @@
+# Pagination
+
+## Overview
+
+Let's see an approach to typed pagination with *Servant* using [servant-pagination](https://hackage.haskell.org/package/servant-pagination).
+
+This module offers opinionated helpers to declare a type-safe and a flexible pagination
+mechanism for Servant APIs. This design, inspired by [Heroku's API](https://devcenter.heroku.com/articles/platform-api-reference#ranges),
+provides a small framework to communicate about a possible pagination feature of an endpoint,
+enabling a client to consume the API in different fashions (pagination with offset / limit,
+endless scroll using last referenced resources, ascending and descending ordering, etc.)
+
+Therefore, client can provide a `Range` header with their request with the following format:
+
+- `Range: <field> [<value>][; offset <o>][; limit <l>][; order <asc|desc>]`
+
+For example: `Range: createdAt 2017-01-15T23:14:67.000Z; offset 5; order desc` indicates that
+the client is willing to retrieve the next batch of document in descending order that were
+created after the fifteenth of January, skipping the first 5.
+
+As a response, the server may return the list of corresponding document, and augment the
+response with 3 headers:
+
+- `Accept-Ranges`: A comma-separated list of fields upon which a range can be defined
+- `Content-Range`: Actual range corresponding to the content being returned
+- `Next-Range`: Indicate what should be the next `Range` header in order to retrieve the next range
+
+For example:
+
+- `Accept-Ranges: createdAt, modifiedAt`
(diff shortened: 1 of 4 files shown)


alpmestan:
@KtorZ Haven't yet looked at the explanations (I won't be too picky though, don't worry), but is there a reason why your patch comes with both a `Main.hs` and a `Pagination.lhs`, with the code duplicated in both?

See https://github.com/haskell-servant/servant/commit/5ac6de82773bcf2f352dda432a07225178feba2b for an example of commit that adds a cookbook recipe with the code and explanations in a single file. The trick is to have:

```
  ghc-options:         -pgmL markdown-unlit
  build-tool-depends: markdown-unlit:markdown-unlit
```

in the .cabal file. This processes the file with markdown-unlit (which strips away all the markdown, only leaving the code) before giving its contents to GHC.

KtorZ:
Ah! You're entirely right, I wrongly committed that. I'll update :+1:

KtorZ:
Done :+1: 
Let me know :)

doc/cookbook/pagination/Pagination.lhs
```diff
@@ -0,0 +1,251 @@
...
+  getFieldValue _ = hex
+
+-- to then define: Ranges '["name", "hex"] Color
+```
+
+
+#### Parsing Options
+
+By default, `servant-pagination` provides an implementation of `getRangeOptions` for each
+`HasPagination` type-class. However, this can be overwritten when defining a instance of that
```
alpmestan: for each `HasPagination` **instance**, I suppose?

KtorZ: Indeed!

*hide in shame*

KtorZ: Fixed in 93838ae9e76ffa18dcf4ff344b52decccd5a6db2

alpmestan:
Thanks @KtorZ!

Your recipe should appear on https://haskell-servant.readthedocs.io/en/latest/cookbook/index.html (note the `latest` in the URL, which reflects the docs for the `master` branch, as opposed to `stable` which has the docs for the latest release) soon (15 mins IIRC? something along those lines).

KtorZ:
Awesome! :tada:

alpmestan:
Here it is: https://haskell-servant.readthedocs.io/en/latest/cookbook/pagination/Pagination.html :-)

ysangkok:
@KtorZ the pagination library is missing a bunch of dependency bound updates, are you still maintaining this? I think it would be best if either the repo became part of this org, or that we removed it.

KtorZ:
I am not directly, though it's part of the [Chordify](https://github.com/chordify/) organization which I am (unfortunately ❤️) no longer working with. 
I don't know if they're still keen on maintaining it but perhaps @jbransen can chime in.

jbransen:
@isomorpheme ^

isomorpheme:
Chiming in as one of the Chordify people with open source responsibility. :) If the dependency bounds can be fixed purely via a Hackage revision then @swamp-agr should be able to do that; if not I'd be happy to review a patch.

In general I expect we'll maintain it just to the degree that we need for internal use - we're open to PRs of course but I've noticed that reviewing those tends to drift down to the bottom of my todo list. So with that in mind I wouldn't mind at all to transfer the repo to this org (but removing the cookbook instead is also fine by me).

ysangkok:
Patch submitted at haskell-servant-pagination/pull/24 , @isomorpheme

ysangkok:
@isomorpheme Please review the linked patch, it seems like swamp-agr is not available.

PR #948, opened 2018-04-18
servant-client: update README.md and test it

domenkozar:
`README.md` was outdated, so I've brought it up in sync with the codebase and made sure it won't happen again.

I've tried to keep example in sync with http://haskell-servant.readthedocs.io/en/stable/tutorial/Client.html but with some simplification so that it's closer to a hello world.

### DIFF
--- a/servant-client/README.lhs
+++ b/servant-client/README.lhs
@@ -1,0 +1,1 @@
+README.md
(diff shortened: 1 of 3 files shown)


alpmestan:
Oh, does that work? Interesting! It does break on CI though, but I suspect this is because you're not listing `markdown-unlit` under `build-tool-depends`... isn't it?

domenkozar:
Seems to work now.

alpmestan:
@phadej Looks good to you?

PR #953, opened 2018-04-25
servant-server: make file-serving functions polymorphic in the monad

alpmestan:
This simply seems possible without breaking anything, so why not?

I've only tried to build servant-server and the tutorial. Let's see what CI says, but it seems to me like the monad is never going to be ambiguous, unless you have a single handler which does file serving, but you still use another monad and call `hoistServer` on your handler, maybe?

Thoughts?


alpmestan:
Alright, CI passes (except 7.8 of course, but that's not new), so it seems like there is no problem with generalising the monad in the typical use cases. I think it's sane that in the case I mentionned above, one has to add some type annotations to let ghc know what the source monad is. If anyone disagrees, please let your voice be heard!

phadej:
@alpmestan feel free to merge when you are satisfied with "discussion".

alpmestan:
Alright, let's merge this then =)

PR #954, opened 2018-04-26
309 handle application exceptions with 500 errors

### LINKED ISSUE #309: Handle application exceptions with 500 errors
In https://github.com/yesodweb/wai/issues/496#issuecomment-167382191 @snoyberg suggests that application errors should not get to the logger middleware and rather get caught by the application itself.

#303 is also related.

I do not have a good understanding of where the handling should happen so I am filing this issue to at least open a discussion.

v0d1ch:
fixes #309

### DIFF
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -125,10 +125,14 @@ import           Servant.Server.Internal
 serve :: (HasServer api '[]) => Proxy api -> Server api -> Application
 serve p = serveWithContext p EmptyContext
 
+type FullyEvaluateResponse = Bool
+
 serveWithContext :: (HasServer api context)
     => Proxy api -> Context context -> Server api -> Application
 serveWithContext p context server =
-  toApplication (runRouter (route p context (emptyDelayed (Route server))))
+  toApplication
+    False
+    (runRouter (route p context (emptyDelayed (Route server))))
 
 -- | Hoist server implementation.
 --
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -122,16 +122,15 @@ import           Servant.Server.Internal
 -- > main :: IO ()
 -- > main = Network.Wai.Handler.Warp.run 8080 app
 --
-serve :: (HasServer api '[]) => Proxy api -> Server api -> Application
-serve p = serveWithContext p EmptyContext
 
-type FullyEvaluateResponse = Bool
+serve :: (HasServer api '[Bool]) => Proxy api -> Server api -> Application
+serve p = serveWithContext p (False :. EmptyContext)
 
 serveWithContext :: (HasServer api context)
     => Proxy api -> Context context -> Server api -> Application
 serveWithContext p context server =
   toApplication
-    False
+    False -- determins if we should fully evaluate response
     (runRouter (route p context (emptyDelayed (Route server))))
 
 -- | Hoist server implementation.
--- a/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
+++ b/servant-server/src/Servant/Server/Internal/RoutingApplication.hs
@@ -86,15 +86,19 @@ instance MonadThrow m => MonadThrow (RouteResultT m) where
     throwM = lift . throwM
 
 toApplication :: Bool -> RoutingApplication -> Application
-toApplication fullyEvaluate ra request respond = ra request routingRespond
- where
-  maybeEval :: (RouteResult Response -> IO ResponseReceived)
-     -> RouteResult Response -> IO ResponseReceived
-  maybeEval resp = if fullyEvaluate then force resp else resp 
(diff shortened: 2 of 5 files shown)


qrilka:
@v0d1ch I've tried out a90b6716072c9b2b1cdc87f1c8c2022cd8267711 and it doesn't seem to work properly, See https://gist.github.com/qrilka/49b0634fce17bcbd2cd4abea503a0f8c

qrilka:
Update doesn't  seem to change the result.
BTW maybe it's worth to fix issue with boolean blindness here?
Using `serveWithContext` and opaque `(True :. EmptyContext)` doesn't look quite user-friendly API

v0d1ch:
Yeah I could probably use some help here.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -84,13 +85,20 @@ instance MonadTransControl RouteResultT where
...
-  routingRespond (Route v)       = respond v
+toApplication :: Bool -> RoutingApplication -> Application
+toApplication fullyEvaluate ra request respond =
+  ra request (maybeEval routingRespond)
+  where
+    maybeEval :: (RouteResult Response -> IO ResponseReceived)
+              -> RouteResult Response -> IO ResponseReceived
+    maybeEval resp =
+      if fullyEvaluate
+        then force resp
```
phadej: http://hackage.haskell.org/package/deepseq-1.4.3.0/docs/src/Control.DeepSeq.html#line-482 - `NFData (a - > b)` probably doesn't do what you expect.

servant-server/src/Servant/Server.hs
```diff
@@ -122,13 +124,24 @@ import           Servant.Server.Internal
 -- > main :: IO ()
 -- > main = Network.Wai.Handler.Warp.run 8080 app
 --
-serve :: (HasServer api '[]) => Proxy api -> Server api -> Application
-serve p = serveWithContext p EmptyContext
 
-serveWithContext :: (HasServer api context)
+serve :: (HasServer api '[Bool]) => Proxy api -> Server api -> Application
```
phadej: I'm against of using `Context` to guide whether to eval or not eval. IMHO single global choice should be enough for now.

Something https://github.com/haskell-servant/servant/issues/309#issuecomment-287222495 is good.

Context won't work for subapis anyway, as we give only single `Context` to `serveWithContext`. And if we use combinators to alter the context, we could be more direct and alter "whether to eval or not".

bergey: The comment suggests both a global setting & a Context.  Are you requesting both, or just the global?

alpmestan: Just global I think.

servant-server/src/Servant/Server/Internal/RoutingApplication.hs
```diff
@@ -93,13 +95,20 @@ instance MonadTransControl RouteResultT where
...
     throwM = lift . throwM
 
-toApplication :: RoutingApplication -> Application
-toApplication ra request respond = ra request routingRespond
- where
-  routingRespond :: RouteResult Response -> IO ResponseReceived
-  routingRespond (Fail err)      = respond $ responseServantErr err
-  routingRespond (FailFatal err) = respond $ responseServantErr err
-  routingRespond (Route v)       = respond v
+toApplication :: Bool -> RoutingApplication -> Application
```
phadej: Last nitpick: let's have

```haskell
data Evaluate = Force | Lazy deriving (Show)
```

so we won't be `Bool`-blind.

servant-server/src/Servant/Server/Internal/Handler.hs
```diff
@@ -46,3 +46,10 @@ instance MonadBaseControl IO Handler where
 
 runHandler :: Handler a -> IO (Either ServantErr a)
 runHandler = runExceptT . runHandler'
+
+-- determins if response should be reduced to NF
+data Evaluate =
```
v0d1ch: @phadej Is this a good place for this type ?

servant-server/src/Servant/Server.hs
```diff
@@ -132,7 +132,7 @@ serve p = serveWithContext p EmptyContext
 serveWithContext :: (HasServer api context)
     => Proxy api -> Context context -> Server api -> Application
 serveWithContext p context server =
-  toApplication (runRouter (route p context (emptyDelayed (Route server))))
+  toApplication Force (runRouter (route p context (emptyDelayed (Route server))))
```
v0d1ch: @phadej Should I alter `serveWithContext` to take in another param that will determine if we want to NF the response or not ? How would you like this to look for the end user in a sense do you want them to be able to control the response evaluation ?

tchoutri:
Hi @v0d1ch! Could you please rebase your PR? :)
If you are willing to finish the work on this, it can be feature in the next major release of Servant!

PR #967, opened 2018-05-28
Allow network-2.7

phadej:


alpmestan:
Finally moving our window of GHC releases? Yes!

phadej:
that's release-0.13, and nope, not moving for it. (haskell-ci reorders newest ghc to be first in the matrix)

PR #968, opened 2018-06-01
Add safeLink'

### LINKED ISSUE #952: safeLink variant with post-compose.
Currently

```haskell
safeLink :: ... => Proxy api -> Proxy endpoint -> MkLInk endpoint
```
`MkLink endpoint` is potentially a chain of arrows `Foo -> Bar -> Link`

Let's change `HasLink endpoint` class so we know that everyhing is known to be `... -> Link`, so we can apply `Link -> a` to it.

We can introduce

```haskell
-- | I don't link primed name.
safeLink' :: ... => (Link -> a) ->  Proxy api -> Proxy endpoint -> MkLink a Link
```

and `safeLink` would change to

```haskell
safeLink :: ... => Proxy api -> Proxy endpoint -> MkLink Link endpoint
```

One can argue that new signature is cleared, as the result `Link` is clearly visible.

phadej:
Resolves #952

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ before_cache:
 
 matrix:
   include:
-    - compiler: "ghc-8.4.2"
+    - compiler: "ghc-8.4.3"
     # env: TEST=--disable-tests BENCH=--disable-benchmarks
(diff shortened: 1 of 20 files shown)


alpmestan:
Don't forget to add it to the changelog :-)

Issue #135, opened 2015-06-20
Redirecting from inside a handler


jhickner:
I just thought I'd post this since it took me a while to figure out. Please comment if there's a better way to do this:

    import Control.Monad.Trans.Either

    redirect303 :: Monad m => ByteString -> EitherT ServantErr m a
    redirect303 url = left err303 { errHeaders = [("Location", url)] }

    -- then, inside your handler you can:
    redirect303 "/login"

alpmestan:
Hello,

I've suggested the inclusion of a generic `Redirect` combinator [here](https://github.com/haskell-servant/servant/issues/117). Maybe that would be of interest to you too?

jhickner:
Would you use that just like `Headers`? That would be great. How would you make it optional though? For instance how do you encode the idea that the route may return some JSON object or it may redirect?

alpmestan:
Here's how it would be used:

``` haskell
"users" :> ReqBody '[FormUrlEncoded] User :> Redirect Post ViewUser UserAPI
-- where ViewUser is, e.g:
type ViewUser = "users" :> Capture "userid" UserId :> Get '[HTML] User
-- and UserAPI contains ViewUser
```

This would describe an endpoint reachable with **POST** requests, which when successful redirects you to a user profile. What you have to do for that is provide an URI (to be used as the value of the `Location` header) as the return value of your handler, and the only way to do that within servant is to use the [safe link module](https://hackage.haskell.org/package/servant-0.4.2/docs/Servant-Utils-Links.html).

Is that clearer now? Let me know if you want some more explanations or an actual complete example you can examine and toy around with.

jkarni:
> How would you make it optional though? For instance how do you encode the idea that the route may return some JSON object or it may redirect?

@jhickner That's not (easily and neatly - i.e., without abusing `Left`) that easy with the current situation. That's a pretty general problem - HTTP status codes are currently *too* static (modulo `Left`). For now, I think your original approach might be best...

jhickner:
The `Redirect` combinator wouldn't cover the use case that prompted this question, but it looks very useful in general. I think `Left` is the right approach or what I needed to do, which was redirect in case of error.

jkarni:
Closing this since we already have an open issue for static redirects, and it seems we all agree `Left` is the right approach for dynamic ones. (Feel free to reopen if anyone disagrees.)

PR #969, opened 2018-06-02
WIP: Issue 841: Rework of Verb

phadej:
this is WIP, but I open this for preview.

Note: `#include "overlapping-compat.h"` is removed from `Servant.Server.Internal`.

*EDIT* If I will work on something in ZuriHac, this will probably be it.

- [ ] add status code to `Redirect`
- [ ] add `Redirect`(s) to `ComprehensiveAPI`
- [x] Use `Verb`, no need for `Verb'`. We can break that.
- [ ] Unify response and request headers
- [x] `contentTypes` in `Verb` or `Result`?
- [x] stream
- [ ] AllCTMimeRender without OverlappingInstances


phadej:
- Redirect combinator https://github.com/haskell-servant/servant/issues/117
- Redirect from inside the handler https://github.com/haskell-servant/servant/issues/135

servant-server/example/greet.hs
```diff
@@ -25,16 +25,24 @@ instance FromJSON Greet
...
 -- API specification
-type TestApi =
-       -- GET /hello/:name?capital={true, false}  returns a Greet as JSON
-       "hello" :> Capture "name" Text :> QueryParam "capital" Bool :> Get '[JSON] Greet
+
+-- GET /hello/:name?capital={true, false)
+-- returns a Greet as JSON or redirects
+type HelloEndpoint =
+    "hello" :> Capture "name" Text :> QueryParam "capital" Bool :>
+        Verb' 'GET (Redirect :> Result 200 '[JSON] Greet)
```
jkarni: Should we maybe use `:<|>` instead of `:>`, since it's *either* one *or* the other, rather than one *and* the other?

servant/src/Servant/API/Redirect.hs
```diff
@@ -0,0 +1,15 @@
...
+                 (Typeable)
+import           Prelude ()
+import           Prelude.Compat
+import           Network.URI
+                 (URI (..))
+
+-- | TODO: TBW...
+-- 
+-- TODO: Add to ComprehensiveAPI
+newtype Redirect = Redirect URI deriving (Typeable, Eq, Ord, Show)
```
jkarni: I think this should take a status code as well (since there are multiple types of redirect) or a type indicating one of the status codes.

servant/src/Servant/API/Verbs.hs
```diff
@@ -21,12 +21,20 @@ import           Network.HTTP.Types.Method
...
 
+import Servant.API.NoContent
+
 -- | @Verb@ is a general type for representing HTTP verbs (a.k.a. methods). For
 -- convenience, type synonyms for each verb with a 200 response code are
 -- provided, but you are free to define your own:
 --
 -- >>> type Post204 contentTypes a = Verb 'POST 204 contentTypes a
-data Verb (method :: k1) (statusCode :: Nat) (contentTypes :: [*]) (a :: *)
+type Verb (method :: k1) (statusCode :: Nat) (contentTypes :: [*]) (a :: *)
```
jkarni: It's very nice that `Verb` can be nearly backwards compatible in this way. But maybe it's `Get`, `Put`, etc that should have this sort of synonym, and we can just redefine `Verb`, since `Verb` is much less user-facing?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -257,31 +316,107 @@ methodRouter splitHeaders method proxy status action = leafRouter route'
...
+
+instance HasServerR Redirect where
+    type ServerR Redirect = Redirect
+
+    routeR _proxy _context _request = routeResultR $ \(Redirect uri) ->
+        Route $ redirectResponse uri
+
+instance HasServerR api => HasServerR (Redirect :> api) where
+    type ServerR (Redirect :> api) = Either Redirect (ServerR api)
+
```
jkarni: It's strange that both this and the above instance are needed. And also a little sad that we might end up with nested `Either`s. If we switch to having both `:>` and `:<|>` inside the `Verb`, I guess we could just map uniformly over `:<|>` into a sumtype? And something to consider is maybe using an extensible sum type?

servant-server/example/greet.hs
alpmestan: It might be just me, but I find both of them equally confusing (not terribly though, I could get used to both being used there). To me, and in very vague terms (i.e don't pay too much attention to this), the `:>` let us chain constraints on the input (request), while `:<|>` lets us "or" constraints, but then neither corresponds to what feels like we're doing under that `Verb`, which would be something like "stating a few properties of the output". `:&` maybe? Not sure.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -34,89 +31,119 @@ module Servant.Server.Internal
...
+        -> ServerT api n
+
+class HasServerR api where
+    type ServerR api :: *
+
+    -- TODO: add headers, method we don't need "Router", but "Leaf", in a sense.
+    routeR
+        :: Proxy api
+        -> Context context
+        -> Request
```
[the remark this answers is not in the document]
phadej: We should, but we can do "bad" things `withRequest` in `HasServer` too.

servant-server/example/greet.hs
phadej: I used `:>` because we already have it.

`:<|>` because of the sum doesn't really make sense, as we can do products too. something like `:><` or `:*` (as something "near enough to" ⊗), but those are also bad as the operator should convey it's fixity (not associative here!), and `:>` does it. `:<` would have "wrong" feeling.

Also having less operators is not a bad thing.

servant/src/Servant/API/Redirect.hs
phadej: Indeed. Will do.

servant/src/Servant/API/Verbs.hs
phadej: You are probably right. I don't remember writing `Verb` that often myself.

servant-server/src/Servant/Server/Internal.hs
phadej: We can have unconditional and optional `Redirect`s. There fore two instances.

One option is to have yet another class for so

```
class (HasClientR api, HasClientRR combinator) => HasClient (combinator :> api)
```

That would be cool as we won't need as meny `FlexibleInstances` (we can do same factoring for `HasClient (combinator :> api)` instances as well. We'll end up with 4 classes then. It's only a naming problem.

Ending up with nested sums (`Redirect`) and products (`Headers`) is inevitable, and I'll keep api simple. It's not even that contrived to have

```haskell
"endpoint" :> Verb' 'Get
  (  Headers '['Header "X-API-Version"]
  :> Redirect
  :> Headers ['Header "X-API-Something"]
  :> Result 200 '[JSON] Value)
```

phadej:
API design question, when `Verb'` is renamed, should we have some aliases to `Verb 'GET`, `Verb 'POST` etc. as `Get ctypes a` is alias to `Verb 'GET (Result 200 ctypes a)`.

So
- `GET` enum value
- `Get` simple `Result`
- ??? `Verb 'GET`, I suspect this to be used often enough.

phadej:
Another question is how to deal with "right extensions". As I commented above I think we don't need to complicate things with `HList` and `HSum` (`NP` / `NS`), however:

- `Headers` uses custom type: `Headers hs a`
- `Redirect` uses "anonymous" `Either`

I think we should be consistent and use one of following options consistently:

### Separate custom product/sum for each combinator

```haskell
-- combinator
data Headers hs

-- term
data ResponseHeaders hs a = ResponseHeaders (HList hs) a -- with field names though
```

Note: that `Headers` and `ResponseHeaders` have different kinds. We have similar happening for
`Header` and `RequestArgument` on the request side.

*Side comment* This change let us unify handling of request and response combinators, and I will do that. Users can decide whether they want single `Header` or plural `Headers`. Former is very simple when you have one header, latter is more convenient when you have many headers to parse/respond.

Note: we'll need two types for right and left:

```haskell
data Headers hs

data ResponseHeaders hs a = MkResponseHeaders (HList hs) a
newtype RequestHeaders hs = MkRequestHeaders (HList hs)

type ServerT (Headers hs :> api) m = RequestHeaders hs -> ServerT api m
type ServerR (Headers hs :> api)   = ResponseHeaders hs (ServerR api)
```

### Embrace `(a, b)` and `Either a b`

This is uniform to how we deal on the left, there we use `->`

```haskell
data Headers hs = MkHeaders (HList hs)

type ServerT (Headers hs :> api) m = Headers hs -> ServerT api m
type ServerR (Headers hs :> api)   = (Headers hs, ServerR api)

-- ^ note: we can still have convinient pair constructor with specific type for that:
mkWithHeaders :: HList hs ->  a -> (Headers hs, a)
```

This in fact let us use same name for combinator and data type.

---

alpmestan:
> API design question, when Verb' is renamed, should we have some aliases to Verb 'GET, Verb 'POST etc. as Get ctypes a is alias to Verb 'GET (Result 200 ctypes a).

On the one hand, it feels like `Get` should just be the general one, and `Get200` or something like that would be what we currently call `Get`, i.e status code fixed to 200, no headers. On the other hand, that'd "uselessly" break quite a lot of code. I'm honestly not sure what the best option is. As witnessed by the hard time I have coming up with a good name for the general `Get`.

> Another question is how to deal with "right extensions".

I really like the simplicity of your second suggestion and how it makes `ServerR` nicely mirror `ServerT`.

servant-server/example/greet.hs
jkarni: > `:<|>` because of the sum doesn't really make sense, as we can do products too. 

What I mean is that *both* `:>` *and* `:<|>` would be allowed inside the `Verb`. The first would correspond to products, the second to sums. That makes a little clearer what the expected handler type would be from the API type, and allows us in the instances to not have to pattern match on individual LHSs.

jkarni:
I agree `NP`/`NS` is a little complicated. But it would be sad to have to have handlers use `return` multiple times (instead of just once) to e.g. return a value from a handler that has multiple optional `Redirect`s. And the handlers won't compose well anymore.

The way I think of it, with sums at least, it'd be nice to have an `mtl`/`extensible-effects`-like interface. So:

```
handler1 :: (ThrowsErr1 m) => Handler m A
handler2 :: (ThrowsErr2 m) => A -> Handler m B

handler3 :: (ThrowsErr1 m, ThrowsErr2 m) => Handler m B
handler3 = handler1 >>= handler2


handler4 :: <throws who knows what m> => Handler m Int
handler4 = return 5

handler5 :: (Redirect, ...) => Handler m A
handler5 = throwError $ inject $ Redirect ...
```

Would be really nice. I know there are problems with this:

- It doesn't really extend in an obvious way to products
- We need types to be concrete at *some* point. If we make it concrete in the associated type synonym (as we've done so far), then the above only works if the end-user gives a more general type than the inferred one to her handlers.

alpmestan:
I agree that having to chain multiple `return`s would be quite annoying. But your last bullet point there makes me thing this is not the sweet spot we're looking for either here.

arianvp:
I think NS/NP is the way to go here honestly. However, perhaps we want a default mode where people don't have to bother with heterogeneous lists. that is, you should be able to use servant as you used to, but only need to bother about heterogeneous lists when  they want to have the return types encoded in the API.

phadej:
Multiple redirects (I guess with different statuses?) can be handler with new combinator `Redirects` (plural) which bundles them into single NS, while `:>` (or whatever) is encoded as `NS`.

We can have `A * (B + (C * D))` like expressions, potentially.

About `:>` and `:<|>`. We *could* have, rewriting my previous example

```haskell
"endpoint" :> Verb' 'Get
  (  Headers '[ Header "X-API-Version"]
  :> (Redirect :<|>
       Headers '[ Header "X-API-Something"]
    :> Result 200 '[JSON] Value))
```

but it opens a possibility to

```haskell
"endpoint" :> Verb' 'Get (Result 200 '[JSON] Value :<|> Result 200 '[CSV] [Text])
```

and it doesn't work out. Client tells it wants `JSON`, so we cannot return `CSV`.  We'd need to have a pair handler there, not an `Either` (it's not our choice, but theirs!)! It might be possible to encode, but it will definitely make implementation more complex, for a little benefit.

I mentioned *⊗*, as it *overloads* *×* and *+*. But actually i'm seeking for a (right-)module like thing of a "tensor production", but I guess there aren't a notation for those.

So in summary:
- `:<|>` is problematic
- `:>` or `:><` or `:<>` or .. choice is irrelevant, but we already have `:>`, but we can add something else too. Note: than we'd need to add `TypeError (Text "use :><") => HasServerR (comb :> api)`, as I'm sure many (me including) will write wrong operator in wrong place).

phadej:
About plurality, examples:

```haskell
Verb' 'Get (Redirect 300 :> Redirect 301 :> Result ...)
Verb' 'Get (Redirects '[300,301] :> Result ...)
Verb' 'Get (Header "foo" Int :> Header "bar" String :> Result ...)
Verb' 'Get (Headers '[Header "foo" Int, Header "bar" String] :> Result ...)
```

And IMO it depends on the details which, nested sums or products, or bundled NS/NP are "better". We can have all of those above.

---

**Different issue** of smashing different combinators into same NS or NP will be also difficult to implement. Returning `Either Redirect (Either (NS Error '[err1, err2] MyValue))` in a handler is not /that/ bad. I think **better** than either

```haskell
NS I '[Redirect, Error err1, Error err2, I MyValue]
Either (NS I '[Redirect, Error err1, Error err2]) MyValue
```

---

I think that term level inconvenience is way easier to abstract than to make some crazy stuff to produce "flatter" handler types.

---

Note I prototyped current design in https://github.com/phadej/servant-tiny/tree/more-comb, feel free to explore desing space based on it (it's simpler), but I won't bike-shedding prevent me proceeding with current design. I.e. MTL / Magical NS-smashing etc is out without "existential proof" it will work. (Note how little extensions are used there!)

phadej:
How's about

```
- `GET` enum value
- `Get ctypes a = Verb 'GET (Result 200 ctypes a)  = Get' (Result 200 ctypes a)`
- `Get' = Verb 'GET`
```

using prime is a little boring, but we already do that for `QueryParam'`...

jkarni:
In the specific case of content types, I think they should be *outside* of
Result, with the Verb. This is because we check for content types *before*
running the Handler (as we should) and therefore don't know what Result (if
any) we'll get.

This also makes people be sure to implement instances for error types that
match what they claim to handle, which is good - the server can thus also
respond with the appropriate format for errors.

Not sure if that solves the problem generally though.

Oleg Grenrus <notifications@github.com> schrieb am Mi., 13. Juni 2018,
13:21:

phadej:
Then we'll need to have a way to bypass Content-Type negotiation for `NoContent` or unconditional `Redirect`. Abusing empty `contentTypes` to mean *any* might work, though feels dirty. Generic programming might become difficult (i.e. one have to be careful with `Filter` like constructs, but I guess one needs to be careful today already).

jkarni:
Is it horrible if we fail with `Not Acceptable` in a handler that only redirects/returns no content if the list of content-types does not include the requested content-type? That is, `NoContent` and `Redirect` would both be instances of all serializations, but you'd still have to mention the ones you want to support.

phadej:
I'll try, maybe having `Verb method ctypes` and `VerbAnyCT method` would be good compromise

phadej:
I refactored `servant-server` part a bit with some of @jkarni suggestions. Now you can write (as in `greet` example):

```haskell
Verb 'GET '[JSON] (Result 400 Text :<|> Result 200 Greet)
```

IMHO it's enough for core `servant` atm, `checked-exceptions` could have own `Result`-like combinator.

phadej:
FWIW

it's possible to have
```haskell
type GET = Verb 'GET
```

but I'm not sure if that will be more confusing than helpful?

phadej:
`servant-client` and

```haskell
Verb 'GET '[JSON] (Result 400 Text :<|> Result 200 Greet)
```

I think we'll need to dispatch on status code, so we'll need some kind of `Alternative` for `ClientM`.

alpmestan:
Yes, we need to do some kind of "inversed/dual routing". Fortunately it's very simple, as the status code is the _only_ thing we will ever look at to decide what to try to decode the response to.

fisx:
I think since we have https://github.com/haskell-servant/servant-uverb now, this can be closed?  Please reopen if I'm wrong.

Issue #931, opened 2018-03-24
streams in constant memory seems invalid


phadej:
`replicate_ n (res void)` is wrong thing to do. `res` should be called only once, and processing down inside the continuaton. See e.g. how `runResultStream` is made.

I suspect that the problem is with test itself, hopefully. I tried run it threaded RTS, still same result. (`startWaiApp` forks a new thread for test server). 

*Note* chunks read by client aren't *5 MB* in size, in fact on my machine they are most *< 1k*.

For some reason when running the test there are **Lazy** `ByteString` chunks visible. That's fishy.

![screenshot from 2018-03-24 12-35-30](https://user-images.githubusercontent.com/51087/37863036-085ad990-2f60-11e8-97d3-823cddfdd8d2.png)

Consuming whole response isn't good, as GC will have opportunity to clean everything. Better approach would be to check maximum residency before the test, and check that it haven't grown (or bigger than something the test expected to use).

cc @jkarni

```diff
diff --git a/servant-client/test/Servant/StreamSpec.hs b/servant-client/test/Servant/StreamSpec.hs
index ad4a266..e4aa823 100644
--- a/servant-client/test/Servant/StreamSpec.hs
+++ b/servant-client/test/Servant/StreamSpec.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns           #-}
 {-# LANGUAGE CPP                    #-}
 {-# LANGUAGE ConstraintKinds        #-}
 {-# LANGUAGE DataKinds              #-}
@@ -83,7 +84,7 @@ server = serve sapi
     lotsGenerator f r = do
       f ""
       withFile "/dev/urandom" ReadMode $
-        \handle -> streamFiveMBNTimes handle 1000 r
+        \handle -> streamFiveMBNTimes handle 10 r
       return ()
 
     streamFiveMBNTimes handle left sink
@@ -93,8 +94,6 @@ server = serve sapi
           sink msg
           streamFiveMBNTimes handle (left - 1) sink
 
-
-
 {-# NOINLINE manager' #-}
 manager' :: C.Manager
 manager' = unsafePerformIO $ C.newManager C.defaultManagerSettings
@@ -126,11 +125,21 @@ streamSpec = beforeAll (CS.startWaiApp server) $ afterAll CS.endWaiApp $ do
        runResultStream res `shouldReturn` (jra, jrb, jra, Nothing)
 
     it "streams in constant memory" $ \(_, baseUrl) -> do
[...]

PR #973, opened 2018-06-09
Use fixed cabal-install-2.2

phadej:


alpmestan:
Fails on 8.4.3:

``` haskell
Failures:
  test/Servant/StreamSpec.hs:142:8: 
  1) Servant.Stream.Servant.Stream streams in constant memory
       predicate failed on: 22020096
Randomized with seed 1509620355
```

phadej:
https://github.com/haskell-servant/servant/issues/931

I'm tempted to delete that test.

alpmestan:
Let's delete it (or expect it to fail) yes, and use #931 to improve on that to get something useful, reliable and that passes. :)

PR #974, opened 2018-06-09
website/tutorial tweaks

alpmestan:
I also wanted to add instructions for `stack` but when I tried `stack build` (before even trying to add the tutorial project to `stack.yaml`) it failed because of some bytestring flag somewhere. If someone wants to figure that out, please send a patch to the tutorial index there.

Here's how it looks:

servant1
servant2


phadej:
bytestring flag is because you have old `stack`. If you have very old `stack`, best way to upgrade is to download new one (i.e. `stack upgrade` won't work)

alpmestan:
Could we use `--allow-newer` for the cookbook? Would fix the CI issue at https://travis-ci.org/haskell-servant/servant/jobs/390243344#L1123

PR #975, opened 2018-06-10
Run `captureAllSpec`

jml:
This was missed due to an oversight.

Adding an explicit export list to `ServantSpec.hs` would allow a computer to catch these problems. I haven't done that here.


jml:
Thanks! I can't push "merge".

alpmestan:
I was just waiting for the green CI (one of the CI scenarios errored out because of the "no output in 10 mins" rule, I wanted to make sure that one was fine as well).

PR #979, opened 2018-06-14
Implement CaptureMany, a non-greedy CaptureAll

### LINKED ISSUE #962: Implement a "CaptureSome" combinator
There [exist in the wild](https://docs.docker.com/registry/spec/api) APIs such as this:

```
GET /v2/<name>/manifests/<reference>
GET /v2/<name>/tags/list
```

where `<name>` comprises _one or more_ path segments. It would be nice to be able to do something like this:

```
"v2" :> CaptureSome "name" [String] :> "manifests" :> Capture "reference" String :> Get ...
"v2" :> CaptureSome "name" [String] :> "tags" :> "list" :> Get ...
```

to support them.

jml:
Takes a pretty simple approach of adding a new router, and having that router try all the routes, starting with the longest and working backwards.

I didn't think very hard about test coverage—I just copied and adjusted the `CaptureAll` tests.

It's very likely I've missed all sorts of things. Feedback very much welcome.

Fixes #962

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -228,7 +228,7 @@ instance (KnownSymbol capture, FromHttpApiData a, HasServer api context)
 -- > server :: Server MyApi
 -- > server = getSourceFile
 -- >   where getSourceFile :: [Text] -> Text -> Handler Book
--- >         getSourceFile pathSegments = ...
+-- >         getSourceFile pathSegments pathLast = ...
 instance (KnownSymbol capture, FromHttpApiData a, HasServer api context)
       => HasServer (CaptureMany capture a :> api) context where
 
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -67,7 +67,7 @@ import           Prelude ()
 import           Prelude.Compat
 import           Servant.API
                  ((:<|>) (..), (:>), Accept (..), BasicAuth,
-                 BoundaryStrategy (..), Capture', CaptureAll, Description,
+                 BoundaryStrategy (..), Capture', CaptureAll, CaptureMany, Description,
                  EmptyAPI, FramingRender (..), Header', If, IsSecure (..),
                  QueryFlag, QueryParam', QueryParams, Raw,
                  ReflectMethod (reflectMethod), RemoteHost, ReqBody',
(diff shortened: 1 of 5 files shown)


servant-server/src/Servant/Server/Internal.hs
```diff
@@ -216,6 +216,41 @@ instance (KnownSymbol capture, FromHttpApiData a, HasServer api context)
...
+-- by simply providing an instance of 'FromHttpApiData' for your type.
+--
+-- Example:
+--
+-- > type MyApi = "src" :> CaptureMany "segments" Text :> Capture "last" Text :> Get '[JSON] SourceFile
+-- >
+-- > server :: Server MyApi
+-- > server = getSourceFile
+-- >   where getSourceFile :: [Text] -> Text -> Handler Book
+-- >         getSourceFile pathSegments = ...
```
alpmestan: getSourceFile is probably missing a `pathLast` argument here? For clarity.

alpmestan:
It would be great if one or more of @phadej @kosmikus @jkarni could take a look at this.

I think the approach is reasonable, but of course the fact that we're slightly repeating ourselves is a little bit bothering. It feels like we might be able to have something a little general behind all those `Capture`s. Perhaps capturing the variability with:

``` haskell
data CaptureKind
  = All Type
  -- matches CaptureAll: we try to capture the rest of the path as the given type,
  -- error out if we don't succeed
  | Many Type
  -- matches CaptureMany: we try to capture as many path components as the given type as we can,
  -- but possibly leave more path components still to be accounted for in the API type.
  -- Could be augmented with a Nat to say "require that at least `n :: Nat` path components are
  -- extracted.
  | Precisely Nat Type  -- matches the standard Capture when the Nat is 1, but is a little more general
```

This might even be summarisable as a single constructor with booleans and naturals, but I'd rather not, the distinction is clear enough here. Then a lot of the infrastructure could be shared and we would special case when it comes to computing the type of the (server or client) argument, the router, etc. But this all feels very `kleene`-y @phadej, at least as far as routing is concerned. You may have more insightful thoughts on the topic. :-)

@jml These are just thoughts, I'm in no way saying this necessarily should happen (if at all) in this PR.

jml:
It does feel a lot like a Kleene algebra. Whether or not we model the user-facing API that way, my guess is we should probably model `Router'` that way.

I'm going to vote for not doing that in this PR, though ;)

fisx:
I would like to propose the more self-explanatory name `CaptureAllNonGreedy`.  It's a bit long, but I think the extra clarity makes it worth it.

Feel free to ignore me, though.  `CaptureMany` is fine, too.  (-:

jml:
It's been about a month. Is there anything I can do to move this forward?

alpmestan:
The branch needs to be rebased against `master`. I was favorable to this patch, @phadej any objection to merging this (until we find a way to avoid repeating ourselves so much with the capture variants) ?

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -285,6 +286,47 @@ captureAllSpec = do
...
+        liftIO $ decode' (simpleBody response) `shouldBe` Just Animal { numberOfLegs = 4, species = "fish" }
+
+      it "can capture when there are no elements in 'pathInfo'" $ do
+        response <- get "/beholder"
+        liftIO $ decode' (simpleBody response) `shouldBe` Just Animal { numberOfLegs = 0, species = "beholder" }
+
+      it "returns 400 if the decoding fails" $
+        get "/notAnInt/fish" `shouldRespondWith` 400
+
+      it "returns 400 if the decoding fails, regardless of which element" $
```
phadej: Is this statement correct, doesn't `CaptureMany` succeed always, but then then we fail because not the whole path is matched?

jml:
I've rebased this PR. The build is failing, but since master is _also_ failing, it's hard to tell whether my branch needs adjusting. I'm going to put this on hold again until there's a known good master to build against.

jml:
```haskell
CaptureMany "many" Integer :> Capture "one more" :> Integer :> Get '[JSON] Int
```

This looks like it's incorrect?

Shouldn't it be?

```haskell
CaptureMany "many" Integer :> Capture "one more" Integer :> Get '[JSON] Int
```

And, just so I follow, the point of the question is that the types of the `Capture*` statements are the same?

jml:
Realistically, I'm not going to work on this further. Closing, but others should feel free to re-use if they want to pick this work up.

PR #980, opened 2018-06-17
CaptureMany client implementation

jml:
Builds on #979 to provide a `CaptureMany` implementation for client. 

Filing this as a separate PR, but can push it to #979 if that's easier from a review POV.


jml:
I'm not going to get to this in the foreseeable future.

PR #988, opened 2018-06-23
Add toStreamGenerator (NonEmpty a) instance

phadej:
I'd like to have

```haskell
"stream" :> StreamGet NewlineFraming '[PlainText] [Text] -- lines
```

Also having `StreamingReqBody` should be in `servant`

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -840,6 +840,24 @@ instance OVERLAPPABLE_
           status = fromInteger $ natVal (Proxy :: Proxy status)
           p = Proxy :: Proxy a
 
+-- | TODO: mention the endpoint is streaming, its framing strategy
+--
+-- Also there are no samples.
+instance OVERLAPPABLE_
+        (MimeRender ct a, KnownNat status
+        , ReflectMethod method)
+    => HasDocs (Stream method status framing ct a) where
+  docsFor Proxy (endpoint, action) DocOptions{..} =
+    single endpoint' action'
+
+    where endpoint' = endpoint & method .~ method'
+          action' = action & response.respTypes .~ allMime t
+                           & response.respStatus .~ status
+          t = Proxy :: Proxy '[ct]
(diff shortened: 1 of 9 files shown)


servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -840,6 +840,24 @@ instance OVERLAPPABLE_
           status = fromInteger $ natVal (Proxy :: Proxy status)
           p = Proxy :: Proxy a
 
+-- | TODO: mentiong it's streaming + framing strategy
```
alpmestan: Typo + TODO :smile:

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -840,6 +840,24 @@ instance OVERLAPPABLE_
...
+-- Also there are no samples.
+instance OVERLAPPABLE_
+        (MimeRender ct a, KnownNat status
+        , ReflectMethod method)
+    => HasDocs (Stream method status framing ct a) where
+  docsFor Proxy (endpoint, action) DocOptions{..} =
+    single endpoint' action'
+
+    where endpoint' = endpoint & method .~ method'
+          action' = action &response.respTypes .~ allMime t
```
alpmestan: spacing between & and response.

servant/src/Servant/API/Stream.hs
```diff
@@ -41,24 +47,61 @@ type StreamGet  = Stream 'GET 200
...
+  buildFromStream x = runResultStream x $ \r -> do
+      e <- r
+      case e of
+          Nothing        -> fail "Empty stream"
+          Just (Left er) -> fail er
+          Just (Right y) -> do
+              ys <- lazyRead r
+              return (y :| ys)
+
+-- | TODO: hide from export
```
alpmestan: TODO to address

servant-docs/src/Servant/Docs/Internal.hs
phadej: Left intentionally, opened an issue already #989

PR #1003, opened 2018-07-04
Update dependencies + servant-generic + s/Utils//

phadej:


servant-server/src/Servant/Server/Generic.hs
```diff
@@ -0,0 +1,51 @@
+{-# LANGUAGE DataKinds            #-}
```
alpmestan: Should we add the module header with the Copyright thing, to indicate that Patrick wrote this? (and same for the other servant-generic modules)

phadej: I added a license note into `Servant.API.Generic`. The only "lines" by Patrick in `Server.Generic` are `Data AsServerT m`, `type AsServerT m :- ..` and `type AsServer`. Not *substantial portion*.

phadej: And `Client` is actually more of your and @domenkozar's work.

alpmestan: OK, thanks. Looks good to me now.

alpmestan:
Thanks @chpatrick!

PR #1005, opened 2018-07-05
Add 'using free client' recipe

phadej:
I'm too easily nerd-snipped, there should be a price tag for that.

### DIFF
--- a/doc/cookbook/using-free-client/UsingFreeClient.lhs
+++ b/doc/cookbook/using-free-client/UsingFreeClient.lhs
@@ -50,8 +50,8 @@ api :: Proxy API
 api = Proxy
 ```
 
-Next we implement a `main`. If passed `server` it will run `server`, if passed
-`client` it will run a small `test` (to be defined next) will be run. This should be
+Next we implement a `main`. If passed `"server"` it will run `server`, if passed
+`"client"` it will run a `test` function (to be defined next). This should be
 pretty straightforward:
 
 ```haskell
--- a/doc/cookbook/using-free-client/UsingFreeClient.lhs
+++ b/doc/cookbook/using-free-client/UsingFreeClient.lhs
@@ -1,0 +1,173 @@
+# Using Free Client (for tests)
+
+Someone asked on IRC about getting `Request` & `Response` of what
+`servant-client` uses, for testing purposes.  My response: to ad-hoc extend
+`servant-client` (as for tests), use `Servant.Client.Free`. This recipe is an
+evidence.
+
+First the module header, but this time We'll comment the imports.
+
+```haskell
+{-# LANGUAGE DataKinds     #-}
+{-# LANGUAGE TypeOperators #-}
+module Main (main) where
+```
+
+We will primarily use `Servant.Client.Free`, it doesn't re-export anything
+from `free` package, so we need to import it as well.
+
+```haskell
+import Control.Monad.Free
+import Servant.Client.Free
+```
+
+Also we'll use `servant-client` internals, which uses `http-client`,
+so let's import them *qualified*
+
+```haskell
+import qualified Servant.Client.Internal.HttpClient as I
+import qualified Network.HTTP.Client                as HTTP
+```
+
+The rest of the imports are for a server we implement here for completeness.
+
+```haskell
+import Servant
+import Network.Wai.Handler.Warp (run)
+import System.Environment       (getArgs)
+```
+
+## API & Main
+
+We'll work with a very simple API:
+
+```haskell
+type API = "square" :> Capture "n" Int :> Get '[JSON] Int
+
+api :: Proxy API
+api = Proxy
+```
+
+Next we implement a `main`. If passed `server` it will run `server`, if
+`client` a small `test` (to be defined next) will be run. This should be pretty
+straigh-forward:
+
+```haskell
+main :: IO ()
+main = do
+    args <- getArgs
+    case args of
+        ("server":_) -> do
+            putStrLn "Starting cookbook-using-free-client at http://localhost:8000"
+            run 8000 $ serve api $ \n -> return (n * n)
+        ("client":_) ->
+            test
+        _ -> do
+            putStrLn "Try:"
+            putStrLn "cabal new-run cookbook-using-free-client server"
+            putStrLn "cabal new-run cookbook-using-free-client client"
+```
+
+## Test
+
+In the actual test, we'll use a `Servant.Client.Free` client.
+Cecause we have a single endpoint API, we'll get a single client function:
+
+```haskell
+fcli :: Int -> Free ClientF Int
+fcli = client api
+```
+
+Next, we can write our small test. We'll pass a value to `fcli` and inspect
+the `Free` structure. First three possibilities are self-explanatory:
+
+```haskell
+test :: IO ()
+test = case fcli 42 of
+    Pure n ->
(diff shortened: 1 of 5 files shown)


doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+        _ -> do
+            putStrLn "Try:"
+            putStrLn "cabal new-run cookbook-using-free-client server"
+            putStrLn "cabal new-run cookbook-using-free-client client"
+```
+
+## Test
+
+In the actual test, we'll use a `Servant.Client.Free` client.
+Cecause we have a single endpoint API, we'll get a single client function:
```
alpmestan: Because*

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+and calling the continuation. We should get a `Pure` value.
+
+```haskell
+        let res = I.clientResponseToResponse res'
+        
+        case k res of
+            Pure n ->
+                putStrLn $ "Expected 1764, got " ++ show n
+            _ -> 
+                putStrLn "ERROR: didn't got a response"
```
alpmestan: didn't g**e**t

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
+# Using Free Client (for tests)
```
alpmestan: "Inspecting and debugging clients (`Servant.Client.Free`)" would I think be a lot clearer and appeal a lot more to people who are looking to do just that =) Thoughts?

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+
+Someone asked on IRC about getting `Request` & `Response` of what
```
alpmestan: I'd suggest:

Someone asked on IRC how one could access the intermediate `Request`s (resp. `Response`s) produced (resp. received) by client functions derived using servant-client.

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+
+```haskell
+type API = "square" :> Capture "n" Int :> Get '[JSON] Int
+
+api :: Proxy API
+api = Proxy
+```
+
+Next we implement a `main`. If passed `server` it will run `server`, if
+`client` a small `test` (to be defined next) will be run. This should be pretty
```
alpmestan: I'd add "if passed" before `client`, for clarity.

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+straigh-forward:
```
alpmestan: straightforward

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+
+In the actual test, we'll use a `Servant.Client.Free` client.
+Cecause we have a single endpoint API, we'll get a single client function:
+
+```haskell
+fcli :: Int -> Free ClientF Int
+fcli = client api
+```
+
+Next, we can write our small test. We'll pass a value to `fcli` and inspect
```
alpmestan: I'd probably try to add a small paragraph here saying that such clients are basically "client functions without a backend", or where the backend has been abstracted out. And that `ClientF` is precisely a functor that represents the operations we need from an http backend. So if we are to emulate one or augment what it does, it will be by interpreting all those operations the way we want to. This also means we get access to the requests and responses and can do anything we want we them as they are produced or consumed, respectively. 

Just as a little intro to the `case ... of` block below. To give it a bit of context.

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+test = case fcli 42 of
+    Pure n ->
+        putStrLn $ "ERROR: got pure result: " ++ show n
+    Free (Throw err) ->
+        putStrLn $ "ERROR: got error right away: " ++ show err
+    Free (StreamingRequest _req _k) ->
+        putStrLn $ "ERROR: need to do streaming request" -- TODO: no Show Req :(
+```
+
+We are interested in `RunRequest`, that's what client should block on:
```
alpmestan: I'd suggest: "that's the only http client backend operation used by a client for a simple API like the one above", or something like that.

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -0,0 +1,173 @@
...
+
+Now we can use `servant-client` internals to convert servant's `Request`
+to http-client's `Request`, and we can inspect it:
+
+```haskell
+        let req' = I.requestToClientRequest burl req
+        putStrLn $ "Making request: " ++ show req'
+```
+
+`servant-client` does a bit more than `httpLbs`, but that's enough for us.
```
alpmestan: the `httpLbs` reference might be a bit obscure for many people no?

phadej:
@alpmestan feel free to amend, that's why it's in `haskell-servant` repo :)

doc/cookbook/using-free-client/UsingFreeClient.lhs
phadej: If you add *client functions without a backend", than you could mention that using `Free` is (probably?) a little easier than implementing own `RunClient m` instance, though it requires to implement only 3 functions too (atm, maybe we'll go do two, let's see). `Free` however is nice especially for single-stepping will do later.

**But** if you only, say need to remember `Request`s made by client, you can write istance for `WriterT [Request] ClientM`. and inspect them afterwards.

phadej: them = Requests

doc/cookbook/using-free-client/UsingFreeClient.lhs
phadej: true.

alpmestan:
@phadej Done. Feel free to take a look :-)

doc/cookbook/using-free-client/UsingFreeClient.lhs
```diff
@@ -1,9 +1,12 @@
-# Using Free Client (for tests)
+# Inspecting, debugging, simulating clients and more
 
-Someone asked on IRC about getting `Request` & `Response` of what
-`servant-client` uses, for testing purposes.  My response: to ad-hoc extend
-`servant-client` (as for tests), use `Servant.Client.Free`. This recipe is an
-evidence.
+or simply put: _a practical introduction to `Servant.Client.Free`_.
+By Oleg Grenrus
```
phadej: nah, I don't need credit here.

phadej: (It's in the `git log`)

phadej:
Let's merge when Travis says green.

alpmestan:
Fixed a few last things. It was green before I fixed the couple of last typos I found. Shall we merge or wait?

PR #1006, opened 2018-07-09
PoC: Make context into simple data

phadej:
Turn `context` into ordinary data:

We have to write a little of boilerplate (class for each combinator using something from context), but OTOH the structure is way simpler. We can still alter context with `NamedContext` like stuff,
though users need to write some boilerplate (or use `vinyl` or other extensible records libraries).
For me this is mostly **win**, as we don't need to reinvent extensible records like thing in `servant`.
Also for simple cases when there's only one thing in a context, usage becomes simpler too,
as that one thing can be used as the context.

Changes needed for `servant-multipart`:

```diff
diff --git a/src/Servant/Multipart.hs b/src/Servant/Multipart.hs
index fb4e4be..66e643f 100644
--- a/src/Servant/Multipart.hs
+++ b/src/Servant/Multipart.hs
@@ -238,13 +238,24 @@ class FromMultipart tag a where
 instance FromMultipart tag (MultipartData tag) where
   fromMultipart = Just
 
+class HasMultipartOptions context tag where
+    getMultipartOptions :: context -> MultipartOptions tag
+
+instance HasMultipartOptions (MultipartOptions tag) tag where
+    getMultipartOptions = id
+
+-- | Note: now we can separate MultipartBackend into backends which have
+-- default options, and which don't!
+instance MultipartBackend tag => HasMultipartOptions () tag where
+    getMultipartOptions () = defaultMultipartOptions Proxy
+
 -- | Upon seeing @MultipartForm a :> ...@ in an API type,
 ---  servant-server will hand a value of type @a@ to your handler
 --   assuming the request body's content type is
 --   @multipart/form-data@ and the call to 'fromMultipart' succeeds.
 instance ( FromMultipart tag a
          , MultipartBackend tag
-         , LookupContext config (MultipartOptions tag)
+         , HasMultipartOptions config tag
          , HasServer sublayout config )
       => HasServer (MultipartForm tag a :> sublayout) config where
 
@@ -261,8 +272,10 @@ instance ( FromMultipart tag a
       psub  = Proxy :: Proxy sublayout
       pbak  = Proxy :: Proxy b
       popts = Proxy :: Proxy (MultipartOptions tag)
-      multipartOpts = fromMaybe (defaultMultipartOptions pbak)
-                    $ lookupContext popts config
+
+      multipartOpts :: MultipartOptions tag
+      multipartOpts = getMultipartOptions config
+
       subserver' = addMultipartHandling pbak multipartOpts subserver
```

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -16,7 +16,7 @@ packages: servant/
   doc/cookbook/file-upload
   doc/cookbook/generic
   doc/cookbook/https
-  doc/cookbook/jwt-and-basic-auth
+  -- doc/cookbook/jwt-and-basic-auth
   doc/cookbook/multiple-contexts
   doc/cookbook/pagination
   doc/cookbook/structuring-apis
--- a/cabal.project
+++ b/cabal.project
@@ -32,3 +32,5 @@ constraints:
 allow-newer:
   servant-pagination:servant,
   servant-pagination:servant-server
+
+packages: servant-multipart
--- a/cabal.project
+++ b/cabal.project
@@ -6,6 +6,8 @@ packages: servant/
   servant-server/
   doc/tutorial/
 
+  servant-server-superrecord/
+
   -- doc/cookbook/*/*.cabal
 
   doc/cookbook/basic-auth
--- a/cabal.project
+++ b/cabal.project
@@ -15,6 +15,7 @@ packages: servant/
   doc/cookbook/generic
   doc/cookbook/https
   doc/cookbook/jwt-and-basic-auth
+  doc/cookbook/multiple-contexts
   doc/cookbook/pagination
   doc/cookbook/structuring-apis
   doc/cookbook/using-custom-monad
--- a/doc/cookbook/basic-auth/BasicAuth.lhs
+++ b/doc/cookbook/basic-auth/BasicAuth.lhs
@@ -132,7 +132,7 @@ newtype BasicAuthCheck usr = BasicAuthCheck
   deriving (Generic, Typeable, Functor)
 ```
 
-This is all great, but how is our `BasicAuth` combinator supposed to know
+**FIXME** This is all great, but how is our `BasicAuth` combinator supposed to know
 that it should use our `checkBasicAuth` from above? The answer is that it
 simply expects to find a `BasicAuthCheck` value for the right user type in
 the `Context` with which we serve the application, where `Context` is just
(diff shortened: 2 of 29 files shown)


jkarni:
Even if users use `vinyl`, don't they still end up with more boilerplate? In particular, don't they need to declare their own instances for e.g. `HasMultiPartOptions`?

I myself don't much mind the `Config` datatype, but also don't object to us using some existing extensible records library instead, if that's the main grievance.

phadej:
I added (a very WIP) cookbook recipe on how to use this version of contexts, also how to make boilerplate minimal using `generics-lens`. Noteworthy change is 

```diff
-data WithNamedContext (name :: Symbol) (subContext :: *) subApi
+data WithNamedContext (name :: Symbol) (subApi :: *)
```

with

```haskell
-- | FunDep!
class GetNamedContext context name sub | context name -> sub where
```

which allows as to omit subcontext in API definitions. It doesn't below there, only the name is relevant for API definition.

phadej:
```haskell
data EmptyContext =  EmptyContext
```

will make error messages even better (and googleable!)

phadej:
TODO:  rename `GetNamedContext` into `HasNamedSubcontext`

phadej:
```haskell
data Context contextTypes where
    EmptyContext :: Context '[]
    (:.) :: x -> Context xs -> Context (x ': xs)
infixr 5 :.
```

```haskell
data NP :: (k -> *) -> [k] -> * where
  Nil  :: NP f '[]
  (:*) :: f x -> NP f xs -> NP f (x ': xs)

infixr 5 :*
```

```haskell
data Rec :: (u -> *) -> [u] -> * where
  RNil :: Rec f '[]
  (:&) :: !(f r) -> !(Rec f rs) -> Rec f (r ': rs)

infixr 7 :&
```

I'd like that `servant` won't re-invent `HList`s. (We need them for `Headers`, but there we should/could re-use e.g. `generics-sop`s `NP`). And others could write the `HasBasicAuthCheck` etc. classes to work directly on `generics-sop` or `vinyl` data structures (and actually on `Context`, but I'd like to remove it).

jkarni:
I still think that having a concrete type is nicer for end-users, since they don't have to write any extra instances and can just `cons` whatever data libs (like `servant-auth`) need onto their context, and for library authors, as they don't need to declare new classes. That said, maybe the actual datatype can be improved. In the back of my mind I always assumed that as soon as support for pre-GHC-8.0 was gone, we could move to a records library (`rawr`, or `labels`, or `superrecord`, or whatever).

phadej:
Added `servant-server-superrecord`. Would it solve @jkarni's concerns (seems `superrecord` have to gain feature for per-type access, but it should be doable).

phadej:
- `servant-0.14` approach rules out not-concrete `Context` approaches
- this approach is general enough to support both: concrete data type + instances, or `superrecord` or ...

PR #1010, opened 2018-07-13
Remove version history from README.md

alpmestan:
It is out of date and we already have enough to do during releases to be updating this table, I think.


phadej:
Yeah. It would been good idea to keep up to date, but it's benefit is debatable indeed.

domenkozar:
How about this: https://github.com/haskell-servant/servant-auth#servant-auth

phadej:
The idea of the table was to show which versions were of the same "epoch", a bit like https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory

Example use would been: "I want to support `servant-0.11`, how low bound should I put on `servant-foreign`.

PR #1029, opened 2018-08-25
Change HasDocs instances for Summary and Description

### LINKED ISSUE #1026: servant-docs: Modify the Description combinator instance of HasDocs
The value of the `Description` combinator can be fairly long unlike that of the similar `Summary` combinator.
 
However, the `HasDocs` instance of the `Description` combinator does the same thing as the `HasDocs` instance of the `Summary` combinator - they both put the entire value of the symbol in the `_noteTitle` of a `DocNote`:

https://github.com/haskell-servant/servant/blob/46973b7ccfe090d7f0097968b9a9e2fc8446d3d2/servant-docs/src/Servant/Docs/Internal.hs#L942

https://github.com/haskell-servant/servant/blob/46973b7ccfe090d7f0097968b9a9e2fc8446d3d2/servant-docs/src/Servant/Docs/Internal.hs#L952

This means that: 

  1. if a long API description is provided, it ends up as a long `_noteTitle`.  If it includes newlines, this does not look so bad when output with the `markdown` function, since the first line of the description becomes the markdown title and the rest of it effectively becomes a body.  This does _not_ work for docs rendered by servant-pandoc.

  2. the generated docs may look odd when both a summary and description are provided.  They are two `DocNotes`  that only have `_noteTitles` mixed in with all the `DocNotes` with no guarantee that summary immediately precedes the description.

Options for fixing this:

a. Simplest:
  * split the Description value on newlines
  * its first line is `_noteTitle`, and its remaining lines are `_noteBody` 
  * solves 1, but not 2

b. And an new instance (an overlapping one?) for when a `Description` immediately follows a `Summary`
   * add a single `DocNote`; the summary provides the `_noteTitle`, the `_noteBody` is the description split on newlines
   * solves 2, only solves 1 when a `Summary` always precedes a `Description`

c. both a and b

adetokunbo:
- if a description follows a summary, the summary is used as a title and the
lines of description form a body.

- if a description occurs without a preceding summary, it's first line is used
  as a title, and the remainder are used as a body.

Fixes #1026

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -930,6 +930,17 @@ instance HasDocs Raw where
   docsFor _proxy (endpoint, action) _ =
     single endpoint action
 
+instance (KnownSymbol summ, KnownSymbol desc, HasDocs api)
+  => HasDocs (Summary summ :> Description desc :> api) where
+
+  docsFor Proxy (endpoint, action) =
+    docsFor subApiP (endpoint, action')
+
+    where subApiP = Proxy :: Proxy api
+          action' = over notes (|> note) action
+          note = DocNote title body
+          title = symbolVal (Proxy :: Proxy summ)
+          body = lines $ symbolVal (Proxy :: Proxy desc)
 
 instance (KnownSymbol desc, HasDocs api)
   => HasDocs (Description desc :> api) where
@@ -939,17 +950,20 @@ instance (KnownSymbol desc, HasDocs api)
 
     where subApiP = Proxy :: Proxy api
           action' = over notes (|> note) action
-          note = DocNote (symbolVal (Proxy :: Proxy desc)) []
+          description = symbolVal (Proxy :: Proxy desc)
+          note = mkNote $ lines $ description
+          mkNote (x:xs) = DocNote x xs
+          mkNote []     = DocNote description []
 
-instance (KnownSymbol desc, HasDocs api)
-  => HasDocs (Summary desc :> api) where
+instance (KnownSymbol summ, HasDocs api)
+  => HasDocs (Summary summ :> api) where
 
   docsFor Proxy (endpoint, action) =
     docsFor subApiP (endpoint, action')
 
     where subApiP = Proxy :: Proxy api
           action' = over notes (|> note) action
-          note = DocNote (symbolVal (Proxy :: Proxy desc)) []
+          note = DocNote (symbolVal (Proxy :: Proxy summ)) []
 
 -- TODO: We use 'AllMimeRender' here because we need to be able to show the
 -- example data. However, there's no reason to believe that the instances of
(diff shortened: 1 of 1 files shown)


adetokunbo:
@alpmestan just checking in - this is approved, so I assume a maintainer will merge it ?

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -931,6 +930,17 @@ instance HasDocs Raw where
   docsFor _proxy (endpoint, action) _ =
     single endpoint action
 
+instance (KnownSymbol summ, KnownSymbol desc, HasDocs api)
+  => HasDocs (Summary summ :> Description desc :> api) where
```
phadej: Write a test. I'm surprised if we don't get into problems with overlapping instances.

phadej: What about `Description desc :> Summary sum :> api`.

I think saner approach is to change `Description` and `Summary` instances so they can find a desc/sum note, and alter it.

phadej:
I added the golden docs for servant-docs, @adetokunbo could you rebase, and `--accept` the test if it fails. Then we'll see the difference, if there's any (ComprehensiveAPI doesn't seem to have both summary and description though, so tests should **not** break)

trskop:
I've stumbled upon this while trying to create a self describing API. In the end I had to build custom combinators for this purpose so I'm glad it's getting some attention. Thank you for that.

What I'm not sure I agree with is the solution. If I understand it correctly then this will create something like:
```Markdown
## GET /blob/:id

### Retrieve a blob with the specified ID

Description omitted for brevity. 
```
The problem I have with that structure is that `Summary "Retrieve a blob with the specified ID"` is turned into a header. This structure doesn't actually allow referencing the section very easily. Especially in presence of multiple notes. One of the hacks that worked nicely for my purposes, before switching to custom combinators, was this:
```Haskell
-- | Retrieve a blob with the specified ID
-- > GET {+api}/blob/{id}
type GetBlob prefix = prefix
    :> Summary "Summary\n\nRetrieve a blob with the specified ID."
    :> Description "Description\n\nDescription omitted for brevity."
    :> "blob"
    :> Capture "id" (Id "blob")
    :> Get '[OctetStream] Blob
```
Which was rendered into:
```Markdown
## GET /blob/:id

### Summary

Retrieve a blob with the specified ID.

### Description

Description omitted for brevity. 
```
This worked nicely as it was easy to integrate into wider documentation. The above structure has the advantage that `HasDocs` instances for `Summary` and `Description` are decoupled and there's no need for either complex logic to find the note or overlapping instances. That said, the best format I've found was actually this one:
```Markdown
## GET /blob/:id

Retrieve a blob with the specified ID.

### Description

Description omitted for brevity. 
```
That required a different rendering function in addition to custom combinators.

tchoutri:
Hi @adetokunbo! Could you please address the comments by phadej? https://github.com/haskell-servant/servant/pull/1029/files#r229139130
If you wish to continue with this PR, it can be featured in the next major release of Servant!

PR #1032, opened 2018-09-05
Add bifunctors instances for Servant.API.Alt

Taneb:
These mirror the corresponding instances for (,)

### DIFF
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -84,6 +84,7 @@ library
   -- Here can be exceptions if we really need features from the newer versions.
   build-depends:
       base-compat            >= 0.10.4   && < 0.11
+    , bifunctors             >= 5.4      && < 5.6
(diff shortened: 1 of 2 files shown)


servant/servant.cabal
```diff
@@ -84,6 +84,7 @@ library
   -- Here can be exceptions if we really need features from the newer versions.
   build-depends:
       base-compat            >= 0.10.4   && < 0.11
+    , bifunctors             >= 5.4      && < 5.6
```
phadej:
```
>= 5.5.3
```

or is there a reason to have lose lower bound? (Stackage nightly has 5.5.3...)

Taneb: Ah! I misread the comment, I understood it to mean make sure we include the latest Stackage nightly but it's OK to be looser. I can change that!

phadej: My bad. I'll improve the wording.

PR #1035, opened 2018-09-17
Implement HasForeign instance

afcady:
This is three patches, to `servant-foreign`, `servant-multipart`, and `servant-js`.

https://github.com/haskell-servant/servant-multipart/compare/master...afcady:multipart-foreign?expand=1

https://github.com/haskell-servant/servant/compare/master...afcady:multipart-foreign?expand=1

https://github.com/haskell-servant/servant-js/compare/master...afcady:multipart-foreign?expand=1

The patch to `servant-foreign` adds a Boolean field to the `Req` type specifying whether the request body is JSON (as is currently assumed).  The patch to `servant-multipart` implements a HasForeign instance that sets this field to `False`.  The patch to `servant-js` changes the output of `vanillaJS` when the flag is false, so that `JSON.stringify()` does not get called on the body and the content-type is not set to JSON.

This address https://github.com/haskell-servant/servant-multipart/issues/5

### DIFF
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -1,5 +1,5 @@
 name:                servant-foreign
-version:             0.11.1
+version:             0.11.2
 x-revision:          3
 synopsis:            Helpers for generating clients for servant APIs in any programming language
 description:
--- a/servant-foreign/servant-foreign.cabal
+++ b/servant-foreign/servant-foreign.cabal
@@ -1,5 +1,5 @@
 name:                servant-foreign
-version:             0.11.2
+version:             0.11.1
 x-revision:          3
 synopsis:            Helpers for generating clients for servant APIs in any programming language
 description:
--- a/servant-foreign/src/Servant/Foreign/Internal.hs
+++ b/servant-foreign/src/Servant/Foreign/Internal.hs
@@ -123,21 +123,24 @@ defUrl = Url [] []
 
 makeLenses ''Url
 
+data ReqBodyContentType = ReqBodyJSON | ReqBodyMultipart
+  deriving (Data, Eq, Show, Read)
+
 data Req f = Req
-  { _reqUrl        :: Url f
-  , _reqMethod     :: HTTP.Method
-  , _reqHeaders    :: [HeaderArg f]
-  , _reqBody       :: Maybe f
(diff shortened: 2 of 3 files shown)


servant-foreign/servant-foreign.cabal
```diff
@@ -1,5 +1,5 @@
 name:                servant-foreign
-version:             0.11.1
+version:             0.11.2
```
phadej: @alpmestan should we copy https://github.com/phadej/github/blob/master/CONTRIBUTING.md#miscellaneous to our `CONTRIBUTING.md`

In this case next version MUST be `0.12`, as `Req` constructror is exported from non-`Internal` module, so this is a *breaking change*. Additionally, if there happen to be another breaking change, we don't want people "blindly" bump it to `0.13` and so on.

servant-foreign/src/Servant/Foreign/Internal.hs
```diff
@@ -130,13 +130,14 @@ data Req f = Req
   , _reqBody       :: Maybe f
   , _reqReturnType :: Maybe f
   , _reqFuncName   :: FunctionName
+  , _reqBodyIsJSON :: Bool
```
phadej: I'd rather have

```haskell
reqBodyContentType :: ReqBodyContentType

data ReqBodyContentType = RBCTJSON | RBCTRaw
```

(better names are welcome). I'd like `servant-foreign` to support other "envelope" formats (cbor, edn, ... - those aren't easy to use from JavaScript, but there are other client languages!)

servant-foreign/servant-foreign.cabal
alpmestan: Yeah, I guess we want to tell people not to update the version and changelogs.

afcady: I assumed the Req constructor was not exported.  It shouldn't be...  https://www.yesodweb.com/blog/2011/10/settings-types .  Anyway it's moot as I just reverted all changes to versions and dependencies.

servant-foreign/src/Servant/Foreign/Internal.hs
afcady: I did that too.

afcady:
Look OK then?

PR #1046, opened 2018-10-08
Fosskers/ghc 86 bounds

phadej:


phadej:
argh. travis is slow for some reason :/

fosskers:
Oh, so everything is actually passing there. It was just a timeout for `8.0.2`.

fosskers:
Setting `sudo` to `true` in the Travis config may also speed up the builds.

PR #1051, opened 2018-10-14
GHC-8.6 for 0.14

phadej:


alpmestan:
Do we really test with 7.8/7.10?

phadej:
For 0.14 yes

alpmestan:
Oh, right. Nevermind my comment, as usual, then :-)

PR #1052, opened 2018-10-14
ResponseHeaders: export GetHeaders'

domenkozar:
Required for https://github.com/plow-technologies/servant-streaming/pull/14/files

PS: there are a few other typeclasses in that module that would be useful to be exported.


domenkozar:
@phadej objections?

PR #1053, opened 2018-10-15
Added type synonym for error in mimeUnrender and ResultStream

jhmcstanton:
Just used one of the type names suggested in the issue, is that name still OK?


tchoutri:
Hi @jhmcstanton! Thank you very much for this PR. In concertation with @gdeest and @akhesaCaro, we have come to the conclusion that this type ought to be a newtype, so that we can better restrict the scope of possible values. As an aside, do you think that the entirety of Shakespear's work in Mandarin could be a valid `BodyDecodingError`? :)

tchoutri (changes requested):
Needs to be newtyped

PR #991, opened 2018-06-23
Streaming refactoring

phadej:
- [x] make `StreamReqBody` combinator
- [x] write proper example (now just cassava copy)
- [x] CPSise SourceT
- [x] Add instances for conduit with resourcet, safe-pipes etc.
- [x] ~potoki~
- [x] ~streaming~
- [x] withFile
- [x] cookbook example for an example how to work without conduit/pipes/machines/... in simple 
examples
- [x] rename things, like `b` to `chunk`.
- [x] update tutorial
- [x] update docs (API.Stream)
- [x] write stub changelog (there're a lot of changes), moved to #1066 

I'll leave implementing docs/foreign for now (i.e. leave dummy instances), as that we'll need to be redone for #841 / #969

PR #1066, opened 2018-11-01
Separate streaming client

phadej:
Follow up to #991

- [ ] write changelog

### DIFF
--- a/servant/src/Servant/Test/ComprehensiveAPI.hs
+++ b/servant/src/Servant/Test/ComprehensiveAPI.hs
@@ -15,38 +15,64 @@ import           Servant.Types.SourceT
 type GET = Get '[JSON] NoContent
 
 type ComprehensiveAPI =
-  ComprehensiveAPIWithoutRaw :<|>
-  "raw" :> Raw
+    ComprehensiveAPIWithoutStreamingOrRaw'
+    (EmptyEndpoint :<|> StreamingEndpoint :<|> RawEndpoint)
+
+type RawEndpoint =
+    "raw" :> Raw
+
+type StreamingEndpoint =
+    "streaming" :> StreamBody NetstringFraming JSON (SourceT IO Int) :> Stream 'GET 200 NetstringFraming JSON (SourceT IO Int)
+
+type EmptyEndpoint =
+    "empty-api" :> EmptyAPI
 
 comprehensiveAPI :: Proxy ComprehensiveAPI
 comprehensiveAPI = Proxy
 
 type ComprehensiveAPIWithoutRaw =
-  GET :<|>
-  "get-int"          :> Get '[JSON] Int :<|>
-  "capture"          :> Capture' '[Description "example description"] "foo" Int :> GET :<|>
-  "header"           :> Header "foo" Int :> GET :<|>
-  "header-lenient"   :> Header' '[Required, Lenient] "bar" Int :> GET :<|>
-  "http-version"     :> HttpVersion :> GET :<|>
-  "is-secure"        :> IsSecure :> GET :<|>
-  "param"            :> QueryParam "foo" Int :> GET :<|>
-  "param-lenient"    :>  QueryParam' '[Required, Lenient] "bar" Int :> GET :<|>
-  "params"           :> QueryParams "foo" Int :> GET :<|>
-  "flag"             :> QueryFlag "foo" :> GET :<|>
-  "remote-host"      :> RemoteHost :> GET :<|>
-  "req-body"         :> ReqBody '[JSON] Int :> GET :<|>
-  "req-body-lenient" :> ReqBody' '[Lenient] '[JSON] Int :> GET :<|>
-  "res-headers"      :> Get '[JSON] (Headers '[Header "foo" Int] NoContent) :<|>
-  "foo"              :> GET :<|>
-  "vault"            :> Vault :> GET :<|>
-  "post-no-content"  :> Verb 'POST 204 '[JSON] NoContent :<|>
-  "post-int"         :> Verb 'POST 204 '[JSON] Int :<|>
-  "streaming"        :> StreamBody NetstringFraming JSON (SourceT IO Int) :> Stream 'GET 200 NetstringFraming JSON (SourceT IO Int) :<|>
-  "named-context"    :> WithNamedContext "foo" '[] GET :<|>
-  "capture-all"      :>  CaptureAll "foo" Int :> GET :<|>
-  "summary"          :> Summary "foo" :> GET :<|>
-  "description"      :> Description "foo" :> GET :<|>
-  "empty-api"        :> EmptyAPI
+    ComprehensiveAPIWithoutStreamingOrRaw'
+    (EmptyEndpoint :<|> StreamingEndpoint)
 
 comprehensiveAPIWithoutRaw :: Proxy ComprehensiveAPIWithoutRaw
 comprehensiveAPIWithoutRaw = Proxy
+
+type ComprehensiveAPIWithoutStreaming =
+    ComprehensiveAPIWithoutStreamingOrRaw'
+    (EmptyEndpoint :<|> RawEndpoint)
+
+comprehensiveAPIWithoutStreaming :: Proxy ComprehensiveAPIWithoutStreaming
+comprehensiveAPIWithoutStreaming = Proxy
+
+-- | @:: API -> API@, so we have linear structure of the API.
+type ComprehensiveAPIWithoutStreamingOrRaw' endpoint =
+    GET
+    :<|> "get-int"          :> Get '[JSON] Int
+    :<|> "capture"          :> Capture' '[Description "example description"] "foo" Int :> GET
+    :<|> "header"           :> Header "foo" Int :> GET
+    :<|> "header-lenient"   :> Header' '[Required, Lenient] "bar" Int :> GET
+    :<|> "http-version"     :> HttpVersion :> GET
+    :<|> "is-secure"        :> IsSecure :> GET
+    :<|> "param"            :> QueryParam "foo" Int :> GET
+    :<|> "param-lenient"    :> QueryParam' '[Required, Lenient] "bar" Int :> GET
+    :<|> "params"           :> QueryParams "foo" Int :> GET
+    :<|> "flag"             :> QueryFlag "foo" :> GET
+    :<|> "remote-host"      :> RemoteHost :> GET
+    :<|> "req-body"         :> ReqBody '[JSON] Int :> GET
+    :<|> "req-body-lenient" :> ReqBody' '[Lenient] '[JSON] Int :> GET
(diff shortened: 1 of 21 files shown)


alpmestan:
@lucasdicioccio See the last commit here, this is quite relevant to your http2 client. We're leaning towards putting the streaming client "runner" in a separate class, where the streaming stuffs is exposed under the `withXXX` pattern, CPS-style.

phadej:
Looking at `servant-http2-client` code, the current [`performStreamingRequest`](http://hackage.haskell.org/package/servant-http2-client-0.1.0.2/docs/src/Network.HTTP2.Client.Servant.html#performStreamingRequest) function already uses `withHttp2Stream`, so adaptation should be straight-forward.

phadej:
```diff
From cd18ac60233a93f76908d121c7ffbb36e35d9489 Mon Sep 17 00:00:00 2001
From: Oleg Grenrus <oleg.grenrus@iki.fi>
Date: Thu, 1 Nov 2018 20:20:18 +0200
Subject: [PATCH] Adapt to stream changes

---
 src/Network/HTTP2/Client/Servant.hs | 27 ++++++++++++++-------------
 test/Spec.hs                        | 16 ++++++----------
 2 files changed, 20 insertions(+), 23 deletions(-)

diff --git a/src/Network/HTTP2/Client/Servant.hs b/src/Network/HTTP2/Client/Servant.hs
index f3d198c..4dd8932 100644
--- a/src/Network/HTTP2/Client/Servant.hs
+++ b/src/Network/HTTP2/Client/Servant.hs
@@ -56,11 +56,13 @@ runH2ClientM cm env = runExceptT $ flip runReaderT env $ unH2ClientM cm
 instance RunClient H2ClientM where
   runRequest :: Request -> H2ClientM Response
   runRequest = performRequest
-  streamingRequest :: Request -> H2ClientM StreamingResponse
-  streamingRequest = performStreamingRequest
   throwServantError :: ServantError -> H2ClientM a
   throwServantError = throwError
 
+instance RunStreamingClient H2ClientM where
+  withStreamingRequest :: Request -> (StreamingResponse -> IO r) -> H2ClientM r
+  withStreamingRequest = performStreamingRequest
+
 data H2ClientEnv = H2ClientEnv ByteString Http2Client
 
 type ByteSegments = (IO ByteString -> IO ()) -> IO ()
@@ -208,32 +210,31 @@ replenishFlowControls icfc isfc len = do
     pure ()
 
 -- | Implementation of requests with streaming replies.
-performStreamingRequest :: Request -> H2ClientM StreamingResponse
-performStreamingRequest req = do
+performStreamingRequest :: Request -> (StreamingResponse -> IO r) -> H2ClientM r
+performStreamingRequest req kont = do
     H2ClientEnv authority http2client <- ask
     let icfc = _incomingFlowControl http2client
     let ocfc = _outgoingFlowControl http2client
     (headersPairs, mBodyIO) <- liftIO $ makeRequest authority req
     let headersFlags = if isJust mBodyIO then id else setEndStream
     ret <- liftIO $ withHttp2Stream http2client $ \stream ->
-        let initStream =
-                headers stream headersPairs headersFlags
+        let initStream = headers stream headersPairs headersFlags
             handler isfc osfc = do
                 -- Send the request
                 flip traverse_ mBodyIO (\bodyIO -> do
                     sendNonEmptySegments http2client stream ocfc osfc bodyIO
                     sendData http2client stream setEndStream "")
+
                 -- Waits for headers and returns the response object to the
                 -- caller.
-                pure $ StreamingResponse (\handleGenResponse -> do
-                    ev <- _waitEvent stream
-                    case ev of
-                        (StreamHeadersEvent fh hdrs) -> handleHeaders stream icfc isfc fh hdrs handleGenResponse
-                        _ -> throwIO $ Servant.Client.Core.ConnectionError $ "unwanted event received in data stream" <> Text.pack (show ev)
-                        )
+                ev <- _waitEvent stream
+                case ev of
+                    (StreamHeadersEvent fh hdrs) -> handleHeaders stream icfc isfc fh hdrs kont
+                    _ -> throwIO $ Servant.Client.Core.ConnectionError $ "unwanted event received in data stream" <> Text.pack (show ev)
+
         in (StreamDefinition initStream handler)
     case ret of
-        Right streamingResp -> pure streamingResp
+        Right r -> return r
         Left (TooMuchConcurrency _) ->
             throwError $ Servant.Client.Core.ConnectionError "too many concurrent streams"
   where
diff --git a/test/Spec.hs b/test/Spec.hs
index fd7419c..01f52db 100644
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -19,6 +19,7 @@ import           Network.HTTP2.Client.Servant
 import qualified Network.TLS as TLS
 import qualified Network.TLS.Extra.Cipher as TLS
 import           Servant.API
+import           Servant.Types.SourceT (foreach)
 import           Servant.Client.Core
 
 data RawText
@@ -33,14 +34,14 @@ instance MimeUnrender OctetStream Text where
 
 type Http2Golang =
        "reqinfo" :> Get '[RawText] Text
-  :<|> "goroutines" :> StreamGet NewlineFraming RawText (ResultStream Text)
+  :<|> "goroutines" :> StreamGet NewlineFraming RawText (SourceIO Text)
   :<|> "ECHO" :> ReqBody '[RawText] Text :> Put '[OctetStream] Text
 
 myApi :: Proxy Http2Golang
 myApi = Proxy
 
 getExample :: H2ClientM Text
-getGoroutines :: H2ClientM (ResultStream Text)
+getGoroutines :: H2ClientM (SourceIO Text)
 capitalize :: Text -> H2ClientM Text
 getExample :<|> getGoroutines :<|> capitalize = h2client myApi
 
@@ -55,15 +56,10 @@ main = do
     runH2ClientM getGoroutines env >>= go
     _close client
   where
-    go :: Either ServantError (ResultStream Text) -> IO ()
+    go :: Either ServantError (SourceIO Text) -> IO ()
     go (Left err) = print err
-    go (Right (ResultStream k)) = k $ \getResult ->
-       let loop = do
-            r <- getResult
-            case r of
-                Nothing -> return ()
-                Just x -> print x >> loop
-       in loop
+    go (Right sourceIO) = foreach fail print sourceIO
+
     tlsParams = TLS.ClientParams {
           TLS.clientWantSessionResume    = Nothing
         , TLS.clientUseMaxFragmentLength = Nothing
-- 
2.17.1
```

lucasdicioccio:
Thanks for the heads-up Alp. And thanks @phadej for the patch.

The `withHttp2Stream` actually doesn't steal the control (a first design did that but now it only steals the thread control for a short while to enforce an HTTP2 invariant on new-stream-IDs). Further, this function doesn't automatically close the stream and a power user can escape everything they want from the 2nd method. Hence, CSP-style should work. I'll need a few more iterations to find a better design (I was thinking adding a module with type-indexed state-machines).

I'll have more time during next week.

phadej:
@lucasdicioccio I realised that (non stealing control). The `with` name is confusing. Anyway, my patch compiles and tests "pass"; but I have no idea if things works as they should. I don't know HTTP2 at all.

phadej:
@lucasdicioccio you should look at this branch `StreamingBody`. *Maybe* you can use it somehow to make duplex communication. That would be very cool!

phadej:
I also run benchmarks (`streaming-benchmark.sh`) and cannot spot the difference between this and before. So I'm heavily inclined to merge this for 0.15

domenkozar:
Really good work (as always) :) 

I have a few questions, although none of those should block the merge.

a) How does one mix&match Servant.Client.Generic with streaming, if streaming endpoints require a different `ClientM`?

b) Does the streaming client handle HTTP redirects?

c) Putting the monad into the API sometimes has devastating results, for example `api` package needs to depend on amazonka for `AWST`. Does the monad that server operate on really depend in the api, or maybe it should've been a type family in the api and attached via `Context` for the servant-server?

phadej:
@domenkozar 

These is something we'd need to add to changelog / tutorial:

a) `Servant.Client.Streaming` works with non-streaming endpoints too. You need to work with all **endpoints** using `withClientM` though.

b) It works as underlying `http-client`s `withResponse` works. So probably yes.

c) I don't understand the question.

domenkozar:
@phadej for (c) see https://github.com/plow-technologies/servant-streaming/issues/11

In short, the monad in which servant-server operates for streaming doesn't really belong in the API description as it's the implementation detail of the server.

phadej:
there are no `StreamBodyMonad` or anything like that.

We have `StreamingBody` now: See https://haskell-servant.readthedocs.io/en/latest/cookbook/basic-streaming/Streaming.html (I don't understand how this is related to `servant-client` refactor?)

phadej:
@alpmestan could you re-review this. I rebased the branch, and there are few small changes because of recent changes in master (I think in good direction).

servant/src/Servant/Test/ComprehensiveAPI.hs
```diff
@@ -15,38 +15,64 @@ import           Servant.Types.SourceT
...
+    :<|> "res-headers"      :> Get '[JSON] (Headers '[Header "foo" Int] NoContent)
+    :<|> "foo"              :> GET
+    :<|> "vault"            :> Vault :> GET
+    :<|> "post-no-content"  :> Verb 'POST 204 '[JSON] NoContent
+    :<|> "post-int"         :> Verb 'POST 204 '[JSON] Int
+    :<|> "named-context"    :> WithNamedContext "foo" '[] GET
+    :<|> "capture-all"      :> CaptureAll "foo" Int :> GET
+    :<|> "summary"          :> Summary "foo" :> GET
+    :<|> "description"      :> Description "foo" :> GET
+    :<|> "alternative"      :> ("left" :> GET :<|> "right" :> GET)
```
phadej: this case is added to make a (small) tree, as otherwise the structure of `ComprehensiveAPI` is linear.

domenkozar:
I'd like to bring up another UX issue, but maybe I don't have the full picture.

Suppose one is developing a big servant application with many routes. Once streaming is added, it changes client calls from `runClientM` to `withClientM`. Previously mixing the two was possible. I think this makes sense if eventually we'd rather get rid of `runClientM` (and rewrite the documentation).

domenkozar:
I'd like to rephrase as my message was too implicit.

Currently default client has "runClientM" and streaming client has "withClientM". We have to think about the story of migration, which is to pick what we recommend to the users of servant to do:

a) if you want to use streaming use withClientM, otherwise runClientM, switch back and forth depending what you need

b) withClientM is the new CPS way, if you happen to use streaming you'll want to migrate right away otherwise take time but `runClientM` will eventually be the deprecated old way

The main question is, do we really still need `runClientM` and should we add `withClientM` also the default client?

phadej:
- `runClientM` is way simpler for tasks where it's enough (i.e. no `Stream`), so I'm :-1: for removing it.
- one can write `withClientM` for `Servant.Client` but I don't want encourage use of it. If one really needs it , it's easy to write inline.

You can use **both**.

See and comment in #1072

Issue #805, opened 2017-09-13
Deprecate Auth?


jkarni:
The work we put into authentication was really useful in getting the internals right; and exposing an API made sense while nothing else existed. But now `servant-auth` exists and seems to me like the better solution. Can we deprecate the auth stuff in this repo, and start pointing people to `servant-auth` instead?

alpmestan:
I'm a little bit bothered by the fact that authentication is a key topic, so much that it's covered in the tutorial, but you're suggesting to move it _all_ to non-core packages. I _want_ auth to be documented in the tutorial, whichever package it ends up being implemented in. servant-auth's README and the tutorial hold different standards wrt depth of explanation, obviously, so I do not want for the go to auth solution to be underdocumented. It'd have to be covered in the tutorial, in reasonable depth.

So yeah, it might require a little bit of work but nothing too scary, so yes from me pending the changes that I suggest.

alpmestan:
There's also the issue of maintenance. Would the auth packages move under the haskell-servant org? If not, what kind of maintenance/development/collaboration scheme would be in place?

jkarni:
I agree with the tutorial needing to be updated; I'd be happy to do it.

I can check whether we can move 'servant-auth' to the 'haskell-servant'
organization.

alpmestan:
The motivation for my question is that there are going to be changes that break servant-auth, and since auth is too important to leave broken for a bit, we need to make sure we'll be able to get the necesary changes in and probably have releases happen together.

3noch:
> The motivation for my question is that there are going to be changes that break servant-auth, and since auth is too important to leave broken for a bit

I agree. However, this cuts both ways. Because it's so important you may want to make it one of the official servant libraries. But also, because its so important, if a new release of servant does not work with servant-auth, you can be very sure that people will get if fixed quickly. Until they do, they'll have to use an older version of servant. That scenario is pretty common in package management and doesn't scare me in the slightest.

alpmestan:
@3noch I'd be in favor of having them as core packages (but just not in the standard ones), but it's not my call =) You make good points. My main request is really to have a detailed and very, very approachable section in the tutorial, like the other sections of it. The rest is just about us trying to make sure both package sets will work well together, hand in hand.

3noch:
@alpmestan You're right about documentation though. That is crucial and it definitely needs to cover auth. I don't care where the package lives, myself. As long as it's easily discoverable by servant neophytes.

[...]

PR #1074, opened 2018-11-09
Deprecate Servant.API.Experimental.Auth and Servant.Auth.BasicAuth

domenkozar:
First phase of #805: recommend using
https://github.com/haskell-servant/servant-auth


domenkozar:
@phadej I was hoping we could get rid of authentication confusion in 0.15 and start the process of the servant-auth merger.

domenkozar:
Blocked by https://github.com/haskell-servant/servant-auth/issues/140 and https://summer.haskell.org/ideas.html#servant-auth-improvements

PR #1076, opened 2018-11-09
Add runClientM for streaming-client

phadej:
ping @domenkozar


servant-client/src/Servant/Client/Internal/HttpClient/Streaming.hs
```diff
@@ -120,6 +123,14 @@ withClientM cm env k =
     let Codensity f = runExceptT $ flip runReaderT env $ unClientM cm
     in f k
 
+-- | A 'runClientM' variant for streaming client.
+--
+-- /Note:/ that we 'force' the result, so likehood of connection leaking
+-- is smaller. Use with care.
```
domenkozar: We should probably document the motivation why this function exists and that for any streaming result `withClientM` should be used.

PR #1080, opened 2018-11-09
Json framing

phadej:
seems that `attoparsec` doesn't support marking, so one could reuse parser from `aeson` to fast-forward in `FramingUnrender` instance. But maybe it's better to do a simple scanner, which recognises enough, as JSON will be parsed properly anyway.

### DIFF
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -150,6 +150,7 @@ test-suite spec
     , base-compat
     , aeson
     , bytestring
+    , mtl
     , servant
     , string-conversions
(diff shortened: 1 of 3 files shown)


phadej:
I want to make `FramingUnrender JSONFraming`; it's not hard, but will take time, maybe for 0.15.1 or 0.16

alpmestan:
Let's remove the commented out stuffs about it then?

phadej:
Cleaning up old PRs.

Issue #819, opened 2017-09-25
RawM data type for using Raw endpoints with Enter


cdepillabout:
I was playing around with using [`Raw`](https://hackage.haskell.org/package/servant-0.11/docs/Servant-API-Raw.html#t:Raw) to create a file server with [`serveDirectoryWebApp`](https://hackage.haskell.org/package/servant-server-0.11/docs/Servant-Utils-StaticFiles.html#v:serveDirectoryWebApp).

However, I am not running in [`Handler`](https://hackage.haskell.org/package/servant-server-0.11/docs/Servant-Server.html#t:Handler), but instead working in my own monad using [`Enter`](https://hackage.haskell.org/package/servant-0.11/docs/Servant-Utils-Enter.html#t:Enter).  I need to get the directory path to use to serve files from my monad, so it is not possible for me to use the normal `Raw` with `serveDirectoryWebApp`.

I whipped up a `RawM` data type that gets turned into an `m Application` by the `ServerT` type family.  This let me write a more generic `serveDirectoryWebApp` function.

It basically looks like this:

```haskell
data RawM deriving Typeable

instance HasServer RawM context where
  type ServerT RawM m = m Application

  route
    :: forall env.
       Proxy RawM
    -> Context context
    -> Delayed env (Handler Application)
    -> Router' env (Request -> (RouteResult Response -> IO ResponseReceived) -> IO ResponseReceived)
  route Proxy _ rawApplication = RawRouter go
    where
      go
        :: env
        -> Request
        -> (RouteResult Response -> IO ResponseReceived)
        -> IO ResponseReceived
      go env request respond =
        runResourceT $ do
          routeRes <- runDelayed rawApplication env request
          liftIO $
            case routeRes of
              (Fail e) -> respond $ Fail e
              (FailFatal e) -> respond $ FailFatal e
              (Route handlerApp) -> do
                eitherApp <- runHandler handlerApp
                case eitherApp of
                  Left err -> respond . Route $ responseServantErr err
                  Right app -> app request (respond . Route)

serveDirectoryWebApp :: Applicative m => FilePath -> ServerT RawM m
serveDirectoryWebApp = pure . staticApp . defaultWebAppSettings . addTrailingPathSeparator

```

Here's a small example of using it:

```haskell
type Api = "serve-directory" :> RawM

serverRoot :: ServerT Api (ReaderT FilePath IO)
serverRoot = rawEndpoint

rawEndpoint :: ReaderT FilePath IO Application
[...]

Issue #785, opened 2017-07-24
Enter for Raw endpoints


mschristiansen:
Agreed with @phadej to open an issue after discussion on IRC.

I have regular Get, Put, Post endpoint, but also need to make use of a raw endpoint and I would like to access my `Reader` environment from the raw handler the same way that I can do it for regular servant handlers, but that is currently not possible. E.g. in the simplified example below from the `undefined` handler I would like to ask for any of the config information e.g. database connection

```
type Api = Raw

api :: ServerT Api MyHandler
api = Tagged $ handler
  where
    handler :: Application
    handler = undefined

server :: Config -> Server Api
server cfg = enter (runReaderTNat cfg :: MyHandler :~> Handler) api
```

phadej:
To expand: `enter` simply retags `Tagged MyHandler Application` to `Tagged Handler Application` in this case. We need something a bit more powerful than `enter`, as I outline in #765

jkarni:
Is this still relevant (with `hoistServer`)?

codedmart:
I don't think so.

afcady:
Well I'm having trouble figuring out how to do this with `hoistServer`.  

I have a `ReaderT Global` and I get this if I try to use `ask` in my `Raw` endpoint:

     • No instance for (MonadReader
                          Global (Tagged (ReaderT Global (ExceptT ServantErr IO))))
         arising from a use of ‘ask’

Specifically this is how I'm using `hoistServer`:

    webApp :: Global -> Application
    webApp global = serve api $ hoistServer api (Handler . flip runReaderT global) server

Not sure if I'm just missing something here, but anyway it's not as easy as just calling `ask` like any other endpoint.  You have to do something about that `Tagged` somehow.

I just ran into this so maybe I'll figure it out soon.  I don't think the bug should be closed without at least having a posted explanation of what to do about this, though.

alpmestan:
`Raw` endpoints are just wai `Application`s in disguise. In particular, you don't get access to your `ReaderT`'s `Global` value in those, you just put a wai `Application` there. However, more and more people are trying to write `Raw` endpoints that use specific operations from specific monad stacks, so maybe it's worth beginning to think about this.

alpmestan:
A new release of that package has shipped, with those fixes.

roberth:
What was the reasoning behind this use of `Tagged`?
[...]

PR #1081, opened 2018-11-11
Add note to custom monad documentation about what to do with Raw endpoints

xaviershay:
Took me a decent amount of googling to figure this out. I believe this is the currently recommended way? As a relatively new user, I was surprised this was not available in the core package.

cc @cdepillabout

Ref #785 and #819.

### DIFF
--- a/doc/cookbook/using-custom-monad/UsingCustomMonad.lhs
+++ b/doc/cookbook/using-custom-monad/UsingCustomMonad.lhs
@@ -115,3 +115,6 @@ Running cookbook-using-custom-monad...
 [Book "To Kill a Mockingbird",Book "Harry Potter and the Order of the Phoenix"]
 [Book "The Picture of Dorian Gray",Book "To Kill a Mockingbird",Book "Harry Potter and the Order of the Phoenix"]
 ```
+
+To use `Raw` endpoints, look at the
+[servant-rawm](http://hackage.haskell.org/package/servant-rawm) package.


cdepillabout:
> As a relatively new user, I was surprised this was not available in the core package.

As was I.

If any of the Servant devs want to pull the code from my `servant-rawm` package into `servant` proper (and then deprecating `servant-rawm`, I'd be interested in helping.

As far as I can tell, the `RawM` type from `servant-rawm` is strictly superior to `Raw`.

phadej:
What's the motivation to using `RawM`. What you have in your custom monads to pass to *raw* `Application`? Is there something missing in `servant` that you *have to* rely on `Raw`? Could you write a cookbook recipe explaining that?

`Raw` is a bit like `unsafePerformIO`, an escape hatch; I want to know where servant isn't good enough.

xaviershay:
> What's the motivation to using RawM. What you have in your custom monads to pass to raw Application?

I want to generate server-side events. My custom monad contains accessors to relevant state:

https://github.com/xaviershay/licross/commit/ac45272ba0e57c0ea5978470af3207bfd6cb3ae1
https://github.com/xaviershay/licross/commit/9b70441c616a5637de56f7d0890f32e41d73b0f3

(Obviously this isn't finished yet!)

cdepillabout:
@alpmestan That's basically what `RawM` does:

https://github.com/cdepillabout/servant-rawm/blob/b4c7d4b2d183a5f83199b0ea85eaffd5e6ad6c83/src/Servant/RawM/Internal/Server.hs#L51

(It also adds a phantom parameter you can use to attach `servant-docs` to.)

From my use of `servant-rawm`, this seems to work well.

Originally I posted about this in https://github.com/haskell-servant/servant/issues/819, but there didn't seem to be any interest, so I just kept it as a separate package.  I'd be happy to deprecate that package if the definition of `ServerT m Raw` changed to `m Application`.  Although it would be nice to have an additional phantom parameter to attach documentation to, in case you use multiple `Raw` in your API.

alpmestan:
Well, in an ideal world, we have less and less `Raw`s, turning many into new, dedicated combinators for well-specified tasks, like file serving.

cdepillabout:
> `RawM` is definitely useful and we should advertise it, so I'm fine merging this documentation patch.

In addition to this, if any of the current (or future?) servant maintainers want push access to the `servant-rawm` respository (and Hackage package), I'd be happy to do that.

PR #1098, opened 2018-12-17
Fix the tutorial app4 curl example

csamak:
Change the app4 curl example to match the API definition.

As defined, it does not return "hi" and instead always returns true.


alpmestan:
Thanks!

PR #1103, opened 2019-01-06
Set http failure code priority explicitly

tstat:
Currently, if all choices fail in a router, then highest http error code
is used as the response. This choice of error code can result in
unexpected behavior. As an example, consider the following:

```haskell
newtype FooId
  = FooId Int
  deriving newtype (ToJSON)

instance FromHttpApiData FooId where
  parseUrlPiece tid = do
    tid' :: Int <- parseUrlPiece tid
    if tid' < 0
      then Left "FooId cannot be negative"
      else Right $ FooId tid'

type API
  =    "foo"
    :> Capture "FooId" FooId
    :> Get '[JSON] Foo
```

A GET request to `/foo/-2` results in a 400 response as
expected. However, expanding the API type with a seemingly unrelated route

```haskell
type API
  =    "foo"
    :> Capture "FooId" FooId
    :> Get '[JSON] Foo
  :<|> "foo"
    :> ReqBody '[JSON] Foo
    :> Post '[JSON] FooId
```

causes the same GET request `/foo/-2` to result in a 404.

This pull request gives an explicit priority to http error codes in
order to avoid unexpected behavior such as the above.


phadej:
Seems fair. Could you add names to the status codes in worseHTTPCode, i.e

    toPriority 404 = 0 -- ^ not found

as I really don't remember other than 404, 403 and 400.

tstat:
Yes, good idea. I have added the names.

PR #1104, opened 2019-01-08
Update CookieJar with intermediate request/responses using Network.HTTP.Client.HistoriedResponse.

mdunnio:
Currently, any cookies that are set prior to a redirect won't be available in the `CookieJar` in `ClientEnv`. Using `HistoriedResponse` allows the updating of the `CookieJar` with all the requests/responses.

### DIFF
--- a/servant-client/src/Servant/Client/Internal/HttpClient.hs
+++ b/servant-client/src/Servant/Client/Internal/HttpClient.hs
@@ -184,8 +184,9 @@ performRequest req = do
     updateWithResponseCookies cj responses = do
         now <- getCurrentTime
         bss <- Client.brConsume $ Client.responseBody fRes
-        let fRes' = fRes { Client.responseBody = BSL.fromChunks bss }
-        mapM_ (atomically . updateCookieJar now) $ Client.hrRedirects responses <> [(fReq, fRes')]
+        let fRes'        = fRes { Client.responseBody = BSL.fromChunks bss }
+            allResponses = Client.hrRedirects responses <> [(fReq, fRes')]
+        atomically $ mapM_ (updateCookieJar now) allResponses
         return fRes'
       where
           updateCookieJar :: UTCTime -> (Client.Request, Client.Response BSL.ByteString) -> STM ()


alpmestan:
LGTM, but it'd be great to add a test that exercises the "with cookie jar" code path to make sure 1/ it works as expected and 2/ it doesn't break in the future. Do you feel like adding this?

michaelsdunn1:
I should be able to add a test case. Let me take a look.

michaelsdunn1:
Added a test case.

alpmestan:
@phadej Any objection to merging this now?

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -158,19 +162,37 @@ performRequest req = do
...
+
+    updateWithResponseCookies :: TVar Client.CookieJar -> Client.HistoriedResponse Client.BodyReader -> IO (Client.Response BSL.ByteString)
+    updateWithResponseCookies cj responses = do
+        now <- getCurrentTime
+        bss <- Client.brConsume $ Client.responseBody fRes
+        let fRes' = fRes { Client.responseBody = BSL.fromChunks bss }
+        mapM_ (updateCookieJar now) $ Client.hrRedirects responses <> [(fReq, fRes')]
+        return fRes'
+      where
+          updateCookieJar :: UTCTime -> (Client.Request, Client.Response BSL.ByteString) -> IO ()
```
phadej: make this `-> STM ()` so we can use only single `atomically`.

michaelsdunn1: Good call, fixed.

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -185,11 +185,11 @@ performRequest req = do
         now <- getCurrentTime
         bss <- Client.brConsume $ Client.responseBody fRes
         let fRes' = fRes { Client.responseBody = BSL.fromChunks bss }
-        mapM_ (updateCookieJar now) $ Client.hrRedirects responses <> [(fReq, fRes')]
+        mapM_ (atomically . updateCookieJar now) $ Client.hrRedirects responses <> [(fReq, fRes')]
```
phadej: `atomically` on the outside. We want single STM transaction, don't we?

michaelsdunn1: Yes, you're correct. We only want to do one STM transaction per chain of request/responses. Fixed.

Issue #1091, opened 2018-11-28
client streaming + http redirect result into runtime exception


domenkozar:
If an http endpoint returns a redirect (for example http -> https upgrade), streaming client errors out:

```
HasClient @StreamBody
CallStack (from HasCallStack):
  error, called at src/Servant/Client/Core/Internal/HasClient.hs:550:44 in servant-client-core-0.15-KKVc3oxuLh47HhicA3CWtB:Servant.Client.Core.Internal.HasClient                                                                                                                                                     
```

https://github.com/haskell-servant/servant/blob/master/servant-client-core/src/Servant/Client/Core/Internal/HasClient.hs#L550

phadej:
Looks like it errors out *always*. I forgot to write an implementation :/

phadej:
I won't be able to look at this issue this week. But if someone can make a patch, then I'll happily review it. (and make the release after merge).

domenkozar:
There's absolutely no hurry from my side.

domenkozar:
This is now blocking https://github.com/cachix/cachix/pull/163

phadej:
Streaming + redirect ought to use something like https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100

Otherwise I don't see how to stream body only once to final target

phadej:
I relabel this as feature, as the bug part is fixed now.

PR #1108, opened 2019-01-31
servant-client: add a test case for StreamBody

domenkozar:
Test case for https://github.com/haskell-servant/servant/issues/1091

```
Failures:                                                                                                        
                                                                                                                 
  src/Servant/Client/Core/Internal/HasClient.hs:550:44:                                                          
  1) Servant.Stream.Servant.Client.Streaming works with Servant.API.StreamBody
       uncaught exception: ErrorCall
       HasClient @StreamBody
         error, called at src/Servant/Client/Core/Internal/HasClient.hs:550:44 in servant-client-core-0.15-KsVeR55
NIJFLZyNMgNW7Dh:Servant.Client.Core.Internal.HasClient
```

cc @phadej

### DIFF
--- a/servant-client/test/Servant/StreamSpec.hs
+++ b/servant-client/test/Servant/StreamSpec.hs
@@ -35,8 +35,9 @@ import qualified Network.HTTP.Client           as C
 import           Prelude ()
 import           Prelude.Compat
 import           Servant.API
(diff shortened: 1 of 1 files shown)


phadej:
thanks

PR #470, opened 2016-04-15
Release

soenkehahn:
I'd like to release `0.7`.

cc: @haskell-servant/maintainers

PR #1109, opened 2019-01-31
client: Preserve request in FailureResponse

### LINKED ISSUE #978: Include Request in FailureResponse
When performing a request which fails because of a 4xx status code, it’s really useful to log the properties of the failed request, in order to be able to reproduce it.

Currently, `FailureResponse` only includes the `Response`. I think it would make sense to also include the `Request` in a `FailureResponse`, in order to be able to log the URL, and optional body, of the failed request.

Is there a way to achieve this without introducing a backwards-incompatible change that adds an argument to the `FailureResponse` constructor?

bgamari:
This was previously implemented in #470 but later unintentionally
reverted in #803. This isn't verbatim the design implemented earlier; we
preserve strictly less information but nevertheless this should be
better than the status quo.

Fixes #978.


servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -169,9 +169,19 @@ performRequest req = do
           status_code = statusCode status
           ourResponse = clientResponseToResponse response
       unless (status_code >= 200 && status_code < 300) $
-        throwError $ FailureResponse ourResponse
+        throwError $ mkFailureResponse request ourResponse
```
phadej: shouldn't we simply use `BaseUrl` and `Request`, then defining a new `UrlReq`?

phadej:
And now there's also a conflict :/

servant-client/src/Servant/Client/Internal/HttpClient.hs
bgamari: I suppose we could but this may end up keeping more data alive than necessary to report the error (e.g. potentially a large request body). Moreover, `ServantError` wants an `NFData` instance, which we cannot provide for `Request` without lying.

bgamari:
Rebased and refactored.

How does this look now, @phadej?

phadej:
I made my changes in #1114

schell:
Why is it that `mkFailureResponse` doesn't include the `RequestBody` in the error's `RequestF`? 
https://github.com/haskell-servant/servant/blob/726380964a742bc614352c9676a9698cb3e36a4f/servant-client/src/Servant/Client/Internal/HttpClient.hs#L202

I'm asking because in one of my projects we're maintaining a module `BetterClientM` that returns the full request body (`RequestF RequestBody ...`) on error to help with debugging. It seems to me that this should be the default behavior, and the body can be discarded down the line if needed.

PR #1110, opened 2019-02-02
Oleg testcase 1091

### LINKED ISSUE #1091: client streaming + http redirect result into runtime exception

phadej:
@domenkozar can you check this works for your use-case?

Fixes #1091

### DIFF
--- a/servant-client-core/src/Servant/Client/Core/Internal/HasClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/Internal/HasClient.hs
@@ -16,6 +16,8 @@ module Servant.Client.Core.Internal.HasClient where
 import           Prelude ()
 import           Prelude.Compat
 
+import           Control.Concurrent.MVar
+                 (modifyMVar, newMVar)
 import qualified Data.ByteString                        as BS
 import qualified Data.ByteString.Lazy                   as BL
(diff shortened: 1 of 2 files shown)


domenkozar:
Something is not right, my backend is returning 502 (although there's nothing in server logs). I'll look into it.

phadej:
@domenkozar thanks, let's make this right this time :)

domenkozar:
Figured out the problem, I was using `OctetStream` while I have my own mimetype. I think there's a bug in servant though, it should return 400 or something similar instead of 503.

Nevertheless, this now works. Could we get a bugfix release for 0.15?

phadej:
@domenkozar let's see. Looks more like that we'll have `servant-client(-core)-0.16`

Issue #714, opened 2017-03-15
servant-client's ConnectionError wraps itself in SomeException


3noch:
Using Servant 0.9.1.1, when I deconstruct a `ServantError` to see a `ConnectionError`, I've found that the `SomeException` inside `ConnectionError` contains, surprisingly, a `ServantError`!

```haskell
tryTlsOrDegrade :: (Scheme -> IO (Either ServantError a)) -> IO (Either ServantError a)
tryTlsOrDegrade req = do
  secureResponse <- req Https
  case secureResponse of
    Left (ConnectionError connE) -> case fromException connE of
      Just (ConnectionError connE2) -> print "Found it" >> pure secureResponse
      _ -> print connE >> pure secureResponse
    _ -> pure secureResponse
```

Both `connE` and `connE` are `SomeException`.

Perhaps I'm doing this wrong.

3noch:
This is what I ended up with:

```
tryTlsOrDegrade :: (Scheme -> IO (Either ServantError a)) -> IO (Either ServantError a)
tryTlsOrDegrade req = do
  secureResponse <- req Https
  case secureResponse of
    Left (ConnectionError connE) -> case fromException connE of
      Just (ConnectionError connE2) -> case fromException connE2 of
        Just (HttpExceptionRequest _ TlsNotSupported) -> req Http
        Nothing -> pure secureResponse
      Nothing -> pure secureResponse
    _ -> pure secureResponse
```

jkarni:
The `ConnectionError` now only contains `Text`. Less informative, but we wanted to not depend on `http-client`-specifics. 

If it feels like that's insufficient (i.e., you need to distinguish `TlsNotSupported` in a more reliable way than text parsing), please open a new issue!

3noch:
@jkarni Yes that's unfortunate. What should my new issue be exactly?

3noch:
Could the error type be parameterized somehow instead?

jkarni:
Agreed - my fault! Maybe we can get this fixed before ever releasing `master`. How about "More informative `ConnectionError`"?

jkarni:
If some backend is less informative than another, there's no actual problem - we can just have as many constructors as we think make sense, and some backends can just choose less informative ones (e.g. a default `OtherConnectionError`)

3noch:
@jkarni Done

ocharles:
[...]

PR #1113, opened 2019-02-05
keep structured exceptions in `ConnectionError`

### LINKED ISSUE #807: More informative ConnectionError in servant-client-core
`servant-client-core` now has an opaque `ConnectionError` type that just stores text. In order to make any sort of reliable decision based on client errors this type needs to preserve as much information as possible. Perhaps it can be parameterized so that underlying client backends can use their native error types. C.f. #714  for an example use-case where we try to examine TLS-not-supported error and downgrade to HTTP connections.

clementd-fretlink:
fixes #807
Previously, there were two levels of `SomeException` (see #714). A
test makes sure there is only one level of wrapping.

### DIFF
--- a/servant-client-core/src/Servant/Client/Core/Internal/Request.hs
+++ b/servant-client-core/src/Servant/Client/Core/Internal/Request.hs
@@ -17,6 +17,8 @@ import           Prelude.Compat
 
 import           Control.DeepSeq
                  (NFData (..))
+import           Control.Exception
+                 (SomeException)
 import           Control.Monad.Catch
                  (Exception)
 import qualified Data.ByteString         as BS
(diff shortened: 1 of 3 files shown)


servant-client-core/src/Servant/Client/Core/Internal/Request.hs
```diff
@@ -56,8 +58,15 @@ data ServantError =
...
-  deriving (Eq, Show, Generic, Typeable)
+  | ConnectionError SomeException
+  deriving (Show, Generic, Typeable)
+
+instance Eq ServantError where
+  (FailureResponse r)                  == (FailureResponse r')            = r == r'
+  (DecodeFailure t r)                  == (DecodeFailure t' r')           = t == t' && r == r'
+  (UnsupportedContentType mt r)        == (UnsupportedContentType mt' r') = mt == mt' && r == r'
+  (InvalidContentTypeHeader r)         == (InvalidContentTypeHeader r')   = r == r'
+  (ConnectionError _)                  == (ConnectionError _)             = True
```
phadej: You forgot non-equals cases:

```haskell
_ == _ = False
```

clementd-fretlink: Ha, I knew there was something missing. I rely too much on `-Wall` for these things.

clementd-fretlink: Fixed, thanks for the heads up

clementd-fretlink:
The build fails on GHC 8.0.2 and 8.2.2 with an error about `FailureResponse` taking two arguments instead of one. I'm not sure why.

phadej:
Merged in #1115

PR #1115, opened 2019-02-05
keep structured exceptions in `ConnectionError`

### LINKED ISSUE #807: More informative ConnectionError in servant-client-core

phadej:
fixes #807
Previously, there were two levels of `SomeException` (see #714). A
test makes sure there is only one level of wrapping.


divarvel:
Thanks for taking the time to help me with this :+1:

PR #1117, opened 2019-02-05
WIP: servant-http-streams

phadej:


servant-http-streams/test/Servant/StreamSpec.hs
```diff
@@ -0,0 +1,184 @@
...
+            testRunSourceIO res `shouldReturn` Right [alice, bob, alice]
+
+    it "works with Servant.API.StreamBody" $ \(_, baseUrl) -> do
+        withClient (getStreamBody (source input)) baseUrl $ \(Right res) ->
+            testRunSourceIO res `shouldReturn` Right output
+        where
+          input = ["foo", "", "bar"]
+          output = ["foo", "bar"]
+
+{-
```
alpmestan: just temporarily commented out? or is it a way to tag this as something to revisit later?

alpmestan: OK, then yes I'd suggest to remove this or alternatively explain why it's commented out right at the beginning of the comment, as you prefer. (depends on how likely we are to revisit the constant-memory testing)

phadej:
Rewrote history, to make it clearer what happened

PR #1121, opened 2019-02-11
Add optimization: False

phadej:


phadej:
Let's see if this noticeably speeds up CI.

PR #1122, opened 2019-02-11
Use ghcjs-dom instead of jsaddle-dom directly.

eskimor:
This way we avoid the dependency on jsaddle-dom on ghcjs and have
slightly better performance on ghcjs.

Seems to work fine on both ghc and ghcjs.

### DIFF
--- a/servant-client-jsaddle/servant-client-jsaddle.cabal
+++ b/servant-client-jsaddle/servant-client-jsaddle.cabal
@@ -62,7 +62,7 @@ library
     , http-media            >= 0.7.1.3  && < 0.8
     , http-types            >= 0.12.2   && < 0.13
     , jsaddle               >= 0.9.5.0  && < 0.10
-    , jsaddle-dom           >= 0.9.2.0  && < 0.10
+    , ghcjs-dom             >= 0.9.2.0  && < 0.10
     , monad-control         >= 1.0.2.3  && < 1.1
     , semigroupoids         >= 5.3.1    && < 5.4
     , string-conversions    >= 0.4.0.1  && < 0.5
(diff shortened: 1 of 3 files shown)


alpmestan:
@phadej Thoughts?

phadej:
I'm confused with `wants to merge 6 commits into haskell-servant:wip-servant-client-jsaddle from eskimor:use-ghcjs-dom`

That branch seems to be my rebase WIP branch to see what stuff is needed CI-wise, there's https://github.com/haskell-servant/servant/pull/1018 however which is not outdated.

There's now a bunch of branches and PRs, mess, that needs to be untangled. Maybe new clean PR to `master` would make stuff clearer (without TODOs).

FWIW, #1018 didn't motivate itself. Is it `servant-client-jsaddle` or is it just `servant-client-ghcjs-dom`, how it's different from the `servant-client-ghcjs` we have (I think I know the answer, but is it the right answer?). Why it should be in `servant` repository at all?

Don't make me think. I nitpick, because I need to understand it well enough, to be able to maintain it. Current state isn't good enough for me to commit into that.

eskimor:
Hi @phadej ! Oh sorry, I guess I picked the wrong branch for my improvements. This PR is meant as an improvement to #1018  -> fixed now.

The reason for #1018 is that servant-client-ghcjs does not work on ghc - therefore it is badly suited for e.g. reflex applications, which are usually built with both ghcjs and ghc - it means authors of web applications would need to maintain two versions of their servant client API. This is not needed with the jsaddle version as it works both on ghc and on ghcjs.

With my PR servant-client-ghcjs, we already have in servant will actually become mostly superfluous, as #1122 is pretty much equivalent to it on ghcjs, but with the added benefit that it also works on ghc.

eskimor:
Closing in favour of: https://github.com/roberth/servant/pull/2

PR #1127, opened 2019-02-18
README: release branches are now protected with a wildcard

domenkozar:


phadej:
This could been `[ci skip]` :)

PR #1130, opened 2019-02-18
Refactor servant-client-core

phadej:
- No more Internal modules
- Remove ClientLike-generic. Let's use Routes-generics
    - Let's see if anyone notices, otherwise we can add it back (there were no cookbook, neither tutorial breakages!)
- Add Makefile for common tasks
    - Fix servant-client-ghcjs


parsonsmatt:
Now that we're finally upgraded to Stackage lts-14, I've been bitten by the missing `ClientLike` stuff! Consider it noticed :wink: Can you point me to a migration path to the `Routes-generics` you mention? Thanks!

phadej:
https://haskell-servant.readthedocs.io/en/stable/cookbook/generic/Generic.html

parsonsmatt:
Thanks! i think I can figure it out from there.

parsonsmatt:
OK, I was not able to figure it out, and ripped out all the generic stuff and reverted to boring code. We used nested clients in a way that I couldn't make work. Is there a blocker towards rereleasing the old generic `ClientLIke` code?

phadej:
It was a mess I don’t want to maintain. Nesting is calling for troubles, it’s not worth it. Make a flat record, or separate records; don’t try to be fancy to begin with.

Also, please don’t use ”boring”-word in this repository; servant is not boring by ”the internet opinion”, i’d recommend picking raw wai or scotty if you want/need boring code.

parsonsmatt:
No worries, I completely understand not wanting to maintain it :)

Apologies for 'boring', didn't mean anything negative by it.

PR #1137, opened 2019-02-26
Add current-route recipe

phadej:


doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
+# How to get current route in servant-server?
+
+This tries to be an answer to the [StackOverflow question](https://stackoverflow.com/questions/41538438/haskell-servant-get-current-route-url-from-handler).
```
alpmestan: I'd suggest s/the/this/ and including the "this" in the link

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+There are two questions in one:
+
+1. how to get current Request or URL?
+2. how to get current "route"?
+
+Note, that URL (e.g. `/route12/42`) is different than route
```
alpmestan: You probably want to remove the ",".

Also, I'd probably add a little something...

> Note that a URL (a string that points to some endpoint or page in your web app, e.g `/route12/42`) is different from a route (the actual endpoint or page, e.g `"route12" :> Capture "id" Int :> Get '[JSON] Int`).

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+Passing current `WAI` `Request` to the handler is actually quite easy.
+This is "lazy" approach, we ask for "everything" in the request,
+and we have to be careful in the handler (e.g. we cannot touch `requestBody`).
+Also this "combinator" ties implementation to the `wai` server implementation,
+which is an implementation detail
+(nothing else in `servant-server` exposes `wai` internals, except of `Raw`).
+
+The idea is to make `Server (Wai.Request :> api) = Wai.Request -> Server api`.
+If we imagine for a second that we have such functionality in place,
+we can write, using `Servant.API.Generic` (see "Using generics" cookbook recipe):
```
alpmestan: Would be great to link to the recipe

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+        return (succ i)
+    }
+
+app1 :: Application
+app1 = genericServe routes1
+```
+
+We define a `Routes1` data type, implement `Routes1 AsServer` value and turn it
+into the `wai`'s `Application`. However, to compile this example, we need an
+additional instance.  We use an *internal* `passToServer` combinator in the
```
alpmestan: we need an additional `HasServer` instance (i.e mention the class of which we need an instance of)

I'd also put the "We use an *internal* [...]" sentence after the code block.

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+    type ServerT (Wai.Request :> api) m = Wai.Request -> ServerT api m
+    
+    hoistServerWithContext _ pc nt s =
+        hoistServerWithContext (Proxy :: Proxy api) pc nt . s
+
+    route _ ctx d = route (Proxy :: Proxy api) ctx $
+        passToServer d id
+```
+
+This solution is good quick fix, but there are arguably better ways.
```
alpmestan: While it's obviously fine not to dive into deep explanations of how the internals work, perhaps a sentence or two + a link to `passToServer`'s haddocks would help?

If you don't feel inspired I can give it a shot.

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+Vault is explained in [Using WAI's vault for fun and profit](https://www.yesodweb.com/blog/2015/10/using-wais-vault) blog post.
+It fills a "dynamic" gap of strongly typed `Request`: we can attach arbitrary data to the request,
+as is common in web frameworks in a dynamically typed languages.
+As `servant-server` is based on `wai`, using vault is the third answer
+to the first part of the question.
+
+We (unsafely) create a key to the vault:
+
+```haskell
+rpiKey :: V.Key BS8.ByteString
```
alpmestan: rpi?

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+rpiKey = unsafePerformIO V.newKey
+```
+
+Then we create a *middleware* which will put `rawPathInfo` into the `vault`.
```
alpmestan: Would be great to have the "middleware" bit link to some place that documents them (well, ideally -- does the yesod book have a section about them?).

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+```haskell
+middleware :: Wai.Middleware
+middleware app req respond = do
+    let vault' = V.insert rpiKey (Wai.rawPathInfo req) (Wai.vault req)
+        req' = req { Wai.vault = vault' }
+    app req' respond
+```
+
+Using this we make third variant of our application.
+Note that we values might not be in the vault,
```
alpmestan: -we

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+that's small functional regression.
+
+```haskell
+data Routes3 route = Routes3
+    { route31 :: route :- Vault :> "route1" :> Get '[JSON] Int
```
alpmestan: Might be nice to mention before/after the code, or in a comment in the code, that `Vault` has an `HasServer` instance that gives you... the vault, and that this is why this "just works".

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+        let p = fromMaybe "?" $ V.lookup rpiKey v
+        BS8.putStrLn p
+        return (succ i)
+    }
+
+app3 :: Application
+app3 = middleware $ genericServe routes3
+```
+
+Note: that `vault` can be used to pass information from middlewares to handlers
```
alpmestan: -that

`vault` => vaults

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+and from handlers to middlewares. For example, the authentication can be done
+completely in the middleware, with a user information stored in the vault for
+handlers to use.
+
+## How to get current *route*?
+
+The second part of a question, is how to get current route.
+Something, we can get `route2/:id` out?
```
alpmestan: I'd rephrase: "Is there something that could give us `/route2/:id` for our example from earlier?"

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+
+The usage is very simple, unfortunately the implementation is not.
+
+### Hairy bits
+
+The implementation of `genericRecServe` is intimidating.
+The missing bit is a function `genericHoist`.
+In short, given a function which can convert `modeA :- api` into `modeB :- api` for all `api`,
+`genericHoist` converts `routes modeA` into `routes modeB`.
+Maybe this function should exist in `Servant.API.Generic`?
```
alpmestan: Note to yourself or do you really mean to ask this question to the reader, hoping one of them will tell us "sure, please add it"? Is this sentence a thunk for the patch that adds this? :-)

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+       , GServantHoist c api modeA modeB (Rep (routes modeA)) (Rep (routes modeB))
+       )
+    => Proxy modeA -> Proxy modeB -> Proxy c -> Proxy api
+    -> (forall api'. c api' => Proxy api' -> (modeA :- api') -> (modeB :- api'))
+    -> routes modeA -> routes modeB
+genericHoist pa pb pc api nt = to . gservantHoist pa pb pc api nt . from
+```
+
+`genericRecServe` is `genericHoist` precomposed with a variant of `genericServe`.
+The implementation of one-liner, given a wall of constraints.
```
alpmestan: the implementation **is a** one-liner, [...]

doc/cookbook/current-route/CurrentRoute.lhs
```diff
@@ -0,0 +1,391 @@
...
+    . toServant
+    . genericHoist
+        (Proxy :: Proxy (AsRecServer (AsLink Link)))
+        (Proxy :: Proxy AsServer)
+        (Proxy :: Proxy HasLink')
+        (genericApi (Proxy :: Proxy routes))
+        (\p f -> f $ safeLink p p)
+```
+
+There we us single-instance-class trick to make partially applicable `HasLink`.
```
alpmestan: Do you have a link we could give at this point to help the reader learn about it?

phadej: Let's see if we can find one

doc/cookbook/current-route/CurrentRoute.lhs
phadej: I'm not sure. It type is quite generic, so it could be in `Servant.API.Generic`, but I don't know if it belongs there with this name and type-signature. 

I'd rephrase that line in

This function might be added into `Servant.API.Generic` later, if it proves to be *generally* useful.

doc/cookbook/current-route/CurrentRoute.lhs
phadej: Hmm. While looking up internals, I was confused about non-granular enough split in `servant-server`. I'll do a PR for that for `servant-0.16` still, and then it will be clearer where to add docs.

E.g. there are no `Delayed` module, which makes finding it difficult.

doc/cookbook/current-route/CurrentRoute.lhs
phadej: How the intra-cookbook linking works?

doc/cookbook/current-route/CurrentRoute.lhs
phadej: yeah, I renamed `/route12`, `/route22` etc to just `/route2`, but not everywhere it seems. Good catch.

doc/cookbook/current-route/CurrentRoute.lhs
alpmestan: Sounds great.

I'm not fond of the name because it's too easy to mix with our existing generic and/or hoist machinery, whereas it's a new kind of transformation here, would be nice to have a dedicated name.

doc/cookbook/current-route/CurrentRoute.lhs
alpmestan: Good question. I don't know if there's a way to give relative links. If not, just link to the stable/latest version, that's fine.

doc/cookbook/current-route/CurrentRoute.lhs
alpmestan: Yeah, Delayed deserves its module.

doc/cookbook/current-route/CurrentRoute.lhs
phadej: Maybe `routesHoist`? It's a `hoist` like function of something we call `routes`.

doc/cookbook/current-route/CurrentRoute.lhs
phadej: R aw P ath I nfo

doc/cookbook/current-route/CurrentRoute.lhs
alpmestan: I don't have a better suggestion, that would do.

doc/cookbook/current-route/CurrentRoute.lhs
alpmestan: aaaaaaaaaaaaaaaaaah

bitc:
Hi @phadej Thank you for this write-up.
I tried using this technique, but have a problem.
Inside the handlers, when the `print l` line runs, then it prints out a [Link](http://hackage.haskell.org/package/servant-0.16.1/docs/Servant-API.html#t:Link) containing the actual URL strings rather than the desired placeholder strings:

```
Link {_segments = ["route2","32"], _queryParams = []}
```

I was expecting it to give me something like `["route2", ":id"]`

I was able to get it working by using a custom Data type with a `ToHttpApiData` instance where `toUrlPiece = const (T.pack ":id")` but this feels like a big hack.

Also, I am eagerly awaiting the next step in your guide: automatically and transparently adding the "route" identifier into the Request Vault for all servant handlers, so that a special wai middleware will then be able to compute aggregate metrics grouped by the route "id"

Thank you

doc/cookbook/current-route/CurrentRoute.lhs
jakequade: Just external opinion, but consider removing the link altogether. Links are temporary, while the issue you're addressing will be around for much longer. Give a brief description instead.

doc/cookbook/current-route/CurrentRoute.lhs
jakequade: write once, read a thousand times. Consider that your audience are relatively new to the subject. Consider using `rawPathInfoKey`.

tfausak:
Apologies for reviving quite an old PR :smiley: 

Like @bitc I am very interested in having this functionality. Being able to get the current route is useful for error reporting, logging, and metrics. In particular this is handy when the variable parts (captures, query params, etc) are replaced with placeholders. That way we can say that `GET /user/123` is the "same route" as `GET /user/456` even though the request path is different. 

Is there anything I can do to help push this along? Should this aim to be a blog post, a cookbook entry, a separate library, part of Servant itself, or something else entirely? 

Thanks!

phadej:
Cleaning up old PRs.

Issue #1129, opened 2019-02-18
Split ClientSpec


phadej:
- let's have common test utilities in own module
- and split specs into separate ones

having huge `ClientSpec.hs` makes it more difficult to add tests, as there are no structure in place.

mhcurylo:
I will help. I enjoy copypasta in the morning.

PR #1154, opened 2019-03-16
ClientSpec split into multiple modules

mhcurylo:
ClientSpec split into seperate Specs, #1129.


servant-client/test/Servant/SucessSpec.hs
```diff
@@ -0,0 +1,140 @@
+{-# LANGUAGE CPP                    #-}
```
phadej: SuccessSpec

mhcurylo: It should be a Success now.

PR #1156, opened 2019-03-18
Feature/capture lenient

### LINKED ISSUE #1155: Capture + Lenient?
I have an endpoint that looks like `/Users/v2/<uuid>`. I would like to use `Capture "id" UUID` in my API types (for documentation purposes), but when the UUID is unparseable, I want the API user to get 401, not 404. It would be nice if I could say `Capture' [Lenient] "id" UUID` and then get an `Either Text UUID` in my handler and deal with it myself.

Does that make sense?

jschaul:
Fixes #1155

### DIFF
--- a/servant-docs/golden/comprehensive.md
+++ b/servant-docs/golden/comprehensive.md
@@ -96,6 +96,28 @@
 
 ```
 
+## GET /capture-lenient/:foo
+
+### Captures:
+
+- *foo*: Capture foo Int
+
+### Response:
+
+- Status code 200
+- Headers: []
+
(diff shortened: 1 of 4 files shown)


phadej:
Please add a route to ComprehensiveApi.

jschaul:
> Please add a route to ComprehensiveApi.

Is this what you mean? https://github.com/haskell-servant/servant/pull/1156/commits/6cbf0d38914522bc5652998c86fcef1de4cb89a5

phadej:
Thanks!

PR #1157, opened 2019-03-21
Try refactored haskell-ci

phadej:


phadej:
argh. Trying to make nice looking `.travis.yml` revealed a bug? https://travis-ci.community/t/multiline-commands-have-two-spaces-in-front-breaks-heredocs/2756

PR #1158, opened 2019-03-26
Add Semigroup and Monoid instances for SourceT and StepT

Sciencei:
This makes building up streams simpler (for, for example, building up a RequestBody including multiple files).

### DIFF
--- a/servant/src/Servant/Types/SourceT.hs
+++ b/servant/src/Servant/Types/SourceT.hs
@@ -93,7 +93,7 @@ instance (Monad m) => Semigroup (SourceT m a) where
     (SourceT l) <> (SourceT r) = SourceT $ \k -> l k >> r k
 
 instance (Monad m) => Monoid (SourceT m a) where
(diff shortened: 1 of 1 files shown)


servant/src/Servant/Types/SourceT.hs
```diff
@@ -89,6 +89,12 @@ instance MFunctor SourceT where
     hoist f (SourceT m) = SourceT $ \k -> k $
         Effect $ f $ fmap (hoist f) $ m return
 
+instance (Monad m) => Semigroup (SourceT m a) where
+    (SourceT l) <> (SourceT r) = SourceT $ \k -> l k >> r k
```
phadej: `>>` ?

Sciencei: You're right; fixed in 27e9d33ad14288b8550bc5ee4b00c4f1806e6aa0. Essentially now it composes `k` with appending the right-hand stream.

phadej:
Merged in #1159

Issue #1065, opened 2018-10-31
Provide a template


Sir4ur0n:
Now that https://github.com/commercialhaskell/stack/issues/4039 is released in Stack 1.9.1, it would be nice to provide Stack templates in this repo.
Maybe start again from https://github.com/commercialhaskell/stack-templates/blob/master/servant.hsfiles and fix it (I think it's broken and not up-to-date)?

**Proposals:**
- A template with a single, minimalistic API. This could contain more or less the result of the [A first example](https://haskell-servant.readthedocs.io/en/stable/tutorial/Server.html#a-first-example) section of the tutorial.
- A production-grade template with several APIs showing the full potential of Servant, including code organization (import API types and handlers from several modules), API Swagger documentation, etc. Maybe even some tests? I'm not sure about this part. Basically, something that I can create using `stack new foobar github:haskell-servant/servant-on-steroids`, build, exec, and just see what can be built/done with Servant in a fraction of minute. Then the curious can start toying with the code, the others will read carefully the documentation. Each its own pace/learning style.

alpmestan:
I certainly wouldn't mind shipping templates in the repo, if that makes the life of stack users (especially beginners) easier. I agree that it would be nice to have a very simple one, and then possibly incrementally fancier ones that set up clients, tests, logging, auth, etc. Are you volunteering to add this @Sir4ur0n? :-)

Sir4ur0n:
I don't mind proposing the simple template (I definitely don't know enough Servant fu for the complete one). However I'm waiting for https://github.com/commercialhaskell/stack/issues/4344 to befixed first 😞

alpmestan:
OK, feel free to use this issue to track anything related to servant stack templates.

Sir4ur0n:
No sooner than 5 months later, here it is 😄 Sorry I had completely forgotten about it, then yesterday a coworker struggled with Servant (because he used the old template) and I remembered ^^'

PR #1163, opened 2019-04-03
feat(template): Provide minimalist Stack template

Sir4ur0n:
The minimalist example for #1065.

Features:
* `package.yaml` unlike the old Servant templates
* An API with 2 endpoints: dumb `/ping` and `/users` taking an optional query param
* Tasty test framework
* QuickCheck property test
* HUnit unit test
* Readme explaining how to build/test/run
* Hindent conf file for pretty format of Servant API type (which can be obnoxious when starting)

### DIFF
--- a/servant-minimalist.hsfiles
+++ b/servant-minimalist.hsfiles
@@ -1,0 +1,232 @@
+{-# START_FILE package.yaml #-}
+name:                {{name}}
+version:             0.1.0.0
+github:              "{{github-username}}{{^github-username}}githubuser{{/github-username}}/{{name}}"
+license:             BSD3
+author:              "{{author-name}}{{^author-name}}Author name here{{/author-name}}"
+maintainer:          "{{author-email}}{{^author-email}}example@example.com{{/author-email}}"
+copyright:           "{{copyright}}{{^copyright}}{{year}}{{^year}}2019{{/year}} {{author-name}}{{^author-name}}Author name here{{/author-name}}{{/copyright}}"
+
+extra-source-files:
+  - README.md
+  - ChangeLog.md
+
+# Metadata used when publishing your package
+# synopsis:            Short description of your package
+# category:            {{category}}{{^category}}Web{{/category}}
+
+# To avoid duplicated efforts in documentation and dealing with the
+# complications of embedding Haddock markup inside cabal files, it is
+# common to point users to the README.md file.
+description:         Please see the README on GitHub at <https://github.com/{{github-username}}{{^github-username}}githubuser{{/github-username}}/{{name}}#readme>
+
+dependencies:
+  - base >= 4.7 && < 5
+  - aeson
+  - servant-server
+  - wai
+  - warp
+
+library:
+  source-dirs: src
+
+executables:
+  {{name}}-exe:
+    main:                Main.hs
+    source-dirs:         app
+    ghc-options:
+      - -threaded
+      - -rtsopts
+      - -with-rtsopts=-N
+    dependencies:
+      - {{name}}
+
+tests:
+  {{name}}-test:
+    main:                Spec.hs
+    source-dirs:         test
+    ghc-options:
+      - -threaded
+      - -rtsopts
+      - -with-rtsopts=-N
+    dependencies:
+      - {{name}}
+      # Property based testing
+      - QuickCheck
+      - quickcheck-instances
+      # Unit testing
+      - HUnit
+      # Test framework (aggregation and discovery)
+      - tasty
+      - tasty-quickcheck
+      - tasty-hunit
+      - tasty-discover
+
+{-# START_FILE Setup.hs #-}
+import           Distribution.Simple
+
+main = defaultMain
+
+{-# START_FILE test/Spec.hs #-}
+{-# OPTIONS_GHC -F -pgmF tasty-discover -optF --tree-display
+ #-}
+-- Driver module responsible for discovering all other modules, and tests inside modules, using tasty-discover preprocessor.
+
+{-# START_FILE test/LibSpec.hs #-}
+module LibSpec where
+
+import           Data.Aeson
+import           Lib
+import           Test.QuickCheck
+import           Test.QuickCheck.Instances ()
+import           Test.Tasty.HUnit
+import           Test.Tasty.QuickCheck
+
+-- Property test
+test_findById_no_user =
+  testProperty "findById for non existing users" $ \(User uid _ _) -> uid /= 1 && uid /= 2 ==> findById uid === Nothing
+
+-- Unit test
+test_findById_users =
+  testCase "findById for existing users" $ do
+    findById (userId albert) @?= Just albert
+    findById (userId isaac) @?= Just isaac
+
+instance Arbitrary User where
+  arbitrary = User <$> arbitrary <*> arbitrary <*> arbitrary
+
+{-# START_FILE src/Lib.hs #-}
+{-# LANGUAGE DataKinds       #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeOperators   #-}
+
+module Lib
+  ( startApp
+  , app
+  , findById
+  , isaac
+  , albert
+  , User(..)
+  ) where
+
+import           Data.Aeson
+import           Data.Aeson.TH
+import           Data.List
+import           Data.Maybe
+import           Network.Wai
+import           Network.Wai.Handler.Warp
+import           Servant
+
+data User = User
+  { userId        :: Int
+  , userFirstName :: String
+  , userLastName  :: String
+  } deriving (Eq, Show)
+
+$(deriveJSON defaultOptions ''User)
+
+-- GET http://localhost:8080/users
+type API
+   = "ping" :> Get '[ JSON] String
+     :<|> "users" :> QueryParam "userId" Int :> Get '[ JSON] [User]
+
+startApp :: IO ()
+startApp = run 8080 app
+
+app :: Application
+app = serve api server
+
+api :: Proxy API
+api = Proxy
+
+server :: Server API
+server = return "pong" :<|> handleUsers
+
+handleUsers :: Maybe Int -> Handler [User]
+handleUsers Nothing    = return users
+handleUsers (Just uid) = return . maybeToList . findById $ uid
+
+isaac = User 1 "Isaac" "Newton"
+
+albert = User 2 "Albert" "Einstein"
+
+users :: [User]
+users = [isaac, albert]
+
+findById :: Int -> Maybe User
+findById requestedId = find (\user -> userId user == requestedId) users
+
+{-# START_FILE app/Main.hs #-}
+module Main where
+
+import           Lib
+
+main :: IO ()
+main = startApp
+
+{-# START_FILE README.md #-}
+# {{name}}
+
+## Build
+`stack build`
+
+## Test
+`stack test`
+
+## Run the server:
+`stack exec {{name}}-exe`
+
+## Try the application
+
+### Ping
+http://localhost:8080/ping
+
+### Users
+http://localhost:8080/users
+http://localhost:8080/users?userId=1
+
+{-# START_FILE ChangeLog.md #-}
+# Changelog for {{name}}
+
+## Unreleased changes
+
+{-# START_FILE .hindent.yaml #-}
+# For Servant API declaration
+line-breaks: [":<|>"]
+
+{-# START_FILE LICENSE #-}
+Copyright {{author-name}}{{^author-name}}Author name here{{/author-name}} (c) {{year}}{{^year}}2019{{/year}}
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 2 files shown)


servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+
+# To avoid duplicated efforts in documentation and dealing with the
+# complications of embedding Haddock markup inside cabal files, it is
+# common to point users to the README.md file.
+description:         Please see the README on GitHub at <https://github.com/{{github-username}}{{^github-username}}githubuser{{/github-username}}/{{name}}#readme>
+
+dependencies:
+  - base >= 4.7 && < 5
+  - aeson
+  - servant-server
```
phadej: There should be proper bounds, on all dependencies, especially on `servant`. We do break something every major release.

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+# synopsis:            Short description of your package
+# category:            {{category}}{{^category}}Web{{/category}}
+
+# To avoid duplicated efforts in documentation and dealing with the
+# complications of embedding Haddock markup inside cabal files, it is
+# common to point users to the README.md file.
+description:         Please see the README on GitHub at <https://github.com/{{github-username}}{{^github-username}}githubuser{{/github-username}}/{{name}}#readme>
+
+dependencies:
+  - base >= 4.7 && < 5
```
phadej: `<4.13`

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
+{-# START_FILE package.yaml #-}
```
phadej: I don't want `package.yaml`. it divides community.

Please change to `{{name}}.cabal`.

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+  - ChangeLog.md
+
+# Metadata used when publishing your package
+# synopsis:            Short description of your package
+# category:            {{category}}{{^category}}Web{{/category}}
+
+# To avoid duplicated efforts in documentation and dealing with the
+# complications of embedding Haddock markup inside cabal files, it is
+# common to point users to the README.md file.
+description:         Please see the README on GitHub at <https://github.com/{{github-username}}{{^github-username}}githubuser{{/github-username}}/{{name}}#readme>
```
phadej: This description is bad. The description should describe, not to link forward.

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+      - quickcheck-instances
+      # Unit testing
+      - HUnit
+      # Test framework (aggregation and discovery)
+      - tasty
+      - tasty-quickcheck
+      - tasty-hunit
+      - tasty-discover
+
+{-# START_FILE Setup.hs #-}
```
phadej: `Setup.hs` isn't required for `build-type: Simple`. IIRC even `stack` doesn't compile those?

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+import           Network.Wai.Handler.Warp
+import           Servant
+
+data User = User
+  { userId        :: Int
+  , userFirstName :: String
+  , userLastName  :: String
+  } deriving (Eq, Show)
+
+$(deriveJSON defaultOptions ''User)
```
phadej: why TH, and not Generics?

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+api = Proxy
+
+server :: Server API
+server = return "pong" :<|> handleUsers
+
+handleUsers :: Maybe Int -> Handler [User]
+handleUsers Nothing    = return users
+handleUsers (Just uid) = return . maybeToList . findById $ uid
+
+isaac = User 1 "Isaac" "Newton"
```
phadej: I'd say, every top-level definition should have a type-signature. I.e. the template should be `-Wall` warning free.

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+{-# START_FILE ChangeLog.md #-}
+# Changelog for {{name}}
+
+## Unreleased changes
+
+{-# START_FILE .hindent.yaml #-}
+# For Servant API declaration
+line-breaks: [":<|>"]
+
+{-# START_FILE LICENSE #-}
```
phadej: I personally against suggesting license to users, is there a way for `stack template` to actually ask for a license (like `cabal init`) and generate the file based on it?

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+{-# START_FILE .gitignore #-}
+.stack-work/
+{{name}}.cabal
+*~
```
phadej: - `dist-newstyle/`
- `.ghc.environment.*`
- `cabal.project.local`

servant-minimalist.hsfiles
```diff
@@ -0,0 +1,232 @@
...
+
+extra-source-files:
+  - README.md
+  - ChangeLog.md
+
+# Metadata used when publishing your package
+# synopsis:            Short description of your package
+# category:            {{category}}{{^category}}Web{{/category}}
+
+# To avoid duplicated efforts in documentation and dealing with the
```
phadej: And I don't agree with this comment.

Sir4ur0n:
Hi @phadej,

Thanks for the thorough review!
I consider myself a newbie in Haskell (started learning it less than a year ago) so sorry if I do obvious mistakes about the language or ecosystem, please bear with me 😅 

I wasn't aware of Hpack dividing the community 😨 However I think I don't understand your statement:

> I'm not against providing project skeletons to the users, but I'm against providing stack-only templates, even the template is interpreted by stack.

Did you mean "even *though* the template is interpreted by stack"?

Since users will need Stack to transform the template into a project skeleton, it's safe to assume all consumers of this template will have it, right?
And since `stack build` on such a skeleton generates the `{{name}}.cabal` file, I thought we didn't need to put it manually. My idea was, if anyone wants to use Cabal directly instead of Hpack afterwards, they can build once, then remove the `package.yaml` file.

Here are the solutions I see:
1. Keep as is, but explain in the documentation that one must `stack build` then remove `package.yaml` file if one doesn't like Hpack. Trivial for users (but Cabal users will have to run a build once using Hpack, which may not be OK for them).
2. Provide both `package.yaml` and `{{name}}.cabal` and let users decide what they want to use. Those who want Cabal have to remove `package.yaml`, those who want Hpack have to add `{{name}}.cabal` to the `.gitignore`. I probably also need to update the documentation. Trivial for users.
3. Provide only `{{name}}.cabal`. Cabal users are happy, but Hpack users have to write `package.yaml` **manually** by looking at `{{name}}.cabal`. Non-trivial for Hpack users.

I think solution 2 is a nice compromise as it doesn't seem that Servant is biased towards either Cabal or Hpack. Would that be OK for you?

As for most other remarks, they are mostly on things I kept from the [old Servant template](https://github.com/commercialhaskell/stack-templates/blob/master/servant.hsfiles), I don't mind fixing/changing them. I'm waiting for your feedback on Hpack vs Cabal before starting though, to avoid duplicate work.

servant-minimalist.hsfiles
Sir4ur0n: I don't have any strong opinion on this question. I usually use TH because it's less code to write/maintain in the end by developers (we often have to change only the Aeson Options, and we have to modify it twice with Generics, while only once with TH). 

To clarify, by Generics, I assume you mean
```haskell
deriving (Eq, Show, Generics)

instance ToJSON User

instance FromJSON User
```
Am I correct? Either way, I'm ok with both, just let me know which you prefer (or refuse).

servant-minimalist.hsfiles
Sir4ur0n: This description comes from the old Servant template. I will try to write an explicit description. What - in your opinion - should appear here? Since it is at the user's discretion, something as simple as `<Fill me>`?

servant-minimalist.hsfiles
Sir4ur0n: This is the moment my newbiness shines 👼 : do you have any idea how I can generate (or find) which dependency boundaries to enforce?

Sir4ur0n: Also: since Stack uses the Resolver/snapshot mecanism, do we really need to enforce boundaries in the Hpack file? Maybe this is only needed for Cabal file?

Sir4ur0n:
I noticed I forgot to reply to your last question:
AFAIK Stack generates the `stack.yaml` file itself if none is found, during detemplating. By default, it seems to use the latest known LTS resolver (so you will get the most up-to-date LTS snapshot).
Here's the `stack.yaml` on a freshly generated project (I removed the 30 lines of comments):
```yaml
resolver: lts-13.17
packages:
- .
```

servant-minimalist.hsfiles
Sir4ur0n: Based on this [StackOverflow answer](https://stackoverflow.com/a/50890044/4345848): I guess we can remove it?

servant-minimalist.hsfiles
phadej: that can be written as

```haskell
  deriving (Eq, Show, Generic, ToJSON, FromJSON)
```

or more explicitly with [`DerivingStrategies`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#deriving-strategies)

```haskell
  deriving stock (Eq, Show, Generic)
  deriving anyclass (ToJSON, FromJSON)
```

servant-minimalist.hsfiles
phadej: We need to enforce the boundaries, because there are also others then `stack` users.

This is quite important, because when we get a bug report with `stack` &`hpack` only based reproducion case (and I'd expect this template to be used to generate ones), I'd still need to do manual work, because I'm not using `stack`.

For that reason, I'd like the template to always explicitly use the latest released version of `servant`. If there's even a small change, that the user will find out that an issue isn't reproducible with the latest version, it's a win.

phadej:
Yes, I meant

> even *though* the template is interpreted by stack"?

---

>  By default, it seems to use the latest known LTS resolver 

Is it documented somewhere? Because that's kind of bad.

I'd rather be very specific, and rather manually update template when new `servant` version is released, so the template is always generated with the most recent `servant` in it. It can be based on some LTS, but when new major version of `servant` is released, we'd need to add `extra-deps` etc.

TL;DR, we would like to make sure which version of `servant` is used in template. And be able to change it.

---

People are lazy and won't remove package.yaml file, they'll stick with the template. I don't honestly believe that optional steps will be done. Thus, `.cabal` file have to be there to begin with. `stack` doesn't require `package.yaml` files, it works well with `.cabal` files too.

servant-minimalist.hsfiles
phadej: I agree with HVR's view on `synopsis`, `description` and `README.md` use cases: https://mail.haskell.org/pipermail/haskell-cafe/2016-November/125541.html

I'd like the placeholders for `synopsis` and `description` to reflect that.

phadej:
I cannot find any documentation on how stack templating works, i.e. what features there are,. https://docs.haskellstack.org/en/v1.9.3/search.html?q=templates gives only pages on how to use them, not how to write them.

Sir4ur0n:
> People are lazy and won't remove package.yaml file, they'll stick with the template. I don't honestly believe that optional steps will be done. Thus, .cabal file have to be there to begin with. stack doesn't require package.yaml files, it works well with .cabal files too.

Is the goal to not provide Hpack support, or to provide an easy way for Cabal users? 

If your only focus is to provide an easy way for Cabal users, we can provide 2 templates:
* `servant-minimalist-cabal`
* `servant-minimalist-hpack` (or `servant-minimalist-stack`)

As for the `servant` version regarding Stack, I'm not sure I understand your concern:

> I'd rather be very specific, and rather manually update template when new servant version is released, so the template is always generated with the most recent servant in it

Precisely, the latest Stackage LTS ensures people will use the latest `servant` version package in Stackage. There's a lot more risk for this template not to be maintained whenever a new Servant version is released, thus people getting an outdated Servant version.

phadej:
> Precisely, the latest Stackage LTS ensures people will use the latest servant version package in Stackage. 

I'd care for people to use latest version of `servant`, which at the moment isn't in Stackage LTS.

Screenshot from 2019-04-16 14-09-26

> There's a lot more risk for this template not to be maintained whenever a new Servant version is released.

I see higher risk in new `servant` version not entering upcoming LTS. That we (= `servant` maintainers) cannot affect directly.

> If your only focus is to provide an easy way for Cabal users, we can provide 2 templates:

- :-1: Two templates is out of question.
-  :+1: Single template which works for everyone. `hpack` is convience tool, and given recent developments in .cabal own format, hardly required for anything. See yourself how the format develops: https://cabal.readthedocs.io/en/latest/file-format-changelog.html

PR #1190, opened 2019-07-02
Add sponsorship button

domenkozar:
I'd recommend that there's contact email for each consultancy, that makes it approachable.


alpmestan:
Where can we see the effect of this change?

arianvp:
A donation button should show up in the UI like the one on https://github.com/nixos/nix

alpmestan:
Works like a charm.

PR #1194, opened 2019-07-16
Testing docs, use Warp.testWithApplication to prevent race condition

rl-king:
Ran into an issue today running tests based on these examples on Gitlab CI. The tests were either tried too soon or the port was unavailable. `Warp.testWithApplication` seems to take care of order and shutdown. 

It should also be fine to use hspecs `parallel` since every test gets its own port. I left it out cause it seemed a bit unnecessary for this example. 

The elasticsearch example is not updated and still uses bracket. It didn't seem entirely straightforward to implement due to the setting of the port with the action.

### DIFF
--- a/doc/cookbook/testing/Testing.lhs
+++ b/doc/cookbook/testing/Testing.lhs
@@ -142,33 +142,31 @@ of it and see how it responds.
 Let's write some tests:
 
 ```haskell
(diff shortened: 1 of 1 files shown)


alpmestan:
Thanks!

PR #1198, opened 2019-07-30
Add mysql cookbook example

jakequade:
Contributes a basic mysql-backed API with basic CRUD functionalities.

### DIFF
--- a/doc/cookbook/db-mysql-basics/MysqlBasics.lhs
+++ b/doc/cookbook/db-mysql-basics/MysqlBasics.lhs
@@ -1,0 +1,236 @@
+# Overview
+
+This doc will walk through a single-module implementation of a servant API connecting to a MySQL database. It will also include some basic CRUD operations.
+
+Once you can wrap your head around this implemenation, understanding more complex features like resource pools would be beneficial next steps.
+
+The only *prerequisite* is that you have a MySQL database open on port 3306 of your machine. Docker is an easy way to manage this.
+
+## Setup
+
+- The mysql database should be up and running on 127.0.0.1:3306
+
+- Our API will be exposed on localhost:8080
+
+## REST actions available
+
+*Get all people*
+
+```
+/people    GET
+```
+
+*Get person by ID*
+
+```
+/people/:id    GET
+```
+
+*Insert a new person*
+
+```
+/people    POST
+
+{
+  "name": "NewName",
+  "age": 24
+}
+```
+
+*Delete a person*
+
+```
+/people/:id    DELETE
+```
+
+## Other notes
+
+At the time of writing this issue may occur when building your project:
+
+```
+setup: Missing dependencies on foreign libraries:
+* Missing (or bad) C libraries: ssl, crypto
+```
+
+If using stack, this can be fixed by adding the following lines to your `stack.yaml`:
+
+```
+extra-include-dirs:
+- /usr/local/opt/openssl/include
+extra-lib-dirs:
+- /usr/local/opt/openssl/lib
+```
+
+Or for cabal, running your builds with these configurations passed as options.
+
+## Implementation: Main.hs
+
+Let's jump in:
+
+```haskell
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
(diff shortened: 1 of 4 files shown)


jakequade:
cc @arianvp @alpmestan hello again :) as a noob I've spent a while trying to get this to work via variations of the existing cookbook recipes. I thought this would help out first-timers. Keen on thoughts!

doc/cookbook/db-mysql-basics/MysqlBasics.lhs
```diff
@@ -0,0 +1,189 @@
...
+Person json
+    Id   Int Primary Unique
+    name Text
+    age  Text
+    deriving Eq Show Generic
+|]
+
+data Error = Error404 | Error503 deriving (Generic, Show)
+
+data ApiResponse =
```
alpmestan: Servant already has a notion of "successful response OR error response". The `Handler` monad has an instance of `MonadExcept ServerError`, which means you can use `throwErrror` with values of type `ServerError` whenever appropriate, and just otherwise return what would be the successful response directly, without going through a type like this. Roughly something like:

``` haskell
type API = "person" :> Capture "name" String :> Get '[JSON] Person

getPerson :: String -> Handler Person
getPerson name = do
  mperson <- findPerson name
  case mperson of
    Nothing -> throwError ...
    Just p    -> return p
```

You can even define handlers in an `ExceptT AppErrorType IO` monad instead, so as to throw nice, clean Haskell values as errors instead of using `ServerError` which is pretty much an arbitrary HTTP response. That really is an idiomatic way to handle errors in web apps, judging from all the servant code I've seen.

doc/cookbook/db-mysql-basics/MysqlBasics.lhs
```diff
@@ -0,0 +1,189 @@
...
+deletePerson id = do runDB $ delete $ PersonKey (fromIntegral id :: Int)
+
+startApp :: IO ()
+startApp = do
+    args <- getArgs
+    let arg1 = if not (null args) then Just (head args) else Nothing
+    case arg1 of
+        Just "migrate" -> doMigration
+        _              -> run 8080 app
+```
```
alpmestan: Do you think you could include at the very end a few `curl` commands that hit this web application? We often do this in cookbook recipes, I think it's a nice touch.

(I can add that myself in a subsequent patch though.)

doc/cookbook/db-mysql-basics/MysqlBasics.lhs
```diff
@@ -0,0 +1,189 @@
...
+
+*Delete a person*
+
+```
+/people/:id    DELETE
+```
+
+## Other notes
+
+At the time of writing this issue may occur when running `stack build`:
```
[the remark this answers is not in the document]
jakequade: Good call - have removed this reference and added the cabal alternative a few lines underneath.

doc/cookbook/db-mysql-basics/MysqlBasics.lhs
```diff
@@ -0,0 +1,189 @@
...
+        personPOST personJson = liftIO $ createPerson personJson
+
+selectPersons :: IO [Person]
+selectPersons = do
+  personList <- runDB $ selectList [] []
+  return $ map (\(Entity _ u) -> u) personList
+
+selectPersonById :: Integer -> IO ApiResponse
+selectPersonById id = do
+  sqlResult <- runDB $ get $ PersonKey (fromIntegral id :: Int)
```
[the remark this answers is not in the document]
jakequade: Great idea - done.

doc/cookbook/db-mysql-basics/MysqlBasics.lhs
jakequade: Ah - cool! Had no idea that was possible. Will make that change 😄 thanks!

doc/cookbook/db-mysql-basics/MysqlBasics.lhs
jakequade: Definitely, will add that in now.

jakequade:
Thanks so much for your help @alpmestan - it's been super helpful. I've made those changes, keen on thoughts!

jakequade:
Just bumping this one @alpmestan @arianvp  - I know you're busy so I wont push any more :) really appreciate all the advice you've given.

alpmestan:
We'll arrange for this recipe to be CI-checked like the others in a subsequent PR. Ping us if we havent' done so within a month or so.

Thanks a lot for your contribution!

PR #1213, opened 2019-09-05
Added a function to create Client.Request in ClientEnv

symbiont-eric-torreborre:
Hi, I am proposing this PR to fix an issue I am currently facing but you might propose a better solution. Basically I need to set "dynamic timeouts" on HTTP requests based on previous attempts (I try a query, if it times out, I try with a bigger timeout).

Unfortunately at the moment when I get to execute a `ClientM` action there is nothing I can do to intercept the code in `performRequest` which is transforming a `Servant.Client.Core.Request` into a `Network.HTTP.Client.Request` (that function is requestToClientRequest). This PR adds such a function into the `ClientEnv` so that I can override it with some fine-tuning behaviour for changing the sent `Network.HTTP.Client.Requests`.

Some other options would be to more massively use `Free ClientF` in our code (instead of `ClientM`) and rely on a specific interpreter. The drawback would be that we would have to embed all the code in `performRequest` (in theory, because in our specific use case we don't rely on cookies so this would be less code).

What do you think? Do you see better alternatives?

### DIFF
--- a/servant-client/src/Servant/Client/Internal/HttpClient.hs
+++ b/servant-client/src/Servant/Client/Internal/HttpClient.hs
@@ -68,16 +68,23 @@ import qualified Servant.Types.SourceT                    as S
 import qualified Network.HTTP.Client         as Client
 
 -- | The environment in which a request is run.
+--   The baseUrl and makeClientRequest function are used to create a `http-client` request.
+--   Cookies are then added to that request if a CookieJar is set on the environment.
+--   Finally the request is executed with the manager.
+--   The 'makeClientRequest' function can be used to modify the request to execute and set values which
+--   are not specified on a `servant` Request like the `responseTimeout` or the `redirectCount`
 data ClientEnv
   = ClientEnv
   { manager :: Client.Manager
   , baseUrl :: BaseUrl
   , cookieJar :: Maybe (TVar Client.CookieJar)
+  , makeClientRequest :: BaseUrl -> Request -> Client.Request
+  -- ^ this function can be used to customize the creation of `http-client` requests from `servant` requests. Default value: 'defaultMakeClientRequest'
   }
 
 -- | 'ClientEnv' smart constructor.
 mkClientEnv :: Client.Manager -> BaseUrl -> ClientEnv
-mkClientEnv mgr burl = ClientEnv mgr burl Nothing
+mkClientEnv mgr burl = ClientEnv mgr burl Nothing defaultMakeClientRequest
 
 -- | Generates a set of client functions for an API.
 --
@@ -148,8 +155,8 @@ runClientM cm env = runExceptT $ flip runReaderT env $ unClientM cm
 
 performRequest :: Request -> ClientM Response
 performRequest req = do
-  ClientEnv m burl cookieJar' <- ask
-  let clientRequest = requestToClientRequest burl req
+  ClientEnv m burl cookieJar' createClientRequest <- ask
+  let clientRequest = createClientRequest burl req
   request <- case cookieJar' of
     Nothing -> pure clientRequest
     Just cj -> liftIO $ do
@@ -158,7 +165,7 @@ performRequest req = do
         oldCookieJar <- readTVar cj
         let (newRequest, newCookieJar) =
               Client.insertCookiesIntoRequest
-                (requestToClientRequest burl req)
+                clientRequest
                 oldCookieJar
                 now
         writeTVar cj newCookieJar
@@ -211,8 +218,11 @@ clientResponseToResponse f r = Response
     , responseHttpVersion = Client.responseVersion r
     }
 
-requestToClientRequest :: BaseUrl -> Request -> Client.Request
-requestToClientRequest burl r = Client.defaultRequest
+-- | Create a `http-client` Request from a `servant` Request
+--    The host, path and port fields are extracted from the BaseUrl
+--    otherwise the body, headers and query string are derived from the `servant` Request
+defaultMakeClientRequest :: BaseUrl -> Request -> Client.Request
+defaultMakeClientRequest burl r = Client.defaultRequest
     { Client.method = requestMethod r
     , Client.host = fromString $ baseUrlHost burl
     , Client.port = baseUrlPort burl
--- a/servant-client/src/Servant/Client/Internal/HttpClient.hs
+++ b/servant-client/src/Servant/Client/Internal/HttpClient.hs
@@ -80,6 +80,10 @@ data ClientEnv
   , cookieJar :: Maybe (TVar Client.CookieJar)
   , makeClientRequest :: BaseUrl -> Request -> Client.Request
   -- ^ this function can be used to customize the creation of @http-client@ requests from @servant@ requests. Default value: 'defaultMakeClientRequest'
+  --   Note that:
+  --      1. 'makeClientRequest' exists to allow overriding operational semantics e.g. 'responseTimeout' per request,
+  --          If you need global modifications, you should use 'managerModifyRequest'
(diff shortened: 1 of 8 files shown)


symbiont-eric-torreborre:
Note: I had to add some dependencies updates to the `stack.yaml` file other wise I would get errors like
```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for servant-0.16:
    http-api-data-0.3.8.1 from stack configuration does not match >=0.4.1 && <0.4.2  (latest matching version is 0.4.1)
needed due to servant-client-0.16 -> servant-0.16

Some different approaches to resolving this:

  * Set 'allow-newer: true' in /Users/etorreborre/.stack/config.yaml to ignore all version constraints and build anyway.

  * Recommended action: try adding the following to your extra-deps in /Users/etorreborre/projects/haskell/servant/stack.yaml:

- http-api-data-0.4.1@sha256:7a3d3782859c7cac74b82cd41ad8ae0ccd6d6ca5279a88b77d4421de71920a4a,4039

Plan construction failed.
```
Anyone knows why?

alpmestan:
Hello! Sorry for the delay.

I'm personally not opposed to this change, especially given that in the absolute majority of situations people won't have to change their code. @phadej Objections?

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -73,11 +73,12 @@ data ClientEnv
   { manager :: Client.Manager
   , baseUrl :: BaseUrl
   , cookieJar :: Maybe (TVar Client.CookieJar)
+  , makeClientRequest :: BaseUrl -> Request -> Client.Request
```
phadej: - this one needs haddocks
    - at leas mentioning that `cookieJar` is applied to the request created
- also we'll need to export default value from non-`Internal` module, as probably many implementations will call it either mangling servant's or http-client's `Request`. I propose `defaultMakeClientRequest` (it also needs haddocks, explaining what it does).

symbiont-eric-torreborre: Thanks for the comment I will do that tomorrow.

symbiont-eric-torreborre:
@phadej I amended my commit with:
 
 - some documentation on `ClientEnv` 
 - a similar change on the `ClientEnv` for streaming requests for consistency
 - a test where I show that it is possible to use the `ClientEnv` to modify the headers before executing a request

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -68,16 +68,23 @@ import qualified Servant.Types.SourceT                    as S
 import qualified Network.HTTP.Client         as Client
 
 -- | The environment in which a request is run.
+--   The baseUrl and makeClientRequest function are used to create a `http-client` request.
```
phadej: Note, haddocks aren't markdown. 

```
The 'baseUrl' and 'makeClientRequest' function are used to create a @http-clietn@ request.

...

on a @servant@

```

In short: use ticks, no backticks. `@` for "code-like".

servant-client/test/Servant/SuccessSpec.hs
```diff
@@ -125,16 +126,26 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
...
-      _ <- runClientM (getRedirectWithCookie HTTP.methodGet) (ClientEnv mgr baseUrl (Just cj))
+      _ <- runClientM (getRedirectWithCookie HTTP.methodGet) (ClientEnv mgr baseUrl (Just cj) defaultMakeClientRequest)
       cookie <- listToMaybe . C.destroyCookieJar <$> atomically (readTVar cj)
       C.cookie_name <$> cookie `shouldBe` Just "testcookie"
       C.cookie_value <$> cookie `shouldBe` Just "test"
 
+    it "Can modify the outgoing Request using the ClientEnv" $ \(_, baseUrl) -> do
+      mgr <- C.newManager C.defaultManagerSettings
+      let createClientRequest url r = (defaultMakeClientRequest url r) { C.requestHeaders = [("X-Added-Header", "XXX")] }
+      let clientEnv = ClientEnv mgr baseUrl Nothing createClientRequest
```
phadej:
```haskell
(mkClientEnv mg baseUrl)} { makeClientRequest = createClientRequest }
```

servant-client/src/Servant/Client/Internal/HttpClient.hs
symbiont-eric-torreborre: Thanks, I fixed the comments in a fixup commit.

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -68,16 +68,23 @@ import qualified Servant.Types.SourceT                    as S
...
+--   Finally the request is executed with the manager.
+--   The 'makeClientRequest' function can be used to modify the request to execute and set values which
+--   are not specified on a `servant` Request like the `responseTimeout` or the `redirectCount`
 data ClientEnv
   = ClientEnv
   { manager :: Client.Manager
   , baseUrl :: BaseUrl
   , cookieJar :: Maybe (TVar Client.CookieJar)
+  , makeClientRequest :: BaseUrl -> Request -> Client.Request
+  -- ^ this function can be used to customize the creation of `http-client` requests from `servant` requests. Default value: 'defaultMakeClientRequest'
```
phadej: add also

`makeClientRequest` exists to allow overriding operational semantics e.g. `responseTimeout` *per request*.
If you need global modifications, you should use [`managerModifyRequest`](https://hackage.haskell.org/package/http-client-0.6.4/docs/Network-HTTP-Client.html#v:managerModifyRequest)

servant-client/test/Servant/SuccessSpec.hs
```diff
@@ -125,16 +126,26 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
...
       cj <- atomically . newTVar $ C.createCookieJar []
-      _ <- runClientM (getRedirectWithCookie HTTP.methodGet) (ClientEnv mgr baseUrl (Just cj))
+      _ <- runClientM (getRedirectWithCookie HTTP.methodGet) (ClientEnv mgr baseUrl (Just cj) defaultMakeClientRequest)
       cookie <- listToMaybe . C.destroyCookieJar <$> atomically (readTVar cj)
       C.cookie_name <$> cookie `shouldBe` Just "testcookie"
       C.cookie_value <$> cookie `shouldBe` Just "test"
 
+    it "Can modify the outgoing Request using the ClientEnv" $ \(_, baseUrl) -> do
+      mgr <- C.newManager C.defaultManagerSettings
+      let createClientRequest url r = (defaultMakeClientRequest url r) { C.requestHeaders = [("X-Added-Header", "XXX")] }
```
phadej: Add a comment:

In proper situation, extra headers should probably visible in API type. However, testing for response timeout is difficult, so we test with something which is easy to observe.

servant-client/src/Servant/Client/Internal/HttpClient.hs
phadej: .. and the note that possible `cookieJar` is applied after.

symbiont-eric-torreborre:
@phadej I added better comments

phadej:
Good. Let me wait for CI, I'll merge them.

Thanks for your contribution.

symbiont-eric-torreborre:
Thanks a lot for reviewing and accepting that change!

symbiont-eric-torreborre:
@phadej I didn't see the docs indeed. The build is passing now.

phadej:
Squashed in https://github.com/haskell-servant/servant/pull/1255

PR #1216, opened 2019-09-07
Add servant-jsaddle

phadej:
- servant-client-jsaddle: Remove some debug printing
- Update .travis.yml with haskell-ci
- servant-client-jsaddle: bump base bounds
- Add libgirepository1.0-dev
- servant-client-jsaddle: bump upper bound on containers
- servant-client-jsaddle: relax upper bound on semigroupoids
- servant-client-jsaddle: bump servant-client-core dependency
- servant-client-jsaddle: fix compatibility
- servant-client-jsaddle: import correct module
- .travis.yml: run xvfb for headless GUI testing
- Use ghcjs-dom instead of jsaddle-dom directly.

  This way we avoid the dependency on jsaddle-dom on ghcjs and have
  slightly better performance on ghcjs.

- Also use ghcjs-dom in tests.
- Ignore exceptions on send - they are handled in toResponse.
- Apparently ghcjs-dom does use the same exception these days.
- Got rid of obsolete comment.
- Make sure response gets handled even in case of exception.
- Update servant-client-jsaddle/servant-client-jsaddle.cabal

Co-Authored-By: Herbert Valerio Riedel <hvr@gnu.org>


kaol:
If someone else is confused and looking for this one, it's been since removed from the main servant repository and it currently resides at https://github.com/haskell-servant/servant-jsaddle

PR #1219, opened 2019-09-07
Add NoContentVerb

### LINKED ISSUE #1028: servant-docs: Incorrect docs generated for NoContent responses
Servant-docs should emit

> ### Response:
>
> - Status code 204
> - Headers: []
> - No response body

but instead emits 

> ### Response:
>
> - Status code 204
> - Headers: []
>
> - Supported content types are:
>
>   - `application/json;charset=utf-8`
>   - `application/json`
>
> - Example (`application/json;charset=utf-8`, `application/json`):
>
>    ```javascript
>
>   ```

because NoContent responses have to set an arbitrary content type. (In this example `'[JSON]`)

Related issues: #498, #683

fierce-katie:
Closes #1028.

The `NoContentVerb` endpoints always return status code 204 and empty body and it is reflected in the docs.

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -75,10 +75,12 @@ import           Servant.API
                  IsSecure (..), QueryFlag, QueryParam', QueryParams, Raw,
                  ReflectMethod (reflectMethod), RemoteHost, ReqBody',
                  SBool (..), SBoolI (..), SourceIO, Stream, StreamBody',
-                 Summary, ToSourceIO (..), Vault, Verb, WithNamedContext)
+                 Summary, ToSourceIO (..), Vault, Verb, NoContentVerb,
+                 WithNamedContext)
 import           Servant.API.ContentTypes
                  (AcceptHeader (..), AllCTRender (..), AllCTUnrender (..),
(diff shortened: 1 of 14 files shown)


fierce-katie:
The only failing build on Travis CI is the one for ghc-8.0.2. It fails with
```
internal error in InstallPlan.completed
  error, called at ./Distribution/Client/InstallPlan.hs:262:25 in main:Distribution.Client.InstallPlan
```
I have not changed the CI configuration.

Probably related to https://github.com/haskell/cabal/issues/6060

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -285,6 +298,15 @@ instance {-# OVERLAPPING #-}
     where method = reflectMethod (Proxy :: Proxy method)
           status = toEnum . fromInteger $ natVal (Proxy :: Proxy status)
 
+instance {-# OVERLAPPABLE #-}
```
phadej: OVERLAPPABLE/OVERLAPPIG (here and elsewhere) are not required, are they?

alpmestan:
I like this patch, because users don't have to specify an arbitrary content type anymore.

phadej:
Thanks. Merged rebased version in https://github.com/haskell-servant/servant/pull/1228 (CI was green there).

PR #1223, opened 2019-09-18
Use TypeApplications

BebeSparkelSparkel:


phadej:
No. I'm not a fan of `AllowAmbiguousTypes` (which one would need to enable). Thanks, but no thanks. Changes like these are better discussed as an issue first.

BebeSparkelSparkel:
That's why I didn't put too much work into it. Thanks for your opinion.

PR #1247, opened 2019-11-26
Some admin (Stack, gitignore)

mheinzel:
Fixes for some small things I encountered when trying to setup and build the project.

1. When running `stack build --fast --pedantic` I got:
    ```
    WARNING: Ignoring out of range dependency (allow-newer enabled): universe-base-1.0.2.1. servant-docs requires: >=1.1.1 && <1.2
    ```

    ```
        /home/shared/projects/servant/servant-docs/src/Servant/Docs/Internal.hs:498:18: error:
        Not in scope: ‘U.cartesianProduct’
        Module ‘Data.Universe.Helpers’ does not export ‘cartesianProduct’.
        |
    498 |   gtoSamples _ = U.cartesianProduct render ps qs
        |                  ^^^^^^^^^^^^^^^^^^
    ```
    This happened because `cartesianProduct` was added in `universe-base-1.1`, which is not in the specified `nightly-2018-09-28` snapshot.

2. When I tried running `stack ghci servant-http-streams`, I noticed it is not part of the stack.yaml (but of the cabal.project).

3. git picked up the `tags` file created by my editor.

Let me know if any of these are not wanted, I'm happy to amend the PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ doc/_build
 doc/venv
 doc/tutorial/static/api.js
 doc/tutorial/static/jq.js
+tags
 
 # local versions of things
 servant-multipart
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -80,7 +80,7 @@ import           Servant.API
 import           Servant.API.ContentTypes
                  (AcceptHeader (..), AllCTRender (..), AllCTUnrender (..),
                  AllMime, MimeRender (..), MimeUnrender (..), canHandleAcceptH,
-                 NoContent (NoContent))
+                 NoContent)
 import           Servant.API.Modifiers
                  (FoldLenient, FoldRequired, RequestArgument,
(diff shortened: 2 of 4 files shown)


mheinzel:
There seems to be problem with GHCJS in CI:
```
The following packages have unmet dependencies:

 ghcjs-8.4 : Depends: npm but it is not going to be installed

E: Unable to correct problems, you have held broken packages.
```

I can try to find some time to see what is going on here, but maybe someone else knows.

.gitignore
```diff
@@ -29,6 +29,7 @@ doc/_build
 doc/venv
 doc/tutorial/static/api.js
 doc/tutorial/static/jq.js
+tags
```
phadej: this really should be part of your global gitignore. I'm against putting all editor temporary files into projects `.gitignore`.

Other stuff looks ok.

phadej:
Cherry-picked parts in https://github.com/haskell-servant/servant/pull/1258.

Thanks!

PR #1249, opened 2019-12-07
use queryString not rawQueryString to enable param  rewrites with

wrover:
WAI middleware provides a handful way to rewrite queries, including query parameters name. Using of `rawQueryString` makes such rewrites useless. Can we use `queryString` instead of it to make it possible?

### DIFF
--- a/servant-server/test/Servant/ServerSpec.hs
+++ b/servant-server/test/Servant/ServerSpec.hs
@@ -35,7 +35,7 @@ import qualified Data.Text                         as T
 import           GHC.Generics
                  (Generic)
 import           Network.HTTP.Types
-                 (Status (..), hAccept, hContentType, imATeapot418,
+                 (Status (..), Query, QueryItem, hAccept, hContentType, imATeapot418,
                  methodDelete, methodGet, methodHead, methodPatch, methodPost,
                  methodPut, ok200, parseQuery)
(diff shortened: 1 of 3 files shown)


phadej:
`master` CI is fixed, please rebase at some point.

phadej:
ping @wrover

wrover:
Sorry for the delay, please take a look on tests.

wrover:
Tests fail on master, code is here: https://github.com/wrover/servant/tree/broken-rewrites

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -339,6 +339,15 @@ qpServer = queryParamServer :<|> qpNames :<|> qpCapitalize :<|> qpAge :<|> qpAge
         queryParamServer (Just name_) = return alice{name = name_}
         queryParamServer Nothing = return alice
 
+-- test query parameters rewriter, commonly implemented using WAI Middleware
+sampleRewrite :: Query -> Query
+sampleRewrite = fmap mapper where
```
phadej: This should be a `Middleware`, `serve queryParamApi qpServer` produces an `Application`, to which it could be applied.

I'll do the change myself

phadej:
Merged in #1262, thanks!

PR #1254, opened 2019-12-14
I guess some sample curl commands are wrong.

natsuki14:
I guess some sample curl commands on the tutorial are wrong.


alpmestan:
Oops, thanks for spotting those!

PR #1288, opened 2020-04-06
Build servant repo with nix.

dmjio:
cc @fizruk @arianvp @fisx 

We can add `pkgs.haskell.lib.sdistTarball` as well for easy generation of release tarballs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 **/*/dist
+*~
 dist-*
 .ghc.environment.*
 /bin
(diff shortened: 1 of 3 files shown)


arianvp:
I think it's useful, but probably not a replacement for the current CI in place.   Currently I think cabal is used against a multitude of GHC versions which will give us guarantee a build plan is always found, whilst nix will only have very specific versions of packages (Even across versions of GHC)

dmjio:
@arianvp agreed, not intended as a CI replacement, but for local nix development.

PR #1301, opened 2020-05-24
Remove dead link to google group from README

fisx:
It looks to me like this link is dead.

Any facts, insights, opinions I'm missing?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,7 @@ The central documentation can be found [here](http://docs.servant.dev/).
 Other blog posts, videos and slides can be found on the
 [website](http://www.servant.dev/).
 
-If you need help, drop by the IRC channel (#servant on freenode) or [mailing
-list](https://groups.google.com/forum/#!forum/haskell-servant).
+If you need help, drop by the IRC channel (#servant on freenode).
 
 ## Contributing
 


dmjio:
@fisx merge away! :)

dmjio:
@fisx, on second thought, are you sure it's dead? It resolves for me https://groups.google.com/forum/#!forum/haskell-servant

fisx:
well, it resolves, but it has this warning in the lower left corner:

2020-06-06-194355_1600x900_scrot

Regardless of whether it's technically still available or not, I would prefer to move away from google.  We have irc and github, and if that's not enough, we can talk about what we want.  But I'd prefer to set up a mailman myself over using google groups.

So: if anybody wants google groups, please speak up soon!  :)

dmjio:
Screen Shot 2020-06-06 at 7 19 15 PM

@fisx, weird.

fisx:
i suppose your google account has more access than mine?  Who owns this group?  Can they see if it can be made more public?

Also, how old are the discussions?  If there is still any reasonably recent traffic, I suppose we shouldn't just kill it.

dmjio:
@fisx yea, I agree, probably best to keep. We can just make you an admin of the google group as well.

PR #1302, opened 2020-05-25
Allow aeson 1.5

felixonmars:
Builds fine and all tests pass here.


dmjio:
@felixonmars can we bump `aeson` for all `servant` packages as well ? Would be nice to keep them all in sync.

felixonmars:
Oops, I forgot about this... Sorry!

PR #1305, opened 2020-06-06
More cookbook improvements

jhrcek:


doc/cookbook/curl-mock/CurlMock.lhs
```diff
@@ -86,7 +86,7 @@ listFromAPI :: (HasForeign lang ftype api, GenerateList ftype (Foreign ftype api
 ```
 
 This looks a bit confusing...
-[Here](https://hackage.haskell.org/package/servant-foreign-0.11.1/docs/Servant-Foreign.html#t:HasForeignType) is the documentation for the `HasForeign` typeclass.
+[Here](https://hackage.haskell.org/package/servant-foreign/docs/Servant-Foreign.html#t:HasForeignType) is the documentation for the `HasForeign` typeclass.
```
jhrcek, line 89: Changed all the cookbook links to point to latest version of the package on hackage. I checked that all the links I changed are alive.

doc/cookbook/curl-mock/CurlMock.lhs
```diff
@@ -130,24 +130,12 @@ generateCurl :: (GenerateList Mocked (Foreign Mocked api), HasForeign NoLang Moc
 generateCurl p host =
   fmap T.unlines body
   where
-  body = foldr (\endp curlCalls -> mCons (generateEndpoint host endp) curlCalls) (return [])
+  body = mapM (generateEndpoint host)
```
jhrcek, line 133: I guess most people dealing with servant will know mapM. The mCons helper function adds unrelated complexity to the recipe.

doc/tutorial/Server.lhs
```diff
@@ -634,7 +634,7 @@ kind and abort early. The next two sections cover how to do just that.
 
 Other important instances from the list above are `MonadIO m => MonadIO
 (ExceptT e m)`, and therefore also `MonadIO Handler` as there is a `MonadIO IO` instance.
-[`MonadIO`](http://hackage.haskell.org/package/transformers-0.4.3.0/docs/Control-Monad-IO-Class.html)
+[`MonadIO`](http://hackage.haskell.org/package/base/docs/Control-Monad-IO-Class.html#t:MonadIO)
```
jhrcek, line 637: `MonadIO` is in base since 4.9.0.0 which corresponds to GHC 8.0. 
Checking travis config in this repo I take it that you test with GHC 8+ only, so it makes sense to link to the current home of MonadIO.

doc/cookbook/curl-mock/CurlMock.lhs
dmjio: Should this be https://hackage.haskell.org/package/servant-foreign-0.15.1/docs/Servant-Foreign.html#t:HasForeign instead?

jhrcek: You're right. I was confused, because the original code lined to HasForeignType, which also exists in the package. But from the context it's clear that link to HasForeign is needed here. Fixed.

PR #1312, opened 2020-06-14
Configurable combinator errors

### LINKED ISSUE #685: Configurable combinator errors
# Problem

Currently combinators throw errors in a non-configurable way. If JSON decoding fails, for example, it's [hard to get the error message to look the way you want](http://stackoverflow.com/questions/41753516/custom-json-errors-for-servant-server). As another example, one would need to use hacky middleware to make 404 pages one likes.

# Proposed approach

One of the exciting things about `Context` is that it makes it much easier to configure combinators, and makes fixing this relatively trivial. The way I conceive of it is:

``` haskell
data BasicErrorFormatting
  = NotFound (String -> ServantErr)  -- argument is requested route
  | BodyParseError (String -> ServantErr  -- argument is parse error string
  | QueryParamParseError ([String] -> ServantErr)  -- argument is all parse error strings
  ...

instance (..., HasConfigEntry ctx BasicErrorFormatting) => symbol :> rest where
  ....
```

Or:

``` haskell
newtype NotFound = NotFound (String -> ServantErr) 
newtype BodyParseError = ...
  ...

instance (..., HasConfigEntry ctx NotFound) => symbol :> rest where
  ....
```

And then we would provide a default `BasicErrorFormatting` (or set of error datatypes) that mimics current behavior, and make `serve` use it. `serveWithContext` would allow configuration.

# Open questions

1. Ideally errors too would be responsive to `Accept` headers, though it's not clear how that would work
2. We could also allow `IO` before the response (i.e., `NotFound (String -> IO ServantErr`). The main advantage that I can see would be logging, though perhaps this is well covered by middleware

maksbotan:
Currently there is no way for Servant users to customize formatting of
error messages that arise when combinators can't parse URL or request
body, apart from reimplementing those combinators for themselves or
using middlewares.

This commit adds a possibility to specify custom error formatters
through Context.

Fixes #685


servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
```diff
@@ -0,0 +1,63 @@
...
+  :. defaultHeaderParseErrorFormatter
+  :. EmptyContext
+
+-- | A custom formatter for errors produced by parsing combinators like
+-- 'Servant.API.ReqBody' or 'Servant.API.Capture'.
+--
+-- A 'TypeRep' argument described the concrete combinator that raised
+-- the error, allowing formatter to customize the message for different
+-- combinators.
+type ErrorFormatter = TypeRep -> String -> ServerError
```
Martoon-00, line 33: I think since servant supports returning response in multiple different formats (e.g. `Post [JSON, PlainText] Response`), being able to specify only one fixed error formatter might be a limitation. Maybe also make this type take the value of `Accept` header?

maksbotan: Good point! That was brought up in #685, but I deliberately ignored it for now. I think that for most applications it would be enough to force one type in the formatters.

There is no sane way to get verb's content-types while parsing URL parts and the body, as far as I can tell. For example:

```
type API = ReqBody '[JSON] Foo :> ("json" :> Get '[JSON] Bar :<|> "text" :> Get '[PlainText] Bar)
```

How would you know what content-type to return while parsing the `ReqBody`?

On the other hand, it's sure possible to add current `Accept` header value to formatter's arguments, should not be too hard to do.

Let's hear what the maintainers think.

maksbotan:
cc @jkarni @epicallan

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
maksbotan: If we were to pass `Accept` to error formatter, one could then write formatter like this:

```haskell
errorFormatter acc rep err = handleAcceptH (Proxy @'[JSON, PlainText]) acc $ object ["combinator" .= show rep, "error" .= err]
```

Modulo the need to wrap that into a `ServerError` of course.
Would that make sense, @fisx?

fisx: > How would you know what content-type to return while parsing the ReqBody?

By inspecting the path in the request?

Passing `Accept` would work, yes.  Seems like a reasonable solution to me.

maksbotan: Okay, I'll do it.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
```diff
@@ -0,0 +1,63 @@
...
+-- Default formatters will just return HTTP 400 status code with error
+-- message as response body.
+type DefaultErrorFormatters = '[BodyParseErrorFormatter, URLParseErrorFormatter, HeaderParseErrorFormatter]
+
+defaultErrorFormatters :: Context DefaultErrorFormatters
+defaultErrorFormatters =
+  defaulyBodyParseErrorFormatter
+  :. defaultURLParseErrorFormatter
+  :. defaultHeaderParseErrorFormatter
+  :. EmptyContext
```
fisx, line 25: What about `404 not found`?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -588,7 +608,7 @@ instance HasServer Raw context where
 -- The @Content-Type@ header is inspected, and the list provided is used to
 -- attempt deserialization. If the request does not have a @Content-Type@
 -- header, it is treated as @application/octet-stream@ (as specified in
--- <http://tools.ietf.org/html/rfc7231#section-3.1.1.5 RFC7231>.
+-- [RFC 7231 section 3.1.1.5](http://tools.ietf.org/html/rfc7231#section-3.1.1.5 RFC7231).
```
fisx, line 611: Can you please keep the haddocks markup?  Or does haddock support markdown these days?

servant-server/src/Servant/Server.hs
```diff
@@ -126,8 +138,8 @@ import           Servant.Server.Internal
 -- > main :: IO ()
 -- > main = Network.Wai.Handler.Warp.run 8080 app
 --
-serve :: (HasServer api '[]) => Proxy api -> Server api -> Application
-serve p = serveWithContext p EmptyContext
+serve :: (HasServer api DefaultErrorFormatters) => Proxy api -> Server api -> Application
+serve p = serveWithContext p defaultErrorFormatters
```
fisx, line 142: Doesn't this mean that if you want to add your own formatters, you need to define variants of `serve`, `hoistServer`, ...?

Also, what happens if the context is to be used for something else?  (I've never used servant contexts and I'm not sure how they work.)

servant-server/src/Servant/Server/Internal.hs
maksbotan: image
It does support this particular syntax, yes.

There was however a problem in my variant, thanks for drawing my attention!

maksbotan: See here: https://haskell-haddock.readthedocs.io/en/latest/markup.html#links

servant-server/src/Servant/Server.hs
maksbotan: Yes, so this is kinda a breaking change.

If one needs to use one's custom context and wants to keep default handlers, custom context should just be cons'ed on top of the default one. An example can be found [here](https://github.com/haskell-servant/servant/pull/1312/files#diff-53827cb3f09ba7c0122b0f4349f0b86aR83):

```haskell
comprehensiveApiContext :: Context (NamedContext "foo" '[] ': DefaultErrorFormatters)
comprehensiveApiContext = NamedContext EmptyContext :. defaultErrorFormatters
```

Thanks to `HasContextEntry` picking the first entry with the correct type, a user can add custom formatter of one type without needing to override others. See example in `greet.hs` (I'm going to revert that change, it's purely for demonstration of the PR).

```haskell
test = serveWithContext testApi (customFormatter :. defaultErrorFormatters) server
```

Here the context will get two formatters of `BodyParseErrorFormatter` type, but user's one will be picked at runtime.

I could've made another typeclass `MaybeHasContextEntry` that will pick something from context if it's present and return `Nothing` otherwise, but I find my approach more explicit.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
maksbotan: As far as I've seen, 404 is usually bodyless, so does not require customization.

But probably you and other maintainers know better. If you think 404 should be customizable as well, I'll add it.

jkarni: I think 404 might be (for people doing server-rendered HTML) one of the most important ones to customize! I think ideally every servant error should be customizable.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
```diff
@@ -0,0 +1,64 @@
...
+
+-- | A custom formatter for errors produced by parsing combinators like
+-- 'Servant.API.ReqBody' or 'Servant.API.Capture'.
+--
+-- A 'TypeRep' argument described the concrete combinator that raised
+-- the error, allowing formatter to customize the message for different
+-- combinators.
+type ErrorFormatter = TypeRep -> AcceptHeader -> String -> ServerError
```
jkarni, line 34: I think the errors showing which combinator threw them is really cool! It has the potential to make a lot of debugging much easier!

jkarni: I'm not sure about the rest of this though. We now have `AcceptHeader`; next release we might want `AcceptLanguage` (so you can do localized errors); release after that something else. Each time, it's going to be a breaking change.

Would `Request` not maybe be better? If there's a reason not to go with that, then maybe a new datatype that looks something like:

```
data ErrorContext = ErrorContext 
   { requestAcceptHeader :: AcceptHeader
   , throwingCombinator :: TypeRep
   , errorMessage :: String
   } 
```
(And then `ErrorFormatter = ErrorContext -> ServerError`).

Note that if there is such a datatype, the constructor should only be exported from an `Internal` module; from eg `Servant` itself, only the field accessors should be exported (that way, we can stay backward compatible while adding new fields).

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
```diff
@@ -0,0 +1,64 @@
...
+  { getHeaderParseErrorFormatter :: ErrorFormatter
+  }
+
+defaultHeaderParseErrorFormatter :: HeaderParseErrorFormatter
+defaultHeaderParseErrorFormatter = HeaderParseErrorFormatter defaultErrorFormatter
+
+-- Internal
+
+defaultErrorFormatter :: ErrorFormatter
+defaultErrorFormatter _ _ e = err400 { errBody = cs e }
```
jkarni, line 64: I think it's a bit misleading to say this is the defaultErrorFormatter. This is more like `default400ErrorFormatter`. 400 might be thrown by more combinators than any other status code, but it's still not really a default.

servant-server/src/Servant/Server.hs
jkarni, line 142: I think it might be a good idea to have `serveWithContext p foo == serveWithContextOld p (foo ++ defaultErrorFormatters)` (`++` is a not-yet-existent version of the list concatenation but for contexts; and serverWithContextOld is the serveWithContext that we currently have).

Or, in words: I think the `serveWithContext` we expose should cons whatever context it's given to the default context (which has error formatters). That way it stays backwards compatible (and will continue to if we keep doing things in this direction); but since it picks the first context, you can still override the behaviour.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
```diff
@@ -0,0 +1,64 @@
...
+
+import           Servant.API.ContentTypes
+import           Servant.Server.Internal.Context
+import           Servant.Server.Internal.ServerError
+
+-- | 'Context' that contains default formatters for all error types.
+--
+-- Default formatters will just return HTTP 400 status code with error
+-- message as response body.
+type DefaultErrorFormatters = '[BodyParseErrorFormatter, URLParseErrorFormatter, HeaderParseErrorFormatter]
```
jkarni, line 19: Would it mabe not make sense to have:


```
data ErrorConfig = ErrorConfig 
   { bodyParseErrorFormatter :: ErrorFormatter
   , ...
  }
```
And then have that be the only thing in the context? (As I mentioned in another context in this PR, the constructor would not be exposed except in an `Internal` module, and only the record fields would. In this case additionally a `defaultErrorConfig :: ErrorConfig` would also be exposed.) It seems like that would help us with backwards-compatibility in the future.

maksbotan:
Thanks for your nice comments, @jkarni!

I will read them thoroughly tomorrow and react.

epicallan:
@maksbotan  thanks for putting this together, I will leave a comment about this on `https://github.com/epicallan/servant-errors` once this gets merged. `Servant-errors` will then only make sense for older versions of Servant.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
maksbotan: Okay! I didn't think that usecase was still popular :)

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
maksbotan: Hmm, actually I think `Request` is the best way to go, I'll make the change.

servant-server/src/Servant/Server.hs
maksbotan: Nice idea! I guess depending on something like `vinyl` for this functionality would be too heavy for `servant`?

I'll roll my own `++` here.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
maksbotan: I wanted to make different error types independent, following the second option from your initial proposal. I do not have a strong preference though.

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
maksbotan: Done.

maksbotan:
@jkarni I believe I've reacted to all of your comments.

maksbotan:
CI fails because of `servant-quickcheck`, it will also have to be updated.

fisx:
please don't feel disencouraged if i don't respond within a few days!  i really want to get this merged, i'll try to make some time for it over the week-end.

maksbotan:
No problem, @fisx!

maksbotan:
@jkarni @fisx ping?

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
jkarni: Would you be okay implementing the single-datatype approach then? I think the difference in how many breaking changes we'll have with one vs the other is pretty significant. (Sorry for the extra work!)

cabal.project
```diff
@@ -32,7 +32,7 @@ packages:
   -- doc/cookbook/jwt-and-basic-auth/
   doc/cookbook/pagination
   -- doc/cookbook/sentry
-  doc/cookbook/testing
+--  doc/cookbook/testing
```
jkarni, line 35: Why this?

jkarni:
Aside from my comment on DefaultErrorFormatters, I'm :+1: on this. Sorry for the delay!
I can likely fix servant-quickcheck this weekend

cabal.project
maksbotan: Because of `servant-quickcheck` :) I hoped that CI will pas with this commented, but it fails anyway. I'll revert.

servant-server/src/Servant/Server.hs
```diff
@@ -126,13 +145,20 @@ import           Servant.Server.Internal
 -- > main :: IO ()
 -- > main = Network.Wai.Handler.Warp.run 8080 app
 --
-serve :: (HasServer api '[]) => Proxy api -> Server api -> Application
+serve :: (HasServer api DefaultErrorFormatters) => Proxy api -> Server api -> Application
 serve p = serveWithContext p EmptyContext
 
-serveWithContext :: (HasServer api context)
+-- | Like 'serve', but allows you to pass custotm context.
```
jkarni, line 151:
```suggestion
-- | Like 'serve', but allows you to pass custom context.
```

servant-server/src/Servant/Server.hs
```diff
@@ -126,13 +145,20 @@ import           Servant.Server.Internal
...
+--
+-- 'defaultErrorFormatters' will always be appended to the end of the passed context,
+-- but if you pass your own formatter, it will override the default one.
+serveWithContext :: (HasServer api (context .++ DefaultErrorFormatters), HasContextEntry (context .++ DefaultErrorFormatters) NotFoundErrorFormatter)
```
jkarni, line 155: Ah, I had thought that we could get away with a `HasServer api context` signature here. But I'm now seeing how the blanket is to short, and if we try to cover this, something else is going to get uncovered (in particular, HasServer will need to variations, or all the class methods will have to change).

maksbotan: Yes, unfortunately looks like there is no sane way to have `HasServer api context` here...

maksbotan:
@jkarni done!

jkarni:
I'm trying to figure out how to make CI happy here and in `servant-quickcheck` work together. I need CPP there, because it exposes the context. But there's still no version bump here, so I can't CPP on version. @phadej any ideas?

maksbotan:
Well, can we bump version here without making a release, update servant-quickcheck using source-repository-package, then update constraint on servant-quickcheck here and finally make a release?

maksbotan:
I suppose that we'll have to go through some intermediate inconsistent state anyway...

servant-server/example/greet.hs
```diff
@@ -17,8 +17,35 @@ import           Network.Wai.Handler.Warp
...
+  case handleAcceptH (Proxy :: Proxy '[JSON]) accH value of
+    Nothing -> err400 { errBody = cs err }
+    Just (ctypeH, body) -> err400
+      { errBody = body
+      , errHeaders = [("Content-Type", cs ctypeH)]
+      }
+
+notFoundFormatter :: NotFoundErrorFormatter
+notFoundFormatter req =
+  err404 { errBody = cs $ "Not found path: " <> rawPathInfo req }
```
fisx, line 41:
```suggestion
  err404 { errBody = cs $ "Path not found: " <> rawPathInfo req }
```

fisx:
(Hum.  I just approved that, but with caveats.  Not sure that's a great habit.)

maksbotan:
Thanks. I'll try writing the cookbook, should not be hard.

Please ping me before merging anyway, I want to prettify the history.

maksbotan:
Thanks for your patch, it's nice :)

Sadly, `servant-quickcheck` problem is still no solved :(

```
src/Servant/QuickCheck/Internal/QuickCheck.hs:43:31: error:
    • Could not deduce (Servant.Server.Internal.Context.HasContextEntry
                          (ctx
                           Servant.Server.Internal.Context..++ Servant.Server.Internal.ErrorFormatter.DefaultErrorFormatters)
                          Servant.Server.Internal.ErrorFormatter.ErrorFormatters)
        arising from a use of ‘serveWithContext’
```

Here `ctx` is unknown and thus GHC can't find `HasContextEntry` instance. And it's right: we don't know here if there are formatters in `ctx` and it would be wrong to pick default ones.

maksbotan:
Hi @fisx!

I've rebased and added a cookbook entry on this feature.

But `servant-quickcheck` problem is still unresolved.

fisx:
> But `servant-quickcheck` problem is still unresolved.

I got the tests pass with this patch:

```
~/src/servant-quickcheck$ git diff
diff --git a/src/Servant/QuickCheck/Internal/QuickCheck.hs b/src/Servant/QuickCheck/Internal/QuickCheck.hs
index 2a1b678..ee40f11 100644
--- a/src/Servant/QuickCheck/Internal/QuickCheck.hs
+++ b/src/Servant/QuickCheck/Internal/QuickCheck.hs
@@ -10,7 +10,9 @@ import qualified Network.HTTP.Client      as C
 import           Network.Wai.Handler.Warp (withApplication)
 import           Servant                  (Context (EmptyContext), HasServer,
-                                           Server, serveWithContext)
+                                           Server, serveWithContext, HasContextEntry, type (.++))
+import           Servant.Server.Internal.ErrorFormatter
+                                          (DefaultErrorFormatters, ErrorFormatters)
 import           Servant.Client           (BaseUrl (..), Scheme (..))
 import           System.IO.Unsafe         (unsafePerformIO)
 import           Test.Hspec               (Expectation, expectationFailure)
@@ -37,7 +39,7 @@ withServantServer api = withServantServerAndContext api EmptyContext
 -- application.
 --
 -- /Since 0.0.0.0/
-withServantServerAndContext :: HasServer a ctx
+withServantServerAndContext :: (HasServer a ctx, HasContextEntry (ctx .++ DefaultErrorFormatters) ErrorFormatters)
   => Proxy a -> Context ctx -> IO (Server a) -> (BaseUrl -> IO r) -> IO r
 withServantServerAndContext api ctx server t
   = withApplication (return . serveWithContext api ctx =<< server) $ \port ->
```

Am I missing something?

maksbotan:
Hm, looks like it!

fisx:
> Hm, looks like it!

looks like i'm missing something, or looks like this is a fix?  if the former: could you give me a hint?  :)

maksbotan:
Sorry, I meant that your diff is OK :)

maksbotan:
@fisx ? :)

fisx:
oh, is it my turn?  you're saying it's ready to merge?

maksbotan:
I would say it's ready, yes.

I propose that we merge, bump version, update `servant-quickcheck`, then make a new PR here re-enabling testing cookbook. What do you say?

servant-server/src/Servant/Server/Internal/ErrorFormatter.hs
```diff
@@ -0,0 +1,79 @@
...
+mkContextWithErrorFormatter :: forall (ctx :: [*]). Context ctx -> Context (MkContextWithErrorFormatter ctx)
+mkContextWithErrorFormatter ctx = ctx .++ (defaultErrorFormatters :. EmptyContext)
+
+-- Internal
+
+err400Formatter :: ErrorFormatter
+err400Formatter _ _ e = err400 { errBody = cs e }
+
+-- These definitions suppress "unused import" warning.
+-- The imorts are needed for Haddock to correctly link to them.
```
fisx, line 73: Just curious: 

- can we give this module an explicit export list, and not export these?
- why `undefined`, and not `Proxy`?  :)

maksbotan: Re explicit list: I was going to but probably forgot :)

maksbotan:
Thanks a lot, @fisx!

maksbotan:
What's your suggestion on the new version? By the rule "if it breaks consumers, bump major" it should be 0.18, but on the other hand, all main servant-* packages have synchronized versions, and it would be strange to bump e.g. `servant` as well.

I'll prepare a PR with a new version, then a PR to `servant-quickcheck` with some CPP.

fisx:
> What's your suggestion on the new version? By the rule "if it breaks consumers, bump major" it should be 0.18, but on the other hand, all main servant-* packages have synchronized versions, and it would be strange to bump e.g. `servant` as well.
> 
> I'll prepare a PR with a new version, then a PR to `servant-quickcheck` with some CPP.

if in doubt, just bump!  so i would say it's servant-0.18 now?

cool that you want to do this, thanks!  i'll take another look at this tomorrow morning CEST.

erewok:
Thanks for working on this and nice work, @maksbotan and @fisx.  I'm looking forward to using this feature!

PR #1314, opened 2020-06-16
union verbs

### LINKED ISSUE #841: Re-engineer Verb
The idea is that

Instead of having 

```haskell
a :> b :> Verb c
```

we could have

```haskell
a :> b :> Verb (c :< d)
```

which is transformed into something like

```haskell
a -> b -> m (c, d)
```

`Verb` glues RHS and LHS together. This way we can 

- Have checked exceptions (here `"date"` endpoint cannot error!)

    ```haskell
    type MyAPI = "date" :> Verb (Full Day)
            :<|> "time" :> Capture TimeZone :> Verb (ServantErr :< Full ZonedTime)
    ```

- Get rid of `{-# OVERLAPPING #-}` instances, as in

    ```haskell
    type NoContentAPI = "tick" :> Verb NoContent
    ```

  `NoContent` doesn't overlap `Full a`. Similarly we can have `Stream a` and
  `Headers ... :< ...`.

---

The devil in the details how to fit this into `Router` and `DelayedIO`
framework we have, but I don't see why it's not possible.

The names of auxiliary class `HasServer'` and `Full` combinator are open to
bikesheding.

---

Comments?

fisx:
Fixes #841 

TODO:

- [x] do not depend on `DerivingVia` ("old" ghcs don't have it).
- [x] should `Servant.Client.UVerb` go to `servant-client-core`?
- [x] add https://github.com/haskell-servant/servant-uverb/blob/4c5d20cdbbea585f69e92351a5264a96c7c62b33/example/Main.hs somewhere
- [x] transform the above Main.hs into tests
- [x] add link to the above Main.hs to haddocks, in lieu of real documentation (for now).
- [x] add contents of https://github.com/haskell-servant/servant-uverb/blob/4c5d20cdbbea585f69e92351a5264a96c7c62b33/README.md to haddocks.
- [x] what is the difference between other-extensions and default-extension (servant.cabal)?  can we clean this up further?
- [x] https://github.com/zerobuzz/servant/pull/1#issuecomment-648784387

TODO elsewhere, outside this PR:
- [ ] is there a better place for [these instances](https://github.com/zerobuzz/servant/blob/e8d2767c9782022cea9cd58ccf6b0773f2daadf1/servant-client/test/Servant/ClientTestUtils.hs#L88-L93)?
- [ ] add https://github.com/haskell-servant/servant-uverb/blob/4c5d20cdbbea585f69e92351a5264a96c7c62b33/servant-uverb-mock/src/Servant/Mock/UVerb.hs to servant-mock: https://github.com/haskell-servant/servant-mock/pull/25
- [ ] add https://github.com/haskell-servant/servant-uverb/blob/4c5d20cdbbea585f69e92351a5264a96c7c62b33/servant-uverb-swagger/src/Servant/Swagger/UVerb.hs to servant-swagger: https://github.com/haskell-servant/servant-swagger/pull/123
- [ ] empty servant-uverb repo, leave a link to the new location(s) there in description and README.
- [ ] release everything to hackage.
- [ ] documentation (beyond tests and example, i mean).
- [ ] servant-foreign
- [ ] *Correct* UVerb support for `RunStreamingClient`, `ClientF` (add `acceptStatus` argument like for `RunClient` class)


maksbotan:
Hi. I would like to hear your opinion about my idea.

Since `UVerb` is (probably) will mostly be used to error-like responses, it may be desirable to be able to early abort handler, like with current servant one would use `throwError` with `ServerError`.

```haskell
newtype UVerbT xs m a = UVerbT { unUVerbT :: ExceptT (Union xs) m a }
  deriving newtype (Functor, Applicative, Monad, MonadTrans)

-- | Deliberately hide 'ExceptT's 'MonadError' instance to be able to use
-- underlying monad's instance.
instance MonadError e m => MonadError e (UVerbT xs m) where
  throwError = lift . throwError
  catchError (UVerbT act) h = UVerbT $ ExceptT $
    runExceptT act `catchError` (runExceptT . unUVerbT . h)

-- | This combinator runs 'UVerbT'. It applies 'respond' internally, so the handler
-- may use the usual 'return'.
runUVerbT :: (Monad m, HasStatus x, IsMember x xs) => UVerbT xs m x -> m (Union xs)
runUVerbT (UVerbT act) = either id id <$> runExceptT (act >>= respond)

-- | Short-circuit 'UVerbT' computation returning one of the response types.
throwUVerb :: (Monad m, HasStatus x, IsMember x xs) => x -> UVerbT xs m a
throwUVerb = UVerbT . ExceptT . fmap Left . respond
```

Example usage:

```haskell
h :: Handler (Union '[Foo, WithStatus 400 Bar])
h = do
  when (something bad) $
    throwUVerb $ WithStatus @400 Bar

  when (really bad) $
    throwError $ err500

  -- a lot of code here...

  return $ Foo 1 2 3
```

erewok:
@maksbotan I like that. I have to admit that I always get really confused with errors: do I want `throwE`, `throwError`, or something else? I am in favor of anything simple and ergonomic. 

Still, I don't tend to think of returning a 400 as "throwing" anything. I handle bad requests all the time and the server is operating fine, so in my mind these are not exceptions. With this in mind, I actually would _not_ recommend making it seem like it's `throw`-ing an error. I'd probably just call it something like `uresponse` because I think `uresponse $ WithStatus @400 Bar` would be easy to teach to my coworkers. 

This is a drive-by opinion, though, by someone who's never bothered to learn all the error/exception stuff in Haskell, so feel free to dismiss.

maksbotan:
Well, I made this for the case when you are doing some validation in your handlers, and let's say there are a lot of cases you need to check and all of them result in some kind of `400` response. So for me it's convenient to have short-circuiting way to return these 400 responses without having to have a lot of nested `case`s.

Anyway, naming is a very subjective matter. I do not have strong preference for `throw*` stuff, but I do think that it goes well along the other short-circuiting stuff.

fisx:
@maksbotan I also like your idea, but I would (slightly) prefer to move it to a separate PR, in the hope this will get merged faster.  Do you have an opinion?

fisx:
Hm...  Shouldn't `runUVerbT` return `Union (a ': xs)`, though?

maksbotan:
Sure, I wasn't suggesting including it into this PR :) It's just that I was experimenting with `servant-uverb` and came up with this.

Re `Union (a ': xs)`: that's actually a nice idea! Let's you separate "main" response from "error" ones.

arianvp:
Having an ExceptT-wrapper for this sounds neat.    I wouldn't want to integrate it into the uverb-mechanism directly though as I personally want to get away from the idea that  "status codes" are "exceptional" and should be "short circuiting" as i do not think that is the case in general.     It should be something a person can opt in to by wrapping in a `UVerbT` .  Anyhow it's good food for thought.  I also see some interesting parallels with  extensible-effect libraries.

But I totally agree having the option to short-circuit is extremely convenient.  Scotty for example has   `raiseStatus` and `respondStatus` for short-circuiting and non-short-circuiting

fisx:
> Having an ExceptT-wrapper for this sounds neat. I wouldn't want to integrate it into the uverb-mechanism directly though as I personally want to get away from the idea that "status codes" are "exceptional" and should be "short circuiting" as i do not think that is the case in general. It should be something a person can opt in to by wrapping in a `UVerbT` . Anyhow it's good food for thought. I also see some interesting parallels with extensible-effect libraries.
> 
> But I totally agree having the option to short-circuit is extremely convenient. Scotty for example has `raiseStatus` and `respondStatus` for short-circuiting and non-short-circuiting

Note that `UVerbT` as sketched here is in no way forcing you to use it.  We can add it to `Servant.Server.UVerb` and clearly mark it as voluntary and optional in the haddocks, and people can decide whether to use it or not on a per-handler basis.

arianvp:
Yep completely agreed. I just wonder if servant is the right place for this .. This feels like a sort of type that must be available somewhere in some library already? If it's not; I wonder why

As said; I see some eer-y parallels with extensible effects libraries

I think that e.g. with the `extensible-effects` library you can model something like this already:

https://hackage.haskell.org/package/extensible-effects-5.0.0.1/docs/Control-Eff-Exception.html
```
myAction :: Eff '[Exc Unauthorised, Exc NotFound] User
myAction = do
  throwError Unauthorised
  throwError NotFound
  pure User 
```

Or in `fused-effects` http://hackage.haskell.org/package/fused-effects-1.0.2.2/docs/Control-Effect-Throw.html


I'm not sure if we should hook into a specific effects library for this; as servant has always been MTL-based so far... but it feels odd to reinvent this behaviour if it's somewhere else already

maksbotan:
I'm sorry for stirring the confusion!
I do not at all insist that this belongs to `servant`.

arianvp:
Don't feel sorry!  It sounds like a super useful pattern. It is _at the least_ worth mentioning in whatever tutorial we come up for with this,

arianvp:
I accidentally pressed the "ready-for-review" button because I have fat fingers. and I cant undo it. sorry @fisx :)

fisx:
> I'm sorry for stirring the confusion!
> I do not at all insist that this belongs to `servant`.

I took the liberty of adding it to the cookbook and making you the author of the commit: 084941a3d918b98e0bada3a99a30506e63d12c2b

I hope that was ok?  Thanks again for dumping the idea here, I think that was very helpful!  :)

(@arianvp feel free to add your thoughts on which libraries could be used for that.)

maksbotan:
Thanks @fisx!

maksbotan:
By the way, what's the status of this one?

fisx:
> By the way, what's the status of this one?

it's all in the description.  :)

i'll try to talk to @arianvp about this today, perhaps we can merge it as well, and then make the ultimate release.

fisx:
Who wants to take a final look?  @arianvp?

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -74,6 +82,22 @@ instance FromForm Person
 instance Arbitrary Person where
   arbitrary = Person <$> arbitrary <*> arbitrary
 
+instance MimeRender PlainText Person where
+  mimeRender _ = LazyByteString.fromStrict . encodeUtf8 . Text.pack . show
```
maksbotan, line 86: I guess you could use `Data.ByteString.Lazy.Char8.pack . show` here.

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -74,6 +82,22 @@ instance FromForm Person
...
+
+instance MimeRender ctype a => MimeRender ctype (WithStatus _status a) where
```
maksbotan, line 88: Shouldn't these instances be in the library itself?

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -74,6 +82,22 @@ instance FromForm Person
...
+  mimeRender contentTypeProxy (WithStatus a) = mimeRender contentTypeProxy a
+
+instance MimeUnrender ctype a => MimeUnrender ctype (WithStatus _status a) where
+  mimeUnrender contentTypeProxy input =
+    WithStatus <$> mimeUnrender contentTypeProxy input
+
+instance MimeUnrender PlainText Person where
+  mimeUnrender _ =
+    -- This does not handle any errors, but it should be fine for tests
+    Right . read . Text.unpack . decodeUtf8 . LazyByteString.toStrict
```
maksbotan, line 98: Likewise, `read . Data.ByteString.Lazy.Char8.unpack`

servant-server/src/Servant/Server/UVerb.hs
```diff
@@ -0,0 +1,103 @@
...
+    Proxy (UVerb method contentTypes as) ->
+    Context context ->
+    Delayed env (Server (UVerb method contentTypes as)) ->
+    Router env
+  route _proxy _ctx action = leafRouter route'
+    where
+      method = reflectMethod (Proxy @method)
+      route' env request cont = do
+        let accH :: AcceptHeader -- for picking the content type.
+            accH = AcceptHeader $ fromMaybe ct_wildcard $ lookup hAccept $ requestHeaders request
```
maksbotan, line 80: AFAIR, there is now a `getAcceptHeader` function that does this, defined in `Servant.Server.Internal`.

servant-server/src/Servant/Server/UVerb.hs
```diff
@@ -0,0 +1,103 @@
...
+
+        runAction action' env request cont $ \(output :: Union as) -> do
+          let encodeResource :: (AllCTRender contentTypes a, HasStatus a) => Identity a -> K (Status, Maybe (LBS, LBS)) a
+              encodeResource (Identity res) =
+                K
+                  ( statusOf $ mkProxy res,
+                    handleAcceptH (Proxy @contentTypes) accH res
+                  )
+              pickResource :: Union as -> (Status, Maybe (LBS, LBS))
+              pickResource = collapse_NS . cmap_NS (Proxy @(IsServerResource contentTypes)) encodeResource
```
maksbotan, line 97: Can you use your `collapseUResp` here?

servant-client/test/Servant/ClientTestUtils.hs
fisx: But that would work only for ASCII, not UTF8, right?

I would prefer to leave it as is, because it's more obvious that it works, and because I think it's not important.  Let me know if I'm missing anything.

maksbotan: Hm, you are right, sorry!

servant-client/test/Servant/ClientTestUtils.hs
fisx: a8734ea20b516d5fd5f37a95ffec58a12bfbd36d

servant-server/src/Servant/Server/UVerb.hs
fisx: f34df7c0c16d64d63e85411824f73cc8521b065c

servant-server/src/Servant/Server/UVerb.hs
fisx: 8a413c28c6fab09c0e62723e5f0847e2bbee527b

servant-client/test/Servant/ClientTestUtils.hs
fisx: sorry, wrong commit hash.  it's this one: a8734ea20b516d5fd5f37a95ffec58a12bfbd36d

thanks for the review!  :)

servant/src/Servant/API/UVerb.hs
```diff
@@ -0,0 +1,123 @@
...
+-- | Convenience function to extract a union element using 'cast', ie. return the value if the
+-- selected type happens to be the actual type of the union in this value, or 'Nothing'
+-- otherwise.
+--
+-- See also: 'collapseUResp'.
+extractUResp ::
+  forall (a :: *) (as :: [*]).
+  (All Typeable as, Typeable a, IsMember a as) =>
+  Union as -> Maybe a
+extractUResp = collapse_NS . cmap_NS (Proxy @Typeable) (K . cast . runIdentity)
```
arianvp, line 123: WE shouldn't use typeable here? it sounds very hacky. We have all the type level info already.

servant/src/Servant/API/UVerb.hs
```diff
@@ -0,0 +1,123 @@
...
+  mimeUnrender contentTypeProxy input =
+    WithStatus <$> mimeUnrender contentTypeProxy input
+
+-- FUTUREWORK:
+-- @type Verb method statusCode contentTypes a = UVerb method contentTypes [WithStatus statusCode a]@
+-- no, wait, this is not the same.  this would mean people would have to use 'respond' instead
+-- of 'pure' or 'return'.
+data UVerb (method :: StdMethod) (contentTypes :: [*]) (as :: [*])
+
+type Union = NS Identity
```
arianvp, line 101: Instead of `Identity` can we use `I` here? https://hackage.haskell.org/package/sop-core-0.5.0.1/docs/Data-SOP-BasicFunctors.html#t:I

servant/src/Servant/API/UVerb/OpenUnion.hs
```diff
@@ -0,0 +1,100 @@
+{-# LANGUAGE ConstraintKinds #-}
```
arianvp, line 1: This is missing a copyright notice that was present in the original prototype

servant/src/Servant/API/UVerb/OpenUnion.hs
```diff
@@ -0,0 +1,100 @@
...
+where
+
+import Data.SOP.Constraint
+import Data.SOP.NS
+import Data.Type.Bool (If)
+import Data.Type.Equality (type (==))
+import GHC.TypeLits
+
+
+-- * Stuff stolen from 'Data.WorldPeace" but for generics-sop
```
arianvp, line 29: Ah the notice is still there. but it's incomplete :)

arianvp:
This introduces `TypeApplication` in the servant codebase; which we didn't use before this.

This means we need to either remove it, or adjust our `base` bounds to drop support for GHC < 8.0

servant/src/Servant/API/UVerb.hs
arianvp: This is just a _very_ complicated way of 

``
match  :: NS f as -> Maybe (f a)
```

which is part of the `IsMember` typeclass ;)

File: servant-client-core/src/Servant/Client/Core/HasClient.hs [hunk 10]
```diff
@@ -7,6 +8,7 @@
 {-# LANGUAGE PolyKinds             #-}
 {-# LANGUAGE RankNTypes            #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TypeApplications      #-}
```
arianvp, line 11: This requires `GHC >= 8.0` which would mean we'd have to adjust cabal bounds; or drop the use of `TypeApplications`

servant/src/Servant/API/UVerb.hs
```diff
@@ -0,0 +1,123 @@
...
+-- | Convenience function to apply a function to an unknown union element using a type class.
+-- All elements of the union must have instances in the type class, and the function is
+-- applied unconditionally.
+--
+-- See also: 'extractUResp'.
+collapseUResp ::
+  forall (c :: * -> Constraint) (a :: *) (as :: [*]).
+  All c as =>
+  Proxy c -> (forall x. c x => x -> a) -> Union as -> a
+collapseUResp proxy render = collapse_NS . cmap_NS proxy (K . render . runIdentity)
```
arianvp, line 112: This can also be replaced with a `match`

servant-client-core/src/Servant/Client/Core/HasClient.hs
arianvp: Though `sop-core` makes heavy use of `TypeApplications` so perhaps we are past this anyway...

servant/src/Servant/API/UVerb/OpenUnion.hs
```diff
@@ -0,0 +1,100 @@
...
+type IsMember (a :: u) (as :: [u]) = (CheckElemIsMember a as, UElem a as (RIndex a as))
+
+type family Contains (as :: [k]) (bs :: [k]) :: Constraint where
+  Contains '[] _ = ()
+  Contains (a ': as) bs = (IsMember a bs, Contains as bs)
+
+data Nat_ = S_ Nat_ | Z_
+
+
+-- | TODO: we probably can make do without this.
```
arianvp, line 41: This TODO should be resolved.  the `Index` part is never used anywhere.  We can get away with:

```haskell
class Elem x xs where
  inject :: f x -> NS f xs
  eject :: NS f xs -> Maybe (f x)

-- base case. we found `x` in `xs`
instance Elem x (x ': xs) where
  inject x = Z x
  eject = \case
    Z x -> Just x
    _ -> Nothing

-- recursive case: We did not find `x` but it might be in `xs`
instance Elem x xs => Elem x (y ': xs) where
  inject = S . inject
  eject = \case
    Z _ -> Nothing
    S z -> eject z
```


This is also now generic enough that we can probably upstream it to `sop-core` as the typeclass:

```
class NSElem (x :: k) (xs :: [k]) where
  injection :: Injection f xs x
  ejection ::  Ejection f xs x
```

@kosmikus  what do you think?  Is such a typeclass useful in `sop-core` ? it seems a bit weird to keep it in `servant` but it's ok for now

arianvp: the pair `inject,eject` form a `Prism`   even . Not sure what to do with that observation :)

arianvp: Ah it actually is a prism in world-peace :) http://hackage.haskell.org/package/world-peace-1.0.2.0/docs/src/Data.WorldPeace.Union.html#UElem

arianvp: From their source code there is a note:

```
-- This is safe because of the 'RIndex' contraint. This 'RIndex' constraint
-- tells us that there /actually is/ an @a@ in @as@ at index @i@.
```
so perhaps I am indeed missing something and it _is_ needed. Type level programming is confusing

arianvp: Without the `RIndex` this complains about overlapping instances. so perhaps it _is_ needed. Let me give it a bit more thought tonight :)

arianvp: Note: I deleted some comments above to avoid confusion. The above code I'll add to the PR to simplify things tomorrow morning :)

arianvp:
I'm trying to build with @dmjio 's nix shell, but I cant get a valid build plan on this PR for some reason:

```
cabal build servant
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cookbook-basic-auth-0.1 (user goal)
[__1] trying: base-4.13.0.0/installed-4.1... (dependency of
cookbook-basic-auth)
[__2] trying: aeson-1.5.2.0 (dependency of cookbook-basic-auth)
[__3] trying: cookbook-uverb-0.0.1 (user goal)
[__4] trying: swagger2-2.6 (dependency of cookbook-uverb)
[__5] trying: QuickCheck-2.14.1 (dependency of swagger2)
[__6] next goal: servant-swagger (dependency of cookbook-uverb)
[__6] rejecting: servant-swagger-1.1.8, servant-swagger-1.1.7.1 (conflict:
aeson==1.5.2.0, servant-swagger => aeson>=1.4.2.0 && <1.5)
[__6] rejecting: servant-swagger-1.1.7 (conflict: aeson==1.5.2.0,
servant-swagger => aeson>=1.4.1.0 && <1.5)
[__6] rejecting: servant-swagger-1.1.6, servant-swagger-1.1.5 (conflict:
aeson==1.5.2.0, servant-swagger => aeson>=0.11.2.0 && <1.5)
[__6] rejecting: servant-swagger-1.1.4, servant-swagger-1.1.3.1,
servant-swagger-1.1.3, servant-swagger-1.1.2.1 (conflict: aeson==1.5.2.0,
servant-swagger => aeson>=0.11.2.0 && <1.3)
[__6] rejecting: servant-swagger-1.1.2 (conflict: aeson==1.5.2.0,
servant-swagger => aeson>=0.11.2.0 && <1.1)
[__6] rejecting: servant-swagger-1.1.1, servant-swagger-1.1 (conflict:
aeson==1.5.2.0, servant-swagger => aeson>=0.11.2.0 && <0.12)
[__6] rejecting: servant-swagger-1.0.3, servant-swagger-1.0.2 (conflict:
base==4.13.0.0/installed-4.1..., servant-swagger => base>=4.7 && <4.9)
[__6] rejecting: servant-swagger-1.0.1, servant-swagger-1.0 (conflict:
base==4.13.0.0/installed-4.1..., servant-swagger => base>=4.8 && <4.9)
[__6] rejecting: servant-swagger-0.1.2 (conflict: swagger2==2.6,
servant-swagger => swagger2>=1 && <2)
[__6] rejecting: servant-swagger-0.1.1, servant-swagger-0.1 (conflict:
swagger2==2.6, servant-swagger => swagger2>=1.0 && <2)
[__6] trying: servant-swagger-0.0.0.1
[__7] next goal: servant (user goal)
[__7] rejecting: servant-0.18 (conflict: servant-swagger => servant>=0.4 &&
<0.5)
[__7] rejecting: servant-0.17, servant-0.16.2, servant-0.16.1,
servant-0.16.0.1, servant-0.16, servant-0.15, servant-0.14.1, servant-0.14,
servant-0.13.0.1, servant-0.13, servant-0.12.1, servant-0.12, servant-0.11,
servant-0.10, servant-0.9.1.1, servant-0.9.1, servant-0.9.0.1, servant-0.9,
servant-0.8.1, servant-0.8, servant-0.7.1, servant-0.7, servant-0.6.1,
servant-0.6, servant-0.5, servant-0.4.4.7, servant-0.4.4.6, servant-0.4.4.5,
servant-0.4.4.4, servant-0.4.4.3, servant-0.4.4.2, servant-0.4.4,
servant-0.4.3.1, servant-0.4.3, servant-0.4.2, servant-0.4.1, servant-0.4.0,
servant-0.2.2, servant-0.2.1, servant-0.2, servant-0.1 (constraint from user
target requires ==0.18)
[__7] fail (backjumping, conflict set: servant, servant-swagger)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: swagger2, QuickCheck, base, aeson,
base-compat, servant-swagger, servant, cookbook-basic-auth, cookbook-uverb
Try running with --minimize-conflict-set to improve the error message.

```

arianvp:
I tried `ghc-8.8.1` now but also no luck.  Any instructions on how to build this would be appreciated. I'm surprised CI isn't catching this?

```
cabal build servant
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cookbook-basic-auth-0.1 (user goal)
[__1] trying: base-4.13.0.0/installed-4.1... (dependency of
cookbook-basic-auth)
[__2] trying: aeson-1.5.2.0 (dependency of cookbook-basic-auth)
[__3] trying: cookbook-uverb-0.0.1 (user goal)
[__4] trying: swagger2-2.6 (dependency of cookbook-uverb)
[__5] trying: QuickCheck-2.14.1 (dependency of swagger2)
[__6] next goal: servant-swagger (dependency of cookbook-uverb)
[__6] rejecting: servant-swagger-1.1.8, servant-swagger-1.1.7.1 (conflict:
aeson==1.5.2.0, servant-swagger => aeson>=1.4.2.0 && <1.5)
[__6] rejecting: servant-swagger-1.1.7 (conflict: aeson==1.5.2.0,
servant-swagger => aeson>=1.4.1.0 && <1.5)
[__6] rejecting: servant-swagger-1.1.6, servant-swagger-1.1.5 (conflict:
aeson==1.5.2.0, servant-swagger => aeson>=0.11.2.0 && <1.5)
[__6] rejecting: servant-swagger-1.1.4, servant-swagger-1.1.3.1,
servant-swagger-1.1.3, servant-swagger-1.1.2.1 (conflict: aeson==1.5.2.0,
servant-swagger => aeson>=0.11.2.0 && <1.3)
[__6] rejecting: servant-swagger-1.1.2 (conflict: aeson==1.5.2.0,
servant-swagger => aeson>=0.11.2.0 && <1.1)
[__6] rejecting: servant-swagger-1.1.1, servant-swagger-1.1 (conflict:
aeson==1.5.2.0, servant-swagger => aeson>=0.11.2.0 && <0.12)
[__6] rejecting: servant-swagger-1.0.3, servant-swagger-1.0.2 (conflict:
base==4.13.0.0/installed-4.1..., servant-swagger => base>=4.7 && <4.9)
[__6] rejecting: servant-swagger-1.0.1, servant-swagger-1.0 (conflict:
base==4.13.0.0/installed-4.1..., servant-swagger => base>=4.8 && <4.9)
[__6] rejecting: servant-swagger-0.1.2 (conflict: swagger2==2.6,
servant-swagger => swagger2>=1 && <2)
[__6] rejecting: servant-swagger-0.1.1, servant-swagger-0.1 (conflict:
swagger2==2.6, servant-swagger => swagger2>=1.0 && <2)
[__6] trying: servant-swagger-0.0.0.1
[__7] next goal: servant (user goal)
[__7] rejecting: servant-0.18 (conflict: servant-swagger => servant>=0.4 &&
<0.5)
[__7] rejecting: servant-0.17, servant-0.16.2, servant-0.16.1,
servant-0.16.0.1, servant-0.16, servant-0.15, servant-0.14.1, servant-0.14,
servant-0.13.0.1, servant-0.13, servant-0.12.1, servant-0.12, servant-0.11,
servant-0.10, servant-0.9.1.1, servant-0.9.1, servant-0.9.0.1, servant-0.9,
servant-0.8.1, servant-0.8, servant-0.7.1, servant-0.7, servant-0.6.1,
servant-0.6, servant-0.5, servant-0.4.4.7, servant-0.4.4.6, servant-0.4.4.5,
servant-0.4.4.4, servant-0.4.4.3, servant-0.4.4.2, servant-0.4.4,
servant-0.4.3.1, servant-0.4.3, servant-0.4.2, servant-0.4.1, servant-0.4.0,
servant-0.2.2, servant-0.2.1, servant-0.2, servant-0.1 (constraint from user
target requires ==0.18)
[__7] fail (backjumping, conflict set: servant, servant-swagger)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: swagger2, QuickCheck, base, aeson,
base-compat, servant-swagger, servant, cookbook-basic-auth, cookbook-uverb
Try running with --minimize-conflict-set to improve the error message.


```

servant/src/Servant/API/UVerb.hs
arianvp: This is exactly `cfoldMap_NS :: forall c proxy f xs m. All c xs => proxy c -> (forall a. c a => f a -> m) -> NS f xs -> m `  which is in `Data.SOP.NS`     

just like `foldMap` it avoids traversing twice.  

we can literally `s/collapseUresp/cfoldMap_NS` everywhere 

https://hackage.haskell.org/package/sop-core-0.5.0.1/docs/Data-SOP-NS.html#v:cfoldMap_NS

servant-server/src/Servant/Server/UVerb.hs
```diff
@@ -0,0 +1,96 @@
...
+            mkProxy _ = Proxy
+
+        runAction action' env request cont $ \(output :: Union as) -> do
+          let encodeResource :: (AllCTRender contentTypes a, HasStatus a) => a -> (Status, Maybe (LBS, LBS))
+              encodeResource res =
+                ( statusOf $ mkProxy res,
+                  handleAcceptH (Proxy @contentTypes) (getAcceptHeader request) res
+                )
+              pickResource :: Union as -> (Status, Maybe (LBS, LBS))
+              pickResource = collapseUResp (Proxy @(IsServerResource contentTypes)) encodeResource
```
arianvp, line 90: `collapseUResp` can be replaced with https://hackage.haskell.org/package/sop-core-0.5.0.1/docs/Data-SOP-NS.html#v:cfoldMap_NS here

doc/cookbook/uverb/UVerb.lhs
```diff
@@ -0,0 +1,206 @@
...
+```
+main :: IO ()
+main = do
+  putStrLn . cs . encodePretty $ toSwagger (Proxy @API)
+  _ <- async . Warp.run 8080 $ serve (Proxy @API) (fisx :<|> arian)
+  threadDelay 50000
+  mgr <- Client.newManager Client.defaultManagerSettings
+  let cenv = mkClientEnv mgr (BaseUrl Http "localhost" 8080 "")
+  result <- runClientM (fisxClient True) cenv
+  print $ collapseUResp (Proxy @Show) show <$> result
```
arianvp, line 126: Can be replaced with https://hackage.haskell.org/package/sop-core-0.5.0.1/docs/Data-SOP-NS.html#v:cfoldMap_NS

doc/cookbook/uverb/UVerb.lhs
```diff
@@ -0,0 +1,206 @@
...
+  print $ extractUResp @FisxUser <$> result
```
arianvp, line 127: Can be replaced with `OpenUnion.eject`

servant/src/Servant/API/UVerb.hs
```diff
@@ -0,0 +1,123 @@
...
+instance MimeRender ctype a => MimeRender ctype (WithStatus _status a) where
+  mimeRender contentTypeProxy (WithStatus a) = mimeRender contentTypeProxy a
+
+instance MimeUnrender ctype a => MimeUnrender ctype (WithStatus _status a) where
+  mimeUnrender contentTypeProxy input =
+    WithStatus <$> mimeUnrender contentTypeProxy input
+
+-- FUTUREWORK:
+-- @type Verb method statusCode contentTypes a = UVerb method contentTypes [WithStatus statusCode a]@
+-- no, wait, this is not the same.  this would mean people would have to use 'respond' instead
```
arianvp, line 97: Do we want to adjust this comment to be more enlightening?  

If you `newtype Verb` instead of `type Verb` you can write a `Monad` instance right? so no need for `respond` instead of `pure`

doc/cookbook/uverb/UVerb.lhs
fisx: I'm wondering if we should still keep the names, just make them aliases?  I'm unsure what I think, and you'll probably disagree?

arianvp: I find the current names more confusing as there is nothing called `UResp` in the library

I think `match` or eject are better names for `extractUresp`


Also because collapseUresp has the same type signature `foldMap` I think keeping the name that `sop-core` uses makes sense?

arianvp: But yeh maybe we can  re-export it  as `collapse`

arianvp: I think we can get away with just using `match` in all the cookbook examples no?  The foldMap stuff is indeed a bit intimidating but I don't think it's needed to be shown off.

maksbotan:
Hi @fisx! Is there any progress on this? I would love to see UVerbs in servant master :)

servant/src/Servant/API/UVerb.hs
```diff
@@ -0,0 +1,123 @@
...
+  statuses _ = statusOf (Proxy :: Proxy a) : statuses (Proxy :: Proxy as)
+
+newtype WithStatus (k :: Nat) a = WithStatus a
+  deriving (Eq, Show, GHC.Generic)
+
+instance (GHC.Generic (WithStatus n a), ToJSON a) => ToJSON (WithStatus n a)
+
+instance (GHC.Generic (WithStatus n a), FromJSON a) => FromJSON (WithStatus n a)
+
+instance MimeRender ctype a => MimeRender ctype (WithStatus _status a) where
```
Taneb, line 88: Are we able to derive this instance (and other related instances) with GeneralizedNewtypeDeriving?

servant/src/Servant/API/UVerb.hs
arianvp: This is still relevant

servant/src/Servant/API/UVerb/OpenUnion.hs
arianvp: @arianvp  will do this

servant-client-core/src/Servant/Client/Core/HasClient.hs
arianvp: @fisx

servant/src/Servant/API/UVerb/OpenUnion.hs
arianvp: @maksbotan what are your thoughts on this? I wrote this quite a while ago; and not sure anymore if it's equivalent with what we have.  I can have a look myself later; but if somebody else could have a look that'd be great

maksbotan: @arianvp looks like your idea works. See here: https://github.com/zerobuzz/servant/compare/uverb...maksbotan:uverb

(feel free to cherry-pick those commits, I've also fixed compilation of the cookbook)

servant/src/Servant/API/UVerb.hs
fisx: 7c16739b4fe978a49dd150d6c163ef5229e6b254

servant/src/Servant/API/UVerb.hs
fisx: 6c53469bcd2f53bf5a8d5728e0b7828846061e07

servant/src/Servant/API/UVerb.hs
fisx: I decided I want the helper to remain here, because it does have a nicer type.  But you're right, the implementation had potential of improvement.  :)  cb008f2b09c042f5ccd9331f47eeef1be1e4b88

doc/cookbook/uverb/UVerb.lhs
fisx: Yeah, so I kept the names, but I'm open to renaming them to something better.  You pick something you like!  I just want to have them defined there.  Even though the definitions are trivial, this discussion is evidence that the sop functions are not trivial to find.  :)

servant/src/Servant/API/UVerb.hs
fisx: Or with DerivingStrategies, good idea.  I tried both, but:

> ‘MimeRender’ is not a unary constraint, as expected by a deriving clause

servant/src/Servant/API/UVerb.hs
fisx: How would the `newtype` look like?  The point of the type alias is to make `Verb` a shortcut for `UVerb`, and to be able to make `Verb`s `HasServer` instance work for `Verb` out of the box, but with `newtype`, you'd have two different types again, so how would you gain anything?

I've extended the comment a little: a024e0bab16da0c5a9ad560ae829a8c3a215a05e

servant-client-core/src/Servant/Client/Core/HasClient.hs
fisx: yes, sop-core, plus I really don't want to have to write:

```haskell
   WithStatus (Foo 3 True) :: WithStatus 201 (Foo Bool)
```

instead of

```haskell
   WithStatus @201 (Foo 3 True)
```

I just tried to make that change, but now I say we drop support for ghc <8.  Any objections?

servant/src/Servant/API/UVerb.hs
maksbotan: You can get away with this:

```haskell
deriving anyclass instance (GHC.Generic (WithStatus n a), ToJSON a) => ToJSON (WithStatus n a)
deriving anyclass instance (GHC.Generic (WithStatus n a), FromJSON a) => FromJSON (WithStatus n a)
deriving newtype instance MimeRender ctype a => MimeRender ctype (WithStatus _status a)
deriving newtype instance MimeUnrender ctype a => MimeUnrender ctype (WithStatus _status a)
```

Requires `GeneralizedNewtypeDeriving`, `DerivingStrategies` and `DeriveAnyClass`.

maksbotan: Oh, and `StandaloneDeriving`.

maksbotan: BTW, actually json stuff is simplified to this:

```haskell
newtype WithStatus (k :: Nat) a = WithStatus a
  deriving (Eq, Show, GHC.Generic)
  deriving newtype (ToJSON, FromJSON)
```

And you don't even need `Generic` constraint here, just rely on GND explicitly using `DerivingStrategies`.

maksbotan: Also also:

```haskell
newtype WithStatus (k :: Nat) a = WithStatus a
  deriving (Eq, Show, GHC.Generic)
  deriving newtype (ToJSON, FromJSON, MimeRender ctype, MimeUnrender ctype)
```

GND solves everything here.

fisx:
Trying to build with cabal now, and I also have no luck.  With cabal 3.2.0.0, ghc 8.8.4, I get a number of entertaining errors, but this is my favorite:

```
Configuring library for hspec-discover-2.7.4..
cabal: ghc-pkg dump failed: dieVerbatim: user error (cabal:
'/home/mf/.ghcup/bin/ghc-pkg' exited with an error:
ghc-pkg: /home/mf/.cabal/store/ghc-8.10.2/package.db/package.cache:
GHC.PackageDb.readPackageDb: inappropriate type (Not a valid Unicode code
point!)
)
```

maksbotan:
That's strange... Maybe you could try nuking your cabal store?

As far as I see, it built successfully on Travis. And I have no problems building with Cabal (via haskell.nix) locally.

fisx:
> That's strange... Maybe you could try nuking your cabal store?

Yes, I wanted to avoid that, but it's probably a good idea...

> As far as I see, it built successfully on Travis. And I have no problems building with Cabal (via haskell.nix) locally.

Ah...  nix!  :)  that probably helps, yes.

@arian can you still reproduce the problem on your machine?

fisx:
(sorry, I pinged the wrong arian.  i meant @arianvp...)

servant/src/Servant/API/UVerb.hs
fisx: For the sake of getting this out the door, I would prefer to not optimize this any further.  It looks tempting, but at least if we start adding DerivingStrategies to the core libraries, we'll have to drop more old releases from support.

Does this make sense?  If you give me any commits that do work with ghc 8.0, I'll probably cherry-pick them.

arianvp (approved):

erewok:
Nice work, @fisk!

bergmark:
This is amazing, thanks!! Error handling was a large gap in servant before this landed.

PR #1324, opened 2020-07-28
Add URI fragment as a separate combinator

### LINKED ISSUE #1322: Add URI Fragment
## Fragment 

I would like to introduce `Fragment` combinator as URI Fragment.

```haskell
data Fragment (a :: *)
    deriving Typeable
```

It represents the URI fragment that quite useful in HTTP, e.g. as HTML anchor, despite it is not a part of REST API spec.

87881364-6ebed680-ca01-11ea-927a-1b699e963760

My implementation also provides following properties:
- Fragment should only have one type parameter (that should fit `ToHttpApiData`/`FromHttpApiData` constraints).
- Fragment should be present only once per API endpoint (implemented via type class and type families).

## Motivation

Consider GitHub, e.g. following URL: https://github.com/haskell-servant/servant/blob/master/servant/src/Servant/API/QueryParam.hs#L46-L50

While the content of the page with and without anchor remains the same, the page markup is different. 

Imagine the situation where I want to introduce:
- server-side rendering for HTML page with similar (highlighting) functionality based on fragment.
- client-side rendering (e.g. custom browser capabilities) with the same motivation.

Would you like to consider PR with this missing part?

## Implementation

Fragment itself has already been implemented. You can check the changes via following link: https://github.com/swamp-agr/servant/commit/546e7d363741d9a91ee9f7b0cea46ffb7482d813

## Usage

Only one fragment per endpoint will be accepted.
```haskell
type PostTrackingApi = "post" :> QueryParam TrackingId :> Fragment Text :> Get '[HTML] Tracking
```

However, if user will specify several fragments per endpoint it will be rejected:
```haskell
>>> type FailExampleAPI = Fragment Bool :> ((Fragment Int :> Get '[HTML] NoContent) :<|> Post '[HTML] NoContent)
>>> instance OnlyOneFragment FailExampleAPI
<interactive>:109:10: error:
    • Only one Fragment allowed per endpoint in api ‘Fragment Bool
                                                     :> ((Fragment Int :> Get '[HTML] NoContent)
                                                         :<|> Post '[HTML] NoContent)’.
```

Thanks,
Andrey

swamp-agr:
## Fragment 

I would like to introduce `Fragment` combinator as URI Fragment.

```haskell
data Fragment' (mods :: [*]) (a :: *)
    deriving Typeable

type Fragment = Fragment' '[Lenient]
```

(updated. see review for more details)

It represents the URI fragment that quite useful in HTTP, e.g. as HTML anchor, despite it is not a part of REST API spec.

87881364-6ebed680-ca01-11ea-927a-1b699e963760

My implementation also provides following properties:
- Fragment should only have one type parameter (that should fit `ToHttpApiData`/`FromHttpApiData` constraints).
- Fragment should be present only once per API endpoint (implemented via type class and type families).
- Fragment should not impact API (i.e. server should return 200, `Nothing` should substitute in case of parsing failed).

## Motivation

Consider GitHub, e.g. following URL: https://github.com/haskell-servant/servant/blob/master/servant/src/Servant/API/QueryParam.hs#L46-L50

While the content of the page with and without anchor remains the same, the page markup is different. 

Imagine the situation where I want to introduce:
- server-side rendering for HTML page with similar (highlighting) functionality based on fragment.
- client-side rendering (e.g. custom browser capabilities) with the same motivation.

## Usage

Only one fragment per endpoint will be accepted.
```haskell
type PostTrackingApi = "post" :> QueryParam TrackingId :> Fragment Text :> Get '[HTML] Tracking
```

However, if user will specify several fragments per endpoint it will be rejected:
```haskell
>>> type FailExampleAPI = Fragment Bool :> ((Fragment Int :> Get '[HTML] NoContent) :<|> Post '[HTML] NoContent)
>>> instance OnlyOneFragment FailExampleAPI
<interactive>:109:10: error:
    • Only one Fragment allowed per endpoint in api ‘Fragment Bool
                                                     :> ((Fragment Int :> Get '[HTML] NoContent)
                                                         :<|> Post '[HTML] NoContent)’.
```

Closes #1322.

## Implementation details

1. [X] Fragment data type
2. [X] HasLink
   1. [X] MkLink
   2. [X] toLink
3. [X] Link
4. [X] Link -> URI
5. [X] [3/3] servant-server: HasServer
   1. [X] ServerT
   2. [X] route
   3. [X] hoistServerWithContext
6. [X] [2/2] servant-client-core
   1. [X] [3/3] HasClient
      1. [X] Client
      2. [X] clientHasRoute
      3. [X] hoistClientMonad
   2. [X] [6/6] RequestF
      1. [X] data type
7. [X] servant-client: defaultMakeClientRequest
8. [X] [4/4] after figuring type families: only one fragment per endpoint!
   1. [X] FragmentNotIn type family
   2. [X] FragmentUnique type family
   3. [X] TypeError on result of (5)
   4. [X] OnlyOneFragment type class
9. [X] [4/4] tests
    1. [X] doctest
    2. [X] [8/8] servant-docs: 
       1. [X] HasDocs
       2. [X] ToFragment
       3. [X] DocFragment
       4. [X] Lens' fragment
       5. [X] Action
       6. [X] defAction
       7. [X] combineAction
       8. [X] golden
    3. [X] [6/6] servant-foreign: HasForeign
       1. [X] extend Url with Maybe Frag
       2. [X] Frag data type
       3. [X] defUrl
       4. [X] lenses
       5. [X] prisms
       6. [X] HasForeign
    4. [X] servant-server:test:ServerSpec
        - [X] Links spec (URI fixed).
        - [X] Client spec.
        - [X] Server spec ("fragment absence should not impact API" part of URI fragment covered as well).

Please point me out if I am missing something.


swamp-agr:
Documentation added in db384303e2ce7b81e57ceec359b66dfb38771d12.

swamp-agr:
- Tests added in eda41d91f8153088a4a3ecd60519d1d24bb918a5:
    - [X] Links spec (URI fixed).
    - [X] Client spec.
    - [X] Server spec ("fragment absence should not impact API" part of URI fragment covered as well).

swamp-agr:
@fisx please provide your feedback on PR!

fisx:
> @fisx please provide your feedback on PR!

I can see you, and I like the idea, but I'll need a little more time.  #1314 goes first, then I'm a bit fuzzy about whose in line next.

I haven't checked, but this seems like something you could provide in a separate package for now?  That would probably be the fastest way into hackage, but I'm definitely interested.

Anybody else with the power of maintainership: feel free to steal this!  :)

swamp-agr:
I did not know about "pipeline" existence. 

- This feature is 100% ready as far as I can see.
- For sure, I will wait as long as needed and will try to keep it up-to-date (in corresponding order).
- Feel free to use check-list from here for #1314. I think, it is partially applicable and might be helpful.

Please let me know, once you or any other contributors are ready to let this feature in.

Thank you!

fisx:
There isn't really a pipeline, except for in my head.  Thanks for the code, it's really appreciated, I'll do my best to merge it as soon as I can!

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -161,6 +161,22 @@ data DocQueryParam = DocQueryParam
...
+  , _fragDesc   :: String -- user supplied
+  } deriving (Eq, Ord, Show)
+
+-- | Combine two Fragments, we can't make a monoid because merging Fragments breaks
+-- the laws.
+--
+-- As such, we invent a non-commutative, left associative operation
+-- 'combineFragment' to mush two together taking the info from the very left.
+combineFragment :: Maybe DocFragment -> Maybe DocFragment -> Maybe DocFragment
+mdocFragment `combineFragment` _ = mdocFragment
```
fizruk, line 178: 1. Picking first occurrence is a valid monoid (see `First` in `Data.Monoid`).
2. I think ``Nothing `combineFragment` Just doc`` should be `Just doc`: if there was no fragment on the left, we should pick the fragment from the right.

```suggestion
-- | There should be at most one `Fragment` per API endpoint.
-- So here we are keeping the first occurrence.
combineFragment :: Maybe DocFragment -> Maybe DocFragment -> Maybe DocFragment
Nothing `combineFragment` mdocFragment = mdocFragment
Just docFragment `combineFragment` _ = Just docFragment
```

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -852,5 +856,53 @@ type HasServerArrowTypeError a b =
...
+instance (HasServer api context, FromHttpApiData a1 )
+#endif
+    => HasServer (Fragment a1 :> api) context where
+  type ServerT (Fragment a1 :> api) m = Maybe a1 -> ServerT api m
+
+  route Proxy context subserver =
+    let parseFragment = join . fmap (either (const Nothing) Just)
+          . join . fmap (either (const Nothing) Just)
+          . fmap (fmap parseQueryParam . T.decodeUtf8')
+          . identifyFragment (not . BC8.null) (crop . rawPathInfo) (crop . rawQueryString)
```
fizruk, line 892: I think instead of `(not . BC8.null)` we should check the presence of `#`. The difference is that `/my/endpoint#` will have an empty fragment while `/my/endpoint` will have no fragment.

servant/src/Servant/API/Fragment.hs
```diff
@@ -0,0 +1,26 @@
...
+                 (Typeable)
+
+-- | Lookup the value associated to the URI fragment
+-- and try to extract it as a value of type @a@.
+--
+-- Example:
+--
+-- >>> -- /post#TRACKING
+-- >>> type MyApi = "post" :> Fragment Text :> Get '[JSON] Tracking
+data Fragment (a :: *)
```
fizruk, line 18: `Fragment` behaves similarly to `QueryParam`, so I think we should add modifiers. I think that we should at least recognise `Lenient`/`Strict` and `Description` as we do for `QueryParam`. I am not sure if `Required` makes sense here, but we could also add `Required`/`Optional`.

servant-docs/src/Servant/Docs/Internal.hs
swamp-agr: Fixed.

servant-server/src/Servant/Server/Internal.hs
swamp-agr: Fixed.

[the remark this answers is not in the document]
swamp-agr: Fixed.

servant/src/Servant/API/Fragment.hs
swamp-agr: - Modifiers added.
- Server restricted to `FoldLenient`.
- Fragment itself set by default to `Lenient` to allow user control of parse errors handling.

swamp-agr:
All comments from Review iteration 1 were resolved three days ago. 

- `Fragment' mods a`  introduced. 
- `Fragment` becomes `Lenient` by default.
- `"#"` and `""` now distinguished.

swamp-agr:
Feature is completed and still waiting for review.
All changes requested by @fizruk implemented month ago. Please provide your review.

@fisx Could we go ahead of UVerb and proceed with this feature?

arianvp:
I was always under the assumption that browsers strip the fragment before sending requests to the server. So I don't see why we would have to handle this in servant.
 Is that assumption wrong?

Your motivating example (Github lines) seems to indeed confirm my suspicion. If I click on the link; the fragment is never sent to the server:
image

Instead. Github parses the fragment in javascript and does the highlighting there.

arianvp:
I guess it's useful for the `HasLink` instances; you you can generate type-safe links to fragments

arianvp:
Indeed HTTP standard makes not mention of fragments in the URI syntax See https://ietf.org/rfc/rfc2616.html#section-3.2.2  and they should be stripped before sending to the server.

I'm ok with having the `Fragment` type merged as a way to generate type-safe links and docs with fragments in them.   But we'd need to make a few changes to this PR to make this spec-complaint.

1. `servant-server` should simply ignore the fragment; and not attempt to parse it. As HTTP-1.1 compliant servers don't know about fragment syntax.  

2. `servant-client` should not include the fragment in the request, as HTT-1.1 compliant clients never include the fragment in the Request URI but strip it on the client-side.

arianvp:
I'm happy to be convinced that is too much of a pedantic take by the way.   Overall the code looks good, and I don't see why we wouldn't merge this.  

@swamp-agr I also agree we can merge this before the Uverb things. They don't seem to influence eachother too much. And I'm happy to rebase the Uverb branch if needed. Lets not get that blocking you.  

Let me know what you think about the "Dont send fragments over the wire" part.  And after that lets get this merged ASAP

swamp-agr:
Hi @arianvp 

> Instead. Github parses the fragment in javascript and does the highlighting there.
> 
> From https://en.wikipedia.org/wiki/URI_fragment:
> > The fragment identifier functions differently to the rest of the URI: its processing is exclusively client-sided with no participation from the web server

Consider,
- Server-side rendering capabilities for HTML 
- and limited JavaScript enabled on the page as well. 
- Under these circumstances, fragment still might be helpful BUT should be reached and processed by server. 
- And this particular case seems to break traditional meaning of URI Fragment,
- i.e. code-highlighting could be recognised on server via fragment.
- Meantime, I realise that the case is extremely specific.

That is why I add `HasClient` and `HasServer` into the scope of changes in order to handle the fragment and allow user to decide what to do with it even when parsing failed on server side.
I realise that this behaviour is not aligned with the standard.

Please let me know what do you think about this specific case? 
If it looks like something really wrong, I will go ahead and restrict `HasClient` and `HasServer` changes as you requested.

Best Regards,
@swamp-agr

maksbotan:
Hello,

I would argue that adding server-side parsing of fragments and client-side sending of them is not the right thing to do, as that contradicts the HTTP spec, and common usage, and behavior of most clients. See also this discussion: https://github.com/OAI/OpenAPI-Specification/issues/1635

Particularly, note that `curl` strips fragment:

```
$ curl -v https://github.com/haskell-servant/servant/blob/master/servant/src/Servant/API/QueryParam.hs#L46-L50
...
> GET /haskell-servant/servant/blob/master/servant/src/Servant/API/QueryParam.hs HTTP/1.1
> Host: github.com
> User-Agent: curl/7.64.1
> Accept: */*
...
```

On the other hand, `ToLink` stuff may be useful to let users denote their usage for fragments for whatever reason.

In other words, I agree with @arianvp: let's leave only `ToLink` part of this PR. What do you think?

swamp-agr:
Hi @arianvp, @maksbotan 
- I absolutely agree that standard should not be violated. 
- You provided enough arguments to make me work around use-case above in another way. 
- I implemented following changes per your requests:

- [X] rebase over `0.18.1`.
- [X] servant-server:
   1. [X] src:
      1. [X] include HTTP spec reference into haddock.
      2. [X] replace current implementation with `skip`. i.e. server should ignore fragment.
      3. [X] restore `Delayed`.
   2. [X] test: servant should ignore `Fragment`.
- [X] servant-client-core:
   1. [X] src:
      1. [X] include HTTP spec reference into haddock.
      2. [X] replace current implementation with `skip`, i.e. client should ignore fragment.
      3. [X] restore `RequestF`.
- [X] servant-client:
   1. [X] src: restore `defaultMakeRequest`.
   2. [X] test: servant should ignore `Fragment`.

servant-client-core/src/Servant/Client/Core/Request.hs
```diff
@@ -52,9 +52,10 @@ import           Network.HTTP.Types
                  (Header, HeaderName, HttpVersion (..), Method, QueryItem,
                  http11, methodGet)
 import           Servant.API
-                 (ToHttpApiData, toEncodedUrlPiece, toHeader, SourceIO)
+                 (SourceIO, ToHttpApiData, toEncodedUrlPiece, toHeader)
```
maksbotan, line 55: Hm, is changing this file really neccessary? Generally, I'd only do formatting in files where I introduce significant changes.

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -65,7 +65,7 @@ import           GHC.Generics
 import           Network.HTTP.Media
                  (renderHeader)
 import           Network.HTTP.Types
-                 (hContentType, renderQuery, statusCode, Status)
+                 (Status, hContentType, renderQuery, statusCode)
```
maksbotan, line 68: Same here.

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -161,6 +161,20 @@ data DocQueryParam = DocQueryParam
...
 
+-- | A type to represent fragment. Holds the name of the fragment and its description.
+--
+-- Write a 'ToFragment' instance for your fragment types.
+data DocFragment = DocFragment
+  { _fragSymbol :: String -- type supplied
+  , _fragDesc   :: String -- user supplied
+  } deriving (Eq, Ord, Show)
+
+-- | There should be at most one `Fragment` per API endpoint.
```
maksbotan, line 172:
```suggestion
-- | There should be at most one 'Fragment' per API endpoint.
```

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -161,6 +161,20 @@ data DocQueryParam = DocQueryParam
...
+-- So here we are keeping the first occurrence.
+combineFragment :: Maybe DocFragment -> Maybe DocFragment -> Maybe DocFragment
+Nothing `combineFragment` mdocFragment = mdocFragment
```
maksbotan, line 175: This is equal (i think) to `combineFragment a b = getFirst $ (First a) <> (First b)`

It's up to you to decide if you want to make this change.

servant/src/Servant/API.hs
```diff
@@ -19,6 +19,8 @@ module Servant.API (
   -- | Retrieving the HTTP version of the request
   module Servant.API.QueryParam,
   -- | Retrieving parameters from the query string of the 'URI': @'QueryParam'@
+  module Servant.API.Fragment,
+  -- | Retrieving parameter from the fragment of the 'URI': @'Fragment'@
```
maksbotan, line 23: Hmm, now you can't /retrieve/ anything, maybe instead write here "Documenting the fragment part of the URI: ..." ?

servant/src/Servant/API/Fragment.hs
```diff
@@ -0,0 +1,31 @@
...
+{-# LANGUAGE PolyKinds          #-}
+{-# LANGUAGE TypeOperators      #-}
+{-# OPTIONS_HADDOCK not-home    #-}
+module Servant.API.Fragment (Fragment, Fragment') where
+
+import           Data.Typeable
+                 (Typeable)
+import           Servant.API.Modifiers
+
+-- | Lookup the value associated to the URI fragment
```
maksbotan, line 12: Same here: you don't lookup anything now as far as I understand, so the doc is misleading.

Furthermore, the modifiers are not needed anymore, right?

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -241,18 +249,53 @@ We might try to factor these our more cleanly, but the type synonyms and type
...
+-- >>> -- type FailAPI = Fragment Bool :> Fragment Int :> Get '[JSON] NoContent
+-- >>> instance OnlyOneFragment FailAPI
+-- ...
+-- ...Only one Fragment allowed per endpoint in api...
+-- ...
+-- ...In the instance declaration for...
+instance OnlyOneFragment (Verb m s ct typ)
+
+type family FragmentUnique api :: Constraint where
+  FragmentUnique (sa :<|> sb)       = Or (FragmentUnique sa) (FragmentUnique sb)
```
maksbotan, line 267: Why `Or` and not `And`? 🤔

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -880,5 +882,31 @@ type HasServerArrowTypeError a b =
...
+-- > server :: Server MyApi
+-- > server = getBooksBy
+-- >   where getBooksBy :: Handler [Book]
+-- >         getBooksBy = ...return all books...
+#ifdef HAS_TYPE_ERROR
+instance ( OnlyOneFragment api, HasServer api context, FromHttpApiData a1
+         , SBoolI (FoldLenient mods)
+         )
+#else
+instance (HasServer api context, FromHttpApiData a1, SBoolI (FoldLenient mods))
```
maksbotan, line 901: `FoldLenient` is not needed anymore, right?

servant/src/Servant/API/Modifiers.hs
```diff
@@ -131,7 +132,8 @@ type RequestArgument mods a =
        (If (FoldLenient mods) (Either Text a) a)
        (Maybe (If (FoldLenient mods) (Either Text a) a))
 
-
+type LenientArgument mods a =
```
maksbotan, line 135: This is not actually used? If so, I think you should just remove it.

maksbotan (commented):
Looks great, thanks! Left a couple of minor comments.

Otherwise I'm in favor of merging this. One thing that bothers me is the change in the `Link` datatype: can we go with a minor release or should it wait for 0.19? cc @fisx @alpmestan.

servant-server/src/Servant/Server/Internal.hs
swamp-agr: Fixed.

servant/src/Servant/API/Modifiers.hs
swamp-agr: Fixed.

servant/src/Servant/API/TypeLevel.hs
swamp-agr: - `Or` here represents difference between several endpoints. 
- The fragment uniqueness check should be applied only per particular API endpoint. Not to the whole API.

servant/src/Servant/API/Fragment.hs
swamp-agr: Fixed.

servant/src/Servant/API.hs
swamp-agr: Fixed.

servant-docs/src/Servant/Docs/Internal.hs
swamp-agr: I left this change as it was before. 
If you insist, I could definitely change it.

servant-docs/src/Servant/Docs/Internal.hs
swamp-agr: Fixed.

servant-client/src/Servant/Client/Internal/HttpClient.hs
swamp-agr: `stylish-haskell` (that was enabled in my configuration) put constructors before the function.
I could disable it and undo changes in imports.

servant-client-core/src/Servant/Client/Core/Request.hs
swamp-agr: `stylish-haskell` (that was enabled in my configuration) put constructors before the function.
I could disable it and undo changes in imports.

swamp-agr:
## Summary from review

- servant-docs: change backticks to apostrophes.
- servant:
    - [X] rephrase haddock section in import.
    - [X] remove `Fragment'`.
    - [X] remove `FoldLenient` usage.
    - [X] remove `LenientArgument`.
    - [X] adopt changes after fixes above.

- several things left (explanations provided):
    - imports.
    - `Or` vs. `And`.
    - `combineFragment`.

servant-client/src/Servant/Client/Internal/HttpClient.hs
maksbotan: Well, I think it makes sense to undo all changes to files where there are no significant changes now. Formatting files you touch for other reasons is fine.

servant/src/Servant/API/TypeLevel.hs
```diff
@@ -241,18 +249,53 @@ We might try to factor these our more cleanly, but the type synonyms and type
...
+
+-- | If fragment appeared in API endpoint twice, compile-time error would be raised.
+--
+-- >>> -- type FailAPI = Fragment Bool :> Fragment Int :> Get '[JSON] NoContent
+-- >>> instance OnlyOneFragment FailAPI
+-- ...
+-- ...Only one Fragment allowed per endpoint in api...
+-- ...
+-- ...In the instance declaration for...
+instance OnlyOneFragment (Verb m s ct typ)
```
maksbotan, line 264: btw, I just realized you need this instance for `UVerb` as well, right?

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -880,5 +882,29 @@ type HasServerArrowTypeError a b =
...
+-- >         getBooksBy = ...return all books...
+#ifdef HAS_TYPE_ERROR
+instance (OnlyOneFragment api, HasServer api context, FromHttpApiData a1)
+#else
+instance (HasServer api context, FromHttpApiData a1)
+#endif
+    => HasServer (Fragment a1 :> api) context where
+  type ServerT (Fragment a1 :> api) m = ServerT api m
+
+  route Proxy context subserver = route (Proxy :: Proxy api) context subserver
```
maksbotan, line 904: You can eta-reduce to have `route _ = route (Proxy :: Proxy api)`, similarly for `hoist`.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -880,5 +882,29 @@ type HasServerArrowTypeError a b =
...
+-- > type MyApi = "books" :> Fragment Text :> Get '[JSON] [Book]
+-- >
+-- > server :: Server MyApi
+-- > server = getBooksBy
+-- >   where getBooksBy :: Handler [Book]
+-- >         getBooksBy = ...return all books...
+#ifdef HAS_TYPE_ERROR
+instance (OnlyOneFragment api, HasServer api context, FromHttpApiData a1)
+#else
+instance (HasServer api context, FromHttpApiData a1)
```
maksbotan, line 899: Hmm... is `FromHttpApiData` actually needed here? Same for `ToHttpApiData` in the client.

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -880,5 +882,29 @@ type HasServerArrowTypeError a b =
...
+#endif
+    => HasServer (Fragment a1 :> api) context where
+  type ServerT (Fragment a1 :> api) m = ServerT api m
```
maksbotan, line 902: You require `OnlyOneFragment api`, shouldn't it be `OnlyOneFragment (Fragment a1 :> api)`? Otherwise you could generate `Fragment a :> (Fragment b :> ...)`, where the thing in `( )` passes the check. Or am I missing something?

servant/src/Servant/API/TypeLevel.hs
maksbotan:
```haskell
type FailApi =
  (Fragment "foo" Text :> Fragment "bar" Int :> Get '[JSON] Int)
  :<|> (Fragment "baz" String) :> Get '[JSON] String)
```

With `Or` your check will pass for this API because it passes for the second branch, ignoring the error in the first one.

Sorry if I'm getting something wrong :)

servant-client/src/Servant/Client/Internal/HttpClient.hs
swamp-agr: Fixed.

servant-client-core/src/Servant/Client/Core/Request.hs
swamp-agr: Fixed.

servant-server/src/Servant/Server/Internal.hs
swamp-agr: Fixed.

servant-server/src/Servant/Server/Internal.hs
swamp-agr: Fixed.

servant-server/src/Servant/Server/Internal.hs
swamp-agr: Fixed.

swamp-agr:
## Summary from review iter. 2

- [X] `Or` to `And` cross-check performed.
- [X] `instance AtLeastOneFragment UVerb`
- [X] simplified HasServer/HasClient instances (eta-reduce)
- [X] removed `From/To-HttpApiData`
- [X] fixed `HasServer`/`HasClient` instances for `FailAPI` example.

servant/src/Servant/API/TypeLevel.hs
swamp-agr: Fixed.

servant/src/Servant/API/TypeLevel.hs
swamp-agr: Fixed.

maksbotan (approved):
OK, now I'm satisfied, thanks!

@fisx @alpmestan are you going to take a final look?

swamp-agr:
@fisx, @fizruk, @arianvp, @maksbotan: Let me share my experience of contributing to `servant` with you: https://an-pro.org/posts/11-servant-fragment.html 

It is a honour for me!
Take care of yourself!

Thanks and Regards,
@swamp-agr

Issue #1212, opened 2019-09-03
Obsolete documentation


nixorn:
https://github.com/haskell-servant/servant/commit/420ebd04759cca34b49b5996b75ad3de15d1b366 I noticed! :smile: 

Update [readme]( https://github.com/haskell-servant/servant/blob/master/servant-client-core/README.md) please.

PR #1335, opened 2020-09-02
Fix servant-docs code sample in README

gege251:
I found a few problems in the code sample of servant-docs, so I fixed them for later users:
- `toSample` function no longer exists
- `DocQueryParam` needs a `ParamKind` argument
- `RQBody` should be `ReqBody`
- added implicit imports for better readability

Related issue:
https://github.com/haskell-servant/servant/issues/1212

### DIFF
--- a/servant-docs/README.md
+++ b/servant-docs/README.md
@@ -2,24 +2,56 @@
 
 ![servant](https://raw.githubusercontent.com/haskell-servant/servant/master/servant.png)
 
-Generate API docs for your *servant* webservice. Feel free to also take a look at [servant-pandoc](https://github.com/mpickering/servant-pandoc) which uses this package to target a broad range of output formats using the excellent **pandoc**.
+Generate API docs for your _servant_ webservice. Feel free to also take a look at [servant-pandoc](https://github.com/mpickering/servant-pandoc) which uses this package to target a broad range of output formats using the excellent **pandoc**.
 
 ## Example
 
 See [here](https://github.com/haskell-servant/servant/blob/master/servant-docs/example/greet.md) for the output of the following program.
 
-``` haskell
+```haskell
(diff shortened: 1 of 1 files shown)


fisx:
Thanks!  I opened an issue that takes this yet a little further: https://github.com/haskell-servant/servant/issues/1336

I was going to merge this PR as a good step in the right direction, but when I copied your `Main.hs` over the example file:

```
example/greet.hs:87:13: error:
    • No instance for (Servant.Docs.Internal.HasDocs
                         (Servant.API.Verbs.Verb
                            'http-types-0.12.3:Network.HTTP.Types.Method.DELETE 200 '[] ()))
        arising from a use of ‘docs’
    • In the expression: docs testApi
      In an equation for ‘greetDocs’: greetDocs = docs testApi
   |
87 | greetDocs = docs testApi
```

What commit did you test this against?

gege251:
I am sorry, I didn't notice that! My IDE auto-formatted the code, but I didn't want to introduce more changes than necessary, so I removed the format related changes, that's when these lines must have slipped out of existence.

I guess the real goal here would be to have running examples somewhere in the repository, and then testing them, and generating README based on them in CI. Is that a good solution? (I am still relatively inexperienced with OSS and library development)

fisx:
> I guess the real goal here would be to have running examples somewhere in the repository, and then testing them, and generating README based on them in CI. Is that a good solution?

Yes, that's what I was trying to suggest in #1336.  I think it may be as simple as copying the code from the README.md to the existing `eamples/greet.hs`, but I haven't looked at it in details.  When I tried, the diff looked non-trivial.

Anyway, your readme compiles and runs, the one from `master` doesn't, so this is an improvement.  :)

If you like, you can give #1336 a shot now, or any time.

Thanks for the fix!

PR #1344, opened 2020-09-28
Fix docs: emptyAPIServer -> emptyServer

brandon-leapyear:
https://github.com/haskell-servant/servant/commit/f5d9983381b96ee1ce03e97e3e1d4982e1c92972


arianvp:
Please don't do this. We don't tolerate hacktoberfest spam. I will report this.

arianvp:
Wow im so sorry @brandon-leapyear i misread . I thought you renamed a function without reason but this actually fixes something. Sorry for the accusation.

PR #1347, opened 2020-10-04
Fix Optional ReqBody' (wrap value into Maybe)

### LINKED ISSUE #1346: Optional ReqBody does not wrap it’s type into Maybe
Here is a minimal reproducible demo that shows the problem: https://gist.github.com/unclechu/347bbc386e0659242de80e8ff38b0ec6 (requires [Nix](https://nixos.org/download.html#nix-quick-install) to run). This demo also shows a trick how to overcome the problem.

Even when I define `ReqBody` with `Optional` mod:
``` haskell
ReqBody' '[Optional, Strict] '[JSON] SomeType
```

In the `Server` type `SomeType` stays the same, without `Maybe` wrap that I would also expect from `QueryParam` for instance.

unclechu:
Closes #1346

Here is a small reproducible demonstration of the fix (requires [Nix](https://nixos.org/download.html#nix-quick-install) to run):
https://gist.github.com/unclechu/77ec314b8651c0c17ea4ac486fdb9804

### DIFF
--- a/servant-server/src/Servant/Server/Internal.hs
+++ b/servant-server/src/Servant/Server/Internal.hs
@@ -32,18 +32,24 @@ module Servant.Server.Internal
   , module Servant.Server.Internal.ServerError
   ) where
 
+import           Control.Applicative
+                 ((<|>))
 import           Control.Monad
                  (join, when)
 import           Control.Monad.Trans
                  (liftIO)
 import           Control.Monad.Trans.Resource
                  (runResourceT)
+import           Data.Bifunctor
+                 (bimap)
 import qualified Data.ByteString                            as B
 import qualified Data.ByteString.Builder                    as BB
 import qualified Data.ByteString.Char8                      as BC8
 import qualified Data.ByteString.Lazy                       as BL
 import           Data.Either
                  (partitionEithers)
+import           Data.Function
+                 ((&))
 import           Data.Maybe
                  (fromMaybe, isNothing, mapMaybe, maybeToList)
 import           Data.Semigroup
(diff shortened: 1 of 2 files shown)


unclechu:
I’m not certain about these two patterns:
``` haskell
(SFalse, STrue,  False) -> return . either (const Nothing) (Just . Right)
(SFalse, SFalse, False) -> return . either (const Nothing) Just
```
But they seem to work as expected. It turned ouit that requests from the tests are having `requestBodyLength request` as `KnownLength 0`. I’m not sure why.

dalpd:
Hi!

Is there something keeping this from going upstream? I was under the impression this was the default behavior (that one would construct a custom type with `'[Optional]` to have optional request bodies) like how `Header` works,  but as pointed out in several places related to the issue this PR aims to resolve that's not the case.

tchoutri:
Yep', there is a conflict to be resolved in this PR. :)

unclechu:
@tchoutri I resolved the conflicts and rebased the change on top of latest master. Also I ran the tests locally with a successful result. Please have a look.

dalpd:
Anything we can do to help maintainers move this PR forward?

tchoutri (approved):

tchoutri:
@unclechu okay, I believe that you should bump the version of HsOpenSSL used, reading its changelog: https://flora.pm/packages/%40hackage/HsOpenSSL/0.11.7.5/changelog

isomorpheme:
I'd like to see this PR merged - is there something I can do to help? I guess if I wanted to rebase it myself I'd have to open a separate PR?

PR #1348, opened 2020-10-12
Loosen bound on base16-bytestring

### LINKED ISSUE #1334: servant-client-core bump base64-bytestring
```
 In the dependencies for servant-client-core-0.18:
     base64-bytestring-1.1.0.0 from stack configuration does not match
 >=1.0.0.1 && <1.1  (latest matching version is 1.0.0.3)
 servant-client-core will need a revision to support latest base64-bytestring

```

serras:
Fixes #1334 

I have checked locally that `servant-client-core` can build with `base16-bytestring` 1.1 without any modifications. According to the PVP I think we could loosen it further to `< 2`, but `< 1.2` works for now.

### DIFF
--- a/servant-client-core/servant-client-core.cabal
+++ b/servant-client-core/servant-client-core.cabal
@@ -77,7 +77,7 @@ library
   build-depends:
       aeson                 >= 1.4.1.0  && < 1.6
     , base-compat           >= 0.10.5   && < 0.12
-    , base64-bytestring     >= 1.0.0.1  && < 1.1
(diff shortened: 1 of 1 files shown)


arianvp (approved):

arianvp:
Thanks!

arianvp:
I should also probably make this into either a  Hackage revision; or publish a new point-release. I'm not too fond of revisions myself; so I'll make sure this gets into a point-release.

PR #1349, opened 2020-10-12
Wrap Application in custom monad for Raw endpoint

maksbotan:
Currently handler `Raw` endpoint for is just an `Application`: a value.
This value has no access to the context of an arbitrary monad that may
be used instead of the default `Handler`.

This commit changes handler type to `m Application`, allowing the user
to do arbitrary actions in `m` prior to generating an `Application`.

This is useful to do custom logic behind `servant`s back while still
having access to the environment of user's monad.

### DIFF
--- a/doc/tutorial/Docs.lhs
+++ b/doc/tutorial/Docs.lhs
@@ -18,6 +18,7 @@ import Data.ByteString.Lazy (ByteString)
 import Data.Proxy
 import Data.Text.Lazy.Encoding (encodeUtf8)
 import Data.Text.Lazy (pack)
+import Control.Monad.IO.Class
 import Network.HTTP.Types
 import Network.Wai
 import Servant.API
@@ -264,9 +265,9 @@ api :: Proxy DocsAPI
 api = Proxy
 
 server :: Server DocsAPI
-server = Server.server3 :<|> return serveDocs where
+server = Server.server3 :<|> serveDocs where
     serveDocs _ respond =
-        respond $ responseLBS ok200 [plain] docsBS
+        liftIO $ respond $ responseLBS ok200 [plain] docsBS
     plain = ("Content-Type", "text/plain")
 
 app :: Application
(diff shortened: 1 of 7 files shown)


maksbotan:
This is a breaking change, so maybe it makes sense to introduce a new combinator, like `RawM`, to have this behavior. However, I think the new behavior is "more right" and the users just have to migrate, the change is not that big (add `return` before your application).

fisx:
I like it, but as you can tell we're a bit behind with processing PRs.  Would you be interested in joining the maintainer team?  :-)

If not, we'll get to this eventually.  I agree with the breaking change, any other opinions?

maksbotan:
Well, maintaining such an important library is a great responsibility...

Can I help with current problems? I'd love to see uverbs finally in master, for example.

maksbotan:
Actually I was able to generalize this even further, please take a look.

amesgen:
JFYI: There is https://github.com/cdepillabout/servant-rawm, which solves this exact problem.

I think it is a very good idea to integrate it/sth similiar in this repo!

In particular, it will make it easier to close cdepillabout/servant-rawm#7, as servant-auth can provide the required instance out of the box (without having to write a servant-auth-rawm compat package).

maksbotan:
Oops, perhaps I should've looked for prior art :)

However, newest version of this PR is actually better than that package, as it allows you to run `m`-based effects depending on the request. I've found it useful in my app.

maksbotan:
@fisx so how about co-maintaining? I could at least help with technical releases for version bounds and stuff.

cc @fizruk.

maksbotan:
I've just released 0.18.1 to Hackage. I believe this change will have to go into 0.19, as it's kinda breaking. What do you think?

amesgen:
I agree that the `Raw` rework should go in 0.19.

One nice feature of servant-rawm is that it [allows to customize the `HasDocs` instance of `RawM'`](https://hackage.haskell.org/package/servant-rawm-1.0.0.0/docs/Servant-RawM.html#t:RawM-39-). This should not be too hard to add to this PR.

PR #1350, opened 2020-10-13
bump "tested-with" ghc versions.

fisx:
I've also collapsed all ghc versions into one line to make it easier to use sed to add another version in the future.


arianvp:
LGTM

PR #1354, opened 2020-11-04
Bump versions

maksbotan:
After we release this on Hackage we can add servant stuff back to Stackage.


fisx (approved):
I would have bumped the bigger version and gone to `0.19`, but this works, too.

Thanks!

PR #1361, opened 2020-11-17
Fix withstatus instances

### LINKED ISSUE #1356: WithStatus is unusable with class-based MimeRender instances
Almost any use of the new UVerb `WithStatus` combinator fails with an `Overlapping instances` error, due to the (in the case of JSON) `MimeRender ctype a => MimeRender ctype (WithStatus status a)` and `ToJSON a => MimeRender JSON a` instances. It's possible to work around this issue by writing an explicit `HasStatus` instance, but this makes UVerbs usage much more verbose.

Due to this issue, the UVerb cookbook page doesn't build at all.

The rest of the new UVerb work is incredible and super helpful, keep up the great work!

arianvp:
Fixes #1356


arianvp:
We should add some test that makes sure at least the cookbook compiles. And probably add a piece of code to the tests that would trigger the overlapping instance. Bit confused that the test-suite didn't catch this oversight

maksbotan:
Looks like travis config was not updated when uverb cookbook was added. Can you please run `haskell-ci regenerate`?

(`haskell-ci` comes from Hackage, you can `cabal new-install` it)

servant/src/Servant/API/UVerb.hs
```diff
@@ -35,6 +36,7 @@ module Servant.API.UVerb
 where
 
 import Data.Aeson (FromJSON, ToJSON)
+import Web.FormUrlEncoded(FromForm, ToForm)
```
arianvp, line 39: Nitpick: indentation

servant/src/Servant/API/UVerb.hs
```diff
@@ -71,18 +73,8 @@ instance (HasStatus a, HasStatuses as) => HasStatuses (a ': as) where
...
-
-instance (GHC.Generic (WithStatus n a), FromJSON a) => FromJSON (WithStatus n a)
-
-instance MimeRender ctype a => MimeRender ctype (WithStatus _status a) where
-  mimeRender contentTypeProxy (WithStatus a) = mimeRender contentTypeProxy a
-
-instance MimeUnrender ctype a => MimeUnrender ctype (WithStatus _status a) where
-  mimeUnrender contentTypeProxy input =
-    WithStatus <$> mimeUnrender contentTypeProxy input
+  deriving stock (Eq, Show, Generic)
```
maksbotan, line 76: this may fail on some old GHC versions we try to support for some reason :)

arianvp: Oh yeh this will fail on `8.0` indeed. Let me remove it for now.

arianvp:
Bit of a thought; but shouldn't we move all these `Status`-related things to `Servant.API.Status` with the rest of the `Status` code? It kind of feels weird to sit in `Servant.API.UVerb` as it has nothing to do with it

servant/src/Servant/API/UVerb.hs
```diff
@@ -70,19 +69,14 @@ instance (HasStatus a, HasStatuses as) => HasStatuses (a ': as) where
...
-instance (GHC.Generic (WithStatus n a), ToJSON a) => ToJSON (WithStatus n a)
-
-instance (GHC.Generic (WithStatus n a), FromJSON a) => FromJSON (WithStatus n a)
-
-instance MimeRender ctype a => MimeRender ctype (WithStatus _status a) where
-  mimeRender contentTypeProxy (WithStatus a) = mimeRender contentTypeProxy a
-
-instance MimeUnrender ctype a => MimeUnrender ctype (WithStatus _status a) where
-  mimeUnrender contentTypeProxy input =
-    WithStatus <$> mimeUnrender contentTypeProxy input
```
fisx, line 85: why did you remove these?

arianvp: Read the commit message and read the issue this PR is trying to fix.

arianvp: (Long story short. These instances already existed; and this instance you defined here was overlapping; which meant code using WithStatus could never compile; but it wasn't caught by the test-suite as we didn't run `haskell-ci regenerate` before merging the uverb PR)

fisx: > why did you remove these?

sorry, found the answer in the commit message.  
38db23b

fisx (approved):
i don't mind this code moving to the status module, otherwise LGTM!

arianvp:
I decided to squash the commits; as they didn't build on their own

servant/src/Servant/API/UVerb.hs
```diff
@@ -49,6 +48,21 @@ class KnownStatus (StatusOf a) => HasStatus (a :: *) where
 statusOf :: forall a proxy. HasStatus a => proxy a -> Status
 statusOf = const (statusVal (Proxy :: Proxy (StatusOf a)))
 
+-- | an instance of this typeclass assigns a HTTP status code to a return type
```
maksbotan, line 51: Small nitpick :)
```suggestion
-- | An instance of this typeclass assigns a HTTP status code to a return type
```

servant/src/Servant/API/UVerb.hs
maksbotan: Are you going to fix this?

maksbotan (approved):
LGTM! Let's wait for what CI has to say :)

arianvp:
It seems many of the compilers we test against do not exist anymore.  Might need to adjust the `Tested-With` things in the `.cabal` files as the CI is now failing for the wrong reasons

servant/src/Servant/API/UVerb.hs
arianvp: Hmmmm I just came up with a simpler solution..

arianvp:
I came up with a better idea. The instance that @fisx wrote actually makes sense; we just need to get rid of the `To/FromJSON` instance for `WithStatus` (as it's unused) and then the overlap should go away as well

doc/cookbook/uverb/uverb.cabal
```diff
@@ -10,7 +10,7 @@ maintainer:          haskell-servant-maintainers@googlegroups.com
 category:            Servant
 build-type:          Simple
 cabal-version:       >=1.10
-tested-with:         GHC==8.4.4, GHC==8.6.5, GHC==8.8.2
+tested-with:         GHC==8.4.4, GHC==8.6.5, GHC==8.8.4, GHC==8.10.1
```
maksbotan, line 13: Rerun `haskell-ci regenerate` :)

arianvp:
I can't figure our why there are two CI builds failing. Nothing seems particularily wrong in the logs.

I tried bumping the GHC versions. Lets see if that changes anything.

maksbotan:
Error is here:

```
1.00s$ if ! $GHCJS ; then ${CABAL} v2-haddock $WITHCOMPILER --with-haddock $HADDOCK ${TEST} ${BENCH} all ; fi
cabal: renderBuildTargetProblem: unexpected status (TargetDisabledByUser,Nothing)
  error, called at ./Distribution/Client/CmdErrorMessages.hs:331:14 in main:Distribution.Client.CmdErrorMessages
The command "if ! $GHCJS ; then ${CABAL} v2-haddock $WITHCOMPILER --with-haddock $HADDOCK ${TEST} ${BENCH} all ; fi" exited with 1.
```

https://travis-ci.org/github/haskell-servant/servant/jobs/744281874#L1276

arianvp:
Looks like a bug in Cabal. Don't really know what to do https://github.com/haskell/cabal/issues/6679

arianvp:
Now only one compiler fails (8.4.4). I again have an extremely hard time parsing the logs again. I don't see the log message for the failure. At least it's not failing in the `haddock` step this time

fisx:
I'm sad that the json instances went away, and everybody will have to add orphans.  But perhaps `WithStatus` is just something people should write themselves to avoid this?  So, fine, no strong objections.

arianvp:
Reread my commit message. There is no need for orphans I think.  Servant only uses the `MimeRender` instance for decoding; not the `To/From` instance directly.

For every `ToJSON a` there is a  `ToJSON a => MimeRender JSON a` instance

and for every `WithStatus a` there is a `MimeRender (WithStatus a)` instance. Ergo:

```
ToJSON a =>  MimeRender JSON (WithStatus a)
```

exists without `WithStatus` having a `ToJSON (WithStatus a)` instance.

No need for orphans

maksbotan:
> Now only one compiler fails (8.4.4). Again with the same error message. Do we consider it a fluke? Lets try reproduce locally perhaps?

My lifehack: open raw log and search for "`exited with 1`". In this case:

```
Configuring executable 'cookbook-uverb' for cookbook-uverb-0.0.1..
Warning: 'license-file: ../../../servant/LICENSE' is a relative path outside of the source tree. This will not work when generating a tarball with 'sdist'.
Warning: The 'license-file' field refers to the file '../../../servant/LICENSE' which does not exist.
Preprocessing executable 'cookbook-uverb' for cookbook-uverb-0.0.1..
Building executable 'cookbook-uverb' for cookbook-uverb-0.0.1..

[;1mUVerb.lhs:15:14: [;1m[31merror:[0m[0m[;1m[0m[0m[;1m Unsupported extension: DerivingVia[0m[0m
[;1m[34m   |[0m[0m
[;1m[34m15 |[0m[0m {-# LANGUAGE [;1m[31mDerivingVia[0m[0m #-}
[;1m[34m   |[0m[0m[;1m[31m              ^^^^^^^^^^^[0m[0m
[0m[0m[0mcabal: Failed to build exe:cookbook-uverb from cookbook-uverb-0.0.1.

travis_time:end:24efbcef:start=1605698235956580326,finish=1605698424533484474,duration=188576904148,event=script
[0K[31;1mThe command "${CABAL} v2-build $WITHCOMPILER ${TEST} ${BENCH} all" exited with 1.[0m
```

I wonder if there is a way to disable building this cookbook for 8.4... Maybe drop it from `cookbook/uverb/*.cabal`? Don't forget `haskell-ci` ;)

maksbotan:
@fisx I think that having only `MimeRender` instances without `ToJSON` instances shows our intent more clearly. Can you think of a situation when someone would need to use `ToJSON` on `WithStatus`?

arianvp:
I noticed that uverb was already pushed in the `0.18.1` releaseNow that we know UVerb didn't even _work_ in `0.18.1` do we want to deprecate that release on hackage and tag a `0.19`  or `0.18.2` instead? http://hackage.haskell.org/package/servant-0.18.1/docs/Servant-API-UVerb.html

maksbotan:
What do you mean didn't work? I'm using 0.18.1 at my work 🤔

arianvp:
Wouldn't any usage trigger an overlapping instance error without this change?

arianvp:
I guess you only run into that issue when you use the `WithStatus` wrapper...  Maybe deprecating is a bit overkill; but uploading `0.18.2` would be nice =)

maksbotan:
Only if you use `WithStatus` in your API. I have hand written `HasStatus` instances and a couple of `deriving HasStatus via WithStatus 200 Foo`. Those work fine.

maksbotan:
Re 0.18.2: I'm ready to upload it, I've contacted @fisx to coordinate efforts, but haven't got a reply yet. I can make a release tomorrow if you don't mind.

arianvp:
It should be fine to tag a release.  I know fisx is a bit busy this week; so might be slow to respond

maksbotan:
Okay, will do tomorrow then.

maksbotan:
@arianvp I've released 0.18.2, btw.

PR #1366, opened 2020-11-24
servant-client-core: depend on 0.18.2

arianvp:
We should make a cabal revision for this too I suppose. Looks like this was an oversight

I don't know how to do this on hackage yet; but would be happy if someone could show me

### DIFF
--- a/servant-client-core/servant-client-core.cabal
+++ b/servant-client-core/servant-client-core.cabal
@@ -64,7 +64,7 @@ library
 
   -- Servant dependencies
   build-depends:
-      servant            >= 0.18.1 && <0.19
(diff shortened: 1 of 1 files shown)


maksbotan:
Oh snap, I must've missed it. I can make Hackage revision if you want me to.

maksbotan (approved):

maksbotan:
How to update it on Hackage:

1. Click this link:

image

2. Click the version you want to change:

image

3. Make a change:

image

4. "Review changes", make sure it's right, then publish.

I've done it now, see here: https://hackage.haskell.org/package/servant-client-core-0.18.2/revisions/

Issue #1367, opened 2020-11-24
WithStatus instances are _still_ overlapping (sorry :( )


arianvp:
From a collegue on `0.18.2`

```
 • Overlapping instances for MimeRender JSON (WithStatus 201 ())
  arising from a use of ‘Servant.serve’
Matching instances:
  two instances involving out-of-scope types
    instance MimeRender ctype a => MimeRender ctype (WithStatus _status a)
    instance [overlappable] ToJSON a => MimeRender JSON a
```

Even without the `ToJSON (WithStatus a)`  instance GHC is still convinced these two overlap.

I am confused with it would think there is an overlap though. Note that there is _no_ `ToJSON (WithStatus 201 ())` instance at all so why it even considers the `  instance [overlappable] ToJSON a => MimeRender JSON a` instance is baffling me.  Maybe the other instance needs an `{-# OVERLAPPING #-}` pragma? Why though?

This is trivially caught if the `uverb` cookbook would call the `serve` function (which it currently doesn't). And I didnt compile the cookbook locally either (as the `cabal.project` file for the cookbooks differs from the one travis generates; and it isn't working).

Action points:

* Make the cookbook fail to compile by calling `serve` on the API type. So that we do not ship these things without proper testing anymore
* Make the cookbooks compile locally again (They currently do not as the `cabal.project` for the cookbooks is self-inconsisent.  CI doesn't mind as it generates its own `cabal.project`)
* Figure out why it overlaps

arianvp:
Apparently instance context is not considered for selection.

```
MimeRender ctype (WithStatus _status a)
MimeRender JSON a
```

and neither are more specific

Taneb:
The downside of that approach is that it necessitates modification of every library that provides a content type for servant to work with `WithStatus`

arianvp:
Indeed a bit of a bummer; but I am afraid there is no way around it.

Note that the `WithStatus` newtype is just there as a convenience; you do not need it to use UVerb.
We should stress this more in the documentation.

You can give your own datatype an instance of `HasStatus`:
```haskell
data MyType = MyType { ... } deriving HasStatus via (WithStatus 300 MyType)

-- or  without derivingvia:
instance HasStatus  MyType where type Status MyType = 300

type MyAPI = UVerb GET  '[MyType, MyOtherType]
```

intolerable:
[...]

PR #1376, opened 2020-12-06
Mess with overlapping instances

fisx:
Fixes https://github.com/haskell-servant/servant/issues/1367, maybe.

Mostly `MimeRender`, but I'm wondering if there are more issues here?

I've tried 2 approaches here (haven't completes either, too much code to touch):

1. Remove `WithStatus`, and leave it to the user to introduce it again (give an example in cookbook).
2. Remove the overlappable `MimeRender` instances that are causing this particular problem.

I don't like either, because they both require the user to write more trivial code, but I haven't thought of anything better.

Option 2 could be made slightly nicer by at least providing simple functions `mime{,Un}Render{JSON,Form,...}`, or by putting them in a separate module `Servant.AwkwardInstances`.

If we can agree on any of this I'm happy to finish the job (including fixing code layout).

### DIFF
--- a/servant/src/Servant/API/UVerb.hs
+++ b/servant/src/Servant/API/UVerb.hs
@@ -28,7 +28,6 @@ module Servant.API.UVerb
     HasStatus (StatusOf),
     statusOf,
     HasStatuses (Statuses, statuses),
-    WithStatus (..),
     module Servant.API.UVerb.Union,
   )
 where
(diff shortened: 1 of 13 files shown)


fisx:
I *now* read the discussion in the issue, and @arianvp your approach indeed seems to work.  Not sure I like it much more than the other two, but it does pass the tests.  :)

servant/src/Servant/API/UVerb.hs
```diff
@@ -105,3 +98,27 @@ instance KnownStatus n => HasStatus (WithStatus n a) where
 -- Backwards compatibility is tricky, though: this type alias would mean people would have to
 -- use 'respond' instead of 'pure' or 'return', so all old handlers would have to be rewritten.
 data UVerb (method :: StdMethod) (contentTypes :: [*]) (as :: [*])
+
+instance {-# OVERLAPPING #-} MimeRender JSON a => MimeRender JSON (WithStatus _status a) where
+  mimeRender contentTypeProxy (WithStatus a) = mimeRender contentTypeProxy a
```
arianvp, line 103: No need for the overlapping pragma.  `MimeRender` is overlappable already

fisx: tried without, didn't' work.  (i know, weird, right?  can you try to reproduce it?  perhaps i bonked it somehow.)

PR #1382, opened 2020-12-10
Bump base64-bytestring limit to 1.3

Philonous:
Compiles and tests ran without error


mimi1vx:
@Philonous  same thing in `servant-client-core.cabal`

PR #1384, opened 2020-12-10
Update upper bound limit for http-client

mimi1vx:
Builds fine with http-client-0.7.3


fisx:
I don't see any error in the travis logs, nor any reason why particular compiler versions should expose an issue in the one-lline change in this PR, so I just merged despite the red CI status.

Thanks!

PR #1386, opened 2020-12-16
Allow bytestring-0.11

Bodigrim:
Tested with 
```
constraints: bytestring >=0.11
allow-newer: *:bytestring

source-repository-package
  type: git
  location: https://github.com/Bodigrim/hspec-wai
  tag: master
```

Since `servant` does not support GHC < 8.0, version bumps are enough.


fisx:
thanks!

PR #1387, opened 2021-01-07
BasicAuth: Allow disabling sending WWW-Authenticate header

agrafix:
Right now, the `BasicAuth` module sends the `WWW-Authenticate` header by default via `servant-server`. This isn't always desirable, since, when called via AJAX, it causes the browser to surface a login prompt on authentication failure.

This PR makes that behavior configurable.


akhesaCaro:
Hi agrafix,

Thanks for your contribution. The CI has been fixed, could you rebase your branch to see if it passes the CI checks?

tchoutri:
@agrafix Hi! Unfortunately we lost CI logs, could you please rebase your branch and push it against? :)

Issue #1315, opened 2020-06-18
Improved composability for Generic APIs


gdeest:
There has been complaints in the past regarding the usability of APIs defined via nested generic products (see #1211 for example).

The problem with generic APIs is that one constantly needs to switch back and forth between generic and “vanilla” servant, at both term- and type-level. This is not am issue for flat APIs that correspond to a tuple of routes, but quickly becomes cumbersome when dealing with deeply nested APIs.

It occurred to me that the conversion machinery could be entirely hidden from the user, by allowing direct embedding of generic APIs via a new Servant combinator:

```haskell
data GenericAPI (routes :: * -> *)
```

For which suitable instances could be defined, directly declaring clients and servers for `GenericApi MyRoutes` as _records_:

```haskell
instance ( … ) => HasServer (GenericApi routes) context where
  type ServerT (GenericApi routes) m = routes (AsServerT m)
  …

instance ( … ) => HasClient m (GenericApi routes) where
  type Client m (GenericApi routes) = routes (AsClientT m)
  …
```

I have taken a stab at it in [this gist](https://gist.github.com/gdeest/27d41541839bae947997bec52c519b51). Please disregard the quality of the code, this has been thrown together in a hurry :)

There is some ugly machinery there, but it works fine with servers. I haven't been able to write the `hoistClientMonad` function for clients though (it looks like we would need additional constraints on the monad type constructors for this to work), but clients in vanilla `ClientM` should work.

Do you think this idea is worth pursuing ? Has this been tried before, or this somehow in violation with the Servant philosophy ?

jkarni:
I think this is a really great idea!

> (it looks like we would need additional constraints on the monad type constructors for this to work),

If you mean a `RunClient` constraint, I don't think we can have it - if we have it, we don't need the method at all! I'm not really sure what to do about that method though.

gdeest:
> If you mean a RunClient constraint, I don't think we can have it - if we have it, we don't need the method at all!

That makes sense.

> I'm not really sure what to do about that method though.

Do you mean that removing `hoistClientMonad` is under consideration ?

Alternatively, we might be able to derive this function automatically with datatype-generic programming. I haven't given it much thought yet.

jkarni:
> Do you mean that removing hoistClientMonad is under consideration ?
I'm no longer sufficiently involved with servant to be making such big decisions!

But what about changing the type of hoistClientMonad to have the constraints?

```
[...]

PR #1388, opened 2021-01-17
Improve API for composing generic routes

gdeest:
**Update (2021-11-17):**

The PR has significantly changed since its inception, so here is an updated example for newcomers that may want to see the current state. Users can now mark part of an API as being *directly* defined by a record using the `NamedRoutes` combinator. Here is an example:

```haskell
type Api = NamedRoutes RootApi

data RootApi mode = RootApi
  { subApi :: mode :- Capture "token" String :> NamedRoutes SubApi
  , hello :: mode :- Capture "name" String :> Get '[JSON] String
  , …
  } deriving Generic

data SubApi mode = SubApi
  { endpoint :: mode :- Get '[JSON] Person
  , …
  } deriving Generic
```

On the surface, it looks exactly the same as what servant's current support for generics provides, just replacing `ToServantApi` by `NamedRoutes`. The nuance here is that `NamedRoutes` is an actual combinator, while `ToServantApi` is type-family application which converts a record type to a “vanilla” servant product using `:<|>`, and thus loses all naming information. Defining `NamedRoutes` as a combinator allows us to define instances for `HasClient`, `HasServer` etc. which define servers / clients **directly as records**.

The most immediate benefits is that nested records of routes actually works, without the need for ugly, error-prone manual conversions between vanilla servant products and records. 

One can serve the above API as easily as:

```haskell
myApp :: Application
myApp = serve (Proxy @Api) server
  where server = 
    RootApi 
      { subApi = \token -> SubApi
          { endpoint = … }
      , hello = \name -> … 
      }
```

Similarly, generating clients produces nested records as needed. A few helpers (initially suggested by @clementd-fretlink ) have been added for convenience:

```haskell
rootClient :: RootApi (AsClientT ClientM)
rootClient = client (Proxy @API)

hello :: String -> ClientM String
hello name = rootClient // hello /: name

endpointClient :: ClientM Person
endpointClient = client // subApi /: "foobar123" // endpoint
```

An additional benefit is that error messages are now *much nicer*.

***END OF UPDATE***

-------

This pull request implements the ideas sketched in #1315 .

It introduces a new combinator:

```haskell
data GApi (routes :: * -> *)
```

which accepts a generic product of routes and can be embedded into a Servant API.

Instances of `HasClient` and `HasServer` are provided for `GApi routes`. The internal machinery requires
the generic routes to be instances of the  `GClient` and `GServer` typeclasses respectively for this to work. ~The necessary code can written (mostly) automatically for us by the compiler~ (EDIT: `deriving (Generic, GClient, GServer)` is now sufficient):

```haskell
data API' as = API
  { bookStore :: as :- "books" :> GApi BookStore
  , apiVersion :: as :- "version" :> Get '[JSON] Version
  }
  deriving (Generic, GClient)

instance
  (HasContextEntry (context .++ DefaultErrorFormatters) ErrorFormatters) =>
  GServer API' context
```

~The need to specify the context of error formatters for `GServer` is a bit unfortunate, I'd like to find a way to fix this.~ EDIT: Done.

Everything is pretty much automatic from there: defining servers and deriving clients works exactly as one would expect, with no added boilerplate and conversion between vanilla / generic servant types.

A complete example program with nested generic APIs can be found [here](https://gist.github.com/gdeest/ae32ab8db1be6e722b9cfd9c5244f583). This should probably be part of the PR as an example / test somewhere ; I am looking for guidance here.

Now, this PR features a change that I expect to be controversial: it removes the `m` parameter of the `HasClient` typeclass, making it more symmetric with `HasServer`, but breaking every third-party implementation.

This parameter is used to be able to constrain the monads for which a client can be defined (to define them in terms of `RunClient` / `RunStreamingClient`). The problem is that `Client m api` then only makes sense if `m` satisfies those constraints, which gets in the way when implementing `hoistClientMonad` with generics (GHC refuses to apply type families any further and cannot prove type equality when converting between vanilla / generic servant). We cannot add those constraints to the source / target monads, as we may want to hoist clients in monads that do not satisfy them. The solution I have found is to define said constraints as an associated type family:

```haskell
  type ClientConstraints (api :: *) (m :: * -> *) :: Constraint
```

The signature of `clientWithRoute` then becomes:

```haskell
  clientWithRoute :: ClientConstraints api m => Proxy m -> Proxy api -> Request -> Client m api
```

### DIFF
--- a/servant-client-core/src/Servant/Client/Core/HasClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/HasClient.hs
@@ -27,6 +27,7 @@ module Servant.Client.Core.HasClient (
     HasClient (..),
     EmptyClient (..),
     AsClientT,
+    (/:),
     foldMapUnion,
     matchUnion,
     ) where
@@ -871,6 +872,37 @@ instance
         toServant @api @(AsClientT ma) clientA
 
 #endif
+
+infixl 1 /:
+
+-- | Convenience function for working with nested record-clients.
+--
+-- Example:
+--
+-- @@
+-- type Api = NamedAPI RootApi
+--
+-- data RootApi mode = RootApi
+--   { subApi :: mode :- NamedAPI SubApi
+--   , …
+--   } deriving Generic
+--
+-- data SubAmi mode = SubApi
+--   { endpoint :: mode :- Get '[JSON] Person
+--   , …
+--   } deriving Generic
+--
+-- api :: Proxy API
+-- api = Proxy
+--
+-- rootClient :: RootApi (AsClientT ClientM)
+-- rootClient = client api
+--
+-- endpointClient :: ClientM Person
+-- endpointClient = client /: subApi /: endpoint
+-- @@
+(/:) :: a -> (a -> b) -> b
+x /: f = f x
 
 
 {- Note [Non-Empty Content Types]
--- a/servant-client-core/src/Servant/Client/Core/HasClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/HasClient.hs
@@ -18,10 +18,15 @@
 #define HAS_TYPE_ERROR
 #endif
 
+#if __GLASGOW_HASKELL__ >= 806
+{-# LANGUAGE QuantifiedConstraints  #-}
+#endif
+
 module Servant.Client.Core.HasClient (
     clientIn,
     HasClient (..),
     EmptyClient (..),
+    AsClientT,
     foldMapUnion,
     matchUnion,
     ) where
@@ -39,6 +44,7 @@ import           Data.ByteString.Builder
 import qualified Data.ByteString.Lazy                     as BL
 import           Data.Either
                  (partitionEithers)
+import           Data.Constraint (Dict(..))
 import           Data.Foldable
                  (toList)
 import           Data.List
@@ -79,7 +85,10 @@ import           Servant.API
                  ReflectMethod (..), RemoteHost, ReqBody', SBoolI, Stream,
                  StreamBody', Summary, ToHttpApiData, ToSourceIO (..), Vault,
                  Verb, WithNamedContext, WithStatus (..), contentType, getHeadersHList,
-                 getResponse, toEncodedUrlPiece, toUrlPiece)
+                 getResponse, toEncodedUrlPiece, toUrlPiece, NamedRoutes)
+import           Servant.API.Generic
+                 (GenericMode(..), ToServant, ToServantApi
+                 , GenericServant, toServant, fromServant)
 import           Servant.API.ContentTypes
                  (contentTypes, AllMime (allMime), AllMimeUnrender (allMimeUnrender))
 import           Servant.API.TypeLevel (FragmentUnique, AtLeastOneFragment)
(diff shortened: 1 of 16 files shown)


gdeest:
I managed to update the branch so that `GServer` can be derived as easily as `GClient`:

```haskell
data BookStore as = BookStore
  { createBook :: as :- ReqBody '[JSON] Book :> Post '[JSON] (ID Book)
  , readBook :: as :- Capture "id" (ID Book) :> Get '[JSON] Book
  }
  deriving (Generic, GClient, GServer)

data API' as = API
  { bookStore :: as :- "books" :> GApi BookStore
  , apiVersion :: as :- "version" :> Get '[JSON] Version
  }
  deriving (Generic, GClient, GServer)

server :: ServerT API ServerMonad
server = API
  { bookStore = BookStore
      { createBook = …
      , readBook = …
      }
  , apiVersion = …
  }

client :: Client ClientM API
client = client (Proxy @API)
```

gdeest:
Any update on this PR ? Is there anything I could do ?

gdeest:
OK, I think this is ready to go.

Summary of changes:

- The refactoring of the `HasClient` class is fortunately not needed ; I still think the occurrence of `m` as a parameter of the typeclass is a design wart, but this issue can at least be discussed independently.
- It is no longer required to manually derived `GServer` and `GClient` ; the (unique) instance is provided by the library, so users only have to derive `Generic`.
- An `HasLink` instance has been added.
- The combinator has been renamed from `GApi` to `NamedRoutes`. This is because the fact that `HasServer` etc. instances depend on `Generic` is an implementation detail that shouldn't, IMHO, be visible in the API type.
- Code has been moved around so that instance of `HasServer`, `HasServer` for the `NamedRoutes` combinator are re-exported from `Servant.Server` and `Servant.Client`. This is more convenient and reduces the risk of users encountering missing instances because they forgot to import `Servant.Server.Generic`, etc.
- A `(/:)` operator has been added to make the use of record-clients nicer.

servant-client-core/src/Servant/Client/Core/HasClient.hs
```diff
@@ -893,10 +898,46 @@ infixl 1 /:
...
+-- api = Proxy
+--
+-- rootClient :: RootApi (AsClientT ClientM)
+-- rootClient = client api
+--
+-- hello :: String -> ClientM String
+-- hello name = rootClient // hello /: name
+--
+-- endpointClient :: ClientM Person
+-- endpointClient = client // subApi /: "foobar123" // endpoint
```
tchoutri, line 937: Awesome!

servant-client-core/src/Servant/Client/Core/HasClient.hs
```diff
@@ -810,6 +810,119 @@ instance HasClient m api => HasClient m (BasicAuth realm usr :> api) where
...
+-- endpointClient :: ClientM Person
+-- endpointClient = client // subApi // endpoint
+-- @@
+(//) :: a -> (a -> b) -> b
+x // f = f x
+
+-- | Convenience function for supplying arguments to client functions when
+-- working with records of clients.
+--
+-- Intended to be use in conjunction with '(//)'.
```
clementd-fretlink, line 894:
```suggestion
-- Intended to be used in conjunction with '(//)'.
```

alpmestan (approved):
- I think we could `unsafeCoerce` our way through proving that e.g `Client m (ToServantApi api) ~ ToServant api (AsClientT m)`, but that probably wouldn't be enough to do away with the need for `GServer`, `GClient` and friends. But at least users never have to worry about any of that now.

- Could we rename the `proof` methods to anything slightly clearer, like perhaps `gserverProof` or whatever in that vein?

- Do you perhaps want to add a tiny teeny paragraph on `NamedRoutes` with a trivial example to the cookbook?

All in all, looks good to me, just need your opinion on the last two bullet points.

Absolutely stellar work, @gdeest. :+1:

akhesaCaro:
:+1: @alpmestan for the cookbook. I am already working on it on another PR.

gdeest:
Thanks @alpmestan !

> Could we rename the proof methods to anything slightly clearer, like perhaps gserverProof or whatever in that vein?

That's entirely reasonable.

> Do you perhaps want to add a tiny teeny paragraph on NamedRoutes with a trivial example to the generic cookbook?

Indeed, that's already planned for the release, but I am gonna ensure that we have a ticket for that before merging.

Issue #1353, opened 2020-11-03
Remove travis and migrate to Github Actions


arianvp:
Travis is dropping support for OSS projects. I have no budget to pay them.

fisx:
not saying we shouldn't, but: https://bugs.chromium.org/p/project-zero/issues/detail?id=2070&can=2&q=&colspec=ID%20Type%20Status%20Priority%20Milestone%20Owner%20Summary&cells=ids

arianvp:
https://github.com/haskell-CI/haskell-ci/pull/416 has landed. This should make our lives a lot easier as we do not have to rewrite our CI from scratch but can keep using the existing CI files!!

gdeest:
Fixed by #1402 .

PR #1396, opened 2021-03-03
Basic GitHub Actions-based CI

gdeest:
The Travis CI has been broken for a while, and it is probably time to fix it so that proper development can continue.

This PR:

- Setups a basic CI based on GitHub actions, with a somewhat limited build matrix.
- Disables cookbook/testing, because servant-quickcheck doesn't build anymore.
- Disables servant-docs on Cabal build, because of some test failures
  - The order of some JSON fields seems to be reversed in the output, need investigation.
- Fix test failures in servant-http-streams when `localhost` points to an IPv6 address rather than 127.0.0.1.

If this PR is merged, I plan on fixing and re-enabling all the disabled tests one-by-one. I think the first step here is to get back to a green CI to prevent further regressions.

Opening as draft for now because it still needs a bit of cleanup. An example run can be witnessed here:

https://github.com/gdeest/servant/runs/2024096343

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,95 @@
+name: CI
+
+# Trigger the workflow on push or pull request, but only for the main branch
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  cabal:
+    name: ${{ matrix.os }} / ghc ${{ matrix.ghc }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
+        cabal: ["3.2"]
+        ghc:
+          - "8.8.4"
+          - "8.10.2"
+        # exclude:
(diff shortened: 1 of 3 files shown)


gdeest:
Noticed the mention of https://github.com/haskell-CI/haskell-ci/pull/416 in #1353 .

I must say I would be in favor of starting with a clean slate rather than inheriting the complexity of the Travis setup.
The goal remaining, of course, feature parity.

Do you have a strong opinion on the matter ?

arianvp:
If we can get away without code generation I'd be happy. People often forget to run haskell-ci when submitting PRs (including me).

 But on the other hand taking the existing .haskell-ci files and simply generating a GitHub Actions file from them sounds like the faster way to success. Given it already works roxay.

I really don't have very strong opinions on the matter. As long as as we keep the ability to build against all supported GHC versions...

gdeest:
Closed in favor of #1402

PR #1397, opened 2021-03-05
Fixing Servant cookbooks part 1 (-testing + sqlite-simple)

akhesaCaro:
The scope of this PR, as the others that will follow, is to fix servant cookbooks.

This PR make the project compiling and propose a first step through that goal.

Here are the tasks that have been made : 

- Fixing the network bug
- Commenting servant-quickcheck cookbook which is trapped into a dependency loop for now (see [this issue](https://github.com/haskell-servant/servant-quickcheck/issues/74))
- Fixing DocSpec failing test
- Bringing sqlite-simple cookbook back to life

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,6 @@ doc/_build
 doc/venv
 doc/tutorial/static/api.js
 doc/tutorial/static/jq.js
-shell.nix
 
 # nix
 result*
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ doc/_build
 doc/venv
 doc/tutorial/static/api.js
 doc/tutorial/static/jq.js
+shell.nix
 
 # nix
 result*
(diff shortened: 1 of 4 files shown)


cabal.project
```diff
@@ -55,8 +55,6 @@ constraints: base-compat ^>=0.11
 constraints: semigroups ^>=0.19
 
 -- MonadFail
```
gdeest, line 57:
```suggestion
```

This line should be removed as well.

.gitignore
```diff
@@ -30,6 +30,7 @@ doc/_build
 doc/venv
 doc/tutorial/static/api.js
 doc/tutorial/static/jq.js
+shell.nix
```
gdeest, line 33: Not strictly part of this PR.

cabal.project
akhesaCaro: :+1:

gdeest (approved):
Looks good ! :)

PR #1398, opened 2021-03-05
Fixing Servant cookbooks part 2 (+hoist-server-with-context, + jwt-and-basic-auth, +sentry, +https)

akhesaCaro:
Following #1397 -> Based on its branch

I may not have the full story here but the following cookbooks are compiling with GHC 8.8.4 :

- hoist-server-with-context
- jwt-and-basic-auth
- sentry
- https

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -29,7 +29,7 @@ packages:
   doc/cookbook/db-sqlite-simple
   doc/cookbook/file-upload
   doc/cookbook/generic
-  -- doc/cookbook/hoist-server-with-context
+  doc/cookbook/hoist-server-with-context
   -- doc/cookbook/https
   -- doc/cookbook/jwt-and-basic-auth/
   doc/cookbook/pagination


doc/cookbook/jwt-and-basic-auth/jwt-and-basic-auth.cabal
```diff
@@ -22,7 +22,7 @@ executable cookbook-jwt-and-basic-auth
                      , servant
                      , servant-client
                      , servant-server
-                     , servant-auth ==0.3.*
+                     , servant-auth
```
gdeest, line 25: I think we'll want to keep the version bound until we move `servant-auth` in the main repository. Could you try with a higher version ? I am guessing that `>= 0.4.0.0" should work.

akhesaCaro: Obviously, that is why I've put this PR back to `draft`.

akhesaCaro:
1 PR per cookbook will be done instead

PR #1402, opened 2021-03-14
Basic GitHub actions-based CI

gdeest:
(New version of #1396 from a Servant repository branch, in order to get GitHub actions actually running here.)

The Travis CI has been broken for a while, and it is probably time to fix it so that proper development can continue.

This PR:

- Setups a basic CI based on GitHub actions, with a somewhat limited build matrix.
- Disables cookbook/testing, because servant-quickcheck doesn't build anymore.
- Disables servant-docs on Cabal build, because of some test failures
  - The order of some JSON fields seems to be reversed in the output, need investigation.
- Fix test failures in servant-http-streams when `localhost` points to an IPv6 address rather than 127.0.0.1.

If this PR is merged, I plan on fixing and re-enabling all the disabled tests one-by-one. I think the first step here is to get back to a green CI to prevent further regressions.

Opening as draft for now because it still needs a bit of cleanup. An example run can be witnessed here:

https://github.com/gdeest/servant/runs/2024096343


gdeest:
For the record, here is what seems to be missing ATM:

- Support for additional GHC versions (it's just a matter of adding them to the build matrix, and potentially fixing associated issues).
- GHCJS tests for the relevant parts of the codebase.
- Doctests.

gdeest:
I noticed the following weirdness in `servant-client` tests:

```
Preprocessing test suite 'readme' for servant-client-0.18.2..
Building test suite 'readme' for servant-client-0.18.2..
Running 1 test suites...
Test suite readme: RUNNING...
Error: ConnectionError (HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 8081
  secure               = False
  requestHeaders       = [("Accept","application/json;charset=utf-8,application/json")]
  path                 = "/books"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.connect: <socket: 3>: does not exist (Connection refused)))
Test suite readme: PASS
Test suite logged to:
/home/runner/work/servant/servant/dist-newstyle/build/x86_64-linux/ghc-8.8.4/servant-client-0.18.2/t/readme/noopt/test/servant-client-0.18.2-readme.log
```

This may be another IPv6 / network-related issue. But I am puzzled by the fact that the test suite is marked as “passing” when it is clearly failing. Needs investigation.

gdeest:
We are hitting: https://github.com/sol/doctest/issues/119

By when doctests the dumb way. There doesn't seem to be an easy way around this (though I am unsure why current setup doesn't seem affected).

I'll give a stab at it using `cabal-doctest`, and writing dedicated test suites for the doctests. This seems to be the simplest way to invoke GHC with _just_ the required dependencies.

`PackageImports` would work as well, but this would be quite an obtrusive change to the codebase for the sole purpose of running tests.

cabal.project
```diff
@@ -3,7 +3,8 @@ packages:
   servant-client/
   servant-client-core/
   servant-http-streams/
-  servant-docs/
+-- Tests failing with Cabal (TODO: investigate)
+--  servant-docs/
```
dfithian, line 7: FWIW I was able to get tests to pass by testing twice, with each field.
https://github.com/haskell-servant/servant/pull/1401/files#diff-1d3cfc9328ae9300f3c558d76c2bab79a5eee646734eac7500f828cd318275f1

gdeest:
The Travis CI works around the ambiguous module problem with doctests with:

```
doctest-filter-packages: base-compat-batteries
```

in `cabal.haskell-ci`. It gives me hope that this seems to be the only package requiring this ; I may be able to reproduce what current `.travis.yml` is doing easily in the Github setup without going down the custom setup / dedicated test-suite road (even though I think this might ultimately be better).

cabal.project
gdeest: That's good to know ! There is a pending PR ( #1397 ) that fixes this problem in a slightly different way. We'll ponder what the best approach is.

gdeest:
The doctests were actually only failing on MacOS, which happened to run faster than Ubuntu and cancelled the entire build.

Disabling the MacOS builds for now since the Travis CI doesn't have them anyway, and we're only aiming for feature-parity at this point.

gdeest:
The `UVerb` cookbook is triggering a failure on GHC < 8.6 because of a usage of `DerivingVia` (which was introduced in GHC 8.6.1).

We will modify the cookbook not to use this extension.

gdeest:
CI is fixed, and (AFAIK) has feature-parity. Since the Travis setup is basically broken, I'll merge now.

A few trivial changes were required:

- New cookbooks did not compile with older GHC versions, due to to the use of extensions that weren't available at that time (`DerivingVia`, `DerivingStrategies`) or to changes in the structure of `base` (missing imports for `Data.Monoid` / `Data.Semigroup`).
- A test was fixed that involved a problem when `localhost` is resolved to `::1` instead of `127.0.0.1`.

A `servant-docs` test was commented out for now. It is inherently buggy, as it relies on the order in which `aeson` will render JSON object fields, but `aeson` uses a `HashMap` from `unordered-containers` for representing objects. A fix is proposed in #1397.

Things to improve on:

- Performance: CI is currently taking in the order of ~20 minutes. There is caching in place, but the cache doesn't seem to always be restored.
- GHCJS setup. We're still using hvr's PPA which doesn't seem to be updated anymore (we're stuck with ghcjs-8.4). We could try to use Nix / `reflex-platform`.
- General ugliness / hacksinherited from Travis CI.

PR #1409, opened 2021-04-05
Update for GHC 9.0.1

maksbotan:


akhesaCaro:
Maybe you should take a look at this : https://gitlab.haskell.org/ghc/ghc/-/issues/19460

maksbotan:
I'm going to release this to Hackage tomorrow, in order to put Servant back to LTS nightly, which has recently switched to GHC 9.

Fellow maintainers, please speak up if there are any objections.

akhesaCaro:
Some tasks have been forgotten in this PR : 

- [ ] nix-shell + its README to provide the new GHC version
- [ ] Cabal fails to build the project over GHC9.0.1. 
```
servant-auth-server-0.2.2.0 (has the same characteristics that caused the
previous version to fail: excludes 'base' version 4.15.0.0)
[__3] fail (backjumping, conflict set: base,
cookbook-hoist-server-with-context, servant-auth-server)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: base, servant-auth-server,
cookbook-basic-auth, cookbook-hoist-server-with-context
```
- [ ] Exclude GHC version (on the CI) for some cookbooks instead of excluding them at all.

I am on it...

Issue #1411, opened 2021-04-10
Fixing cookbooks


akhesaCaro:
Some cookbooks have been commented because they didn't compile anymore with new versions of GHC or dependencies.

**Cookbooks to fix**

- [ ] hoist-server-with-context
-- PR : #1131 
- [X] https
-- Commenting PR : #1224
-- Fixing PR : https://github.com/haskell-servant/servant/pull/1412
- [ ] jwt-and-basic-auth
-- PR : #1131 
- [ ] sentry
-- PR : #1131 
- [ ] testing
-- PR : #1402
-- Issue https://github.com/haskell-servant/servant-quickcheck/issues/74
- [X] uverb  
-- Fixing PR #1407
- [ ] open-id-connect  
-- PR : #1182

## HTTPS 
- Commenting PR : #1224 
- Errors with GHC 8.0.2 :

``` haskell
[1 of 1] Compiling Network.Wai.Handler.WarpTLS ( Network/Wai/Handler/WarpTLS.hs, dist/build/Network/Wai/Handler/WarpTLS.o )
cabal: Failed to build warp-tls-3.2.8 (which is required by exe:cookbook-https

Error:     • Constructor ‘Connection’ does not have the required strict field(s): connHTTP2
from cookbook-https-0.1). See the build log above for details.

    • In the expression:
        Connection
          {connSendMany = TLS.sendData ctx . L.fromChunks,
           connSendAll = sendall, connSendFile = sendfile, connClose = close',
           connFree = freeBuffer writeBuf, connRecv = recv ref,
           connRecvBuf = recvBuf ref, connWriteBuffer = writeBuf,
           connBufferSize = bufferSize}
      In an equation for ‘conn’:
          conn ctx writeBuf ref
            = Connection
                {connSendMany = TLS.sendData ctx . L.fromChunks,
                 connSendAll = sendall, connSendFile = sendfile, connClose = close',
                 connFree = freeBuffer writeBuf, connRecv = recv ref,
                 connRecvBuf = recvBuf ref, connWriteBuffer = writeBuf,
                 connBufferSize = bufferSize}
            where
                sendall = TLS.sendData ctx . L.fromChunks . return
                sendfile fid offset len hook headers
                  = readSendFile
                      writeBuf bufferSize sendall fid offset len hook headers
                close' = void (tryIO sendBye) `finally` TLS.contextClose ctx
                sendBye
                  = handleJust
[...]

PR #1131, opened 2019-02-18
Rename ServantError to ClientError, ServantErr to ServerError

phadej:

PR #1413, opened 2021-04-10
enabling hoist-server-with-context cookbook

akhesaCaro:
Related to https://github.com/haskell-servant/servant/issues/1411

Enabling hoist-server-with-context cookbook

Compability with GHC 8.0.2 has been dropped because `servant-auth-server-0.4.4.0` isn't compatible with it anymore. See : haskell-servant/servant-auth@90d14b4

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -29,7 +29,7 @@ packages:
   doc/cookbook/db-sqlite-simple
   doc/cookbook/file-upload
   doc/cookbook/generic
-  -- doc/cookbook/hoist-server-with-context
+  doc/cookbook/hoist-server-with-context
   doc/cookbook/https
   -- doc/cookbook/jwt-and-basic-auth/
   doc/cookbook/pagination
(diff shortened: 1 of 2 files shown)


akhesaCaro:
With GHC 8.0.2

```haskell
src/Servant/Auth/Server/Internal/ThrowAll.hs:5:48: error:
Error:     Module ‘Servant’ does not export ‘ServantErr’
cabal: Failed to build servant-auth-server-0.1.0.1 (which is required by
exe:cookbook-hoist-server-with-context from
cookbook-hoist-server-with-context-0.0.1).

Error: Process completed with exit code 1.
```

- servant-auth-server 0.4.4.0 or higher is compatible with this PR : https://github.com/haskell-servant/servant/pull/1131/files
- servant-auth-server needs to be at least 0.4.4.0, but it has a constraint on servant that needs GHC 8.2.1. So both aren't compatible with GHC 8.0.2
- This PR has dropped the compatibility with 8.0.2 , but why? haskell-servant/servant-auth@90d14b4

gdeest (approved):
LGTM

gdeest:
@akhesaCaro You may want to clean history a bit. I am not sure why it is showing your changes as a merge commit…

PR #1414, opened 2021-04-11
WIP : enabling jwt-and-basic-auth coobook

akhesaCaro:
enabling jwt-and-basic-auth coobook


akhesaCaro:
Is also broken by `servant-auth-server` 
-> see #1413 

GHC 8.0.2 must be excluded.

PR #1419, opened 2021-04-29
Add MimeRender instances for Headers

### LINKED ISSUE #1373: UVerb doesn't understand Headers
`Headers` types can't be included in UVerb unions as there's no equivalent to the `HasServer (Verb meth st cts (Headers h a))` instance for UVerb.

pcapriotti:
This should fix #1373.

/cc @akshaymankar


pcapriotti:
This makes it possible to compile UVerb APIs with `Headers`, but does not actually generate response headers. I created a new pull request (#1420) that does implement header generation. Closing this one.

PR #1420, opened 2021-05-07
Add response header support to UVerb

### LINKED ISSUE #1373: UVerb doesn't understand Headers

pcapriotti:
This implements response header generation for UVerb. It uses an internal `HasResponseHeaders` class with a general overlappable instance.

This should fix #1373.

### DIFF
--- a/servant-client/test/Servant/SuccessSpec.hs
+++ b/servant-client/test/Servant/SuccessSpec.hs
@@ -32,6 +32,7 @@ import           Data.Foldable
 import           Data.Maybe
                  (listToMaybe)
 import           Data.Monoid ()
+import           Data.SOP (NS (..), I (..))
 import           Data.Text
                  (Text)
 import qualified Network.HTTP.Client                as C
@@ -128,6 +129,14 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
       case res of
         Left e -> assertFailure $ show e
         Right val -> getHeaders val `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
+
+    it "Returns headers on UVerb requests" $ \(_, baseUrl) -> do
+      res <- runClient getUVerbRespHeaders baseUrl
+      case res of
+        Left e -> assertFailure $ show e
+        Right (Z (I (WithStatus val))) ->
+          getHeaders val `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
+        Right (S _) -> assertFailure "expected first alternative of union"
 
     it "Stores Cookie in CookieJar after a redirect" $ \(_, baseUrl) -> do
       mgr <- C.newManager C.defaultManagerSettings
--- a/servant-server/src/Servant/Server/UVerb.hs
+++ b/servant-server/src/Servant/Server/UVerb.hs
@@ -28,7 +28,7 @@ import Data.SOP (I (I))
 import Data.SOP.Constraint (All, And)
 import Data.String.Conversions (LBS, cs)
 import Network.HTTP.Types (Status, HeaderName, hContentType)
-import Network.Wai (responseLBS)
+import Network.Wai (responseLBS, Request)
 import Servant.API (ReflectMethod, reflectMethod)
 import Servant.API.ContentTypes (AllCTRender (handleAcceptH), AllMime)
 import Servant.API.ResponseHeaders (GetHeaders (..), Headers (..))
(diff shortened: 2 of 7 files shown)


servant-server/src/Servant/Server/UVerb.hs
```diff
@@ -81,16 +98,18 @@ instance
             mkProxy _ = Proxy
 
         runAction action' env request cont $ \(output :: Union as) -> do
-          let encodeResource :: (AllCTRender contentTypes a, HasStatus a) => a -> (Status, Maybe (LBS, LBS))
+          let encodeResource :: (AllCTRender contentTypes a, HasStatus a, HasResponseHeaders a)
```
fisx, line 101:
```suggestion
          let encodeResource :: IsServerResource contentTypes a
```

servant-server/test/Servant/ServerSpec.hs
```diff
@@ -684,6 +687,31 @@ responseHeadersSpec = describe "ResponseHeaders" $ do
...
+type UVerbHeaderResponse = '[
+  WithStatus 200 (Headers '[Header "H1" Int] String),
+  WithStatus 404 String ]
+
+type UVerbResponseHeadersApi =
+       Capture "ok" Bool :> UVerb 'GET '[JSON] UVerbHeaderResponse
+
+uverbResponseHeadersServer :: Server UVerbResponseHeadersApi
+uverbResponseHeadersServer True = pure . Z . I . WithStatus $ addHeader 5 "foo"
+uverbResponseHeadersServer False = pure . S . Z . I . WithStatus $ "bar"
```
fisx, line 703: I'm not sure, but:

```suggestion
uverbResponseHeadersServer True = respond . WithStatus $ addHeader 5 "foo"
uverbResponseHeadersServer False = respond . WithStatus $ "bar"
```

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -65,6 +68,18 @@ data Headers ls a = Headers { getResponse :: a
 instance (NFDataHList ls, NFData a) => NFData (Headers ls a) where
     rnf (Headers x hdrs) = rnf x `seq` rnf hdrs
 
+instance {-# OVERLAPPING #-} MimeRender PlainText a => MimeRender PlainText (Headers _ls a) where
```
fisx, line 71: How are these related to `UVerb`?  Of why did yo introduce them?  Having default non-overlappable instances seems problematic, but I can't think of a reason why anybody would want to overlap these, so maybe it's fine?

fisx (commented):
Nice, thanks!  Do you want to also add the instances for servant-client?  (It's good that you wrote the test like you did, though, so you can catch server bugs that are cancelled out by the dual client bug.)

servant/src/Servant/API/ResponseHeaders.hs
pcapriotti: As far as I understand, they are needed because the `HasServer` instance for `UVerb ... as` has an `AllCTRender` on the elements of `as`, and I don't see how this could be changed to a constraint on the types inside potential `Headers h` wrappers.

It is possible to get away without these instances in the case of `Verb`, because there is a special instance for `HasServer (Verb ... (Headers h a))` that has an `AllCTRender` constraint on `a` (rather than the whole `Headers h a`).

pcapriotti:
I found a way to avoid those instances by making the ad-hoc type class handle generating the body as well, and I also changed its name to `IsServerResource`. The original `IsServerResource` constraint is now called `IsServerResourceWithStatus`.

I used a similar technique for the client. In that case, having `MimeUnrender` instances would not have worked, because it is not possible to construct a value of type `Headers h a` from the body only.

servant-server/test/Servant/ServerSpec.hs
fisx: That's very strange.  Why does `WithStatus k1 a1` not match both of the expected types?

Also, I think I remembered correctly, and [respond](https://hackage.haskell.org/package/servant-server-0.18.2/docs/Servant-Server-UVerb.html#v:respond) is indeed the idiom of choice here.  So something is wrong.  I probably botched the `UVerb` code that has been sitting on master already for a while in yet another interesting subtle way?

servant/src/Servant/API/ResponseHeaders.hs
fisx: Ah, I see.  This LGTM then.

servant-client/test/Servant/SuccessSpec.hs
```diff
@@ -129,6 +130,14 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
         Left e -> assertFailure $ show e
         Right val -> getHeaders val `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
 
+    it "Returns headers on UVerb requests" $ \(_, baseUrl) -> do
+      res <- runClient getUVerbRespHeaders baseUrl
+      case res of
+        Left e -> assertFailure $ show e
+        Right (Z (I (WithStatus val))) ->
+          getHeaders val `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
+        Right (S _) -> assertFailure "expected first alternative of union"
```
fisx, line 139: There are helpers for this, too: https://hackage.haskell.org/package/servant-0.18.2/docs/Servant-API-UVerb-Union.html#v:foldMapUnion, https://hackage.haskell.org/package/servant-0.18.2/docs/Servant-API-UVerb-Union.html#v:matchUnion

servant-server/test/Servant/ServerSpec.hs
pcapriotti: It seems `respond` does not work very well when there are still metavariables left to be unified. I replaced `respond` with its definition `pure . inject . I` to get more meaningful error messages, and I got (among other errors):
```
servant-server      >     • Overlapping instances for Servant.API.UVerb.Union.UElem
servant-server      >                                   (WithStatus k1 a1)
servant-server      >                                   '[WithStatus 200 (Headers '[Header "H1" Int] String),
servant-server      >                                     WithStatus 404 String]
servant-server      >         arising from a use of ‘inject’
```
My understanding is that the type-class based mechanism for checking list membership cannot perform unification of the type variables involved.

One solution, besides manually lifting to the appropriate constructor with `S` and `Z` like now, is to fully annotate the argument to respond with its type:
```haskell
uverbResponseHeadersServer True = respond (WithStatus (addHeader 5 "foo") :: WithStatus 200 (Headers '[Header "H1" Int] String))
uverbResponseHeadersServer False = respond (WithStatus "bar" :: WithStatus 404 String)
```
Which one do you prefer?

servant-client/test/Servant/SuccessSpec.hs
pcapriotti: There is a similar inference problem here. Is this better?
```haskell
      case res of
        Left e -> assertFailure $ show e
        Right val -> case matchUnion val of
          Just (WithStatus val' :: WithStatus 200 (Headers TestHeaders Bool))
            -> getHeaders val' `shouldBe` [("X-Example1", "1729"), ("X-Example2", "eg2")]
          Nothing -> assertFailure "expected first alternative of union"
```

servant-server/test/Servant/ServerSpec.hs
fisx: Ah, I remember now.  Yes, you figured out the problem and the solution.  I personally prefer `respond` with the type annotation, but I think you should make the call.

fisx: Ah, I saw that you finished this, never mind.  I'll take another look.

fisx (approved):
Looks good to me!  (CI issue looks unrelated.)

After thinking it through, I prefer the type annotations and `respond` over explicit value construction; it's a bit verbose and redundant, but doesn't expose internals of the types that shouldn't matter in the application logic.  If you want, you can make another PR and tidy that up, I'll merge this one now.

PR #1423, opened 2021-05-19
use Capture Description if available

### LINKED ISSUE #1422: Include Capture's Description in the generated docs
[The documentation for `Capture'`](https://hackage.haskell.org/package/servant-0.18.2/docs/Servant-API-Capture.html#t:Capture-39-) and [the example `ComprehensiveAPI`](https://github.com/haskell-servant/servant/blob/d06b65c4e6116f992debbac2eeeb83eafb960321/servant/src/Servant/Test/ComprehensiveAPI.hs#L51) both hint that we can use `Capture' '[Description "the id of the Foo to retrieve"] "id" Text]` instead of `Capture "id" Text` in order to get documentation for our capture parameters. However, [the `HasDocs` instance for `Capture`](https://github.com/haskell-servant/servant/blob/d06b65c4e6116f992debbac2eeeb83eafb960321/servant-docs/src/Servant/Docs/Internal.hs#L864-L865) ignores all modifiers including `Description`, in favour of a `ToCapture` instance.

I would prefer to use this `Description` than a `ToCapture` instance, but if that's not the chosen solution, then I think references to this idiom should be dropped from the documentation for `Capture'` and the example `ComprehensiveAPI`.

gelisam:
fixes #1422

When a capture includes a description, e.g. `Capture' '[Description "the id of the Foo to retrieve"] "id" Text]` instead of `Capture "id" Text`, the description is now used in the generated documentation. Previously, the `Description` was ignored and the user had to specify an instance repeating both strings at the value level:

    instance ToCapture (Capture "id" Text) where
      toCapture _ = DocCapture "id" "the id of the Foo to retrieve"

[The example `ComprehensiveAPI`](https://github.com/haskell-servant/servant/blob/d06b65c4e6116f992debbac2eeeb83eafb960321/servant/src/Servant/Test/ComprehensiveAPI.hs#L51) already uses both `Capture' '[Description "example description"] "foo" Int` and `Capture "foo" Int`, and ./servant-docs/test/Servant/DocsSpec.hs provides a `ToCapture (Capture "foo" Int)` instance which covers both cases. With this PR, the first capture now uses `example description` instead of that `ToCapture` instance.

This is thus technically a breaking change, because if a `Capture` has both a `Description` and a `ToCapture` instance, the `Description` now takes precedence. Since this `Description` wasn't doing anything before, I am guessing that most projects currently only use `Description` to describe their endpoints and not their `Capture`s, and thus that few people will be affected by this breaking change.

The case I am more interested in is the case in which there is only a `Description` and no `ToCapture` instance; I want to make sure that a `ToCapture` instance is not required. I have thus changed the `ComprehensiveAPI` example from `Capture' '[Description "example description"] "foo" Int` to `Capture' '[Description "example description"] "bar" Int`, this way this parameter has a `Description` but no `ToCapture (Capture "bar" Int)` instance, while the `Capture "foo" Int` parameter still tests the case in which there is a `ToCapture (Capture "foo" Int)` instance but no `Description`.

### DIFF
--- a/servant-docs/src/Servant/Docs/Internal.hs
+++ b/servant-docs/src/Servant/Docs/Internal.hs
@@ -862,7 +862,7 @@ instance HasDocs EmptyAPI where
 -- | @"books" :> 'Capture' "isbn" Text@ will appear as
 -- @/books/:isbn@ in the docs.
 instance (KnownSymbol sym, ToCapture (Capture sym a), HasDocs api)
-      => HasDocs (Capture' mods sym a :> api) where
+      => HasDocs (Capture' '[] sym a :> api) where
 
   docsFor Proxy (endpoint, action) =
     docsFor subApiP (endpoint', action')
@@ -873,6 +873,28 @@ instance (KnownSymbol sym, ToCapture (Capture sym a), HasDocs api)
           action' = over captures (|> toCapture captureP) action
           endpoint' = over path (\p -> p ++ [":" ++ symbolVal symP]) endpoint
           symP = Proxy :: Proxy sym
+
+instance (KnownSymbol descr, KnownSymbol sym, HasDocs api)
+      => HasDocs (Capture' (Description descr ': mods) sym a :> api) where
+
+  docsFor Proxy (endpoint, action) =
+    docsFor subApiP (endpoint', action')
+
+    where subApiP = Proxy :: Proxy api
+
+          docCapture = DocCapture (symbolVal symP) (symbolVal descrP)
+          action' = over captures (|> docCapture) action
+          endpoint' = over path (\p -> p ++ [":" ++ symbolVal symP]) endpoint
+          descrP = Proxy :: Proxy descr
+          symP = Proxy :: Proxy sym
+
+instance {-# OVERLAPPABLE #-} HasDocs (Capture' mods sym a :> api)
+      => HasDocs (Capture' (mod ': mods) sym a :> api) where
+
+  docsFor Proxy =
+    docsFor apiP
+
+    where apiP = Proxy :: Proxy (Capture' mods sym a :> api)
 
 
 -- | @"books" :> 'CaptureAll' "isbn" Text@ will appear as
--- a/servant-docs/golden/comprehensive.md
+++ b/servant-docs/golden/comprehensive.md
@@ -52,11 +52,11 @@
 
 ```
 
-## GET /capture/:foo
+## GET /capture/:bar
 
 ### Captures:
 
-- *foo*: example description
+- *bar*: example description
 
 ### Response:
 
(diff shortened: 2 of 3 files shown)


servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -874,6 +874,28 @@ instance (KnownSymbol sym, ToCapture (Capture sym a), HasDocs api)
...
+    where subApiP = Proxy :: Proxy api
+
+          docCapture = DocCapture (symbolVal symP) (symbolVal descrP)
+          action' = over captures (|> docCapture) action
+          endpoint' = over path (\p -> p ++ [":" ++ symbolVal symP]) endpoint
+          descrP = Proxy :: Proxy descr
+          symP = Proxy :: Proxy sym
+
+instance {-# OVERLAPPABLE #-} HasDocs (Capture' mods sym a :> api)
+      => HasDocs (Capture' (mod ': mods) sym a :> api) where
```
gelisam, line 892: I know I can use `reflectDescription` to extract the `Description` string from `mods` without having to recur myself, but if I did that I would not be able to treat the `'[]` case specially and ask for a `ToCapture` instance in that case.

servant-docs/src/Servant/Docs/Internal.hs
```diff
@@ -874,6 +874,28 @@ instance (KnownSymbol sym, ToCapture (Capture sym a), HasDocs api)
...
+instance (KnownSymbol descr, KnownSymbol sym, HasDocs api)
+      => HasDocs (Capture' (Description descr ': mods) sym a :> api) where
+
+  docsFor Proxy (endpoint, action) =
+    docsFor subApiP (endpoint', action')
+
+    where subApiP = Proxy :: Proxy api
+
+          docCapture = DocCapture (symbolVal symP) (symbolVal descrP)
+          action' = over captures (|> docCapture) action
```
gelisam, line 886: this is the same as the `'[]` instance above, except I construct and use `docCapture` instead of calling `toCapture captureP`.

gelisam:
Is the guidance in [CONTRIBUTING.md](https://github.com/haskell-servant/servant/blob/master/CONTRIBUTING.md#changelog-entries) to add a file to `changelog.d` still accurate? I don't see any such files in the current `changelog.d` folder, nor do I see such files being added in other recent PRs.

gelisam:
As per the guidance in CONTRIBUTING.md, I am pinging @haskell-servant/maintainers because I haven't received feedback yet. I am guessing that the lack of feedback means the maintainers team is very busy; if so, are you accepting applications to join the maintainers team?

gelisam:
Hmm, `@haskell-servant/maintainers` didn't get recognized by github, so I guess that's not how I'm supposed to ping them. Pinging every single member on https://github.com/orgs/haskell-servant/people seems overly aggressive, so I'll ping a random one: @ dmjio

dmjio:
@gelisam this is great.

gelisam:
Thanks!

PR #1432, opened 2021-06-27
Addresses problem with URL encodings

### LINKED ISSUE #1418: Double encoding of percent signs (servant-client)
First of all, this is a wonderful project and very useful.  Now, down to business.

When using defaults, `servant-client` code can double-encode percent signs in query parameters.  Consider this (incomplete) example:

```haskell
newtype UrlEncoded = UrlEncoded { unUrlEncoded :: ByteString }

instance ToHttpApiData UrlEncoded where
  toQueryParam = decodeUtf8 . urlEncode True . unUrlEncoded

type ExampleEndpoint = QueryParam "value" UrlEncoded :> GetNoContent

exampleEndpoint :: UrlEncoded -> ClientM ()
exampleEndpoint = client $ Proxy @ExampleEndpoint

problemCall :: ClientM ()
problemCall = exampleEndpoint $ UrlEncoded "\x01"
```

Running `problemCall` creates a request with query string `?value=%2501`.  I believe the reason is that [the `HasClient` instance of `QueryParam` uses `toQueryParam`][hcqp] when adding the parameter to the (servant) `Request` it builds up, then [`defaultMakeClientRequest` calls `renderQuery` which re-encodes the string][dmcr] while marshalling a (http-client) `Request`.

[hcqp]: https://github.com/haskell-servant/servant/blob/448c444db665730d40df46f86f04d1b04dcfb32f/servant-client-core/src/Servant/Client/Core/HasClient.hs#L536
[dmcr]: https://github.com/haskell-servant/servant/blob/448c444db665730d40df46f86f04d1b04dcfb32f/servant-client/src/Servant/Client/Internal/HttpClient.hs#L243

It is my understanding that `ToHttpApiData (toQueryParam)` is meant to produce _escaped_ output.   Therefore, it seems that `defaultMakeClientRequest` should use a custom version of `renderQuery` which does not escape values.

GambolingPangolin:
This changes the way URL encoding for query parameters is handled,
making it possible to correctly encode arbitrary binary data into query
parameter values.

Closes #1418


alpmestan:
@GambolingPangolin Hello, sorry for the delay, I haven't had much time this summer to carefully follow up on servant stuffs. Would you mind rebasing so that we can see if this passes CI?

There's a bit of unnecessary churn in your patch (imports etc), but

GambolingPangolin:
@alpmestan I shrank the patch and rebased.  It would be a nice quality of life improvement to impose formatting rules on the sources so that I could defer code layout more to an automated tool.  I totally get the objection to churn, but want to share my 2c.

alpmestan:
@GambolingPangolin I understand the motivation, having had these discussions countless times, but it's going to be tough to agree on a particular style with all the folks involved in the project =) CI should be running, let's see how this goes.

akhesaCaro:
To support older version of GHC you need to explicitly import the `<>` operator.

`import           Data.Semigroup (<>)`

GambolingPangolin:
@akhesaCaro Thanks for the tip about testing with nix.  I'm confused about the CI failure, though.  It looks like the error message concerns a module that I did not touch with my change.  Should I change the imports there, or is this a sign of some other problem?

GambolingPangolin:
@alpmestan @akhesaCaro did you see my previous comment?

akhesaCaro:
I stopped the support of GHC < 8.6.5 since yesterday. You just have to rebase and it will pass.

GambolingPangolin:
@akhesaCaro I rebased this patch and it's ready for another try 😁

Issue #1438, opened 2021-08-11
Improve serve functions


brandon-leapyear:
It'd be nice to have a helper for the common case (in my experience) of serving all routes in a custom monad, especially with a custom context. Currently, I do this with
```hs
app = serveRoutes @MyAPI myContext runMyMonad routes

serveRoutes ::
  forall api context m.
  HasServer api context =>
  Context context ->
  (forall x. m x -> Handler x) ->
  ServerT api m ->
  Application
serveRoutes ctx f routes =
  serveWithContext (Proxy @api) ctx $
    hoistServerWithContext (Proxy @api) (Proxy @context) f routes
```

This seems like a pretty general helper that can be exported upstream. I would think `servant-server` could expose an API like:
```hs
serve :: HasServer api '[] => Proxy api -> Server api -> Application
serve p = serveWithContextT p EmptyContext id

-- related: https://github.com/haskell-servant/servant/issues/1437
serveWithContext :: HasServer api context => Proxy api -> Context context -> Server api -> Application
serveWithContext p ctx = serveWithContextT p ctx id

serveWithContextT ::
  HasServer api context =>
  Proxy api ->
  Context context ->
  (forall x. m x -> Handler x) ->
  ServerT api m ->
  Application
```

It would also be nice if `serve` and `serveWithContext` used type applications instead of using proxies, but I'm not too hung up about that.

alpmestan:
Hello.

Regarding your central suggestion, I definitely agree. I think we could perhaps even reuse the `Server`/`ServerT` naming convention (i.e the `T` suffix), and perhaps rename your `serveWith` to `serveWithContextT`? No strong opinion on that, just throwing it out there.

---

Regarding:

> It would also be nice if `serve` and `serveWithContext` used type applications instead of using proxies, but I'm not too hung up about that.

I'm not overly enthusiastic about this. Since `ServerT` is a non injective type family, am I mistaken in thinking that this would essentially force everyone to enable `AllowAmbiguousTypes`? Which is not something I'm delighted to be encouraging, even if it's pretty benign all things considered.

brandon-leapyear:
> Regarding your central suggestion, I definitely agree. I think we could perhaps even reuse the `Server`/`ServerT` naming convention (i.e the `T` suffix), and perhaps rename your `serveWith` to `serveWithContextT`? No strong opinion on that, just throwing it out there.

Yeah, that sounds like a good idea. Updated the description

[...]

Issue #1437, opened 2021-08-11
Simplify serveWithContext ErrorFormatters constraint


brandon-leapyear:
`servant-server-0.18` added the `HasContextEntry (context .++ DefaultErrorFormatters) ErrorFormatters` constraint to `serveWithContext`, which is annoying to rewrite if one writes a helper function around `serveWithContext`.

It seems like this constraint is redundant if you just call `mkContextWithErrorFormatter` before passing the `context` to `route`.

alpmestan:
@brandon-leapyear Do you have a suggestion to improve the experience of writing helpers around `serveWithContext`, or for giving a less cumbersome constraint there in the first place?

brandon-leapyear:
hm it seems like my idea didn't work. i guess servant could provide a type alias, so it's at least not so cumbersome, like
```hs
type IsValidContext context =
  ( HasContextEntry (context .++ DefaultErrorFormatters) ErrorFormatters
  , ... -- in the future, could add other constraints needed, if more default context entries are added
  )
```

alpmestan:
I find `IsValidContext` to be a bit too cryptic without conveying any illuminating information. `ServerContext`?

brandon-leapyear:
yeah, `ServerContext` sounds good!

PR #1441, opened 2021-08-19
Add serveWithContextT, ServerContext

brandon-leapyear:
Resolves https://github.com/haskell-servant/servant/issues/1437
Resolves https://github.com/haskell-servant/servant/issues/1438


servant-server/src/Servant/Server.hs
```diff
@@ -128,6 +131,11 @@ import           Servant.Server.UVerb
 
 -- * Implementing Servers
 
+-- All constraints on a context passed to serveWithContext.
+type ServerContext context =
```
alpmestan, line 135: Could you turn this into a proper haddock docstring that concisely explains what this synonym is about?

stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-16.24
+resolver: lts-18.5
```
alpmestan, line 1: Looks like this isn't quite sufficient to get the stack CI build to work. I've not been using stack much lately, do you think you could look into that failure?

alpmestan (commented):
2 minor things to address and this should be good to go.

stack.yaml
brandon-leapyear: Sorry, I should've done this in another PR. I just wanted to avoid building an old snapshot on my local computer when working on this. But yeah, I'll get this working.

.github/workflows/master.yml
```diff
@@ -95,8 +95,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        stack: ["2.3.1"]
-        ghc: ["8.8.4"]
+        stack: ["2.7.3"]
+        ghc: ["8.10.4"]
```
brandon-leapyear, line 99: maybe we should just avoid setting the GHC version and remove `--system-ghc` in all the `stack` calls?

alpmestan: I have no idea how that stack version of the CI workflow works, no idea what GHC version would be picked if we leave it out (if any). All I know is that we don't want to specify many of them since the cabal jobs already test several versions of GHC.

alpmestan:
@brandon-leapyear Is this good to go now?

.github/workflows/master.yml
brandon-leapyear: The `ghc` version here just specifies what version of ghc will be installed globally. Calling `stack` with `--system-ghc` will tell it to find some `ghc` on `PATH` and use it.

If GHC weren't specified, `stack` would use the version of GHC in the resolver of the `stack.yaml` file. This would reduce the duplication between setting GHC version both in `stack.yaml` (via the resolver) and this line.

brandon-leapyear:
yup everything's good with me

alpmestan:
Alright, it's in. Thanks Brandon!

Issue #1435, opened 2021-07-11
Raw files served from root override other routes.


bChiquet:
I run into 405 `Only GET or HEAD is supported` errors when trying to setup a Raw file service directly on root.

Minimum reproduction example:

In this example, `Raw` is served directly from root.
There is a second run-of-the-mill `Post` route.
`Post` is called with an empty list as a body.
This nets a 405 `Only GET or HEAD is supported` error.

```haskell
{-# language TypeOperators, DataKinds #-}
module Lib
    ( api
    ) where

import Servant.API (Raw, Post, ReqBody, JSON, (:>), (:<|>)(..))
import Servant.Server.StaticFiles (serveDirectoryWebApp)
import Servant.Server (serve)
import Data.Proxy (Proxy(Proxy))
import Data.Aeson (Value)

type Api 
  = Raw -- serve files from parent directory
  :<|> "reports" :> ReqBody '[JSON] Value :> Post '[JSON] () 

appServer
  = serveDirectoryWebApp "../"
  :<|> \body -> pure ()

api = serve (Proxy :: Proxy Api) appServer
```
Running this gives: 
```
POST /reports
  Request Body: []
  Accept: application/json
  Status: 405 Method Not Allowed 0.00001707s
```

In this second example, the raw files are behind a path parameter (`"example"`)
The `Post` method is not changed.
It is still called with an empty list as a body.
This nets a 200 ok.

```haskell
{-# language TypeOperators, DataKinds #-}
module Lib
    ( api
    ) where

import Servant.API (Raw, Post, ReqBody, JSON, (:>), (:<|>)(..))
import Servant.Server.StaticFiles (serveDirectoryWebApp)
import Servant.Server (serve)
import Data.Proxy (Proxy(Proxy))
import Data.Aeson (Value)

type Api 
  = "example" :> Raw -- serve files from parent directory
  :<|> "reports" :> ReqBody '[JSON] Value :> Post '[JSON] () 

appServer
  = serveDirectoryWebApp "../"
  :<|> \body -> pure ()

api = serve (Proxy :: Proxy Api) appServer
```
Running this gives: 
```
POST /reports
  Request Body: []
  Accept: application/json
  Status: 200 OK 0.000114646s
```

Possibly related to https://github.com/haskell-servant/servant/issues/194

alpmestan:
[...]

PR #1455, opened 2021-09-29
Document Raw's behaviour when composing APIs

bChiquet:
Patch for #1435.

### DIFF
--- a/doc/tutorial/ApiType.lhs
+++ b/doc/tutorial/ApiType.lhs
@@ -389,3 +389,30 @@ One example for this is if you want to serve a directory of static files along
 with the rest of your API. But you can plug in everything that is an
 `Application`, e.g. a whole web application written in any of the web
 frameworks that support `wai`.
+
+Be mindful! This library works by pattern-matching the different routes that are
+composed using `:<|>`. `Raw`, as an escape hatch, matches any route that hasn't
+been matched by previous patterns. Therefore, any subsequent route will be silently
+ignored.
+
+``` haskell
+type UserAPI14 = Raw
+            :<|> "users" :> Get '[JSON] [User]
+                 -- In this situation, the /users endpoint
+                 -- will not be reachable because the Raw
+                 -- endpoint matches requests before
+```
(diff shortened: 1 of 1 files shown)


doc/tutorial/ApiType.lhs
```diff
@@ -389,3 +389,30 @@ One example for this is if you want to serve a directory of static files along
 with the rest of your API. But you can plug in everything that is an
 `Application`, e.g. a whole web application written in any of the web
 frameworks that support `wai`.
+
+Be mindful! This library works by pattern-matching the different routes that are
```
alpmestan, line 393: I'd probably say

> `servant-server`'s router

in place of "this library".

doc/tutorial/ApiType.lhs
```diff
@@ -389,3 +389,30 @@ One example for this is if you want to serve a directory of static files along
...
+
+``` haskell
+type UserAPI14 = Raw
+            :<|> "users" :> Get '[JSON] [User]
+                 -- In this situation, the /users endpoint
+                 -- will not be reachable because the Raw
+                 -- endpoint matches requests before
+```
+A simple way to avoid this pitfall is to either use `Raw` as the last
+definition, or to always have it under a directory.
```
alpmestan, line 406: Here and in the Haskell comment below, I'd probably use a slightly more generic vocabulary than "directory", as there's usually no guarantee that this indeed corresponds to any sort of directory. "under a static path", "under a specific endpoint", or something along those lines...?

alpmestan (commented):
That seems like a very useful addition to me. Some nitpicks in the review, but nothing major at all.

Thanks a lot!

doc/tutorial/ApiType.lhs
bec-fretlink: I actually didn't know what to call it, `static path` is a great suggestion.
I will also be more explicit in the example that the static path is the `/file/` bit, since there doesn't seem to be a widely accepted term for that.

alpmestan (approved):

PR #1462, opened 2021-10-09
Upgrade GHCJS

gdeest:
This is an attempt to get rid of GHCJS 8.4.

We're stuck using CPP tricks to support our CI workflow ; this does not make sense, since we deprecated this version for vanilla GHC.

`reflex-platform` seems the only sensible way to install GHCJS. Unfortunately, it is stuck to GHCJS 8.6: this will have to do, as we can't afford building the compiler ourselves.


gdeest:
TODO: We need to update all the `tested-with` fields in Cabal files, and generally hunt down references to GHC 8.4 all over the project.

tchoutri (approved):
_C'est tipar_

PR #1469, opened 2021-10-19
Derive HasClient good response status from Verb status

marinelli:
Hi Everyone,
There's a limitation/bug in the `HasClient` instances for the `Verb` datatype.
If I try to perform a client request with `runClientM`, I'll obtain a successful response if and only if the backend endpoint returns a response with status code >=200 and <300. This is due to the use of `runRequest` in the `clientWithRoute` definitions.

This PR brings the following changes:
- Replace `runRequest` with `runRequestAcceptStatus` in the `Verb` instances for the `HasClient` class (as suggested by @alpmestan).
- Derive the good response status from the `Verb` status.
- Use `statusIsSuccessful` from `Network.HTTP.Types.Status` to lighten the code (not required to solve the issue).

### DIFF
--- a/servant-client/test/Servant/SuccessSpec.hs
+++ b/servant-client/test/Servant/SuccessSpec.hs
@@ -59,11 +59,15 @@ spec = describe "Servant.SuccessSpec" $ do
 
 successSpec :: Spec
 successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
-    it "Servant.API.Get root" $ \(_, baseUrl) -> do
-      left show <$> runClient getRoot baseUrl  `shouldReturn` Right carol
+    describe "Servant.API.Get" $ do
+      it "get root endpoint" $ \(_, baseUrl) -> do
+        left show <$> runClient getRoot baseUrl  `shouldReturn` Right carol
 
-    it "Servant.API.Get" $ \(_, baseUrl) -> do
-      left show <$> runClient getGet baseUrl  `shouldReturn` Right alice
+      it "get simple endpoint" $ \(_, baseUrl) -> do
+        left show <$> runClient getGet baseUrl  `shouldReturn` Right alice
+
+      it "get redirection endpoint" $ \(_, baseUrl) -> do
+        left show <$> runClient getGet307 baseUrl `shouldReturn` Right "redirecting"
 
     describe "Servant.API.Delete" $ do
       it "allows empty content type" $ \(_, baseUrl) -> do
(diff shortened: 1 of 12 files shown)


alpmestan (changes requested):
Looks great thanks, and sorry for the delay!

Do you perhaps feel like adding a test case where the server returns e.g 201 but the client wants 200, to make sure this doesn't break in the future?

@tchoutri I suspect we'll want to be very loud and clear about this when announcing the release that ships this, including in the changelog. This might even upset some users ("but the response is a 2xx, it should be fine!"), perhaps? I don't know.

tchoutri:
You got it boss

servant-client/test/Servant/SuccessSpec.hs
```diff
@@ -83,6 +83,9 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
       let p = Person "Clara" 42
       left show <$> runClient (getBody p) baseUrl `shouldReturn` Right p
 
+    it "Servant.API.Get redirection" $ \(_, baseUrl) -> do
```
alpmestan, line 86: The idea of this test being that before your patch, we'd get a `ClientError` because we don't get a 2xx status code?

marinelli: Yes, that's the idea. I also checked the same test in `master` and it returns, as expected, a `FailureResponse`.

marinelli:
> Do you perhaps feel like adding a test case where the server returns e.g 201 but the client wants 200, to make sure this doesn't break in the future?

I had a look at how to add this test, and I was not sure how to proceed.
If we want to test that the right status code for the client is the one derived from the public _api_ signature of a server, well, what I would like to do is to change the status code in the server response before the client receive it. The only way I found to do that is to create a copy of the public _api_ signature of the server and create an other one specific for the _client_. they would be identical except for the status code in the endpoint we want to test. Eg:
```haskell
type ServerAPI = Verb GET 201 '[JSON] ()
type ClientAPI = Verb GET 200 '[JSON] ()
```
I tried to use also the free monad client in the `Servant.Client.Free`. At the end, that `client` is usefull to _simulate_ a client, not to verify if the real client reports errors as expected.

alpmestan:
For testing this precise thing, I'd do exactly what you said, no need to appeal to the big common API type with a bunch of constructs in it.

Just the two tiny almost-identical, differing-only-in-status-code endpoint types you gave seem perfect. They're even short enough to write them inline, `client (Proxy :: Proxy (Verb GET 200 '[JSON] ()))`, if you want.

servant-client/test/Servant/SuccessSpec.hs
marinelli: Btw, this is similar to the `"Redirects when appropriate"` test, that uses `uverbGetSuccessOrRedirect True`.

alpmestan: OK, sounds good. Since it wasn't 100% obvious to me initially, that might be the case for other contributors, perhaps indicating the purpose in the test label or a comment nearby would help. But the test otherwise looks good.

marinelli:
> For testing this precise thing, I'd do exactly what you said, no need to appeal to the big common API type with a bunch of constructs in it.
> 
> Just the two tiny almost-identical, differing-only-in-status-code endpoint types you gave seem perfect. They're even short enough to write them inline, `client (Proxy :: Proxy (Verb GET 200 '[JSON] ()))`, if you want.

I added the `Servant.BrokenSpec` module for checking client/server api inconsitencies.

alpmestan (approved):
LGTM. Thanks!

PR #1486, opened 2021-11-18
Custom errors for HasClient, HasServer

gdeest:
This PR applies the approach from #1289 to `HasServer` and `HasClient`.

The `context` parameter is missing from the missing instance message in one case, but I am not sure it is a big enough problem to warrant making the code more complex.

In the process, I realized there was already a similar mechanism in place to detect usage errors for `HasServer`. The mechanism #1289 supersedes it, as it produces more informative messages, displays the number of missing parameters, etc.

### DIFF
--- a/servant-client-core/src/Servant/Client/Core/HasClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/HasClient.hs
@@ -65,7 +65,7 @@ import           Data.Text
 import           Data.Proxy
                  (Proxy (Proxy))
 import           GHC.TypeLits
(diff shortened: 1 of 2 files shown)


gdeest:
OK, I consider the issue of the missing `context` parameter relatively benign: we are not pattern-matching on the context, but on the API ; the context is irrelevant to the error message, so I'd rather avoid making the error generation machinery more complex for little reason.

Moreover, since we are talking about type errors, we can decide to change how these are generated at any point without consequence on our users, so I am going to merge.

Issue #1497, opened 2021-12-21
the license in *.cabal is not parsed correctly by cabal for some servant packages


teto:
while running [cabal2nix](https://github.com/NixOS/cabal2nix) on servant  (to be able to reference the master version from nix), I hit the following issue:
```
cabal2nix: Distribution.Nixpkgs.Haskell.FromCabal.License.fromCabalLicense: unknown licenseUnknownLicense "BSD-3"
Choose one of: GPL, GPL-2, GPL-3, LGPL, LGPL-2.1, LGPL-3, AGPL, AGPL-3, BSD2, BSD3, MIT, ISC, MPL-2.0, Apache, Apache-2.0, PublicDomain, AllRightsReserved, OtherLicense
CallStack (from HasCallStack):
  error, called at src/Distribution/Nixpkgs/Haskell/FromCabal/License.hs:46:62 in cabal2nix-2.18.0-inplace:Distribution.Nixpkgs.Haskell.FromCabal.License
```
I first thought this was a cabal2nix issue but a few debug statements later, I think it's servant that incorrectly specifies its license. Cabal supports two types of licenses, native and SPDX, which can be seen here https://hackage.haskell.org/package/Cabal-3.6.2.0/docs/Distribution-Types-PackageDescription.html#v:licenseRaw

Several packages use `BSD-3-Clause` as a license, in conjonction with `cabal-version:       >=1.10` which cabal parses as `Right (UnknownLicense "BSD-3")`.
If I change teh cabal-version to `cabal-version:       2.2` , cabal correctly identifdies the license `License (ELicense (ELicenseId BSD_3_Clause))`.

Remediation:
- either update cabal-version from `>= 1.1.` to `2.2`
- or change BSD-3-Clause to BSD3 

My preference goes to the first solution (because SPDX is good) but the second doesn't need a bump in cabal-version.

I can provide a PR once a consensus is reached.

tchoutri:
@teto Hi, and thank for reporting this! 
As a co-maintainer, my opinion is to specify 2.2 as the minimum version for Cabal files, but I'll ask the other @haskell-servant/maintainers for their opinion.

alpmestan:
I think it's reasonable.

tchoutri:
Fantastic.
@teto You may proceed with a PR, thank you very much for having noticed this.

PR #1498, opened 2021-12-21
bumped cabal-version field

teto:
Fix https://github.com/haskell-servant/servant/issues/1497

### DIFF
--- a/servant-swagger/servant-swagger.cabal
+++ b/servant-swagger/servant-swagger.cabal
@@ -1,3 +1,4 @@
+cabal-version:       2.2
 name:                servant-swagger
 version:             1.1.11
 synopsis:            Generate a Swagger/OpenAPI/OAS 2.0 specification for your servant API.
@@ -27,7 +28,6 @@ maintainer:          haskell-servant-maintainers@googlegroups.com
 copyright:           (c) 2015-2018, Servant contributors
 category:            Web, Servant, Swagger
 build-type:          Custom
-cabal-version:       1.18
 tested-with: GHC ==8.6.5 || ==8.8.4 || ==8.10.2
 
 extra-source-files:
--- a/servant-swagger/servant-swagger.cabal
+++ b/servant-swagger/servant-swagger.cabal
@@ -21,7 +21,7 @@ description:
   * and [many others](http://swagger.io/open-source-integrations/).
 homepage:            https://github.com/haskell-servant/servant/servant-swagger
 bug-reports:         https://github.com/haskell-servant/servant/issues
-license:             BSD3
+license:             BSD-3-Clause
 license-file:        LICENSE
 author:              Servant Contributors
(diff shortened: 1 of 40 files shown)


akhesaCaro:
Should we change it in the sub projects cabal file, such as `servant-swagger.cabal` for instance?

tchoutri:
I think we should

teto:
I've done the requested change. Failures dont seem linked to PR ?

akhesaCaro:
I have to fix that first, I am sorry.

akhesaCaro:
It seems that some other Cabal files have been forgotten.

akhesaCaro:
for instance, servant-auth-swagger is still missing...

servant-swagger/servant-swagger.cabal
```diff
@@ -1,5 +1,5 @@
 name:                servant-swagger
-version:             1.1.11
+version:             2.2
```
tchoutri, line 2: You've changed `version` (which is the library version) instead of `cabal-version` (which is the format used by the Cabal file.

tchoutri (changes requested):

servant-swagger/servant-swagger.cabal
```diff
@@ -20,7 +20,7 @@ description:
   * and [many others](http://swagger.io/open-source-integrations/).
 homepage:            https://github.com/haskell-servant/servant/servant-swagger
 bug-reports:         https://github.com/haskell-servant/servant/issues
-license:             BSD3
+license:             BSD-3-Clause
```
akhesaCaro, line 23: For the record : Why BSD3 license has to be changed in this PR?

tchoutri: Because using such a Cabal file format mandates the usage of SPDX-compliant identifiers, which are listed at https://spdx.org/licenses.

servant-swagger/servant-swagger.cabal
teto: damn sorry. Fixed.

servant-swagger/servant-swagger.cabal
teto: i think both are supported so it could remain BSD3 but I believe the intent of cabal is to favor the SPDX format so I think it's fine to change it here. I haven't changed it everywhere though as I prefer to keep changes minimal.

akhesaCaro:
- servant-auth-client
- servant-auth-server
- servant-auth-swagger

also need to be updated...

servant-swagger/servant-swagger.cabal
tchoutri: I'be more comfortable if you got that out of the way in this PR (maybe in a different commit, for ease of review). The next versions of Cabal will force us to comply

teto: done

tchoutri (approved):
LGTM

PR #1502, opened 2022-01-06
Add HeadNoContent to Servant.API.Verbs

supersven:
As the head method isn't allowed to contain any response body, no
general Head Verb is added. (This may easily lead to wrong usages...)

(https://httpwg.org/specs/rfc7231.html#HEAD)


tchoutri:
Excellent, thank you.

tchoutri (approved):
LGTM

PR #1503, opened 2022-01-08
Update docs: #servant is now on libera.chat

supersven:
Looks like `#servant` is now hosted on libera.


tchoutri:
@haskell-servant/maintainers Currently we have `#haskell-servant` and `#servant` on Libera. Which one do we keep? I would be in favour of `#haskell-servant` personally.

fisx:
Plus one on `#haskell-servant`.

alpmestan:
Likewise.

gdeest:
Agreed, `#haskell-servant` sounds like a better name for discussions about the servant ecosystem (and not just the servant project per se).

tchoutri:
Perfect, thanks everyone.
@supersven We're going with `#haskell-servant` then. :)

tchoutri (approved):

PR #1510, opened 2022-01-20
WIP: Add prometheus metrics to the servant router

arianvp:
Annotates the routing tree with a middleware that instruments each endpoint in your servant application.
For each endpoint we will keep track of a histogram with request latency + request count

Captures are automatically filtered; such that  /users/13813 and /users/24774
get mapped to the same metric.

Unfortunately the name of the capture is not preserved; as we throw away that
data when we're at the point of routing. We could perhaps change this but it's
more significant change.

This was sketched out on the train without a compiler at hand. So definitely doesn't work yet. Just need a place to keep the draft around.

### DIFF
--- a/servant-server/src/Servant/Server/Internal/Metrics.hs
+++ b/servant-server/src/Servant/Server/Internal/Metrics.hs
@@ -1,0 +1,96 @@
+module Servant.Server.Internal.Router.Prometheus where
+
+-- | The 'RoutingApplication' variant of 'instrumentHandlerValue'
+instrumentRoutingHandlerValue :: (Request -> Text) -> RoutingApplication -> RoutingApplication
+instrumentRoutingHandlerValue getHandlerLabel app = \req respond -> do
+    start <- getTime Monotonic
+    app req $ \res -> do
+        end <- getTime Monotonic
+        let method = Just $ decodeUtf8 (Wai.requestMethod req)
+        let status = Just $ T.pack (show (HTTP.statusCode (Wai.responseStatus res')))
+        observeSeconds (getHandlerLabel req) method status start end
+        respond res
+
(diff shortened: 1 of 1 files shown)


tchoutri:
@arianvp I could see the more significant change land in the next major.

Thanks a bunch for the PR. :)

Any chance you could also throw in a cookbook in the mix? ;)

servant-server/src/Servant/Server/Internal/Metrics.hs
```diff
@@ -0,0 +1,96 @@
...
+-- Note that the time it takes to execute the routing to the endpoint is not included in the metrics.
+-- If we want that, we'd have to reimplement 'runRouterEnv'. 
+-- It builds up the labels as the routes are being traversed, such that we can label the metric correctly.
+routerWithMetrics :: Router ([Text], env)  -> Router ([Text],env) 
+routerWithMetrics (StaticRouter children leafs) =
+    StaticRouter (Map.map routerWithMetrics children) (List.map instrumentLeaf leafs)
+routerWithMetrics (CaptureRouter router) =
+    CaptureRouter (routerWithMetrics router)
+routerWithMetrics (CaptureAllRouter router) =
+    CaptuerAllRouter (routerWithMetrics router)
```
puellanivis, line 92:
```suggestion
    CaptureAllRouter (routerWithMetrics router)
```

arianvp:
I ended up going a completely different route. Reusing the `HasEndpoint` logic from https://github.com/haskell-servant/servant-ekg (which needs a new release btw!).

We might want to break out the `HasEndpoint` typeclass from `servant-ekg`. I'll write a little cookbook how to use it with prometheus soon

PR #1514, opened 2022-01-24
Fix GHC 9.2 build

guibou:
Close #1513.

GHC 9.2 needs explicit kind signature here, I don't really understand
why.

This kind signature is correct and not too restritive, because `HasLink`
is technically defined `class HasLink endpoint` which means that it is
infered as `k -> Constraint`. In the instance signature, we have
`HasLink ((arr :: a -> b) :> sub)`, so here the `k` is the same kind as
the one of `:>` which is not polykinded.

Note: build tested with GHC 9.2 and the following workflow:

```
$ nix-shell ./ --argstr compiler 'ghc921' --arg pkgs 'import <nixpkgs> {}' -A servant

[nix-shell:~/srcs/servant]$ nix-shell -p cabal-install

[nix-shell:~/srcs/servant]$ cabal build servant
Up to date
```

I had to bump a few bounds:

```diff
diff --git a/cabal.project b/cabal.project
index ac4a4e3b..e2573fa9 100644
--- a/cabal.project
+++ b/cabal.project
@@ -57,7 +57,7 @@ constraints:
   foundation >=0.0.14,
   memory <0.14.12 || >0.14.12
 
-constraints: base-compat ^>=0.11
+constraints: base-compat ^>=0.12
 constraints: semigroups ^>=0.19
 
 -- allow-newer: sqlite-simple-0.4.16.0:semigroups
@@ -74,3 +74,5 @@ allow-newer: servant-js:servant
 
 -- ghc 9
 allow-newer: tdigest:base
+
+allow-newer: *
diff --git a/servant/servant.cabal b/servant/servant.cabal
index f5f973bb..14193e18 100644
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -80,7 +80,7 @@ library
   --
   -- note: mtl lower bound is so low because of GHC-7.8
   build-depends:
-      base                   >= 4.9      && < 4.16
+      base                   >= 4.9      && < 4.17
     , bytestring             >= 0.10.8.1 && < 0.12
     , constraints            >= 0.2
     , mtl                    >= 2.2.2    && < 2.3
@@ -98,7 +98,7 @@ library
   -- Other dependencies: Lower bound around what is in the latest Stackage LTS.
   -- Here can be exceptions if we really need features from the newer versions.
   build-depends:
-      base-compat            >= 0.10.5   && < 0.12
+      base-compat            >= 0.12   && < 0.13
     , aeson                  >= 1.4.1.0  && < 3
[...]

Issue #1513, opened 2022-01-24
Type error with GHC 9.2


guibou:
Build with GHC 9.2 leads to the following type error

```
Build profile: -w ghc-9.2.1 -O0
In order, the following will be built (use -v for more details):
 - servant-0.18.3 (lib) (first run)
 - servant-0.18.3 (test:spec) (first run)
Preprocessing library for servant-0.18.3..
Building library for servant-0.18.3..
[31 of 33] Compiling Servant.Links    ( src/Servant/Links.hs, /home/guillaume/srcs/servant/dist-newstyle/build/x86_64-linux/ghc-9.2.1/servant-0.18.3/noopt/build/Servant/Links.o, /home/guillaume/srcs/servant/dist-newstyle/build/x86_64-linux/ghc-9.2.1/servant-0.18.3/noopt/build/Servant/Links.dyn_o )

src/Servant/Links.hs:652:10: error:
    • Uninferrable type variable k0 in
      type family equation right-hand side: (TypeError ...)
    • In the type instance declaration for ‘MkLink’
      In the instance declaration for ‘HasLink ((arr :: a -> b) :> sub)’
    |
652 |     type MkLink (arr :> sub) _ = TypeError (PartialApplication HasLink arr)
    |          ^^^^^^
cabal: Failed to build servant-0.18.3 (which is required by test:spec from
servant-0.18.3).
```

Associated with the following code:

```haskell
instance TypeError (PartialApplication HasLink arr) => HasLink ((arr :: a -> b) :> sub)
  where
    type MkLink (arr :> sub) _ = TypeError (PartialApplication HasLink arr)
    toLink = error "unreachable"
```

I solved the problem by just removing the instance (That's just better error, isn't it?.)

It may or may not come from: https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2#kind-inference-for-data-family-instances-is-pickier

guibou:
This may also be related: https://gitlab.haskell.org/ghc/ghc/-/issues/20675 however it discuss about the activation of `PolyKinds` which is already activated for this file.

guibou:
```diff
--- a/servant/src/Servant/Links.hs
+++ b/servant/src/Servant/Links.hs
@@ -649,7 +649,7 @@ simpleToLink _ toA _ = toLink toA (Proxy :: Proxy sub)
 -- Erroring instance for 'HasLink' when a combinator is not fully applied
[...]

PR #1525, opened 2022-02-04
servant-server: Fix build on GHC 9.2.1

k0001:
The issue is similar to the one in #1513:

```
src/Servant/Server/Internal.hs:824:10: error:
    • Uninferrable type variable k0 in
      type family equation right-hand side: (TypeError ...)
    • In the type instance declaration for ‘ServerT’
      In the instance declaration for
        ‘HasServer ((arr :: a -> b) :> sub) context’
    |
824 |     type ServerT (arr :> sub) _ = TypeError (PartialApplication HasServer arr)
    |
```

This fix is similar to the one in #1514.


ysangkok:
I think it is good idea to apply smaller patches like this instead of a giant PR with lots of `allow-newer`.

gdeest:
Thanks !

PR #1529, opened 2022-02-11
Fix servant-client-ghcjs for servant 0.19

mikatammi:
I fixed servant-client-ghcjs to be compatible with servant-client-core version 0.19, also bumped max-bound on base to be able to use base 4.14


tchoutri (approved):
LGTM

PR #1533, opened 2022-02-15
feat(SourceT): add fromConsumableActionStep that creates a StepT from…

### LINKED ISSUE #1448: Make `fromAction` and `fromActionStep` able to pass a value from an action to the next one
## Context
On `servant-server`, I must serve some data that I query from a store. The persistance interface returns me some [`Stream`](https://hackage.haskell.org/package/streaming-0.2.3.0/docs/Streaming-Prelude.html) of resources, so actual queries to the store are done only when the stream is consumed.

### Unsatisfactory solution
Fold the stream into a list and respond with that list in a normal `Servant` endpoint. This is unsatisfactory because by folding the stream, all resources are read from the persistance layer and accumulated in memory before being sent, which leads me to huge memory usage.

## Objective
Pipe the resources directly from the `Prelude.Streaming.Stream` into a `Servant.Types.SourceT`, so that the persistance layer is queried only when the resource must be sent in the response stream, and thus avoid to accumulate many resources in memory.

To achieve this, I can generate a `SourceIO` with a [`fromAction`](https://hackage.haskell.org/package/servant-0.18.3/docs/Servant-Types-SourceT.html#v:fromAction) or `fromActionStep`, that reads the next element of the stream.

## Issue
Since all the actions run by `fromAction` are actually the same, I cannot consume the stream by accessing the next values.

## Solution
Tweak `fromActionStep` to make it pass a value from an action to the next one:
```haskell
fromActionStep' :: Functor m => (c -> m (Maybe (a,c))) -> c -> StepT m a
fromActionStep' action = loop where
  loop c = Effect $ step <$> action c
  step Nothing = Stop
  step (Just (x,t)) = Yield x $ loop t
```
And then call it on `Prelude.Streaming.uncons`.

Of course, `Maybe (a,c)` can be generalized to any type `v`, provided that we pass additionally some `stop :: v -> Bool`, `value :: v -> a` and `rest :: v -> v` functions.

## Question
Was there an easier out-of-the-box solution I did not see?

Otherwise, is it worth adding such `fromActionStepWithConsumable` method to the library?

giulioforesto:
… an action that consumes some data

Closes #1448

### DIFF
--- a/changelog.d/generalize-fromActionStep-to-consumable
+++ b/changelog.d/generalize-fromActionStep-to-consumable
@@ -1,0 +1,22 @@
+synopsis: Add a `StepT` constructor method that consumes a resource
+prs: #1533
+issues: #1448
+
+description: {
+
+`fromActionStep` always runs the same action, which makes it impossible with such monadic actions to
+"consume" a resource (that is to say to pass the modified resource to the following action), or
+"unfold" an input structure.
+
+`unfoldStep` gives this possibility.
+
+This allows for example to build a `StepT m` directly from a `Streaming.Prelude.Stream`, by passing
+`unfoldStep` the following argument:
+
+```haskell
+import Streaming.Prelude as S
+
+action :: Stream (Of a) m r -> m (Maybe ( a, Stream (Of a) m r ))
+action = S.uncons
+```
+}
--- a/servant/src/Servant/Types/SourceT.hs
+++ b/servant/src/Servant/Types/SourceT.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
 module Servant.Types.SourceT where
 
 import           Control.Monad.Except
@@ -311,6 +312,28 @@ fromActionStep stop action = loop where
         | stop x    = Stop
         | otherwise = Yield x loop
 {-# INLINE fromActionStep #-}
+
+-- | Create a `StepT' from a consumable @c@, that is to say from an input and an action that returns
+-- and is called again on an updated version of that input.
+--
+-- >>> import qualified Streaming.Prelude as S
+-- >>> foreachStep mempty print (unfoldStep S.uncons $ S.each [1..3] :: StepT IO Int)
+-- 1
+-- 2
+-- 3
+-- 
+unfoldStep :: Functor m
+                     => (c -> m (Maybe (a,c)))
+                     -- ^ Action. Return @Nothing@ to stop or @Just (a,c)@ where @a@ is the
+                     -- output element of the action and @c@ the updated input
+                     -> c
+                     -- ^ Input
+                     -> StepT m a
+unfoldStep action = loop where
+  loop c = Effect $ step <$> action c
+  step Nothing = Stop
+  step (Just (x,t)) = Yield x $ loop t
+{-# INLINE unfoldStep #-}
 
 -------------------------------------------------------------------------------
 -- File


changelog.d/generalize-fromActionStep-to-consumable
```diff
@@ -0,0 +1,21 @@
+synopsis: Add a `StepT` constructor method that consumes a resource
+prs:
```
alpmestan, line 2: you can put #1533 in this field :)

servant/src/Servant/Types/SourceT.hs
```diff
@@ -305,13 +306,28 @@ fromAction stop action = SourceT ($ fromActionStep stop action)
...
+fromActionStep stop action = fromConsumableActionStep action' () where
+  action' = const $ flip fmap action $
+    \x -> if stop x
+          then Nothing
+          else Just (x,())
 {-# INLINE fromActionStep #-}
 
+-- | Create a `StepT' from a consumable @c@, that is to say from an input and an action that returns
+-- and is called again on an updated version of that input.
+fromConsumableActionStep :: Functor m
```
alpmestan, line 318: I like the function, but the type immediately brings `unfold`-style functions in mind. This makes me think some name based around `unfold` might be clearer here? "Consumable" on the contrary doesn't really have a standard intuition we can hang onto.

servant/src/Servant/Types/SourceT.hs
```diff
@@ -305,13 +306,28 @@ fromAction stop action = SourceT ($ fromActionStep stop action)
 {-# INLINE fromAction #-}
 
 fromActionStep :: Functor m => (a -> Bool) -> m a -> StepT m a
-fromActionStep stop action = loop where
-    loop = Effect $ fmap step action
-    step x
-        | stop x    = Stop
-        | otherwise = Yield x loop
+fromActionStep stop action = fromConsumableActionStep action' () where
```
alpmestan, line 309: I'd probably leave the definition we had before, as it didn't allocate a bunch of pairs with `()`s in them, which might start mattering when building up large streams.

servant/src/Servant/Types/SourceT.hs
```diff
@@ -305,13 +306,28 @@ fromAction stop action = SourceT ($ fromActionStep stop action)
...
+  action' = const $ flip fmap action $
+    \x -> if stop x
+          then Nothing
+          else Just (x,())
 {-# INLINE fromActionStep #-}
 
+-- | Create a `StepT' from a consumable @c@, that is to say from an input and an action that returns
+-- and is called again on an updated version of that input.
```
alpmestan, line 317: Could you perhaps include a tiny teeny example in the haddocks here? That just unfolds a stream of 10 numbers or something.

alpmestan (changes requested):
Thanks for the patch!

LGTM, just a little bit of nitpicking.

servant/src/Servant/Types/SourceT.hs
giulioforesto: Thank you. Is `fromUnfoldActionStep` better?

servant/src/Servant/Types/SourceT.hs
giulioforesto: It seems that the `doctest` suite passes even if I write faulty examples, so I'm not sure. But I checked in a `ghci` and this should be ok.

servant/src/Servant/Types/SourceT.hs
alpmestan: Yes that is better IMO. I'd perhaps even drop `from` and (`Action` or `Step`). `unfoldStep`? `unfoldAction`?

giulioforesto:
> This seems ready to land, modulo the hardest part: the name of your function :-)
> 
> I think I'm warming up to `unfoldStep`.

Haha go for `unfoldStep` then. I trust you, I don't master all the subtleties of Haskell function naming yet ;)

alpmestan (approved):

ysangkok:
@giulioforesto The doctest isn't passing for me. Was this only meant as a hint/suggestion, or was it supposed to be actually be an executable doctest?

PR #1534, opened 2022-02-15
Cookbook named routes

akhesaCaro:
Introducing `NamedRoutes` cookbook


tchoutri:
Do you think it should replace the previous generics cookbook? (I think it should but I'd like your opinion as well)

akhesaCaro:
> Do you think it should replace the previous generics cookbook? (I think it should but I'd like your opinion as well)

It think it should yes. 
I will remove it from the repo.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,314 @@
...
+                <:|>  "movies" :> Get '[JSON] [Movie]
+```
+
+does not follow the `Handler` implementation order
+
+```haskell,ignore
+apiHandler :: (AsServerT Handler)
+apiHandler =   getMovies
+                 <:|> getVersion
+```
```
arianvp, line 29: Maybe for illustration add the annoying error message that you'd get in this case?

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,314 @@
...
+```haskell,ignore
+data API mode = API
+    { *version* :: "version" :> Get '[JSON] Version
+    , **movies* ::  "movies" :> …..
+    }
+```
+
+So, it is more readable for a human and helps GHC to give you more accurate error messages.
+
+[TODO GHC error examples]
```
arianvp, line 42: Ah you already mention that here. NVM :')

doc/cookbook/namedRoutes/NamedRoutes.lhs
akhesaCaro: I totally agree with that :)

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
+# NamedRoutes - Using records to define APIs
+
+Servant offers a very natural way of constructing APIs with nested records, called `NamedRoutes`.
```
gdeest, line 3: Maybe mention that it is available since 0.19.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+
+This cookbook explains how to implement such nested-record-APIs using `NamedRoutes` through the example of a Movie Catalog.
+
+First, we start by constructing the domain types of our Movie Catalog.
+After, we show you how to implement the API type with the NamedRoutes records.
+Lastly, we make a Server and a Client out of the API type.
+
+However, it should be understood that this cookbook does _not_ dwell on the built-in servant combinators as the [<Structuring APIs> cookbook ](<https://docs.servant.dev/en/stable/cookbook/structuring-apis/StructuringApis.html>) already covers that angle.
+
+## Why should we use `NamedRoutes` over the alternative `<:|>` operator?
```
gdeest, line 13: I'd tone that down a trifle:

```suggestion
## Why would I want to use `NamedRoutes` over the alternative `:<|>` operator?
```

(Also note the typo fix for `:<|>`)

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+
+```haskell,ignore
+type API1  =   "version" :> Get '[JSON] Version
+                :<|>  "movies" :> Get '[JSON] [Movie]
+```
+
+does not follow the `Handler` implementation order
+
+```haskell,ignore
+apiHandler :: (AsServerT Handler)
```
gdeest, line 26: I think you meant:

```suggestion
apiHandler :: ServerT API1 Handler
```

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+                      getMovieHandler :<|> updateMovieHandler :<|> deleteMovieHandler
+    |
+226 | server = versionHandler
+```
+
+On the contrary, with the `NamedRoutes` technique, we refer to the routes by their name:
+
+```haskell,ignore
+data API mode = API
+    { *list*   :: "list" :> ...
```
gdeest, line 63: Unfortunately, I don't think the asterisks will put the text in bold within a code block ­— they will likely just appear verbatim.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+``` haskell
+data Movie = Movie
+    { movieId :: MovieId
+    , title :: Title
+    , year :: Year
+    }
+    deriving stock Generic
+    deriving anyclass (FromJSON, ToJSON)
+
+type MovieId = String -- Very bad way of implementing an ID
```
gdeest, line 145:
```suggestion
type MovieId = String
```

It's just an example. It doesn't really matter.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+            Capture MovieId - Put Movie
+                            \
+                              Delete MovieId
+
+In this example, we create a very simple endpoint for the Version,
+and several complex endpoints that use nested records for the CRUD part of the movie.
+
+So, the URLs would look like
+
+- GET .../version
```
gdeest, line 169:
```suggestion
- GET …/version
```

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+    { version :: mode :- "version" :> Get '[JSON] Version
+    , movies :: mode :- "movies" :> NamedRoutes MoviesAPI
+    } deriving stock Generic
+
+type Version = String -- This will do for the sake of example.
+
+```
+Here, we see the first node of our tree. It contains the two branches “version” and “movies” respectively:
+
+The “version” branch is very simple and sufficient.
```
gdeest, line 191:
```suggestion
The “version” branch is very simple and self-explanatory.
```

What do you think ?

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+The “movies” branch will contain another node, represented by another record (see above). That is why we need the `NameRoutes` helper.
+
+Note:
+
+The `mode` type parameter indicates into which implementation the record’s `Generic` representation will be transformed–as a client or as a server. We will discuss that later.
```
gdeest, line 196:
```suggestion
The `mode` type parameter indicates into which implementation the record’s `Generic` representation will be transformed — as a client or as a server. We will discuss that later.
```

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+
+That's it, we have our `MovieCatalogAPI` type!
+
+Let's make a server and a client out of it!
+
+### The Server
+
+As you know, we can’t talk about a server, without addressing the handlers.
+
+First, we take our handlers…
```
gdeest, line 251:
```suggestion
First, we write our handlers…
```

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+movieHandler :: MovieId -> MovieAPI (AsServerT Handler)
+movieHandler movieId = MovieAPI
+    { get = getMovieHandler movieId
+    , update = updateMovieHandler movieId
+    , delete = deleteMovieHandler movieId
+    }
+```
+As you might have noticed, we assemble our handlers with the same API’s records: `MoviesAPI` and `MovieAPI`.
```
gdeest, line 310:
```suggestion
As you might have noticed, we build our handlers out of the same record types we used to define our API: `MoviesAPI` and `MovieAPI`. What kind of magic is this ?
```

doc/cookbook/namedRoutes/NamedRoutesOperator.lhs
```diff
@@ -0,0 +1,259 @@
...
+server = versionHandler
+      :<|> movieListHandler
+      :<|> getMovieHandler
+      :<|> updateMovieHandler
+      :<|> deleteMovieHandler
+
+```
+As you might have noticed, we assemble our handlers with the same API’s records: `MoviesAPI` and `MovieAPI`.
+
+Remember the `mode` type parameter we saw earlier? Since we need to transform our API type into a _serve_r, we need to provide a server `mode`, which is `AsServerT` _handler_ here.
```
gdeest, line 235:
```suggestion
Remember the `mode` type parameter we saw earlier? Since we need to transform our API type into a _server_, we need to provide a server `mode`, which is `AsServerT Handler` here.
```

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,371 @@
...
+type MovieId = String
+type Title = String
+type Year = Int
+
+```
+
+
+Let’s forget about the deriving stuff for now and think about the API that we want to make.
+
+  "version" -> Get Version
```
gdeest, line 156: Now that I think of it, we probably want this diagram in a code block (otherwise it won't be formatted properly).

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,371 @@
...
+
+First, we start by constructing the domain types of our Movie Catalog.
+After, we show you how to implement the API type with the NamedRoutes records.
+Lastly, we make a Server and a Client out of the API type.
+
+However, it should be understood that this cookbook does _not_ dwell on the built-in servant combinators as the [<Structuring APIs> cookbook ](<https://docs.servant.dev/en/stable/cookbook/structuring-apis/StructuringApis.html>) already covers that angle.
+
+## Why would I want to use `NamedRoutes` over the alternative `:<|>` operator?
+
+With `NamedRoutes`, we don’t need to care about the declaration order of the endpoints.
```
alpmestan, line 17: I'd say just go straight into "this gives better errors". And perhaps you can indeed mention that we can afford not to care about order in the "record API" scenario, by using record syntax to construct the record of handlers, for example. `ConstructorName f1 f2` (order matters) vs `ConstructorName { field2 = f2, field1 = f1 }` (order doesn't matter, names do). But only after the better, "per-field" errors, IMO.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,371 @@
...
+            = versionHandler
+                :<|>
+                  movieListHandler
+                    :<|>
+                      getMovieHandler :<|> updateMovieHandler :<|> deleteMovieHandler
+    |
+226 | server = versionHandler
+```
+
+On the contrary, with the `NamedRoutes` technique, we refer to the routes by their name:
```
alpmestan, line 61: Not about `NamedRoutes` per se, but "just" a virtue of the record-based API specs.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,371 @@
...
+          movieHandler movieId
+            = MovieAPI
+                {get = getMovieHandler movieId,
+                 update = updateMovieHandler movieId,
+                 delete = deleteMovieHandler movieId}
+    |
+252 |     , delete = deleteMovieHandler movieId
+```
+
+So, NamedRoutes is more readable for a human, and GHC gives you more accurate error messages.
```
alpmestan, line 92: API records* (or whatever we want to call this)

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+    , movies :: mode :- "movies" :> NamedRoutes MoviesAPI
+    } deriving stock Generic
+
+type Version = String -- This will do for the sake of example.
+
+```
+Here, we see the first node of our tree. It contains the two branches “version” and “movies” respectively:
+
+The “version” branch is very simple and self-explanatory.
+The “movies” branch will contain another node, represented by another record (see above). That is why we need the `NameRoutes` helper.
```
alpmestan, line 196: `NamedRoutes` *

alpmestan: I'd also probably suggest expanding with another sentence, along the lines of: "It allows us to reference a record-based API where a 'standard' API is expected."

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+
+Note:
+
+The `mode` type parameter indicates into which implementation the record’s `Generic` representation will be transformed—as a client or as a server. We will discuss that later.
```
alpmestan, line 200: s/into//

If I may suggest some rewording:

> The `mode` type parameter will eventually indicate how to interpret the (API) type of each of these fields: a handler for that endpoint or group of endpoints, clients, link-making functions? The choice of `mode` is how we pick which interpretation we're interested in, and `:-` will apply the right type-level computations to turn the endpoint spec into a suitable handler/client/etc type for the corresponding field of your record.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+  } deriving stock Generic
+```
+
+As you can see, we end up implementing the deepest routes of our API.
+
+Small detail: as our main API tree is also a record, we need the `NamedRoutes` helper.
+To improve readability, we suggest you create a type alias:
+
+``` haskell
+type MovieCatalogAPI = NamedRoutes API
```
alpmestan, line 244: Why articulate everything around a top-level `NamedRoute` instead of using `genericServe` and friends? There is a way for users to just not care about "the old way" at all. Perhaps a final section could explain how to convert between "old style" and "record style" (back and forth) at the level of API types, server handlers, clients, etc (all the `fromServant`/`toServant` business). I can write that if needed.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+deleteMovieHandler _ =
+   -- delete the movie from the database...
+   pure NoContent
+
+```
+
+And assemble them together with the record structure, which is the glue here.
+
+```haskell
+server :: ServerT MovieCatalogAPI  Handler
```
alpmestan, line 293: I'd probably have the type be `API AsServer`, to keep the record-centric view all along, as per my comment about articulating things around the "old style". I won't repeat it but this obviously applies to the rest of this example.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+movieHandler :: MovieId -> MovieAPI (AsServerT Handler)
+movieHandler movieId = MovieAPI
+    { get = getMovieHandler movieId
+    , update = updateMovieHandler movieId
+    , delete = deleteMovieHandler movieId
+    }
+```
+As you might have noticed, we build our handlers out of the same record types we used to define our API: `MoviesAPI` and `MovieAPI`. What kind of magic is this ?
```
alpmestan, line 314: Might just be me, but I think "magic" and "servant" are already linked together too often, and not in a positive way I'd say, so I suggest replacing that final sentence. I'd probably instead go for "What's going on?" or "How does that work?".

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+
+Remember the `mode` type parameter we saw earlier? Since we need to transform our API type into a _server_, we need to provide a server `mode`, which is `AsServerT Handler` here.
```
alpmestan, line 316: Probably worth adding a little sentence or two here:

> You can alternatively use the `AsServer` (= `AsServerT Handler`) type alias. If you need to define handlers in some specific `App` monad from your codebase, the `mode` would simply be changed to `AsServerT App`.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+
+``` haskell
+api :: Proxy MovieCatalogAPI
+api = Proxy
+
+main :: IO ()
+main = run 8081 app
+
+app :: Application
+app = serve api server
```
alpmestan, line 328: I'd suggest using `genericServe`, as suggested in earlier comments. With a comment saying that there's `genericServeT` for when using a custom monad.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+
+```
+Yay! That’s done and we’ve got our server!
+
+## The Client
+
+The client, so to speak, is very easy to implement:
+
+``` haskell
+movieCatalogClient :: API (AsClientT ClientM)
```
alpmestan, line 338: Don't we have an `AsClient` type synonym, for when the client monad is the standard `ClientM`?

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+movieCatalogClient = client api -- remember: api: Proxy MovieCatalogAPI
```
alpmestan, line 339: `api :: Proxy MovieCatalogAPI`

alpmestan: And also, in line with what I said about the server side, I'd recommend using `genericClient` etc.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,373 @@
...
+
+deleteMovie :: MovieId -> ClientM NoContent
+deleteMovie movieId = movieCatalogClient // movies // movie /: movieId // delete
+```
+
+Done! We’ve got our client!
+
+## Conclusion
+
+We hope that you found this workbook helpful, and that you now feel more confident using the `NamedRoutes` technique.
```
alpmestan, line 369: cookbook*

And the wrap up should probably say "confident about using record-based APIs, nested or not".

alpmestan (commented):
This is a great start! I have some feedback on the contents of the cookbook recipe, but there's a larger topic I'd like to bring up: if this replaces the `generic` cookbook, it must cover the "non-nested" cases in a first section, otherwise the only examples people will see of generic-based API specs will be the nested subAPIs ones.

Along these lines, I'd suggested stealing the "generic" name, or perhaps finding something even clearer, "records", it's a lot more to the point: we specify APIs as records, that's what this is all about, and `NamedRoutes` is just one of those things that we have to get to know to use "API records" effectively. So a cookbook that describes that, through 1/ simple case, no nesting (possibly stealing lots of contents from `generic`, if appropriate) 2/ nested sub-APIs like described in this new cookbook recipe, would probably be the most consistent way to present this. Or should this even be moved to a final, dedicated section in the tutorial? To promote it a bit more.

PR #1535, opened 2022-02-15
Remove unnecessary constraint/allow-newer

ysangkok:
Removing these bounds adjustments doesn't seem to impede building for any GHC version. The
disadvantage of them is that they make people not notice when bounds need
updating, which will make each project just replicate the same allow-newer
list. Better remove them when unnecessary such that a potential future upstream issue can be
found.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -52,28 +52,9 @@ tests: True
 optimization: False
 -- reorder-goals: True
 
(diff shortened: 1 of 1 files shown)


gdeest (approved):
I am all for it. Thanks !

Issue #1540, opened 2022-02-19
Tutorial installation instructions didn't work


mjdominus:
The instructions at https://docs.servant.dev/en/stable/tutorial/install.html failed for me in multiple ways.

## cabal-install

The `cabal new-build tutorial` command failed, saying 

          /home/mjd/src/haskell-job/servant/cabal.project:59:
          Parse of field 'allow-newer' failed.

I figured out that this was because the Ubuntu Cabal package is from the stone age and installed cabal-3.4.  But it still failed:

```
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cookbook-basic-auth-0.1 (user goal)
[__1] trying: base-4.9.1.0/installed-4.9.1.0 (dependency of
cookbook-basic-auth)
[__2] next goal: servant-auth (user goal)
[__2] rejecting: servant-auth-0.4.1.0 (conflict:
base==4.9.1.0/installed-4.9.1.0, servant-auth => base>=4.10 && <4.16)
[__2] skipping: servant-auth-0.4.0.0 (has the same characteristics that caused
the previous version to fail: excludes 'base' version 4.9.1.0)
[__2] rejecting: servant-auth-0.3.2.0, servant-auth-0.3.1.0,
servant-auth-0.3.0.1, servant-auth-0.3.0.0, servant-auth-0.2.7.0,
servant-auth-0.2.6.1, servant-auth-0.2.6.0, servant-auth-0.2.1.0,
servant-auth-0.2.0.0, servant-auth-0.1.0.1, servant-auth-0.1.0.0 (constraint
from user target requires ==0.4.1.0)
[__2] fail (backjumping, conflict set: base, servant-auth)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: base, servant-auth,
cookbook-basic-auth
```

I didn't know how to proceed from there so I moved on to the `stack` instructions.

## stack

The instructions say to do `stack new myproj servant`.  I got this error:

```
Downloading lts-18.25 build plan ...RedownloadFailed Request {
  host                 = "raw.githubusercontent.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/fpco/lts-haskell/master//lts-18.25.yaml"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
[...]

PR #1541, opened 2022-02-21
Update documentation

mjdominus:
1. Add a note to the tutorial install doc that addresses problems caused by the ancient toolchain on Ubuntu (resolves issue #1540)
2. Update error message in `doc/Makefile`
3. Update copyright date from 2018 to 2022

### DIFF
--- a/doc/tutorial/install.rst
+++ b/doc/tutorial/install.rst
@@ -40,3 +40,29 @@ nix
 `Nix <https://nixos.org/nix/>`_ users should feel free to take a look at
 the `nix/shell.nix` file in the repository and use it to provision a suitable
 environment to build and run the examples.
+
+Note for Ubuntu users
+--------
+
+Ubuntu's packages for `ghc`, `cabal`, and `stack` are years out of date.
+If the instructions above fail for you,
+try replacing the Ubuntu packages with up-to-date versions.
+First remove the installed versions:
+
+.. code-block:: bash
+
+   # remove the obsolete versions
+   $ sudo apt remove ghc haskell-stack cabal-install
+
+Then install fresh versions of the Haskell toolchain
+using the `ghcup <https://www.haskell.org/ghcup/install/>`_ installer.
+
+As of February 2022, one easy way to do this is by running a bootstrap script:
+
+.. code-block:: bash
+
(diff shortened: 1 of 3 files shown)


doc/tutorial/install.rst
```diff
@@ -40,3 +40,24 @@ nix
...
+   # remove the obsolete versions
+   $ sudo apt remove ghc haskell-stack cabal-install
+
+Then install fresh versions of the Haskell toolchain
+using the `ghcup <https://www.haskell.org/ghcup/install/>`_ installer.
+As of February 2022, an easy way to do this is by running a bootstrap script:
+
+.. code-block:: bash
+
+   $ curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
```
alpmestan, line 63: I'd either leave out all commands and point to some official ghcup instructions, or additionally include one that shows how to install `ghc` and `¢abal`/`stack` when `ghcup` is installed. This seems to sit between the two approaches without that last final step =)

alpmestan (approved):
LGTM, modulo a tiny teeny nitpick.

Thanks!

doc/tutorial/install.rst
mjdominus: I considered both of those options and chose this specifically.  Regarding #1, I really dislike instructions that say “now go off and perform an unknown number of operations as directed by a completely different web site”; I like to get some idea of what is coming next and whether there is an easy option.  Often the target web site has been reorganized and it's not clear where to find further details.

Regarding #2, the indicated installer is conversational.  It prompts the user for what to do next, so there's no reason to try to embed its output here.  Any such embedded output would be soon out of date.

Perhaps it would suffice to add a single sentence that explains “from there the installation process is interactive”.

alpmestan: Yes that final sentence would be very helpful I think, I had forgotten the interactive part with `ghcup\ even though I've already used it a few times, for example =)

Thanks!

mjdominus: Updated per your suggestion.

alpmestan (approved):

gdeest (approved):

PR #1542, opened 2022-02-22
NamedRoutes Cookbook

akhesaCaro:


doc/cookbook/generic/Generic.lhs
```diff
@@ -1,4 +1,14 @@
-# Using generics
+# Record-based APIs: the simple case
+
+This cookbook explains how to implement an API with a simple record-based
+structure. We only deal with non-nested APIs in which every endpoint is on the same
+level.
+
+If a you need nesting because you have different branches in your API tree, you
```
alpmestan, line 7: "If a" => "If"

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
+# Record-based APIs: the nested records case
```
alpmestan, line 1: nested API*

It does indeed result in "nested records", but this merely mirrors the structure of the API, which is the central element that should be in the title IMO

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
...
+
+*Available in Servant 0.19 or higher*
+
+Servant offers a very natural way of constructing APIs with nested records, called `NamedRoutes`.
```
alpmestan, line 5: Do you think it'd be good to clarify what "nested" means here? Describing the general pattern dividing up APIs in sub-APIs and being able to combine those easily, etc.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
...
+
+First, we start by constructing the domain types of our Movie Catalog.
+After, we show you how to implement the API type with the NamedRoutes records.
+Lastly, we make a Server and a Client out of the API type.
+
+However, it should be understood that this cookbook does _not_ dwell on the
+built-in servant combinators as the [Structuring APIs
+](<../structuring-apis/StructuringApis.html>) cookbook already covers that angle.
+
+## Why would I want to use `NamedRoutes` over the alternative `:<|>` operator?
```
alpmestan, line 23: I think this whole section can go. Or at least be integrated into the `generic` cookbook. This has nothing to do with `NamedRoutes` per se but the record approach, so would be a great fit for the `generic` cookbook. By the time we learn about the nested case and how to handle it, it's in a way "too late" to explain basics about records vs `:<|>`-separated stuffs, I think.

("This section" = until "Boilerplate time!")

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
...
+```
+As you might have noticed, we build our handlers out of the same record types we used to define our API: `MoviesAPI` and `MovieAPI`. What kind of magic is this ?
+
+Remember the `mode` type parameter we saw earlier? Since we need to transform our API type into a _server_, we need to provide a server `mode`, which is `AsServerT Handler` here.
+
+You can alternatively use the AsServer (= AsServerT Handler) type alias. If you need to define handlers in some specific App monad from your codebase, the mode would simply be changed to AsServerT App.
+
+Finally, we can run the server and connect the API routes to the handlers as usual:
+
+``` haskell
```
alpmestan, line 330: don't we want to use `genericServe` here? to stay in the "record world".

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
...
+```
+Yay! That’s done and we’ve got our server!
+
+## The Client
+
+The client, so to speak, is very easy to implement:
+
+``` haskell
+movieCatalogClient :: API (AsClientT ClientM)
+movieCatalogClient = client api -- remember: api :: Proxy MovieCatalogAPI
```
alpmestan, line 349: `genericClient` ?

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+movieHandler :: MovieId -> MovieAPI AsServer
+movieHandler movieId = MovieAPI
+    { get = getMovieHandler movieId
+    , update = updateMovieHandler movieId
+    , delete = deleteMovieHandler movieId
+    }
+```
+As you might have noticed, we build our handlers out of the same record types we used to define our API: `MoviesAPI` and `MovieAPI`. What kind of magic is this ?
```
alpmestan, line 322: Might fit well in the `generic` cookbook? Where we introduce the whole record-API machinery.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,383 @@
...
+## The Client
+
+The client, so to speak, is very easy to implement:
+
+``` haskell
+movieCatalogClient :: API (AsClientT ClientM)
+movieCatalogClient = client api -- remember: api :: Proxy MovieCatalogAPI
+```
+
+Have you noticed the `mode` `AsClient ClientM`?
```
alpmestan, line 352: Those explanations and tips, do you think they might fit well in the generic cookbook? Where we introduce the whole record-API machinery.

doc/cookbook/namedRoutes/NamedRoutes.lhs
arianvp: There's no need as `NamedRoutes` has a `HasServer` instance

gdeest: I really don't see the point of introducing `genericServe` here — it makes the overall story less uniform.

We checked yesterday, and `genericServe` produces the exact same error message as `serve` if `Generic` instances are missing (a relatively unhelpful message mentioning `Generic`) so we are not gaining anything.

doc/cookbook/namedRoutes/NamedRoutes.lhs
gdeest: The two cookbooks should IMHO be merge into a single cookbook, the non-nested / flat API case being the first section, and the nested records one expanding upon the first. So we can move this section a bit up when we do so ; for now I think it is better if the `NamedRoutes` cookbook remains standalone.

doc/cookbook/namedRoutes/NamedRoutes.lhs
alpmestan: Well that forces people to use `NamedRoutes` at the top, which is completely unnecessary if your entire API is made of records. The most direct way to serve a record of handlers is `genericServe`, hence my suggestion.

It'd be good to have some place that really explains those ":<|> separated" and "record" worlds and how we can go from one to the other, at all levels. But in the meantime, in the interest of giving the simplest code possible in the cookbook, without having to explain that named routes is an "adaptator" of sorts to plug records into old-style APIs, I'd recommend personally recommend `genericServe` (but do as you wish, this is just my opinion).

doc/cookbook/namedRoutes/NamedRoutes.lhs
alpmestan: That is what I have advocated in favor of as well :-) And even putting the resulting thing in the tutorial. But in the interest of splitting that up into a bunch of much smaller steps, we're first consolidating the named routes one to make it stand as a nice follow up to the `generic` cookbook. Then slowly but surely proceeding with the rest of the plan.

> for now I think it is better if the NamedRoutes cookbook remains standalone
We'd end up with 2 cookbooks which half-explain the same basics, before covering simple APIs and nested APIs respectively. Might not be the best option for users, no?

doc/cookbook/namedRoutes/NamedRoutes.lhs
gdeest: I think:

```haskell
serveGeneric (Proxy @MyApi) myServer
```

isn't so much shorter than:

```haskell
serve (Proxy @(NamedRoutes MyAPI)) myServer
```

as to warrant the introduction of another function in the cookbook. It gives off the feeling that record APIs still somewhat live in a different world — which they now don't.

alpmestan: That function is already introduced in the introductory cookbook: https://docs.servant.dev/en/stable/cookbook/generic/Generic.html#server

(and even the `ServerT` version)

And requires appending a few chars to `serve` vs another pair of parens and the `NamedRoutes` thing.

But I also suggested to @akhesaCaro at some point that another possibility could just be to pick one and show the equivalent with the other, in a commented-out (or not, could just be with a different name) line right below.

doc/cookbook/namedRoutes/NamedRoutes.lhs
akhesaCaro: Generic cookbook should have clarified this beforehand.

doc/cookbook/namedRoutes/NamedRoutes.lhs
akhesaCaro: I agree, it doesn't belong here.. We will find where to put it in another PR.

doc/cookbook/namedRoutes/NamedRoutes.lhs
```diff
@@ -0,0 +1,296 @@
...
+
+Done! We’ve got our client!
+
+## Conclusion
+
+We hope that you found this cookbook helpful, and that you now feel more confident using the record-based APIs, nested or not.
+
+If you are interested in further understanding the built-in Servant combinators, see [Structuring APIs](../structuring-apis/StructuringApis.html).
+
+Since `NamedRoutes` is based on the Generic mechanism, you might want to have a look at [Sandy Maguire’s _Thinking with Types_ book](https://doku.pub/download/sandy-maguire-thinking-with-typesz-liborgpdf-4lo5ne7kdj0x).
```
ysangkok, line 296: Not nice to put a piracy link. Better link the official website: https://thinkingwithtypes.com/

tchoutri:
Superceded by https://github.com/haskell-servant/servant/pull/1726

PR #1543, opened 2022-02-22
Remove overlapping instance `Verb … (Headers x a)`

gdeest:
It is possible to apply a bit of elbow grease to remove some of the overlapping instances in the codebase. Those instances often go in the way of refactoring, and sometimes creep up to the user in confusing messages, so it would be helpful to get rid of them.

The main challenge is backward-compatibility: we want to avoid breaking our users code, so we cannot change the API.

This PR experiments with an approach to do so, using a closed type-family definition. Closed type family instances are ordered, which lets us work around the need for overlapping instances.


tchoutri:
@gdeest what's the state of this PR? Shall we wait for 0.21.0.0 or is it ok for 0.20.3.0?

gdeest:
@tchoutri I had completely forgotten about that PR ... :-) I'd say it is urgent to wait and experiment a bit more with the approach before moving forward.

I do remember overlapping instances in `servant` being a hindrance to many of my type-level endeavors, though, so it is probably worth considering.

tchoutri:
@gdeest perfect, I'd be more than happy to pair with you on it in order to fully understand the approach. :)

PR #1548, opened 2022-03-01
fix broken links

marcosh:
some links which appear on Hackage point to non-existing pages. This PR should fix them


tchoutri (approved):
Thanks a lot!

Issue #1532, opened 2022-02-15
VerbNoContent should allow status codes to be set not hardcode to 204


arianvp:
e.g. 301, 302, 303 are perfectly valid status codes for a Get with No Content.

alpmestan:
Agreed, want to make the patch? We can still have some default definition(s?) that hardcore the status to 204, but indeed we don't want the general construct to hardcode that.

PR #1550, opened 2022-03-03
Refactor NoContentVerb into NoContentVerbWithStatus (#1532)

nbacquey:
There are several HTTP status codes that correspond to a response body with `NoContent`. This PR introduces `NoContentVerbWithStatus` which generalizes `NoContentVerb` to cases when the return status may be
different from 204. The former replaces the latter anywhere possible.    
`NoContentVerb` is kept as a special case of `NoContentVerbWithStatus` for backwards compatibility.

This PR also adds a test case for `NoContentVerbWithStatus` in `ServerSpec.hs`

### DIFF
--- a/servant/src/Servant/API/Verbs.hs
+++ b/servant/src/Servant/API/Verbs.hs
@@ -29,11 +29,16 @@ import           Network.HTTP.Types.Method
 data Verb (method :: k1) (statusCode :: Nat) (contentTypes :: [*]) (a :: *)
   deriving (Typeable, Generic)
 
--- | @NoContentVerb@ is a specific type to represent 'NoContent' responses.
--- It does not require either a list of content types (because there's
--- no content) or a status code (because it should always be 204).
-data NoContentVerb  (method :: k1)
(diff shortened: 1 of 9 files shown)


servant/src/Servant/API/Verbs.hs
```diff
@@ -29,12 +29,17 @@ import           Network.HTTP.Types.Method
...
   deriving (Typeable, Generic)
 
--- | @NoContentVerb@ is a specific type to represent 'NoContent' responses.
--- It does not require either a list of content types (because there's
--- no content) or a status code (because it should always be 204).
-data NoContentVerb  (method :: k1)
+-- | @NoContentVerbWithStatus@ is a specific type to represent 'NoContent' responses.
+-- It does not require either a list of content types (because there's no content).
+-- It still requires a status code, because several statuses may have no content.
+-- (e.g. 204, 301, 302, or 303).
```
gdeest, line 35: I do have a philosophical question here: when using 30X status codes, the `Location` header must be set. We are not enforcing this here — and in fact, your test does not do that either.

Wouldn't it be better to provide redirect combinators that do enforce this constraint ?

PR #1554, opened 2022-03-08
Use cabal-install to invoke doctest

ysangkok:
Doctests can be made more robust by making cabal-install invoke doctest with the right configuration. This is a new feature added in doctest 0.20. This makes the environment files unnecessary, and the CI configuration becomes simpler.

This also inserts a space such that `$requestHeaders` isn't mistaken for a TemplateHaskell splice.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -47,10 +47,7 @@ jobs:
 
     - name: Configure
       run: |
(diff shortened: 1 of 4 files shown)


gdeest:
Awesome ! I am quite happy to get rid of those hacks.

PR #1482, opened 2021-11-15
Add GHC 9.2 support

brandon-leapyear:
Related: #1475

PR #1555, opened 2022-03-08
Allow GHC 9.2 for all packages

### LINKED ISSUE #1496: Fails to build on ghc 9.2.1
It's related to the FlexibleContexts bug that got fixed. E.g.:

```
[ 4 of 31] Compiling Servant.API.ContentTypes ( src/Servant/API/ContentTypes.hs, dist/build/Servant/API/ContentTypes.o, dist/build/Servant/API/ContentTypes.dyn_o )

src/Servant/API/ContentTypes.hs:185:10: error:
    • Non type-variable argument
        in the constraint: AllMimeRender (ct : cts) a
      (Use FlexibleContexts to permit this)
    • In the instance declaration for ‘AllCTRender (ct : cts) a’
    |
185 |          (Accept ct, AllMime cts, AllMimeRender (ct ': cts) a) => AllCTRender (ct ': cts) a where
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/Servant/API/ContentTypes.hs:281:10: error:
    • Non type-variable argument
        in the constraint: AllMimeRender (ctyp' : ctyps) a
      (Use FlexibleContexts to permit this)
    • In the instance declaration for
        ‘AllMimeRender (ctyp : ctyp' : ctyps) a’
    |
281 |          ( MimeRender ctyp a
    |          ^^^^^^^^^^^^^^^^^^^...

src/Servant/API/ContentTypes.hs:302:10: error:
    • Non type-variable argument
        in the constraint: AllMime (ctyp : ctyp' : ctyps)
      (Use FlexibleContexts to permit this)
    • In the instance declaration for
        ‘AllMimeRender (ctyp : ctyp' : ctyps) NoContent’
    |
302 |          ( AllMime (ctyp ': ctyp' ': ctyps)
```

ysangkok:
Fixes #1496.

This is an alternative to #1482 that just needs a single additional allow-newer. Instead of overriding dependencies, this simply marks packages as unbuildable if they haven't been updated for GHC 9.2 yet (e.g. due to Cryptonite). ~~This still allows many packages to add compatibility, and it is verified in CI that it compiles. By using `buildable: False` instead of e.g. base bounds, we can continue listing all projects in `cabal.project`, and cabal-install will not complain if a project can't be built.~~ EDIT: Now adds all packages except one cook book, since Cryptonite was updated.

I haven't added a Stack build because I think it makes most sense to use Stackage snapshots are they are, instead of e.g. replacing the compiler version. There is still no snapshot with GHC 9.2.2.

I haven't added the Nix configuration, since as far as I can see, Nixpkgs hasn't made a tag with GHC 9.2.2 yet. But I'd be happy to add that in a later PR when such a release becomes available.

For now, I think that this provides a reasonable improvement for users and no notable dependency burden.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -19,6 +19,7 @@ jobs:
           - "8.8.4"
           - "8.10.7"
           - "9.0.2"
+          - "9.2.2"
 
     steps:
     - uses: actions/checkout@v2
--- a/cabal.project
+++ b/cabal.project
@@ -55,3 +55,6 @@ optimization: False
 -- https://github.com/chordify/haskell-servant-pagination/pull/12
 allow-newer: servant-pagination-2.2.2:servant
 allow-newer: servant-pagination-2.2.2:servant-server
+
+allow-newer: servant-multipart:bytestring
+allow-newer: servant-multipart-api:bytestring
--- a/cabal.project
+++ b/cabal.project
@@ -51,10 +51,3 @@ packages:
 tests: True
 optimization: False
 -- reorder-goals: True
-
--- https://github.com/chordify/haskell-servant-pagination/pull/12
-allow-newer: servant-pagination-2.2.2:servant
-allow-newer: servant-pagination-2.2.2:servant-server
-
-allow-newer: servant-multipart:bytestring
-allow-newer: servant-multipart-api:bytestring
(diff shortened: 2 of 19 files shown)


maksbotan:
Published https://hackage.haskell.org/package/servant-server-0.19.1 with 9.2 fixes, updated the rest of packages with revisions.

Thanks for your work!

PR #1556, opened 2022-03-08
Display capture hints in router layout

nbacquey:
This PR introduces a `CaptureHint` type, which is passed as an extra argument to the `CaptureRouter` and `CaptureAllRouter` constructors for the `Router'` type.
`CaptureHint` values are then used in `routerLayout`, to display the name and type of captured values, instead of just `<capture>` previously.
    
N.B.:
Because the `choice` smart constructor for routers can aggregate `Capture` combinators with different capture hints, the `Capture*Router` constructors actually take a *list* of `CaptureHint`, instead of a single one.

This PR also introduces `Spec` tests for the `routerLayout` function.

**Warning:**
This change is potentially breaking, because it adds the constraint `Typeable a` to all types that are to be captured. Because all types are typeable since GHC 7.10, this is not as bad as it sounds ; it only break expressions where `a` is quantified in an expression with `Capture a`.
In those cases, the fix is easy: it suffices to add `Typeable a` to the left-hand side of the quantification constraint.

### DIFF
--- a/changelog.d/1556
+++ b/changelog.d/1556
@@ -1,0 +1,81 @@
+synopsis: Display capture hints in router layout
+prs: #1556
+
+description: {
+
+This PR enhances the `Servant.Server.layout` function, which produces a textual description of the routing layout of an API. More precisely, it changes `<capture>` blocks, so that they display the name and type of the variable being captured instead.
+
+Example:
+
+For the following API
+```haskell
+type API =
+       "a" :> "d" :> Get '[JSON] NoContent
+  :<|> "b" :> Capture "x" Int :> Get '[JSON] Bool
+  :<|> "a" :> "e" :> Get '[JSON] Int
+```
+
+we previously got the following output:
+
+```
+/
+├─ a/
+│  ├─ d/
+│  │  └─•
+│  └─ e/
+│     └─•
+└─ b/
+   └─ <capture>/
+      ├─•
+      ┆
+      └─•
+```
+
+now we get:
+
+```
+/
+├─ a/
+│  ├─ d/
+│  │  └─•
+│  └─ e/
+│     └─•
+└─ b/
+   └─ <x::Int>/
+      ├─•
+      ┆
+      └─•
+```
+
+This change is achieved by the introduction of a CaptureHint type, which is passed as an extra argument to the CaptureRouter and CaptureAllRouter constructors for the Router' type.
+CaptureHint values are then used in routerLayout, to display the name and type of captured values, instead of just `<capture>` previously.
+
+N.B.:
+Because the choice smart constructor for routers can aggregate Capture combinators with different capture hints, the Capture*Router constructors actually take a list of CaptureHint, instead of a single one.
+
+This PR also introduces Spec tests for the routerLayout function.
+
+Warning:
+This change is potentially breaking, because it adds the constraint `Typeable a` to all types that are to be captured. Because all types are typeable since GHC 7.10, this is not as bad as it sounds ; it only break expressions where `a` is quantified in an expression with `Capture a`.
+In those cases, the fix is easy: it suffices to add `Typeable a` to the left-hand side of the quantification constraint.
+
+For instance, the following code will no longer compile:
+```haskell
+type MyAPI a = Capture "foo" a :> Get '[JSON] ()
+
+myServer :: forall a. Server (MyAPI a)
+myServer = const $ return ()
+
+myApi :: forall a. Proxy (MyAPI a)
+myApi = Proxy
+
+app :: forall a. (FromHttpApiData a) => Application
+app = serve (myApi @a) (myServer @a)
+```
+
+Indeed, `app` should be replaced with:
+```haskell
+app :: forall a. (FromHttpApiData a, Typeable a) => Application
+app = serve (myApi @a) (myServer @a)
+```
+}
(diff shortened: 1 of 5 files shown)


servant-server/src/Servant/Server/Internal.hs
```diff
@@ -173,7 +173,7 @@ instance (HasServer a context, HasServer b context) => HasServer (a :<|> b) cont
 -- > server = getBook
 -- >   where getBook :: Text -> Handler Book
 -- >         getBook isbn = ...
-instance (KnownSymbol capture, FromHttpApiData a
+instance (KnownSymbol capture, FromHttpApiData a, Typeable a
```
gdeest, line 176: This is a significant breaking change ; I am not sure we can require all captures to have a `Typeable` instance for debugging purposes.

tchoutri: It's a pity that we didn't have it before but I don't know if we can afford having it not, indeed.

arianvp:
Interesting. Could this be used as a replacement for https://hackage.haskell.org/package/servant-ekg-0.3.1/docs/src/Servant.Ekg.html#HasEndpoint  e.g. to implement prometheus metrics directly in `servant-server` (https://github.com/haskell-servant/servant-ekg/issues/30 ) ?

servant-server/src/Servant/Server/Internal.hs
gdeest: Fair :) I feel old.

gdeest: I am guessing it still precludes `a` from being e.g. a function type, but maybe that's OK ?

nbacquey: It would still break if someone wrote something like
```
genericAPI :: forall a. (FromHttpApiData a) => Proxy API
genericAPI = (Proxy :: Capture "x" a :> Get '[JSON] (Foo a)) 
```
They would need to add `Typeable a` to the constraints.

gdeest:
After giving it some thought, I am in favor of re-introducing the `Typeable` constraint. While it is indeed a breaking change, 

- It should indeed affect a relatively small number of users.
- It should always be possible (and trivial) to fix the error.
- This can be documented thoroughly in a changelog / migration guide.
- It makes router layout output much more useful (and allows nice features such as #1561 which are also useful for debugging and monitoring purposes).

It is possible that carrying `Typeable` dictionaries around somehow slows down routing a bit — I don't think this should be significant, but may be worth mentioning.

nbacquey:
> After giving it some thoughts, I am in favor of re-introducing the `Typeable` constraint. While it is indeed a breaking change,
> 
>     * It should indeed affect a relatively small number of users.
> 
>     * It should always be possible (and trivial) to fix the error.
> 
>     * This can be documented thoroughly in a changelog / migration guide.
> 
>     * It makes router layout output much more useful (and allows nice features such as [New combinator to return routed path in response headers #1561](https://github.com/haskell-servant/servant/pull/1561) to also include type information for debugging and monitoring purposes).
> 
> 
> It is possible that carrying `Typeable` dictionaries around somehow slows down routing a bit — I don't think this should be significant, but may be worth mentioning.

I've reinstated the `Typeable` constraint, and also added a spec test to check that `CaptureAll "foos" Int` produces "<foos::[Int]>"

changelog.d/1556
```diff
@@ -0,0 +1,18 @@
+synopsis: Display capture hints in router layout
+prs: #1556
+
+description: {
+
+This PR introduces a CaptureHint type, which is passed as an extra argument to the CaptureRouter and CaptureAllRouter constructors for the Router' type.
```
gdeest, line 6: I think this changelog entry is a bit too technical from the start — it does not inform bystanders very well on what the end result is.

Maybe starting with an before / after example would be a good idea ?

gdeest, line 6: I think this changelog entry is a bit too technical from the start — it does not inform bystanders very well on what the end result is.

Maybe starting with an before / after example would be a good idea ?

servant-server/src/Servant/Server/Internal/Router.hs
```diff
@@ -111,11 +127,21 @@ routerStructure (Choice r1 r2) =
     (routerStructure r1)
     (routerStructure r2)
 
--- | Compare the structure of two routers.
+-- | Compare the structure of two routers. Ignores capture hints.
```
gdeest, line 130: Why ?

changelog.d/1556
nbacquey: Added a short introduction and an example for the new feature

changelog.d/1556
```diff
@@ -0,0 +1,62 @@
...
+CaptureHint values are then used in routerLayout, to display the name and type of captured values, instead of just `<capture>` previously.
+
+N.B.:
+Because the choice smart constructor for routers can aggregate Capture combinators with different capture hints, the Capture*Router constructors actually take a list of CaptureHint, instead of a single one.
+
+This PR also introduces Spec tests for the routerLayout function.
+
+Warning:
+This change is potentially breaking, because it adds the constraint `Typeable a` to all types that are to be captured. Because all types are typeable since GHC 7.10, this is not as bad as it sounds ; it only break expressions where `a` is quantified in an expression with `Capture a`.
+In those cases, the fix is easy: it suffices to add `Typeable a` to the left-hand side of the quantification constraint.
```
gdeest, line 60: This probably deserves an example, which could serve as a migration guide.

gdeest (approved):
Just another minor nitpick on the changelog, otherwise this is good to go for me. :+1:

changelog.d/1556
nbacquey: Done

gdeest (approved):

PR #1557, opened 2022-03-09
Use Stack 2.7.5, cleanup allow-newer/CI

ysangkok:
servant-pagination 2.5.0 is out, and it allows servant 0.19. So we do not need this allow-newer any more. Stack 2.7.5 is out, and it only downloads extra-deps with multiple subdirs once, and also contains other bugfixes. A stale CI configuration line is removed, since GHC 9.0.1 is no longer run in CI, we use GHC 9.0.2.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -58,9 +58,6 @@ jobs:
         cabal test all
 
     - name: Run doctests
-      # doctests are broken on GHC 9 due to compiler bug:
-      # https://gitlab.haskell.org/ghc/ghc/-/issues/19460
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yml
```diff
@@ -58,9 +58,6 @@ jobs:
         cabal test all
 
     - name: Run doctests
-      # doctests are broken on GHC 9 due to compiler bug:
```
alpmestan, line 61: is the bug gone?

alpmestan: Thanks for the clarification :) And your decision makes sense to me.

alpmestan (approved):

PR #1560, opened 2022-03-14
[WIP] Servant auth server PoC

Vlix:
An attempt at making a better Auth instance of `HasServer`.
Also added an example of how this would make for easy adding of different auth procedures.

Any comments, critiques, etc. are very much welcome and appreciated.

### DIFF
--- a/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal.hs
+++ b/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal.hs
@@ -1,24 +1,38 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Servant.Auth.Server.Internal where
 
+import           Control.Monad.Except (runExceptT, join)
 import           Control.Monad.Trans (liftIO)
-import           Servant             ((:>), Handler, HasServer (..),
-                                      Proxy (..),
-                                      HasContextEntry(getContextEntry))
-import           Servant.Auth
+import           Data.Kind           (Type)
+import           Data.Typeable       (Typeable, typeRep)
+import           Network.Wai         (Request, queryString)
+import           Servant
+import           Servant.API.Modifiers (FoldLenient, FoldRequired, RequestArgument, unfoldRequestArgument)
+import           Servant.Auth        (Auth)
 import           Servant.Auth.JWT    (ToJWT)
+import           Data.Text           (Text)
 
 import Servant.Auth.Server.Internal.AddSetCookie
 import Servant.Auth.Server.Internal.Class
 import Servant.Auth.Server.Internal.Cookie
 import Servant.Auth.Server.Internal.ConfigTypes
-import Servant.Auth.Server.Internal.JWT
+-- import Servant.Auth.Server.Internal.JWT
 import Servant.Auth.Server.Internal.Types
 
-import Servant.Server.Internal (DelayedIO, addAuthCheck, withRequest)
+import Servant.Server.Experimental.Auth (AuthHandler (..))
+import Servant.Server.Internal (DelayedIO, addAuthCheck, delayedFail, delayedFailFatal, mkContextWithErrorFormatter, withRequest, MkContextWithErrorFormatter)
+import qualified Data.Text as T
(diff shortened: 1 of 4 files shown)


servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal.hs
```diff
@@ -68,3 +82,137 @@ instance ( n ~ 'S ('S 'Z)
...
+
+instance
+    ( HasServer api ctxs
+    , HasContextEntry (MkContextWithErrorFormatter ctxs) ErrorFormatters
+    , SBoolI (FoldRequired mods)
+    , SBoolI (FoldLenient mods)
+    , AllAuth auths a
+    , HasContextEntry ctxs (AuthHandler Request (NewAuthResult a))
+    ) => HasServer (NewAuth mods auths a :> api) ctxs where
+    type ServerT (NewAuth mods auths a :> api) m =
```
gdeest, line 115: Would it make sense define the instance for `NewAuth mods (auth ': auths) a :> api` to statically ensure that _at least_ one auth mode is defined ? We could define a custom type error for the case where the list of auths is empty.

gdeest:
> In the current implementation, all of the 'auths' are checked one by one, and the first that is present is tried and will either be returned when successful, or when not, throw an error given the appropriate 'ErrorFormatter' or return a 'Left err' if the 'Lenient' modifier is set. Only if NONE of the auths are present will it either throw an 'err401' or, if the 'Optional' modifier is set, return a 'Nothing'. This error might also be customizeable if we make it be required from the context.

Just to clarify my understanding here (I am not overly familiar with `servant-auth`). With the current `Auth` combinator, if two modes of authentication are defined (say, `Cookie` and `JWT`, in that order):

- If a cookie is set, but authentication fails for some reason (e.g., invalid XSRF token), an error is raised.
- If NO cookie is set, then `JWT` is tried.
- If there is no JWT token, a 401 error is raised.
- If there is a JWT token, but it proves to be invalid, some other error is raised.

Whereas you would prefer for subsequent auth schemes to be tried anyhow, and only return a 401 error when _all_ of them fail ?

gdeest:
I very much like that the `HasServer` instance no longer mentions JWT or Cookie settings explicitly: this is a big wart in current design.

servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal.hs
Vlix: Hmm, probably a good idea, yeah.

Vlix:
> Just to clarify my understanding here (I am not overly familiar with `servant-auth`). With the current `Auth` combinator, if two modes of authentication are defined (say, `Cookie` and `JWT`, in that order):
> 
> * If a cookie is set, but authentication fails for some reason (e.g., invalid XSRF token), an error is raised.
> * If NO cookie is set, then `JWT` is tried.
> * If there is no JWT token, a 401 error is raised.
> * If there is a JWT token, but it proves to be invalid, some other error is raised.
> 
> Whereas you would prefer for subsequent auth schemes to be tried anyhow, and only return a 401 error when _all_ of them fail ?

No, my implementation works the same way. If the first `auth` returns a `Nothing` (meaning it's not there), it will try the next. But if it IS present, it is tried and taken as either succeeding (`Right`) or failing (`Left`) and it will return a 401 error if the modifier is `Strict` (or absent), but return the `Left errMsg` if the modifier `Lenient` is set.
-> _The `go` helper function in `processAllAuthHandlers` shows it looping on `Nothing` and returning on `Just x`_

In this way, if you'd want more than one authentication to be present and tried, you'd have to use two `NewAuth mods auths a`. i.e. `"path" :> NewAuth '[] [Auth1] MyType :> NewAuth '[] [Auth2] MyType2 :> Get '[JSON] Result`

But if you accept a JWT in the query parameters, a header or BasicAuth, it would just look like this and only one has to succeed:
`"path" :> NewAuth '[] [JWTQueryParam, JWTHeader, BasicAuth] MyAuth :> Get '[JSON] Result`

Vlix:
Note to self:
* make stack build every Servant package with `lts-18.27`
* make two instances of `HasServer (NewAuth ...`
  * `TypeError ... => NewAuth '[]`
  * `NewAuth (auth ': auths)`

domenkozar:
@Vlix do you need any help with this?

Vlix:
> do you need any help with this?

I haven't spent time on this at all, so if you want to pick it up, or just give your own crack at it, please do :)

PR #1561, opened 2022-03-14
New combinator to return routed path in response headers

### LINKED ISSUE #1553: Expose some information about API resolution to middlewares
Recently, I tried to write a [WAI Middleware](https://hackage.haskell.org/package/wai-3.2.3/docs/Network-Wai.html) that was aware of the API my application was using. In particular, given an incoming request url, I wanted to be able to tell which endpoint was used, and which parts of the url were captured.

For instance, consider the server api below:
```haskell
type Api =
       "foo" :> Capture "id" Int :> Get '[JSON, PlainText] Foo
  :<|> "bar" :> Capture "id" Int :> Get '[JSON, PlainText] Bar
```

Let us request it with the following command:
```bash
curl -X GET "http://localhost/foo/12345" -H "Accept: application/json"
```
If the request succeeds, we should get a Foo object, encoded as JSON.
The additional data I would like to get from Servant for my middleware would be something like:
```json
{
  "effective-url": "/foo/<id::Int>",
  "content-type": "application/json"
}
```
The `"content-type"` part is easy to get from the response header, but the `"url"` part doesn't seem easily available.

I can think of several ways for Servant to provide this information:
* Return the "effective type" that was used for each request, to be processed by the middleware. For instance, the request above would return
```haskell
type EffectiveApi = "foo" :> Capture "id" Int :> Get '[JSON] Foo
```
* Introduce a combinator that would have the effect of including routing information in the response headers. The API above would then be written as
```haskell
type Api = WithRouting :>
       "foo" :> Capture "id" Int :> Get '[JSON, PlainText] Foo
  :<|> "bar" :> Capture "id" Int :> Get '[JSON, PlainText] Bar
```
  then the response headers would include `"Effective-Url": "/foo/<id::Int>"`
* Introduce a new `Servant.Middleware` that would be aware of API resolution mechanisms.

nbacquey:
This commit introduces a new type-level combinator, `WithRoutingHeader`.
It modifies the behaviour of the following sub-API, such that all endpoints of said API return an additional routing header in their response.

A routing header is a header that specifies which endpoint the incoming request was routed to.
Endpoint are designated by their path, in which `Capture'` and `CaptureAll` combinators are replaced by a capture hint.

This header can be used by downstream middlewares to gather information about individual endpoints, since in most cases
a routing header uniquely identifies a single endpoint.
    
Example:
```haskell
type MyApi =
  WithRoutingHeader :> "by-id" :> Capture "id" Int :> Get '[JSON] Foo
-- GET /by-id/1234 will return a response with the following header:
--   ("Servant-Routed-Path", "/by-id/<id::Int>")
```
    
To achieve this, two refactorings were necessary:
* Introduce a type `RouterEnv env` to encapsulate the `env` type (as in `Router env a`), which contains a tuple-encoded list of url  pieces parsed from the incoming request. This type makes it possible to pass more information throughout the routing process, and the computation of the `Delayed env c` associated with each request.
* Introduce a new kind of router, which only modifies the RouterEnv, and doesn't affect the routing process otherwise: `EnvRouter (RouterEnv env -> RouterEnv env) (Router' env a)`.
  This new router is used when encountering the `WithRoutingHeader` combinator in an API, to notify the endpoints of the sub-API that they must produce a routing header (this behaviour is disabled by default).

This PR also introduces `Spec` tests for the `WithRoutingHeader` combinator, which showcase some of its possible uses.

This PR is based upon #1556, and should remain WIP until it is merged.

Closes #1553

### DIFF
--- a/changelog.d/1561
+++ b/changelog.d/1561
@@ -1,0 +1,33 @@
+synopsis: New combinator to return routed path in response headers
+prs: #1561
+issues: #1553
+
+description: {
+
+This commit introduces a new type-level combinator, `WithRoutingHeader`.
+It modifies the behaviour of the following sub-API, such that all endpoints of said API return an additional routing header in their response.
+
+A routing header is a header that specifies which endpoint the incoming request was routed to.
+Endpoint are designated by their path, in which `Capture'` and `CaptureAll` combinators are replaced by a capture hint.
+
+This header can be used by downstream middlewares to gather information about individual endpoints, since in most cases
+a routing header uniquely identifies a single endpoint.
+
+Example:
+
+```haskell
+type MyApi =
+  WithRoutingHeader :> "by-id" :> Capture "id" Int :> Get '[JSON] Foo
+-- GET /by-id/1234 will return a response with the following header:
+--   ("Servant-Routed-Path", "/by-id/<id::Int>")
+```
+
+To achieve this, two refactorings were necessary:
+
+* Introduce a type `RouterEnv env` to encapsulate the `env` type (as in `Router env a`), which contains a tuple-encoded list of url pieces parsed from the incoming request. This type makes it possible to pass more information throughout the routing process, and the computation of the `Delayed env c` associated with each request.
+* Introduce a new kind of router, which only modifies the `RouterEnv`, and doesn't affect the routing process otherwise: `EnvRouter (RouterEnv env -> RouterEnv env) (Router' env a)`.
+  This new router is used when encountering the `WithRoutingHeader` combinator in an API, to notify the endpoints of the sub-API that they must produce a routing header (this behaviour is disabled by default).
+
+This PR also introduces `Spec` tests for the `WithRoutingHeader` combinator, which showcase some of its possible uses.
+
+}


tchoutri:
I would love to read a bit more on practical uses of this feature! 
Do you use this in tracing scenarios? Observability and Telemetry?

servant/src/Servant/API/Environment.hs
```diff
@@ -0,0 +1,31 @@
...
+-- individual endpoints, since in most cases a routing header uniquely identifies a
+-- single endpoint.
+--
+-- Example:
+--
+-- >>> type MyApi = WithRoutingHeader :> "by-id" :> Capture "id" Int :> Get '[JSON] Foo
+-- >>>     -- GET /by-id/1234 will return a response with the following header:
+-- >>>     --   ("Servant-Routed-Path", "/by-id/<id:CaptureSingle>")
+data WithRoutingHeader
+    deriving (Typeable)
```
gdeest, line 24: We probably don't need this anymore, since every datatype already derives it implicitly in all the GHC versions we support.

tchoutri:
@nbacquey This is music to my ears.

servant-server/src/Servant/Server.hs
```diff
@@ -235,7 +235,7 @@ hoistServer p = hoistServerWithContext p (Proxy :: Proxy '[])
 -- > │  └─ e/
 -- > │     └─•
 -- > ├─ b/
--- > │  └─ <capture>/
+-- > │  └─ <x::CaptureSingle>/
```
tchoutri, line 238: Unrelated, but do you have any intuition on how hard it would be to add HTTP method information next to each leaf? :)

servant-server/src/Servant/Server/Internal.hs
```diff
@@ -238,6 +241,21 @@ instance (KnownSymbol capture, FromHttpApiData a
     where
       rep = typeRep (Proxy :: Proxy CaptureAll)
       formatError = urlParseErrorFormatter $ getContextEntry (mkContextWithErrorFormatter context)
+      hint = CaptureHint (T.pack $ symbolVal $ Proxy @capture) CaptureList
+
+-- | Using 'WithRoutingHeaders' in one of the endpoints for your API,
+-- will automatically add routing headers to the response generated by the server.
```
tchoutri, line 247: Could you add a `@since` annotation?

servant-server/src/Servant/Server/Internal/RouterEnv.hs
```diff
@@ -0,0 +1,65 @@
+{-# LANGUAGE DeriveFunctor     #-}
```
tchoutri, line 1: Could you add

* [ ] Module-level haddocks
* [ ] `@since` annotations on these elements?

tchoutri: And there are extraneous empty lines between declarations

servant/src/Servant/API/Environment.hs
```diff
@@ -0,0 +1,26 @@
+{-# OPTIONS_HADDOCK not-home    #-}
+-- | Define API combinator that modify the behaviour of the routing environment.
+module Servant.API.Environment (WithRoutingHeader) where
+
+-- | Modify the behaviour of the following sub-API, such that all endpoint of said API
```
tchoutri, line 5: Could you add a `@since` annotation?

tchoutri (changes requested):
Wonderful PR, thank you very much.
I left some requests. 

For the `@since` annotations, you can target 0.20.0.0 (informally known as 0.20)

nbacquey:
I don't remember my specific usecase, but I hadn't thought about using `tweakResponse` like that, thanks for the tip!

servant-server/src/Servant/Server.hs
nbacquey: I don't think it would be really hard ; it would look a lot like what I've done in #1556, except that you would have to add another argument to `StaticRouter`, and pass it in the `HasServer` instance for `Verb`. I expect the new router constructor to look like
```haskell
StaticRouter  (Map Text (Router' env a)) [(RouterEnv env -> a, HTTPMethod)]
```
Then the information would get passed along `routerStructure`, then to `routerLayout`

tchoutri: Excellent, thank you :)

tchoutri (approved):
My comments have been addressed

alpmestan:
> With this PR merged, this behavior would be really easy to implement: you'd only need a new combinator that uses an EnvRouter, a new field in RouterEnv env, and the appropriate switch in runChoice

Do I understand correctly that this would require changing a type in servant and that the new combinator you describe would need to be based off a patched servant?

nbacquey:
> Do I understand correctly that this would require changing a type in servant and that the new combinator you describe would need to be based off a patched servant?

You are correct, this hypothetical new combinator would need to be based off a patched servant. My point was that the patch would be quite simple to write.

nbacquey:
I rebased on the new version of #1556, which reinstates "real" type hints, instead of `CaptureSingle`/`CaptureList`

nbacquey:
Added a fix to prevent `WithRoutingHeader` from breaking `IsElem`

tchoutri:
@alpmestan @gdeest I would appreciate if we could bring this PR to a conclusion :)

PR #1564, opened 2022-03-17
Resolve build warning

nbacquey:
This PR solves all current compile time warnings in servant, so that mistakes can be more easily spotted when interacting with the code base.

The only nontrivial fix is the one of "erroring instances" of `HasLink`, `HasClient`, and `HasServer`. Those erroring instances, by construction, do not define all methods of their type class, thus generating warnings. It wouldn't have been reasonable to silence all such warnings in their respective files, because those files contain multiple other instance declarations, and having `missing methods` warnings for those is useful.

Therefore, I chose to move the erroring instances in separate files (thus refactoring a bit the structure of the code base), and silence "missing methods" warning in those new files only.


ysangkok:
@nbacquey Have you noticed that doctests are failing?

servant-client/test/Servant/StreamSpec.hs
```diff
@@ -21,16 +21,18 @@
...
 import           Control.Monad.Codensity
                  (Codensity (..))
 import           Control.Monad.IO.Class
                  (MonadIO (..))
+-}
 import           Control.Monad.Trans.Except
 import qualified Data.ByteString               as BS
 import           Data.Proxy
-import qualified Data.TDigest                  as TD
+-- import qualified Data.TDigest                  as TD
```
gdeest, line 35:
```suggestion
```

servant-client/test/Servant/StreamSpec.hs
```diff
@@ -46,19 +48,23 @@ import           System.Entropy
                  (getEntropy, getHardwareEntropy)
 import           System.IO.Unsafe
                  (unsafePerformIO)
+{-
```
gdeest, line 51: It looks like the tests that used this import have been commented out for 4 years ; maybe we can just remove them ?

servant-client/test/Servant/StreamSpec.hs
```diff
@@ -78,9 +84,8 @@ api :: Proxy StreamApi
 api = Proxy
 
 getGetNL, getGetNS :: ClientM (SourceIO Person)
-getGetALot :: ClientM (SourceIO BS.ByteString)
 getStreamBody :: SourceT IO BS.ByteString -> ClientM (SourceIO BS.ByteString)
-getGetNL :<|> getGetNS :<|> getGetALot :<|> getStreamBody = client api
+getGetNL :<|> getGetNS :<|> _ :<|> getStreamBody = client api
```
gdeest, line 88:
```suggestion
getGetNL :<|> getGetNS :<|> _getGetALot :<|> getStreamBody = client api
```

This will suppress the warning while being explicit about what we are ignoring.

servant-client/test/Servant/StreamSpec.hs
nbacquey: Done

nbacquey:
> @nbacquey Have you noticed that doctests are failing?

Thanks for noticing ; doctests should be fine now

nbacquey:
@tchoutri @alpmestan 
Could I have your position about this PR?

tchoutri:
@nbacquey Hi! After a cursory review, I'd like to ask for a changelog entry, as it is my understanding that you add a new module `Servant.Server.TypeErrors` that you have split from `Servant.Server.Internal`. :)

nbacquey:
> @nbacquey Hi! After a cursory review, I'd like to ask for a changelog entry, as it is my understanding that you add a new public module `Servant.Server.TypeErrors` that you have split from `Servant.Server.Internal`. :)

Instead of adding a changelog entry, I preferred to declare `Servant.Server.TypeErrors` in `other-modules` instead. Its introduction is just a technicality, users should only use `Servant.Server` directly to benefit from the custom TypeErrors

tchoutri:
Ah I see indeed. :)

tchoutri (approved):
Unless something else pops up, approved!

PR #1569, opened 2022-03-21
Use toEncodedUrlPiece directly when encoding captures

### LINKED ISSUE #1511: How to add an asterisk in the URL?
I'd like to make a request to an URL with path `/1/indexes/*/queries`, that is it has a verbatim asterisk in it. By default, servant url encodes an url so the resulting path is `/1/indexes/%2A/queries`. But that's not supported by the API endpoint I'm trying to query.

For reference, I'm trying to query https://www.algolia.com/doc/rest-api/search/#search-multiple-indices.

I tried adding a new type, see [this commit](https://github.com/ibizaman/servant/commit/74e2e5c1f8dc0b21dce2b65574feb0904d76dfaa), but although unit test works, sevrant still url encodes the asterisk (as can be seen if using https://mitmproxy.org/ for example):
```haskell
newtype Verbatim = Verbatim { unVerbatim :: ByteString }

instance ToHttpApiData Verbatim where
    toEncodedUrlPiece = byteString . unVerbatim
    toUrlPiece = decodeUtf8 . unVerbatim

instance FromHttpApiData Verbatim where
    parseUrlPiece = pure . Verbatim . encodeUtf8
```

Any idea how I can achieve this?

gdeest:
Current implementation of captures uses the `toUrlPiece` method from the
`ToHttpApiData` typeclass, and encodes the resulting `Text` using `toEncodedUrlPiece`
when appending to the request path.

The problem with this approach is that the instance for `Text` percent-encodes
characters that are perfectly valid in URLs, such as `*`.

This patch makes direct use of `toEncodedUrlPiece`, which lets users implement
encoding according to their needs.

Closes #1511

### DIFF
--- a/changelog.d/1569
+++ b/changelog.d/1569
@@ -1,0 +1,13 @@
+synopsis: Encode captures using toEncodedUrlPiece
+prs: #1569
+issues: #1511
+
+description: {
+The `servant-client` library now makes direct use of `toEncodedUrlPiece` from `ToHttpApiData`
+to encode captured values when building the request path. It gives user freedom to implement
+URL-encoding however they need.
+
+Previous behavior was to use `toUrlPiece` and URL-encode its output using `toEncodedUrlPiece`
+from the `Text` instance of `ToHttpApiData`. The issue with this approach is that
+`ToHttpApiData Text` is overly zealous and also encodes characters, such as `*`, which are perfectly valid in a URL.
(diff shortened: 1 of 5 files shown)


arianvp:
This probably warrants a changelog entry :)

tchoutri (approved):

PR #1572, opened 2022-03-26
Derive MonadMask for ClientM

shlevy:


tchoutri:
Hi! Thank you very much for this PR. :)
Could you tell us a bit more about the usecase that made you need MonadMask? :)

shlevy:
I'm working on an event-oriented observability library and I have a generic combinator

```haskell
withEvent :: (MonadMask m) => m (Event m r es) -> (Event m r es -> m a) -> m a
```

which ensures that the `Event` is finalized and if there's an exception that the information is attached to the `Event.

Now I'm writing a `RunClient` instance for a wrapper around `ClientM`that emits an event for every `runRequestAcceptStatus` call, and currently I have to do this to work around the lack of `MonadMask`:

```haskell
ClientM $ withEvent (hoistEvent unClientM <$> unClientM doE) \ev' -> unClientM $ do
    let ev = hoistEvent ClientM ev'
    -- Do stuff with ev in ClientM
```

tchoutri:
Okay cool! I don't see any downside to this addition.
Keep us updated if you find cool stuff to do with Servant. :)

PR #1580, opened 2022-04-14
Allow IO in JWTSettings' validationKeys

### LINKED ISSUE #1579: Support updating JWKSet
Originally reported by @axman6 in [the old servant-auth repo](https://github.com/haskell-servant/servant-auth/issues/153):

We have a service which periodically needs to update the set of known keys which will be used to sign requests. The simplest way I can see to allow this is to have

```haskell
validationKeys :: JWKSet
```

become

```haskell
validationKeys :: Either (IO JWKSet) JWKSet
```

or something isomorphic to this. For our particular use case this happens frequently enough that it would be better to not have to shoot the service in the head to periodically fetch the latest keys.

jkarni:
Fixes #1579 

There have been two PRs in the old servant-auth repo doing similar things. [This one](https://github.com/haskell-servant/servant-auth/pull/193/files) by [dnikolovv](https://github.com/dnikolovv) which uses IORefs, and [this one](https://github.com/haskell-servant/servant-auth/pull/169/files) by me which allows IO as here (but that PR also does a lot of other things).

I personally prefer `IO` rather than `IORef` since it's a bit more general (and works with libraries that either don't use or don't expose the `IORef` - e.g. [cached-io](https://hackage.haskell.org/package/cached-io)).

This is a breaking change.

### DIFF
--- a/changelog.d/1580
+++ b/changelog.d/1580
@@ -1,0 +1,10 @@
+synopsis: Allow IO in validationKeys
+prs: #1580
+issues: #1579
+
+description: {
+
+Currently validationKeys are a fixed JWKSet. This does not work for setups such
+as AWS Cognito or Okta, where there is a regular key rotation. This change
+alters the type of validationKeys from JWKSet to IO JWKSet.
+}
--- a/changelog.d/1580
+++ b/changelog.d/1580
@@ -4,7 +4,9 @@ issues: #1579
 
 description: {
 
-Currently validationKeys are a fixed JWKSet. This does not work for setups such
-as AWS Cognito or Okta, where there is a regular key rotation. This change
-alters the type of validationKeys from JWKSet to IO JWKSet.
+Currently validationKeys are a fixed JWKSet. This does not work with OIDC
+providers such as AWS Cognito or Okta, which regularly fetching jwks_uri to
+discover new and expired keys.
+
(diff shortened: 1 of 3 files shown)


changelog.d/1580
```diff
@@ -0,0 +1,10 @@
+synopsis: Allow IO in validationKeys
+prs: #1580
+issues: #1579
+
+description: {
+
+Currently validationKeys are a fixed JWKSet. This does not work for setups such
```
arianvp, line 7: Can we generalise this to:

This does not work with OIDC providers like for example Okta and AWS Cognito, which require regularly fetching the `jwks_uri` to discover expired and new keys. 

(As it's not Okta specific; it's OIDC specific)

jkarni: Done!

gdeest (approved):
LGTM. CI is failing though, but it is likely unrelated (other PRs are affected as well, such as #1561 — I honestly don't know why, but it looks like an encoding issue.)

gdeest:
@jkarni Could you rebase on `master` ? The CI failure should be fixed.

jkarni:
Rebased!

jkarni:
Should I merge this myself? It's been a while since I've contributed - am a bit wary of using my commit bit without catching up on processes first!

tchoutri:
If it's approved and you are able to merge, I guess you can. :)

PR #1587, opened 2022-05-02
Evaluate NoContent before (not) rendering it.

NorfairKing:
This way it shows up as covered in coverage reports,
 
Currently any `return NoContent` is marked at "not covered" in coverage reports because it is not evaluated.


NorfairKing:
Now with a test as well @arianvp :D

NorfairKing:
I figured I'd also make a success-case test while I'm at it.

tchoutri (approved):

NorfairKing:
Is there anything else that needs to happen before this can be merged?

tchoutri:
Nope, I just waited for the CI to pass and then got sidetracked. Let's merge this!

PR #1588, opened 2022-05-03
Add HasSwagger instance for NamedRoutes

LightAndLight:
I need an instance like this to generate API docs for my API that uses NamedRoutes.

I'm using the API record type name to tag the contained endpoints, so that they appear grouped in the Swagger UI. I'm open to changing this.


tchoutri:
It's a good change! It is very useful in OpenAPI3, so thank you for backporting this to swagger :)

tchoutri (approved):

gdeest (approved):

PR #1589, opened 2022-05-15
Only add a ? when query string is nonempty

GambolingPangolin:
In #1432 we changed the way that servant composes query strings.  Among other problems, the new scheme always adds a `?` even when the query parameter set is empty.  This fixes the scheme so that when there are no query parameters, the resulting query string is empty.


arianvp (approved):

tchoutri (changes requested):
Please add a changelog entry for your PR. :)

GambolingPangolin:
@tchoutri I just added it.

tchoutri (approved):

PR #1593, opened 2022-06-04
[servant] Generalize type of Servant.Types.SourceT.source to any foldable

Daniel-Diaz:
I didn't see any reason to restrict this function to only work on lists, so I changed the type signature from

```haskell
source :: [a] -> SourceT m a
```

to

```haskell
source :: Foldable f => f a -> SourceT m a
```


alpmestan (approved):

Daniel-Diaz:
Thanks for the merge!

PR #1595, opened 2022-06-17
servant-client: Run ClientEnv's makeClientRequest in IO

Minnozz:
Use cases:
- [applyDigestAuth](https://hackage.haskell.org/package/http-client-tls-0.3.6.1/docs/Network-HTTP-Client-TLS.html#v:applyDigestAuth)
- Sign requests

### DIFF
--- a/servant-client/test/Servant/SuccessSpec.hs
+++ b/servant-client/test/Servant/SuccessSpec.hs
@@ -162,7 +162,8 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
       mgr <- C.newManager C.defaultManagerSettings
       -- In proper situation, extra headers should probably be visible in API type.
       -- However, testing for response timeout is difficult, so we test with something which is easy to observe
-      let createClientRequest url r = (defaultMakeClientRequest url r) { C.requestHeaders = [("X-Added-Header", "XXX")] }
+      let createClientRequest url r = fmap (\req -> req { C.requestHeaders = [("X-Added-Header", "XXX")] })
+                                           (defaultMakeClientRequest url r)
           clientEnv = (mkClientEnv mgr baseUrl) { makeClientRequest = createClientRequest }
       res <- runClientM (getRawSuccessPassHeaders HTTP.methodGet) clientEnv
       case res of
--- a/changelog.d/1595
+++ b/changelog.d/1595
@@ -1,0 +1,2 @@
+synopsis: Run ClientEnv's makeClientRequest in IO.
+prs: #1595
--- a/doc/cookbook/using-free-client/UsingFreeClient.lhs
+++ b/doc/cookbook/using-free-client/UsingFreeClient.lhs
@@ -119,7 +119,7 @@ Now we can use `servant-client`'s internals to convert servant's `Request`
 to http-client's `Request`, and we can inspect it:
 
 ```haskell
-        let req' = I.defaultMakeClientRequest burl req
+        req' <- I.defaultMakeClientRequest burl req
         putStrLn $ "Making request: " ++ show req'
 ```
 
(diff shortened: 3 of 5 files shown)


tchoutri:
@Minnozz Hi! Has there been a discussion about this PR? :)

Minnozz:
@tchoutri Hi! This change was suggested by `alp__` on IRC today in response to a question from me. I took a shot at implementing it; should it also be discussed elsewhere?

tchoutri:
Great, then I'll leave @alpmestan review your PR. :)

servant-client/test/Servant/SuccessSpec.hs
```diff
@@ -162,7 +162,8 @@ successSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
       mgr <- C.newManager C.defaultManagerSettings
       -- In proper situation, extra headers should probably be visible in API type.
       -- However, testing for response timeout is difficult, so we test with something which is easy to observe
-      let createClientRequest url r = (defaultMakeClientRequest url r) { C.requestHeaders = [("X-Added-Header", "XXX")] }
+      let createClientRequest url r = fmap (\req -> req { C.requestHeaders = [("X-Added-Header", "XXX")] })
+                                           (defaultMakeClientRequest url r)
```
Minnozz, line 166: I'm not sure how ergonomic this is; any suggestions for improving that?

alpmestan: It's fine, don't worry.

alpmestan (approved):
This looks fine to me. We'll want to add an entry in the relevant changelog. Also, this will force a major version bump.

Let's perhaps leave a few days to give others a chance to give feedback on this, and for you to add a changelog entry (see how other PRs do it, with entries in the `changelog.d` directory).

Minnozz:
@alpmestan I added a changelog entry, do you think it needs more explanation?

alpmestan:
@Minnozz Should be fine, I think people using servant are familiar enough with `IO` =)

If you ever feel like writing down a little something that demonstrates how you're using this to your advantage (cookbook recipe, blog post, whatever), then it can serve as a reference for users who might need this in the future. But that's more work that adding a few lines in the changelog :-)

tchoutri:
Considering the approval of this PR, I will merge, but a cookbook or a blog post would be very appreciated, @Minnozz  :)

unclechu:
Please have a look at https://github.com/haskell-servant/servant/issues/1787 trying to read this change through I can’t say why this change is necessary. Why was it worth to break the API? Why does it have to be exactly `IO`? Etc. etc.

Minnozz:
@unclechu The change was necessary for my project to be able to call [applyDigestAuth](https://hackage.haskell.org/package/http-client-tls-0.3.6.1/docs/Network-HTTP-Client-TLS.html#v:applyDigestAuth), as is noted in the description of this PR. As for requiring precisely IO: as a Haskell novice I did not understand the difference between IO and MonadIO when I made this PR; feel free to propose a PR with a better API :)

tchoutri:
+1, submit a PR to depend on MonadIO, it would be an improvement.

unclechu:
@Minnozz 
> The change was necessary for my project to be able to call [applyDigestAuth](https://hackage.haskell.org/package/http-client-tls-0.3.6.1/docs/Network-HTTP-Client-TLS.html#v:applyDigestAuth)

I don’t think it was necessary, you could always call `defaultMakeClientRequest` by prefixing it with `pure`/`return`. Or am I missing something?

`MonadIO` constraint is also an overkill. `Applicative` is enough as a dependency to do the same thing.

I’ll make a pull request.

BartSchuurmans:
It's not about defaultMakeClientRequest, it's about being able to configure ClientEnv with a different makeClientRequest function that requires IO

unclechu:
I would just mean the only change required is for `ClientEnv` but `defaultMakeClientRequest` didn’t have to be changed.

unclechu:
The change for the `defaultMakeClientRequest` return type: https://github.com/haskell-servant/servant/pull/1789

PR #1596, opened 2022-07-17
servant-auth-swagger: allow base-4.16

maksbotan:


maksbotan:
Published a revision: https://hackage.haskell.org/package/servant-auth-swagger-0.2.10.1/revisions/

PR #1586, opened 2022-04-29
Add failing test case to demonstrate query param issue

tomsmalley:
See issue #1584

PR #1597, opened 2022-07-19
Escape special chars in QueryParams

### LINKED ISSUE #1584: Query parameter encoding
In v0.18, query parameters were converted with [toQueryParam](https://hackage.haskell.org/package/servant-0.19/docs/Servant-API.html#v:toQueryParam) and later encoded as part of [renderQuery](https://hackage.haskell.org/package/http-types-0.12.3/docs/Network-HTTP-Types.html#v:renderQuery).

In v0.19, this was changed so that conversion and encoding of query parameters is done at once by [toEncodedUrlPiece](https://hackage.haskell.org/package/servant-0.19/docs/Servant-API.html#v:toEncodedUrlPiece), and the parts are joined without doing extra encoding.

At first glance this seems like a good change since it allows more control over how encoding is done, but there's a subtlety:

The default implementation of toEncodedUrlPiece uses [encodePathSegmentsRelative](https://hackage.haskell.org/package/http-types-0.12.3/docs/Network-HTTP-Types.html#v:encodePathSegmentsRelative), which uses [urlEncodeBuilder](https://hackage.haskell.org/package/http-types-0.12.3/docs/Network-HTTP-Types.html#v:urlEncodeBuilder) but with the first argument set to `False`: i.e., denoting the input as a path element, not a query string.

So, when using servant "end to end" on the same API with `String` or `Text` query params, the server no longer gets what the client sent if any of the extra query special characters are present `&=+$,`. I _think_ this change wasn't intentional, thus this ticket.

I suppose a more correct change would be to add `toEncodedQuery` or something to the class, but that's a much larger change.

lagunoff:
closes #1584

I was investigating a bug in our application and it turned out it was caused by servant-client not escaping plus signs in query params. After some digging down the code in servant-client I noticed servant-client uses `toEncodedUrlPiece` to encode the values of QueryParams, which is wrong because it should be used for path pieces only and it does not escape characters `:@&=+$,`. Later I found https://github.com/haskell-servant/servant/issues/1584 and https://github.com/haskell-servant/servant/pull/1586. Here I'm suggesting a workaround — using `urlEncodeBuilder True` directly instead of `toEncodedUrlPiece`.

At first I thought the `toEncodedUrlPiece` is itself the source of the problem and created a PR to https://github.com/fizruk/http-api-data/pull/119. The conclusion there was that it is wrong to encode query params with `toEncodedUrlPiece` and that this should be changed in servant-client, even though it's weird that ToHttpAPIData only has method for encoding path pieces

### DIFF
--- a/servant-client-core/src/Servant/Client/Core/Request.hs
+++ b/servant-client-core/src/Servant/Client/Core/Request.hs
@@ -51,9 +51,9 @@ import           Network.HTTP.Media
                  (MediaType)
 import           Network.HTTP.Types
                  (Header, HeaderName, HttpVersion (..), Method, QueryItem,
-                 http11, methodGet)
+                 http11, methodGet, urlEncodeBuilder)
 import           Servant.API
-                 (ToHttpApiData, toEncodedUrlPiece, toHeader, SourceIO)
+                 (ToHttpApiData, toEncodedUrlPiece, toQueryParam, toHeader, SourceIO)
 
 import Servant.Client.Core.Internal (mediaTypeRnf)
 
@@ -165,7 +165,8 @@ appendToQueryString pname pvalue req
 -- | Encode a query parameter value.
 --
 encodeQueryParamValue :: ToHttpApiData a => a  -> BS.ByteString
-encodeQueryParamValue = LBS.toStrict . Builder.toLazyByteString . toEncodedUrlPiece
+encodeQueryParamValue = LBS.toStrict . Builder.toLazyByteString
+  . urlEncodeBuilder True . encodeUtf8 . toQueryParam
 
 -- | Add header to the request being constructed.
 --


jkarni:
Thanks! And sorry this took so long.

jkarni:
@lagunoff @tomsmalley it ike this PR causes a failure (https://github.com/haskell-servant/servant/actions/runs/4752900295/jobs/8443824136#step:8:352) - not exactly sure why it didn't fail in the previous run though. Any thoughts on how to fix this?

tomsmalley:
The failing test (binary data in query param) was introduced by the change which broke query param text encoding (#1432). I think it's either one or the other unless we add `toEncodedQuery` to `ToHttpApiData`. If the decision is to walk back #1432 then removing the test case seems fine.

lagunoff:
Hi, I'm afraid I can't provide any help with this issue. It's not clear what's going on under the hood in the test case it seems it sends a request and compares the response body, but I'm not familiar with the codebase can't dig deeper where exactly encoding/decoding fails

ysangkok:
For any subscribers on this issue, the failing test is proposed to be removed in #1679

PR #1604, opened 2022-08-28
Full query string helpers

divarvel:
The goal of this PR is to adress a use-case that is not currently covered by servant: query strings with dynamic parameter names.

`QueryParam` and its friends are super useful, but extract the parameter name from a type-level string, so while this is useful in many cases, it cannot cover every use-case. A typical use-case would be implementing a search filter where filters are not static and can't be enumerated ahead of time.

I'll propose two helpers, directly derived from a use case that came up at work:

- one is an escape hatch that extracts the whole query string as a `[(ByteString, Maybe ByteString)]` value. It cannot fail and provides raw access to the query string: `QueryString :> Get '[JSON] [Book]`.  
  This was the first version that i implemented for my concrete use-case at work, to get things going.
- the other one is a bit more constrained and implements a pattern commonly known as [_deep objects_](https://swagger.io/docs/specification/serialization/). It builds upon the convention of using `[]` for a list of parameters: `books?filter[search]=value&filter[author][name]=value`. The corresponding type would be `DeepQuery "filter" BookQuery :> Get '[JSON] [Book]`.  
 This is based on what's currently used for my use-case at work.
 

## Raw query string capture

- [x] Servant.API
- [x] Servant.Server
- [x] Servant.Server Specs
- [x] Servant.Server Docs
- [x] Servant.Client
- [x] Servant.Client Specs
- [x] Servant.Client Docs

## Deep object query parsing

- [x] Servant.API
- [x] Servant.Server
- [x] Servant.Server Specs
- [x] Servant.Server Docs
- [x] Servant.Client
- [x] Servant.Client Specs
- [x] Servant.Client Docs

### DIFF
--- a/changelog.d/1604
+++ b/changelog.d/1604
@@ -1,0 +1,10 @@
+synopsis: Full query string helpers
+prs: #1604
+
+description: {
+  This PR introduces `DeepQuery`, a route combinator that implements a pattern commonly known as deep objects.
+  It builds upon the convention of using `[]` for a list of parameters: 
+  `books?filter[search]=value&filter[author][name]=value`.
+  The corresponding type would be `DeepQuery "filter" BookQuery :> Get '[JSON] [Book]`.
+
+}
--- a/servant-client-core/src/Servant/Client/Core/HasClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/HasClient.hs
@@ -22,7 +22,7 @@ module Servant.Client.Core.HasClient (
     (//),
     (/:),
     foldMapUnion,
-    matchUnion,
+    matchUnion
     ) where
 
 import           Prelude ()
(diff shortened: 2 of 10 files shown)


tchoutri:
Considering the limitations of what we already have, these additions are most welcome. Ping me if you ever want to pair on a Cookbook addition for this.

divarvel:
I'm not using haskell at work any more and I don't have time to revive this PR myself, but I still think it's a worthwhile addition to servant and am available to help anyone who wants to get it over the finish line.

tchoutri (approved):

ysangkok:
@tchoutri Since this PR broke backwards compatibility, we need to do a major version bump. Do we do that now, or just before the release? The advantage of doing it now is that then we don't have to remember that this PR changed the interface.

tchoutri:
@ysangkok yep let's do it right away. If you can't do it in the next 10 hours I'll do it in my morning. :)

PR #1614, opened 2022-09-26
Allow lens 5.2

TristanCacqueray:
See https://hackage.haskell.org/package/lens-5.2/changelog


ysangkok:
This only changes four occurrences, so I think #1607 is better.

PR #1618, opened 2022-10-05
Allow http-api-data 0.5

felixonmars:
Builds fine and all tests pass.


servant-server/servant-server.cabal
```diff
@@ -73,7 +73,7 @@ library
   -- strict dependency as we re-export 'servant' things.
   build-depends:
       servant             >= 0.19     && < 0.20
-    , http-api-data       >= 0.4.1    && < 0.4.4
+    , http-api-data       >= 0.4.1    && < 0.5.1
```
endgame, line 76:
```suggestion
    , http-api-data       >= 0.4.1    && < 0.6
```

servant/servant.cabal
```diff
@@ -92,7 +92,7 @@ library
   -- We depend (heavily) on the API of these packages:
   -- i.e. re-export, or allow using without direct dependency
   build-depends:
-      http-api-data          >= 0.4.1    && < 0.4.4
+      http-api-data          >= 0.4.1    && < 0.5.1
```
endgame, line 95:
```suggestion
      http-api-data          >= 0.4.1    && < 0.6
```

endgame (commented):
Under the PVP, versions `>=0.5 && <0.6` aren't allowed to make breaking changes, so I recommend relaxing the upper bound to `<0.6`.

servant-server/servant-server.cabal
felixonmars: Updated.

endgame: It appears you only updated the bounds for `servant` but not `servant-server`.

endgame:
Reposting here because I can't unresolve the conversation above: It looks like the upper bound for `servant-server` is still on `<0.5.1`. Should it be `<0.6`?

*****

As a Hackage Trustee, I have made the following revisions:

* `servant-0.19:` `>=0.4.1 && <0.4.4` -> `>=0.4.1 && <0.6`
* `servant-server-0.19.1`: `>=0.4.1 && <0.4.4` -> `>=0.4.1 && <0.5.1` (I think this could be looser, but I'm revising it to match the PR as it stands.)

felixonmars:
Updated to 0.6 everywhere and resolved conflict.

ysangkok:
Thank you for the PR, a similar change was merged as part of #1663, so I'll close this. Please reopen if there are any issues.

PR #1624, opened 2022-10-30
Remove postgresql-simple allow-newer

TeofilC:
No longer needed as a new version of the library has been released


tchoutri (commented):
nice.

TeofilC:
See 0fc6e395cbf808a30e57025ac1b8a5649fe75ec6

Issue #1626, opened 2022-11-01
QueryParam doesn't conform to HTTP


cblp:
Instance HasClient QueryParam' uses this function

```hs
encodeQueryParam :: ToHttpApiData a => a  -> BS.ByteString
encodeQueryParam = BL.toStrict . toLazyByteString . toEncodedUrlPiece
```

where `toEncodedUrlPiece` is from `http-api-data` package, where it is documented to be applicable to path segments only, not to query parameters. As a result, if we try to send a simple text string as a parameter,

```hs
type API = QueryParam "p" Text
```

this will be encoded as `?p=a%20+%20b` which is wrong, because `+` is the encoding for the space, and `+` must be encoded as `%2B`.

`?p=a%20+%20b` is interpreted by a standard-conforming server as `p = "a   b"`.

`p = "a + b"` must be encoded as `?p=a+%2B+b`

cblp:
If you think http-api-data must be fixed, see https://github.com/fizruk/http-api-data/pull/123

ysangkok:
@cblp Has this been addressed by #1597, which was released as part of servant-0.20?

cblp:
I think, this is already fixed in http-api-data, but I can't recheck right now.

tchoutri:
@cblp Is this still an ongoing problem?

cblp:
Look like it's fixed, but I'm not sure.

PR #1628, opened 2022-11-06
Use toEncodedQueryParam for QueryParam

cblp:


cblp:
This fixes #1626

cblp:
A complete solution needs those changes to be released; this is just an idea of solution.

Issue #1645, opened 2023-01-29
MimeRender/MimeUnrender instances for String do not agree with PlainText instance for Accept


plredmond:
<details><summary>imports</summary>

```haskell
import Servant.API
import qualified Data.ByteString.Lazy.Char8 as C8
import qualified Data.ByteString.Lazy as BSL
import qualified Data.Text.Lazy.Encoding as TLE
import qualified Data.Text.Lazy as T
```

</details>

The [`Accept PlainText` instance](https://hackage.haskell.org/package/servant-0.19.1/docs/Servant-API-ContentTypes.html#t:PlainText) gives the mimetype `text/plain;charset=utf-8`, but the [`MimeRender PlainText String` instance](https://hackage.haskell.org/package/servant-0.19.1/docs/Servant-API-ContentTypes.html#t:PlainText) uses the [`Data.ByteString.Lazy.Char8.pack` function](https://hackage.haskell.org/package/bytestring-0.11.4.0/docs/Data-ByteString-Lazy-Char8.html#v:pack), which produces ASCII. Furthermore, `Data.ByteString.Lazy.Char8.pack` mangles unicode characters.

```haskell
ghci> let unicode = "\53575" :: String -- A hangul character
ghci> putStrLn unicode
텇
ghci> TLE.encodeUtf8 (TL.pack unicode) :: BSL.ByteString -- Expected result (UTF-8 bytes)
"\237\133\135"
ghci> mimeRender (Proxy :: Proxy PlainText) unicode :: BSL.ByteString -- Actual result (ASCII bytes)
"G"
```

A similar problem exists with the [`MimeUnrender PlainText String` instance](https://hackage.haskell.org/package/servant-0.19.1/docs/Servant-API-ContentTypes.html#t:PlainText) which uses the corresponding `Data.ByteString.Lazy.Char8.unpack` function which parses ASCII and doesn't recognize UTF-8 data.

```haskell
ghci> let bytes = TLE.encodeUtf8 $ TL.pack ("\53575" :: String) :: BSL.ByteString -- UTF-8 bytes (the expected result from above)
ghci> TL.unpack (TLE.decodeUtf8 bytes) :: String -- Expected result (A hangul character)
"\53575"
ghci> mimeUnrender (Proxy :: Proxy PlainText) bytes :: Either String String -- Actual result (ASCII mojibake)
Right "\237\133\135"
```

[...]

PR #1646, opened 2023-01-29
change `Mime(Un)Render PlainText String` instances to encode/decode UTF-8

plredmond:
Commit message: change `Mime(Un)Render PlainText String` instances to encode/decode UTF-8 data as specified by `Accept PlainText` instance

Details: https://github.com/haskell-servant/servant/issues/1645


jkarni:
Could you add an entry into the changelog? Otherwise, looks good to me. (And sorry for the delay.)

plredmond:
No worries. Thank you for responding. :)

PR #1653, opened 2023-02-26
Fix AddSetCookies instance

tchoutri:
Depends on https://github.com/haskell-servant/servant/pull/1652

Please review this carefully.

cc @hasufell @alpmestan @gdeest


tchoutri:
@mgajda Your input is also valued. :)

endgame:
I don't think this is right, because if you have this instance then it's impossible to change the types on either side of a `:<|>`, which doesn't seem like what you want?

tchoutri:
@endgame great, thank you!

PR #1660, opened 2023-03-17
Add servant-io-streams

shlevy:


servant-io-streams/src/Servant/IO/Streams.hs
```diff
@@ -0,0 +1,31 @@
...
+  toSourceIO src = S.SourceT ($ go)
+    where
+      go = S.Effect $ trans <$> IOS.read src
+
+      trans Nothing = S.Stop
+      trans (Just c) = S.Yield c go
+
+instance FromSourceIO a (IOS.InputStream a) where
+  {-# NOINLINE fromSourceIO #-}
+  fromSourceIO src = unsafePerformIO $ S.unSourceT src $ IOS.fromGenerator . gen
```
shlevy, line 25: Doesn't seem that outlandish for a streaming interface to require some initial stateful allocation... Any chance we could have variants of these typeclasses operating in a `MonadIO`?

shlevy:
Any thoughts?

shlevy:
Will update after #1661 is merged

servant-io-streams/servant-io-streams.cabal
```diff
@@ -0,0 +1,57 @@
...
+source-repository head
+  type: git
+  location: http://github.com/haskell-servant/servant.git
+
+library
+  exposed-modules:     Servant.IO.Streams
+  build-depends:
+      base        >=4.9  && <5
+    , io-streams ^>=1.5
+    , servant     >=0.15 && <0.20
```
shlevy, line 33: These bounds are not correct, will need to update to whatever the new major version will be that includes #1661

PR #1662, opened 2023-03-17
Add SameSite attribute to XSRF token cookie

Minnozz:
It is currently missing, which causes a [warning](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#cookies_without_samesite_default_to_samesitelax) in modern browsers.

By moving it to `applyCookieSettings`, the value specified in `cookieSameSite` is used for both the session cookie and the XSRF token cookie.


tchoutri (approved):
Awesome, thank you :+1:

PR #1664, opened 2023-03-21
Allow semigroupoids < 6.1

Tristano8:
Relaxes bounds on `semigroupoids` for `servant-client` and `servant-http-streams`


Tristano8:
Gentle ping - would it be possible to do a metadata release for this?

PR #1666, opened 2023-03-21
Fix multiple headers with the same name

### LINKED ISSUE #1665: servant-client doesn't handle multiple Set-Cookie headers correctly
`servant-client` seems to concatenate all `Set-Cookie` headers instead of treating them as separate values.

The problem most likely is related to the handling of multiple headers with the same name.

I've added a test to the test suite in order to demonstrate this: https://github.com/dgaw/servant/commit/7b4a39f1e80092c9d1ac918dc3d2b323a0f8a855

Here is the result:
```
Failures:

  test/Servant/SuccessSpec.hs:157:37: 
  1) Servant.Success.Servant.SuccessSpec Returns multiple Set-Cookie headers appropriately
       expected: [("Set-Cookie","cookie1"),("Set-Cookie","cookie2")]
        but got: [("Set-Cookie","cookie1\ncookie2"),("Set-Cookie","cookie1\ncookie2")]
```

dgaw:
Fix recursion in the instance `BuildHeadersTo (Header h v ': xs)` so that multiple headers with the same name work in `servant-client`.

Fixes #1665

### DIFF
--- a/servant/src/Servant/API/ResponseHeaders.hs
+++ b/servant/src/Servant/API/ResponseHeaders.hs
@@ -102,6 +102,8 @@ class BuildHeadersTo hs where
 instance {-# OVERLAPPING #-} BuildHeadersTo '[] where
     buildHeadersTo _ = HNil
 
+-- The current implementation does not manipulate HTTP header field lines in any way,
+-- like merging field lines with the same field name in a single line.
 instance {-# OVERLAPPABLE #-} ( FromHttpApiData v, BuildHeadersTo xs, KnownSymbol h )
          => BuildHeadersTo (Header h v ': xs) where
     buildHeadersTo headers = case L.find wantedHeader headers of
--- a/servant/src/Servant/API/ResponseHeaders.hs
+++ b/servant/src/Servant/API/ResponseHeaders.hs
@@ -39,6 +39,7 @@ import           Control.DeepSeq
 import           Data.ByteString.Char8     as BS
                  (ByteString, init, pack, unlines)
 import qualified Data.CaseInsensitive      as CI
+import qualified Data.List                 as L
 import           Data.Proxy
 import           Data.Typeable
                  (Typeable)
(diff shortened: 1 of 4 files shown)


tchoutri:
@dgaw Thanks a lot! Could you also include a changelog entry? :) We collect entries in different files in https://github.com/haskell-servant/servant/tree/master/changelog.d and amalgamate them later during release

dgaw:
@tchoutri I've added a changelog entry and removed an old comment.

tchoutri (approved):
Awesome, thank you

servant/src/Servant/API/ResponseHeaders.hs
```diff
@@ -97,24 +98,19 @@ type family HeaderValMap (f :: * -> *) (xs :: [*]) where
...
-          Left _err -> UndecodableHeader (BS.init $ BS.unlines xs)
-             `HCons` buildHeadersTo headers
-          Right h   -> Header h `HCons` buildHeadersTo headers
+    buildHeadersTo headers = case L.find wantedHeader headers of
+      Nothing -> MissingHeader `HCons` buildHeadersTo headers
+      Just header@(_, val) -> case parseHeader val of
+        Left _err -> UndecodableHeader val `HCons` buildHeadersTo (L.delete header headers)
+        Right h   -> Header h `HCons` buildHeadersTo (L.delete header headers)
+      where wantedHeader (h, _) = h == wantedHeaderName
+            wantedHeaderName = CI.mk . pack $ symbolVal (Proxy :: Proxy h)
```
marinelli, line 113: I had a look at multiple RFCs regarding HTTP header fields, i.e.:
- [1] https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2
- [2] https://www.rfc-editor.org/rfc/rfc9110#name-field-order
- [3] https://www.rfc-editor.org/rfc/rfc6265.html
- [4] https://www.rfc-editor.org/rfc/rfc8941.html
- [5] https://httpwg.org/http-extensions/draft-ietf-httpbis-retrofit.html

At the end, we do not still have a precise definition for which field lines with the same field name are allowed to be present in the HTTP message or not (except for the `Set-Cookie`), and which ones are allowed to be _merged_ in some way.
We have something about it, see [4] and [5], and those RFCs are pretty new or are still drafts.

I think that the current (fixed) approach of not manipulating the _requested_ header fields is ok.
We'll be able to create also _not valid_ messages, and this is good for testing (here being _not valid_ means that we could generate HTTP messages with header fields with the same field name, for which we require to have just one line, and not multiple ones).

I would just add a short comment that summarise this behaviour, like: _The current implementation do not manipulate HTTP header field lines in any way, like merging field lines with the same field name in a single line._

dgaw: Thanks for doing the research @marinelli! I've added the suggested comment.

marinelli (approved):

tchoutri (approved):

ysangkok:
This is now released on Hackage and is visible in [the changelog for servant-client-0.20](https://hackage.haskell.org/package/servant-client-0.20/changelog#020).

PR #1677, opened 2023-04-16
Chore: update extensions list

### LINKED ISSUE #1202: Check other-extensions field

michivi:
This PR:

- Updates the extensions list in the `servant.cabal` file (should fix #1202).
- Removes some unnecessary extensions.


servant/servant.cabal
```diff
@@ -121,8 +121,8 @@ library
                   , ConstraintKinds
                   , DataKinds
                   , DeriveDataTypeable
+                  , DeriveFunctor
```
ysangkok, line 124: Why did this need to get turned on?

ysangkok: I don't think there is really any system to the extensions enabled. I think it would be best to fix the failing query string test on master first. But I do think this looks sensible. Is TypeFamilyDependencies added for the same reason?

michivi: Yes. This one is explicitly enabled for `servant/src/Servant/API/ResponseHeaders.hs`.

PR #1687, opened 2023-06-20
Add attoparsec-aeson dependency

9999years:
This makes `servant` buildable with the upcoming aeson-2.2 release.

### DIFF
--- a/servant/servant.cabal
+++ b/servant/servant.cabal
@@ -102,6 +102,7 @@ library
       base-compat            >= 0.10.5   && < 0.14
     , aeson                  >= 1.4.1.0  && < 3
     , attoparsec             >= 0.13.2.2 && < 0.15
+    , attoparsec-aeson       >= 2.2      && < 3
     , bifunctors             >= 5.5.3    && < 5.7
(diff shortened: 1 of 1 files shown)


tchoutri:
@9999years Hi! Is there a reason why you wish to add attoparsec-aeson to the dependencies of `servant`? :)

9999years:
Yes, it's for the upcoming aeson-2.2 release, which moves the `Data.Aeson.Parser` module from `aeson` to `attoparsec-aeson`.

```
src/Servant/API/ContentTypes.hs:79:1: error:
    Could not find module ‘Data.Aeson.Parser’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
79 | import           Data.Aeson.Parser
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

aeson-2.2 isn't released yet, so we don't need to merge this yet, but I'm working on supporting it internally and created this PR to reference in our build.

tchoutri:
I see, thank you :)

ysangkok:
Thanks for the contribution, but we merged #1695 instead, so this is no longer necessary.

PR #1690, opened 2023-06-26
Allow newer hspec, free

ysangkok:
Especially the free bound is important for inclusion into Stackage Nightly.

I have also bumped the min bound for servant-auth, since jose-0.10 requires aeson-2, we might as well.

Also included is a bump for monad-time.

### DIFF
--- a/servant-auth/servant-auth/servant-auth.cabal
+++ b/servant-auth/servant-auth/servant-auth.cabal
@@ -35,7 +35,7 @@ library
   build-depends:
       base                    >= 4.10     && < 4.19
     , containers              >= 0.6      && < 0.7
-    , aeson                   >= 1.3.1.1  && < 3
+    , aeson                   >= 2.0      && < 3
     , jose                    >= 0.10     && < 0.11
     , lens                    >= 4.16.1   && < 5.3
     , servant                 >= 0.15     && < 0.21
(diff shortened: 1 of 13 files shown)


tchoutri (approved):

maksbotan:
@ysangkok sure, done!

do other packages have to be revised as well?

ysangkok:
It's not strictly necessary, as hspec is only for running tests. But if you like, we can bump all these hspec bounds, and then Nightly users will be able to run the test suites.

maksbotan:
afair, stackage runs tests when building lts, so I will make those revisions. later this day

wonder if that can be automated 🤔

servant-auth/servant-auth/servant-auth.cabal
```diff
@@ -35,7 +35,7 @@ library
   build-depends:
       base                    >= 4.10     && < 4.19
     , containers              >= 0.6      && < 0.7
-    , aeson                   >= 1.3.1.1  && < 3
+    , aeson                   >= 2.0      && < 3
```
maksbotan, line 38: @ysangkok is this really necessary?

maksbotan:
I've revised (hopefully) everything on Hackage except `servant-auth`.

servant-auth/servant-auth/servant-auth.cabal
ysangkok: the version of `jose` we use doesn't permit aeson 1.x. So I think it is better to set the tighten the bound here to reflect what can actually be chosen. So no, it is not necessary, but having 1.3.1.1 is confusing, and therefore I think we should get rid of it.

PR #1695, opened 2023-07-11
Allow aeson-2.2

maksbotan:


maksbotan:
@theophile-fl @ysangkok @jkarni and co, I need your advice on this. Servant.API.ContentTypes uses `Data.Aeson.Parser`:

https://github.com/haskell-servant/servant/blob/16cfd03e74557d8fd8ef6e2ebfb9c26bd645d756/servant/src/Servant/API/ContentTypes.hs#L374-L391

`Data.Aeson.Parser` module was removed in aeson-2.2, so there is no easy way to upgrade our version constraint.

However, I think that this function is not needed anymore. Haddock mentions that it parser any json value, contrary to aeson's behavior to parse only list and object. See haddock for 'json' in aeson-2.1: https://hackage.haskell.org/package/aeson-2.1.2.1/docs/Data-Aeson-Parser.html#v:json

> This function is an alias for [value](https://hackage.haskell.org/package/aeson-2.1.2.1/docs/Data-Aeson-Parser.html#v:value). In aeson 0.8 and earlier, it parsed only object or array types, in conformance with the now-obsolete RFC 4627.

Our lower bound on aeson is 1.4 for a long time, so I propose to remove our `eitherDecodeLenient` and use aeson's native `eitherDecode` instead. What do you think? I'm not sure if we have some tests to verify that the change is OK.

I will make the change in this PR a bit later.

tchoutri:
I support your proposition, in any case we will want to make sure that a pre-release is uploaded to Hackage so that users (me included) can test the behaviour. :)

maksbotan:
There was a test that servant's `eitherDecodeLenient` behaves exactly like `eitherDecode` from aeson:

https://github.com/haskell-servant/servant/blob/104a8ad2f387b68d6844101e01b585d106f4d0d3/servant/test/Servant/API/ContentTypesSpec.hs#L222-L229

I've pushed an update to check if everything works. I've also had to change the tutorial, as it also mentioned old `aeson` behavior.

servant/test/Servant/API/ContentTypesSpec.hs
```diff
@@ -219,16 +219,6 @@ spec = describe "Servant.API.ContentTypes" $ do
                 handleCTypeH (Proxy :: Proxy '[JSONorText]) "image/jpeg"
                     "foobar" `shouldBe` (Nothing :: Maybe (Either String Int))
 
-    -- aeson >= 0.9 decodes top-level strings
-    describe "eitherDecodeLenient" $ do
-
-        it "parses top-level strings" $ do
```
ysangkok, line 225: It would be best if we could retain this test, ideally by having it use the `MimeUnrender JSON` instance that was modified. Just to make sure the docs are not mistaken.

maksbotan:
I've returned the test

maksbotan:
Do you want me to upload candidate to Hackage to test?

tchoutri:
That would be lovely yes

maksbotan:
It's here: https://hackage.haskell.org/package/servant-0.20.1/candidate

You can try it by adding https://hackage.haskell.org/package/servant-0.20.1/candidate/servant-0.20.1.tar.gz to your `cabal.project` or `stack.yaml`.

tchoutri:
@maksbotan Wonderful. I think servant-server & servant-client-core are going to need a pre-release as well, I'm getting:

```

src/Servant/Server/Internal.hs:686:26: error:
    • Couldn't match type ‘a’ with ‘IO a0’
      Expected: SourceIO chunk -> a
        Actual: SourceIO chunk -> IO a0
      ‘a’ is a rigid type variable bound by
        the instance declaration
        at src/Servant/Server/Internal.hs:(673,5)-(675,68)
    • In the first argument of ‘return’, namely ‘fromSourceIO’
      In the expression: return fromSourceIO
      In an equation for ‘ctCheck’: ctCheck = return fromSourceIO
```

and

```
src/Servant/Client/Core/HasClient.hs:430:35: error:
    • Couldn't match type ‘a’ with ‘IO a1’
      Expected: Client m (Stream method status framing ct a)
        Actual: m (IO a1)
      ‘a’ is a rigid type variable bound by
        the instance declaration
        at src/Servant/Client/Core/HasClient.hs:(422,3)-(424,54)
    • In the expression:
        withStreamingRequest req'
          $ \ gres
              -> do let ...
                    return $ fromSourceIO $ framingUnrender' $ responseBody gres
      In an equation for ‘clientWithRoute’:
          clientWithRoute _pm Proxy req
            = withStreamingRequest req'
                $ \ gres
                    -> do let ...
                          ....
            where
                req'
                  = req
                      {requestAccept = fromList [...],
                       requestMethod = reflectMethod (Proxy :: Proxy method)}
      In the instance declaration for
        ‘HasClient m (Stream method status framing ct a)’
    • Relevant bindings include
        clientWithRoute :: Proxy m
                           -> Proxy (Stream method status framing ct a)
                           -> Request
                           -> Client m (Stream method status framing ct a)
          (bound at src/Servant/Client/Core/HasClient.hs:430:3)
    |
430 |   clientWithRoute _pm Proxy req = withStreamingRequest req' $ \gres -> do
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...

```

tchoutri:
I will also send a call for beta-testers on social media to make sure our users can take the time to test things on their own applications

maksbotan:
Okay, so let's wait for a bit for more tests, and I'll merge & release.

tchoutri:
@maksbotan servant-lucid would need a bound bump, I can do it for you if you give me permission on Hackage. :)

ysangkok:
@maksbotan Has this been sufficiently tested already?

Vekhir:
@maksbotan 
Is there anything blocking this from getting merged?

Publishing this patch in a new version would allow the Arch package to be updated without carrying an additional patch.

larskuhtz:
I would also love to see this merged and released. 

(I think, as a web framework, servant should try to always support the latest version of aeson.)

.github/workflows/master.yml
```diff
@@ -3,6 +3,8 @@ name: CI
 on:
   pull_request:
   push:
+    branches:
+      - master
```
larskuhtz, line 7: This seems unrelated to this PR.

.github/workflows/master.yml
```diff
@@ -20,6 +22,7 @@ jobs:
           - "9.2.8"
           - "9.4.5"
           - "9.6.2"
+      fail-fast: false
```
larskuhtz, line 25: This change seems unrelated to this PR.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -65,7 +65,6 @@ module Servant.API.ContentTypes
     , AllMime(..)
     , AllMimeRender(..)
     , AllMimeUnrender(..)
-    , eitherDecodeLenient
```
larskuhtz, line 68: From the comments below it seems that with currently supported aeson versions `eitherDecodeLenient` and `eitherDecode` have the same behavior. 

What about making `eitherDecodeLenient` an alias of `eitherDecode` and just mark it as deprecated? That way it would not break the API.

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -371,28 +364,9 @@ instance NFData NoContent
 --------------------------------------------------------------------------
 -- * MimeUnrender Instances
```
larskuhtz, line 366:
```suggestion
-- | Deprecated: since aeson version 0.9 `eitherDecode` has lenient behavior.
--
 eitherDecodeLenient :: FromJSON a => ByteString -> Either String a
 eitherDecodeLenient = eitherDecode
 {-# DEPRECATED eitherDecodeLenient "use eitherDecode instead" #-}
```

servant/src/Servant/API/ContentTypes.hs
```diff
@@ -65,7 +65,6 @@ module Servant.API.ContentTypes
     , AllMime(..)
     , AllMimeRender(..)
     , AllMimeUnrender(..)
-    , eitherDecodeLenient
     , canHandleAcceptH
```
larskuhtz, line 68: Keep `eitherDecodeLenient` as alias of `eitherDecde`, but mark it as deprecated (cf. below).

```suggestion
    , eitherDecodeLenient
    , canHandleAcceptH
```

servant/test/Servant/API/ContentTypesSpec.hs
```diff
@@ -219,15 +219,13 @@ spec = describe "Servant.API.ContentTypes" $ do
                 handleCTypeH (Proxy :: Proxy '[JSONorText]) "image/jpeg"
                     "foobar" `shouldBe` (Nothing :: Maybe (Either String Int))
 
-    -- aeson >= 0.9 decodes top-level strings
     describe "eitherDecodeLenient" $ do
```
larskuhtz, line 222:
```suggestion
    describe "eitherDecode is lenient" $ do
```

larskuhtz (commented):
I propose to keep `eitherDecodeLenient` as an deprecated alias of `eitherDecode`, since both have (probably) the same behavior since version 0.9 of aeson. (I haven't checked possible differences is the handling of of trailing characters, but I expect that to be a negligible issue.)

This would reduce the potential of the PR to break users and will hopefully help this change to be release soon.

TeofilC:
Let me know if you'd appreciate help getting this over the line, but no pressure

Vekhir:
The CI fails because it tries to build `warp-3.3.25` which is incompatible with `http2-4.2.0`. There is a newer version, `warp-3.3.29` which is compatible, but not chosen.

maksbotan:
Thank you @Vekhir. I'll look into this.

ysangkok:
> Thank you @Vekhir. I'll look into this.

You might want to remove or flip `constraints: warp < 3.3.26` in `cabal.project`.

maksbotan:
Due to backwards-incompatible changes to master since this PR (see #1697), I will have to make Hackage release from this branch.

@ysangkok @tchoutri @larskuhtz please check out this release candidate: https://hackage.haskell.org/package/servant-0.20.1/candidate

Direct tar gz link: https://hackage.haskell.org/package/servant-0.20.1/candidate/servant-0.20.1.tar.gz

If everything's fine, I'll publish a release.

tchoutri:
Alright, except for a usage of `eitherDecodeLenient` and odd-jobs lagging a bit behind for aeson-2.2 compat, I haven't encountered any problem.

maksbotan:
What do you mean by "odd-jobs lagging"?

tchoutri:
It still does not support aeson-2.2 https://github.com/saurabhnanda/odd-jobs/pull/101

maksbotan:
Ah, didn't get that it's a name of a package :)

Great, I'll publish that release on Hackage.

maksbotan:
Published!

PR #1697, opened 2023-07-19
Use full header type in response header instances

worm2fed:
This is required to be able provide `Description` into headers, for example like this:
```
type PaginationTotalCountHeader =
  Header'
    '[ Description "Indicates to the client total count of items in collection"
     , Optional
     , Strict
     ]
    "Total-Count"
    Int
```

This PR is related to another one: https://github.com/biocad/servant-openapi3/pull/48 and related to issue https://github.com/biocad/servant-openapi3/issues/45

### DIFF
--- a/changelog.d/full-header-type
+++ b/changelog.d/full-header-type
@@ -1,0 +1,23 @@
+synopsis: Use `Header'` in response headers.
+prs: #1697
+
+description: {
+
+Use `Header'` instead of `Header` in response, so it's possible to provide
+`Description`, for example:
+
+```
+type PaginationTotalCountHeader =
+  Header'
+    '[ Description "Indicates to the client total count of items in collection"
+     , Optional
+     , Strict
+     ]
+    "Total-Count"
+    Int
+```
+
+Note: if you want to add header with description you should use `addHeader'`
+or `noHeader'` which accepts `Header'` with all modifiers.
+
+}
--- a/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal/AddSetCookie.hs
+++ b/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal/AddSetCookie.hs
@@ -7,11 +7,9 @@ module Servant.Auth.Server.Internal.AddSetCookie where
 
 import           Blaze.ByteString.Builder (toByteString)
 import qualified Data.ByteString          as BS
-import           Data.Tagged              (Tagged (..))
 import qualified Network.HTTP.Types       as HTTP
 import           Network.Wai              (mapResponseHeaders)
 import           Servant
-import           Servant.API.UVerb.Union
 import           Servant.API.Generic
 import           Servant.Server.Generic
 import           Web.Cookie
@@ -76,7 +74,7 @@ instance (orig1 ~ orig2) => AddSetCookies 'Z orig1 orig2 where
 instance {-# OVERLAPPABLE #-}
   ( Functor m
   , AddSetCookies n (m old) (m cookied)
-  , AddHeader "Set-Cookie" SetCookie cookied new
+  , AddHeader mods "Set-Cookie" SetCookie cookied new
   ) => AddSetCookies ('S n) (m old) (m new)  where
   addSetCookies (mCookie `SetCookieCons` rest) oldVal =
     case mCookie of
--- a/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal/AddSetCookie.hs
+++ b/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal/AddSetCookie.hs
@@ -78,8 +78,8 @@ instance {-# OVERLAPPABLE #-}
   ) => AddSetCookies ('S n) (m old) (m new)  where
   addSetCookies (mCookie `SetCookieCons` rest) oldVal =
     case mCookie of
-      Nothing -> noHeader <$> addSetCookies rest oldVal
-      Just cookie -> addHeader cookie <$> addSetCookies rest oldVal
+      Nothing -> noHeader' <$> addSetCookies rest oldVal
+      Just cookie -> addHeader' cookie <$> addSetCookies rest oldVal
 
 instance {-# OVERLAPS #-}
   (AddSetCookies ('S n) a a', AddSetCookies ('S n) b b')
--- a/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal/Cookie.hs
+++ b/servant-auth/servant-auth-server/src/Servant/Auth/Server/Internal/Cookie.hs
@@ -2,6 +2,7 @@
 module Servant.Auth.Server.Internal.Cookie where
 
 import           Blaze.ByteString.Builder (toByteString)
+import           Control.Monad (MonadPlus(..), guard)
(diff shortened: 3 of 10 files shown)


tchoutri:
@worm2fed thanks a lot for this PR. Does this unlock nice things beyond the openapi feature? If so we should probably have a changelog entry for it. :)

worm2fed:
@tchoutri thank you)
I hope yes, it allows to bring more information into generated openapi3 spec.

As I can see there a few things have to be done:
- add entry to changelog
- update documentation to reflect that this is possible to provide descriptions to `Headers` via modifiers
- review more instances where limited; I found that it's not possible to functions with `AddHeader`, for example
  `addHeader :: [AddHeader](https://hackage.haskell.org/package/servant-0.20/docs/Servant-API.html#t:AddHeader) h v orig new => v -> orig -> new`
 
 Maybe it will be better to define helper type
 ```haskell
 -- | Helper type to define 'Header' with 'Description'.
type HeaderWithDescription name a description = 
  Header' '[Description description, Optional, Strict] name a
```

**Comment by worm2fed** [comment]
Seems like it will be not so trivial, a lot of code relies on current `Header`.

How about redefine current `Header` like this `type Header = Header' '[Description "", Optional, Strict]` so, it will be same, without any description.
and provide alternative: `type HeaderWithDescription description = Header' '[Description description, Optional, Strict]` for those who wants to add it

wdyt?

**Comment by worm2fed** [comment]
@tchoutri maybe you can explain what is going on?

My current understanding is that 
```haskell
-- We need all these fundeps to save type inference
class AddHeader h v orig new (mods :: [*])
    | h v orig mods -> new, new -> h, new -> v, new -> orig, new -> mods where
  addOptionalHeader :: ResponseHeader h v -> orig -> new  -- ^ N.B.: The same header can't be added multiple times
```
in this class all these dependencies required to validate that there is only one exact header. And by using `mods` we break this rule, cause `Header' '[Optional, Strict] "MyHeader" Int` is not same as `Header' '[Description "Just header", Optional, Strict] "MyHeader" Int`

And in sense of logic it is _same_ because description is just "documentation label", but current implementation of providing description to headers via modifiers limits to implement this.
Any way, it can raise another question, how would we have two "same" headers but with different description.

So, to perform such validation we need to add `mods` to `AddHeader`.

worm2fed:
Oh, I see in test
`addHeader @"H1" (5 :: Int) $ ("foo" :: String)` must become `addHeader @"H1" @_ @_ @_ @('[Optional, Strict]) (5 :: Int) $ ("foo" :: String)`

blackheaven:
I think you may consider mimicking [Summary](https://hackage.haskell.org/package/servant-0.20/docs/Servant-API.html#t:Summary):

```
Describe "Indicates to the client total count of items in collection"
  (Header "Total-Count" Int)
```

worm2fed:
@blackheaven I do no see how it will work with `servant-openapi3` this way, only by providing an instance there to extract that description.

I will take a look

worm2fed:
so, smith like this:

```
data Describe (sym :: Symbol) (a :: *)

instance (KnownSymbol sym, ToParamSchema a, KnownSymbol description) => ToResponseHeader (Describe description (Header sym a)) where
  toResponseHeader _ = (hname, header)
    where
      hname = Text.pack (symbolVal (Proxy :: Proxy sym))
      header = mempty
        & description ?~ Text.pack (symbolVal (Proxy :: Proxy description))
        & schema ?~ Inline (toParamSchema (Proxy :: Proxy a))
```
which can be used not only for response headers, but for any

@blackheaven is it what you are suggesting?

blackheaven:
Exactly, thanks.

worm2fed:
@blackheaven I think it will be good to include `Describe` into `servant` package. Or, do you think it will be better to keep all in `servant-openapi3`?

blackheaven:
I would say, in a perfect world, it should be in `servant-openapi3`, but

1. There is already [`Summary`](https://hackage.haskell.org/package/servant-0.20/docs/Servant-API.html#t:Summary)
2. It not a big patch
3. It contribute to a better API readability

worm2fed:
@blackheaven it is a little more difficult than it might seem at first

I've added test and it's failing, `addHeader` cannot add `Header` wrapped in `Describe`.
So it's required to add `AddHeader` instance / change current. I left a commented version there, but it's not working.
Do you know how to solve this problem without introducing new param to `AddHeader`?

```
    • Couldn't match type: Describe "desc" (Header "D" Int)
                     with: Servant.API.Header.Header'
                             '[Servant.API.Modifiers.Optional, Strict] h0 v0
        arising from a use of ‘addHeader’
    • In the first argument of ‘return’, namely ‘(addHeader 5 alice)’
      In the first argument of ‘(:<|>)’, namely
        ‘return (addHeader 5 alice)’
      In the second argument of ‘(:<|>)’, namely
        ‘return (addHeader 5 alice)
           :<|>             
             (return alice :<|> return "B")
               :<|> return (S.source ["bytestring"])’
    |                       
137 |           :<|> return (addHeader 5 alice)
```

This showed, that it will be impossible to add this just to `servant-opeanapi3`, because `servant` changes required as well

blackheaven:
I tried also, I didn't find a solution, it doesn't seems a good idea in the end, sorry for that :/

worm2fed:
I will leave diff here, maybe someone else will want to try
but I will revert this PR to initial version

[haskell-servant_servant_pull_1697.diff.txt](https://github.com/haskell-servant/servant/files/12136722/haskell-servant_servant_pull_1697.diff.txt)

worm2fed:
Strange... tests are fine localy

worm2fed:
@tchoutri should I add `servant-auth` targets to `stack.yaml`? currently they are only it `.cabal`

tchoutri:
@worm2fed good question, I must admit I haven't used stack in some years now. If you add them and it works, I'll approve it. :)

worm2fed:
mm.. seems like `MonadPlus` and `guard` not in base for ghc 9.6.2

ysangkok:
They are in base, you just don't get them automatically through mtl. See [changelog](https://hackage.haskell.org/package/mtl-2.3.1/changelog)

worm2fed:
@tchoutri can you please review this PR?

tchoutri (approved):
LGTM, thanks a lot!

worm2fed:
@tchoutri I think we had to squash fixups before merge..

Any way, thanks)

tchoutri:
@worm2fed I'm pretty sure I did! Please do check though :) https://github.com/haskell-servant/servant/commit/72f5d5c86ef3db22888b34450abe8339e1f5e6b1

worm2fed:
@tchoutri oh, I see, sorry
Was it merge-with squash?

tchoutri:
Yep

maksbotan:
Hi @worm2fed!

As far as I understand, your change to `AddHeader` (adding `mods` type argument) is not backwards-compatible and therefore would require a new servant version, e.g. 0.21.

Can we instead add `AddHeader'` with new argument and make `AddHeader` an alias to it? So that we can release this in 0.20 series without breaking public API.

@tchoutri what do you think?

tchoutri:
A compat alias could be nice yes!

PR #1698, opened 2023-08-01
Add AddSetCookies instance for when the left tree is the same before and after the transformation

ohri-anurag:
Adding code suggested by @mastarija on issue #1601

PR #1699, opened 2023-08-02
Add missing IsElem instance for NamedRoutes

### LINKED ISSUE #1674: Making query parameters optionals in allLinks/safeLink with `NamedRoute`s
There currently does not seem to be a clean way to generate links without query parameters. I tried using safeLink, but there's a missing type family instance of `IsElem` for `NamedRoute`s.

There's a partial workaround of using `ToServantApi` to remove the `NamedRoute`, but it won't work for nested NamedRoutes. Ie, like this:

```
{-# LANGUAGE AllowAmbiguousTypes #-}
myApiSafeLink :: forall endpoint. (IsElem endpoint (ToServantApi MyAPI), HasLink endpoint) => MkLink endpoint a
myApiSafeLink = safeLink' someFunction (Proxy @(ToServantApi MyAPI)) (Proxy @endpoint)
```

RaoulHC:
Added various tests for NamedRoutes too.

In writing the tests I realised that the client combinators `//` and `/:` can be used to create links in the same natural way.
I've added the combinators to the test suite for now, but it probably makes sense to put them somewhere in the `servant` package, though I wasn't sure where. Maybe just in the root `API` module, and then ensure that `HasClient` and `Reexport` in `servant-client-core` reexport them for any backwards compatibility?

Fixes #1674.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -58,6 +58,10 @@ optimization: False
 constraints: crypton < 0, crypton-connection < 0, crypton-x509 < 0, crypton-x509-store < 0, crypton-x509-system < 0, crypton-x509-validation < 0
 constraints: warp < 3.3.26
 
+-- wreq-0.5.4.1 doesn't seem to work with ghc-8.6.5
+if (impl(ghc < 8.8))
+  constraints: wreq == 0.5.4.0
+
 allow-newer: servant-js:base
 
 -- Print ticks so that doctest type querying is consistent across GHC versions.
--- a/changelog.d/1699
+++ b/changelog.d/1699
@@ -1,0 +1,7 @@
+synopsis: Add NamedRoutes instance to IsElem
+prs: #1699
+issues: #1674
+description: {
+Add missing IsElem instance for NamedRoutes, this allows links to be checked
+with `safeLink`.
+}
--- a/servant/src/Servant/API/TypeLevel.hs
+++ b/servant/src/Servant/API/TypeLevel.hs
@@ -64,6 +64,10 @@ import           Servant.API.QueryParam
                  (QueryFlag, QueryParam, QueryParams)
 import           Servant.API.ReqBody
                  (ReqBody)
+import           Servant.API.NamedRoutes
+                 (NamedRoutes)
+import           Servant.API.Generic
+                 (ToServantApi)
 import           Servant.API.Sub
                  (type (:>))
 import           Servant.API.Verbs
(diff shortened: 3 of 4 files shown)


RaoulHC:
The CI build failure seems unrelated to my change and exactly the same as the one in #1698.
Couldn't reproduce it locally, though I wasn't able to get ghc-8.6.5 working on my system, so I've added a constraint to use the previous version for 8.6.5, though I'm not sure this will fix it. Running the same version of cabal locally (3.10.1.0), so it seems odd that it's failing to parse `wreq.cabal`.

cabal.project
```diff
@@ -58,6 +58,10 @@ optimization: False
 constraints: crypton < 0, crypton-connection < 0, crypton-x509 < 0, crypton-x509-store < 0, crypton-x509-system < 0, crypton-x509-validation < 0
 constraints: warp < 3.3.26
 
+-- wreq-0.5.4.1 doesn't seem to work with ghc-8.6.5
+if (impl(ghc <= 8.6))
```
ysangkok, line 62: Upper bounds should never be set with `<=` since that is excluding minor versions (which normally would only contain fixes). So as I read this, 8.6.5 is not matching this predicate.

Better use `< 8.8`.

ysangkok:
This looks good but it is missing a changelog entry. See the other PRs for how to make one.

cabal.project
RaoulHC: Ah didn't realise that, hopefully this will get CI to work for 8.6.5 now.

RaoulHC:
Added a changelog entry, any thoughts on where the `//` and `/:` might belong in `servant`?

ysangkok:
I think that people will only appreciate them if they see examples. And if they see examples, it almost doesn't matter what the name of the combinator is. Personally, I think it is unnecessary since their definitions are trivial. There is no reason to abstract away function application.

changelog.d/1699
```diff
@@ -0,0 +1,7 @@
+synopsis: Add NamedRoutes instance to IsElem
+prs: #1699
+issues: #1674
+description: {
+Add missing IsElem instance for NamedRoutes, this allows links to be checked
+`safeLink`.
```
ysangkok, line 6: Is this a stray line?

RaoulHC:
> I think that people will only appreciate them if they see examples. And if they see examples, it almost doesn't matter what the name of the combinator is. Personally, I think it is unnecessary since their definitions are trivial. There is no reason to abstract away function application.

While `//` is just a redefinition of `&`, `/:` does make it more ergonomic imo. Either way, these helpers are already defined in [`Servant.Client.Core.HasClient`](https://github.com/haskell-servant/servant/blob/master/servant-client-core/src/Servant/Client/Core/HasClient.hs#L887-L955), I just think it might sense to move them to the servant package. If we do that then the documentation should be updated to have examples for both links and clients.

changelog.d/1699
RaoulHC: Just wrapped at that point, though I realised it was missing a 'with'.

ysangkok:
You could submit a separate pull request for that, for now I will merge this since it fixes CI and the instance is useful even without the combinators.

Issue #1601, opened 2022-08-13
AddSetCookies missing an instance for when the left tree is the same before and after the transformation


mastarija:
So, I've got this error

```
server/Server/SRV.hs:63:14: error:
    • Overlapping instances for Servant.Auth.Server.Internal.AddSetCookie.AddSetCookies
                                  ('Servant.Auth.Server.Internal.AddSetCookie.S
                                     ('Servant.Auth.Server.Internal.AddSetCookie.S
                                        'Servant.Auth.Server.Internal.AddSetCookie.Z))
                                  ((Data.Tagged.Tagged Handler Network.Wai.Application
                                    Servant.API.Alternative.:<|> ((Data.Text.Internal.Text
                                                                   -> Data.Tagged.Tagged
                                                                        Handler
                                                                        Network.Wai.Application)
                                                                  Servant.API.Alternative.:<|> Data.Tagged.Tagged
                                                                                                 Handler
                                                                                                 Network.Wai.Application))
                                   Servant.API.Alternative.:<|> (Data.Tagged.Tagged
                                                                   Handler Network.Wai.Application
                                                                 Servant.API.Alternative.:<|> ((Data.Text.Internal.Text
                                                                                                -> Data.Tagged.Tagged
                                                                                                     Handler
                                                                                                     Network.Wai.Application)
                                                                                               Servant.API.Alternative.:<|> Server.API.PdxfAPI.PdxfAPI
                                                                                                                              Flouble
                                                                                                                              (AsServerT
                                                                                                                                 Handler))))
                                  ((Data.Tagged.Tagged Handler Network.Wai.Application
                                    Servant.API.Alternative.:<|> ((Data.Text.Internal.Text
                                                                   -> Data.Tagged.Tagged
                                                                        Handler
                                                                        Network.Wai.Application)
                                                                  Servant.API.Alternative.:<|> Data.Tagged.Tagged
                                                                                                 Handler
                                                                                                 Network.Wai.Application))
                                   Servant.API.Alternative.:<|> (Data.Tagged.Tagged
                                                                   Handler Network.Wai.Application
                                                                 Servant.API.Alternative.:<|> ((Data.Text.Internal.Text
                                                                                                -> Data.Tagged.Tagged
                                                                                                     Handler
                                                                                                     Network.Wai.Application)
                                                                                               Servant.API.Alternative.:<|> Data.Tagged.Tagged
                                                                                                                              Handler
                                                                                                                              Network.Wai.Application)))
[...]

PR #1702, opened 2023-08-22
Add add set cookies instance

ohri-anurag:
Adding code suggested by @mastarija on issue #1601

### DIFF
--- a/changelog.d/1698
+++ b/changelog.d/1698
@@ -1,0 +1,3 @@
+synopsis: Add AddSetCookies instance for when the left tree is the same before and after the transformation.
+prs: #1698
+issues: #1601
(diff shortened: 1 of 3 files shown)


tchoutri:
Fantastic, thank you!

tchoutri (approved):

adinapoli:
@endgame @tchoutri Folks, unless I'm being completely blind (which could just as well be the case and might need more ☕ ) this instance hasn't landed yet on Hackage, and the last uploaded version of the package ( `0.4.8.0` ) doesn't seem to have it.

It would be great if we could have a `0.4.8.1` release published on Hackage at some point 😉 

Thanks for your time!

tchoutri:
Hi @adinapoli, we are preparing a 9.10.1 compat release that will certainly have it. Thank you for reaching out.

PR #1707, opened 2023-09-18
Add `IsIn` instance for `NamedRoutes`

Friede80:
This was needed for adding custom tags using the [subOperations](https://hackage.haskell.org/package/servant-swagger-1.2/docs/Servant-Swagger.html#v:subOperations) combinator. `IsSubAPI` constraint could not resolve for APIs using the new `NamedRoutes`.


tchoutri (approved):
Thank you for your patience!

PR #1720, opened 2024-01-19
Client middleware

thought2:
At work we had the following requirement: Every endpoint call with servant-client should log it's raw request and response data.

At first, it looked like this may be possible to implement with [`managerModifyRequest`](https://hackage.haskell.org/package/http-client-0.7.16/docs/Network-HTTP-Client.html#v:managerModifyRequest) and [`managerModifyResponse`](https://hackage.haskell.org/package/http-client-0.7.16/docs/Network-HTTP-Client.html#v:managerModifyResponse).

However, this turned out to be not a good idea. The docs itself mention for the first function that "This function may be called more than once during request processing". And generally it was important that the logging of the request and response happens at the exact time when they orccur. And those function don't seem to give guarantees about this.

In the end our solution was roughly to create a newtype wrapper around [`ClientM`](https://hackage.haskell.org/package/servant-client-0.20/docs/Servant-Client-Internal-HttpClient.html#t:ClientM) with a custom instance for [`RunClient`](https://hackage.haskell.org/package/servant-client-core-0.20/docs/Servant-Client-Core.html#t:RunClient), which calls [`performRequest`](https://hackage.haskell.org/package/servant-client-0.20/docs/src/Servant.Client.Internal.HttpClient.html#performRequest) but wrapped with some middleware functions.

I wonder if there could be a way to support defining "real" middleware of type `(Request -> ClientM Response) -> (Request -> ClientM Response)` in servant-client.

This PR is an implementation sketch of how I'd imagine this to work. I contains a simple testcase which should clarify the main idea.

I'd be curious to hear what people think about this approach.

Note: This PR does not cover yet how errors thrown in the middleware should be handled. One way to do this would be to extend the `ClientError` type with a new constructor `MiddlewareError SomeException`.


tchoutri:
Hi, thank you for this contribution! As this is a breaking change (adding a publicly exposed field to a record breaks pattern-matching in consumers) and not trivial addition, would you mind adding a changelog entry? :)

tchoutri (approved):

thought2:
@tchoutri
Sure, I added a changelog entry and another test which shows how errors can be thrown in the middleware.

servant-client/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 
 Package versions follow the [Package Versioning Policy](https://pvp.haskell.org/): in A.B.C, bumps to either A or B represent major versions.
 
+0.21
+----
+
+- Clients now support real middleware of type `(Request -> ClientM Response) -> Request -> ClientM Response` which can be configured in `ClientEnv`. This allows access to raw request and response data. It can also be used to control how/when/if actual requests are performed.
```
tchoutri, line 9: Ah, apologies, please use the format described here: https://github.com/haskell-servant/servant/blob/master/CONTRIBUTING.md#changelog-entries

tchoutri:
@abailly May I ask for your opinion on this?

abailly:
@tchoutri Sure, always happy to help but it's been a while since I have used servant in anger :)

servant-client/CHANGELOG.md
thought2: Alright, I reverted the entry in CHANGELOG.md and added a file to the `changelog.d` folder.

thought2:
I also added a test for "chaining middleware".

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -86,11 +86,16 @@ data ClientEnv
...
   }
 
+type ClientApplication = Request -> ClientM Response
+
+type ClientMiddleware = ClientApplication -> ClientApplication
+
 -- | 'ClientEnv' smart constructor.
 mkClientEnv :: Client.Manager -> BaseUrl -> ClientEnv
-mkClientEnv mgr burl = ClientEnv mgr burl Nothing defaultMakeClientRequest
+mkClientEnv mgr burl = ClientEnv mgr burl Nothing defaultMakeClientRequest id
```
abailly, line 98: suggestion: use record fields assignment to make things clearer and know what we are assigning `id` to

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -153,15 +158,18 @@ instance Alt ClientM where
   a <!> b = a `catchError` \_ -> b
 
 instance RunClient ClientM where
-  runRequestAcceptStatus = performRequest
+  runRequestAcceptStatus statuses req = do
+    ClientEnv _ _ _ _ mid <- ask
```
abailly, line 162: suggestion: Use `NamedFieldPuns` to remove the need for positional pattern matching which will make code more robust in the face of change
suggestion: `mid` -> `middleware` will make it clearer what this function is about

servant-client/src/Servant/Client/Internal/HttpClient.hs
```diff
@@ -153,15 +158,18 @@ instance Alt ClientM where
...
+    mid oldApp req
   throwClientError = throwError
 
 runClientM :: ClientM a -> ClientEnv -> IO (Either ClientError a)
 runClientM cm env = runExceptT $ flip runReaderT env $ unClientM cm
 
 performRequest :: Maybe [Status] -> Request -> ClientM Response
 performRequest acceptStatus req = do
-  ClientEnv m burl cookieJar' createClientRequest <- ask
+  ClientEnv m burl cookieJar' createClientRequest _ <- ask
```
abailly, line 172: I would like to suggest use of `namedFieldPuns` here too, but that's perhaps too big a change.

servant-client/test/Servant/MiddlewareSpec.hs
```diff
@@ -0,0 +1,124 @@
...
+import Servant.Client.Core (RequestF (..))
+import Servant.Client.Internal.HttpClient (ClientMiddleware)
+import Servant.ClientTestUtils
+import Test.Hspec
+import Prelude ()
+import Control.Exception (Exception, throwIO, try)
+import Data.IORef (newIORef, modifyIORef, readIORef)
+
+spec :: Spec
+spec = describe "Servant.MiddlewareSpec" $ do
```
abailly, line 37: suggestion: Remove `describe ....` call as it's already added by default by `spec` which namespaces the tests according to their module names

servant-client/test/Servant/MiddlewareSpec.hs
```diff
@@ -0,0 +1,124 @@
...
+runClientWithMiddleware x mid baseUrl' =
+  runClientM x ((mkClientEnv manager' baseUrl') {middleware = mid})
+
+data CustomException = CustomException deriving (Show, Eq)
+instance Exception CustomException
+
+middlewareSpec :: Spec
+middlewareSpec = beforeAll (startWaiApp server) $ afterAll endWaiApp $ do
+  describe "middleware" $ do
+    it "Raw request and response can be accessed in middleware" $ \(_, baseUrl) -> do
```
abailly, line 50: suggestion: remove  `describe` calls, there's only one per `it` statement and either it does not add any information or it could be incorporated in the `it "..."` description

abailly (approved):
I think this change completely makes sense and I have often wanted to have that functionality available and, as the proposer, resorted to custom `ClientM` or `RunClient` implementations for purposes that could have been well served by a `ClientMiddleware` like authentication, logging, etc.

I have made general suggestions which seem to make sense to me but might be inappropriate because of the coding conventions in force in the codebase, so take those with a grain of salt.

servant-client/src/Servant/Client/Internal/HttpClient.hs
thought2: I'd rather not change this one, I think it's not related enough.

thought2:
@abailly
I applied all the changes that you suggested, except the one where I left a comment.
Thanks for the review!

tchoutri:
Looks like sqlite-simple-0.4.19.0 can't be compiled with GHC 8.6.5. Since it's an obsolete version it can be safely dropped from CI.

thought2:
Great to see this merged, thanks for the support! @abailly @tchoutri

PR #1725, opened 2024-03-04
Use tested-with for CI

tchoutri:
In order to minimise drift, CI testing matrix can be derived from the `tested-with` stanza of the cabal file, which in turn is also displayed by [Flora.pm](https://flora.pm/packages/@hackage/servant)

### DIFF
--- a/servant-client-core/servant-client-core.cabal
+++ b/servant-client-core/servant-client-core.cabal
@@ -16,7 +16,7 @@ author:              Servant Contributors
 maintainer:          haskell-servant-maintainers@googlegroups.com
 copyright:           2014-2016 Zalora South East Asia Pte Ltd, 2016-2019 Servant Contributors
 build-type:          Simple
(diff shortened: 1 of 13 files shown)


servant-client-core/servant-client-core.cabal
```diff
@@ -16,7 +16,7 @@ author:              Servant Contributors
 maintainer:          haskell-servant-maintainers@googlegroups.com
 copyright:           2014-2016 Zalora South East Asia Pte Ltd, 2016-2019 Servant Contributors
 build-type:          Simple
-tested-with: GHC==8.6.5, GHC==8.8.4, GHC ==8.10.7, GHC ==9.0.2, GHC ==9.2.7, GHC ==9.4.4
+tested-with: GHC ==8.10.7, GHC ==9.0.2, GHC ==9.2.8, GHC ==9.4.8, GHC ==9.6.4, GHC ==9.8.1
```
ysangkok, line 19: would be better to test on GHC 9.8.2 instead of 9.8.1

Kleidukos: ah yes indeed, thank you

ysangkok (approved):

Issue #1732, opened 2024-04-14
Export a function of type `IO (Either ServerError a) -> Handler a` from `Servant.Server`


danidiaz:
Currently we have `runHandler :: Handler a -> IO (Either ServerError a)`, and it would be good to have a function going in the other direction.

Of course, it's easy to lift an `IO a` to `Handler a` using `liftIO` but, unless I'm mistaken, going from `IO (Either ServerError a)` to a `Handler` requires a dependeny on `transformers` and importing the `ExceptT` constructor.

Instead of a special-purpose function, re-exporting the `ExceptT` constructor could be another solution.

ysangkok:
Sounds fairly innocent, I'd review a PR for this. But `transformers` is such a popular package, I am not sure I fully understand why one would avoid it. Is it just because it's cumbersome for something you think is something Servant should have out of the box?

danidiaz:
For small enough cases, I do like to use `Servant` without `tranformers` and `mtl`. I just stay in the `IO` monad and lift the result to `Handler` at the last minute. I guess you could call it "simple Servant" 😆 

I've just realised that I already can write a function `IO (Either ServerError a) -> Handler a` without those dependencies, because [`throwError`](https://hackage.haskell.org/package/servant-server-0.20/docs/Servant.html#v:throwError) is re-exported by the `Servant` module. Still, I believe having an explicit inverse to [`runHandler`](https://hackage.haskell.org/package/servant-server-0.20/docs/Servant-Server.html#v:runHandler) could make sense.

tchoutri:
@danidiaz Are you happy with the PR? If yes, I'll let you close this ticket. :)

PR #1733, opened 2024-04-20
Add `MkHandler` pattern synonym

danidiaz:
This bidirectional pattern synonym makes it easier to build `Handler a` values from `IO (Either ServerError a)` values, and is a complement to the already existing `runHandler` function.

image

As an alternative to this, I also considered re-exporting the `ExceptT` constructor, but I think it's better to have an explicit pattern synonym. `ExceptT` is not exactly an internal implementation detail, but it's sufficiently close to being one.

As for naming the pattern synonym, I considered `HandlerT` but it would be a bit misleading because `Handler` is not really a monad transformer. The prefix is `Mk` instead of `Make` because other parts of the Servant ecosystem (like safe links) have already settled on `Mk`.

Solves #1732.

### DIFF
--- a/changelog.d/1733
+++ b/changelog.d/1733
@@ -1,0 +1,7 @@
+synopsis: Add `MkHandler` pattern synonym
+prs: #1733
+issues: #1732
+description: {
+Add a bidirectional pattern synonym to construct `Handler a` values from `IO
+(Either ServerError a)` ones, and match in the other direction.
+}
--- a/servant-server/src/Servant/Server.hs
+++ b/servant-server/src/Servant/Server.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 {-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE PatternSynonyms     #-}
 
 -- | This module lets you implement 'Server's for defined APIs. You'll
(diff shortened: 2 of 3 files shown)


ysangkok:
I think this is a breaking change per PVP, not sure if we are still making changelog.d entries or not. Another maintainer might want to weigh in.

tchoutri:
Yes please produce a changelog.d entry! :)

tchoutri:
Regarding the PVP breaking change, @ysangkok would I love to hear your rationale, this isn't immediately clear to me why it would be such a change.

ysangkok:
Oh you're right @tchoutri , it's not a breaking change according to the PVP flow chart to only add constructors/functions/non-orphan instances. And I suppose a pattern is like a constructor.

tchoutri:
So, in my book, adding a (public) constructor can lead to breakages when consumers do total case matches on the value. Patterns are less prone to provoke breakages when added.

tchoutri (changes requested):
Please write a changelog.d entry. :)

tchoutri:
> I guess it _could_ be a breaking change, if some existing set of pattern synonyms were declared `COMPLETE` and a new pattern synonym were added to that set. That would be analogous to adding a "real" constructor to a datatype. But this is not the case here.

yep, that's not our case here :)

tchoutri (approved):

PR #1744, opened 2024-04-25
Remove base-compat dependency from servant

tchoutri:
First batch of removing extraneous dependencies.

### DIFF
--- a/servant/src/Servant/Links.hs
+++ b/servant/src/Servant/Links.hs
@@ -1,19 +1,21 @@
 {-# LANGUAGE AllowAmbiguousTypes    #-}
-{-# LANGUAGE CPP    #-}
-{-# LANGUAGE ConstraintKinds        #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE ConstraintKinds #-}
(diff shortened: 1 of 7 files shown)


servant/src/Servant/Links.hs
```diff
@@ -661,19 +661,19 @@ simpleToLink _ toA _ = toLink toA (Proxy :: Proxy sub)
...
     type MkLink (arr :> sub) _ = TypeError (PartialApplication (HasLink :: Type -> Constraint) arr)
     toLink = error "unreachable"
 
 -- Erroring instances for 'HasLink' for unknown API combinators
 instance {-# OVERLAPPABLE #-} TypeError (NoInstanceForSub
-
+#if __GLASGOW_HASKELL__ >= 904
                                          @(Type -> Constraint)
-
+#endif
```
MangoIV, line 676: why 904?

tchoutri: You're watching the second commit, which reinstates these lines after they were mistakenly removed in the first commit.

ysangkok (approved):

PR #1748, opened 2024-04-29
Added a note about cookie settings for browser clients

LaurentRDC:
Hi,

I was following this [cookbook recipe](https://docs.servant.dev/en/stable/cookbook/hoist-server-with-context/HoistServerWithContext.html), but it turns out that it doesn't quite work with web browser clients due to XSRF protection. 

This PR directs readers to the appropriate explanation.


tchoutri (approved):

tchoutri:
Thanks!

PR #1755, opened 2024-05-15
Allow base-compat-0.14

ysangkok:
Tested using

    cabal build all -c 'base-compat>=0.14'  --allow-newer=servant-openapi3:base-compat,openapi3:base-compat-batteries,swagger2:base-compat-batteries,servant-js:base-compat


maksbotan:
Will make new releases for servant-client* and servant-quickcheck. Rest done with revisions.

maksbotan:
servant-client{,-core} stuff released with 0.20.1 versions.

servant-quickcheck pending in #1757.

maksbotan:
I've verified that servant-client* 0.20 from Hackage build with base-compat 0.14 and made revisions on Hackage.

PR #1766, opened 2024-07-01
Integrate MultiVerb into the `servant` packages

theophile-scrive:
This commit is Part 1 of the integration, where only the packages of this monorepo are touched. 

`MultiVerb` is integrated alongside of the other Verb machinery, so that people may take their time to transition to it. 

---

IsWaiBody will not be able to live in its current form, as it mixes concepts from the Server and Client packages.

This impacts SomeResponse, which itself impacts the HasClient and HasServer instances.
ResponseF can be replicated internally, even without the httpVersion field.
SomeResponse can be kept on the server, but the client can move to use ResponseF.

---

The `IsResponse` and `IsResponseList` classes are split between client-side `Reponse{,List}Unrender` and server-side `Response{,List}Render`.

---

This is a collaborative work between Scrive AB and Wire Swiss GmbH.


servant/src/Servant/API/Alternative.hs
```diff
@@ -35,7 +35,6 @@ instance (Semigroup a, Semigroup b) => Semigroup (a :<|> b) where
 
 instance (Monoid a, Monoid b) => Monoid (a :<|> b) where
     mempty = mempty :<|> mempty
```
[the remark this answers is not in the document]
theophile-scrive: You're right, it could be derived too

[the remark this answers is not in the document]
theophile-scrive: Actually a deriving statement doesn't work, it raises an error abound unimplemented methods. I'll leave it as such, then.

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -145,6 +145,24 @@ instance ToDeepQuery Filter where
...
+
+type GetUserVerb =
+  Capture "personId" Int
+  :> MultiVerb
+    'GET
+    '[JSON]
+    '[ ErrorResponse UserNotFound,
+       Respond 200 "User found" Person
+     ]
+    (Maybe Person)
```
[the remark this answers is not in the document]
theophile-scrive: I am looking at https://github.com/wireapp/wire-server/blob/0e57e6be7eccaf23f4b69569855d8893e33a5c18/libs/wire-api/src/Wire/API/User.hs#L783-L822

If I understand well: 

* I have to define the errors for the endpoint ([`data RegisterError`](https://github.com/wireapp/wire-server/blob/0e57e6be7eccaf23f4b69569855d8893e33a5c18/libs/wire-api/src/Wire/API/User.hs#L783))
* I have to define another error (?) as defined in [Error/Brig.hs](https://github.com/wireapp/wire-server/blob/0e57e6be7eccaf23f4b69569855d8893e33a5c18/libs/wire-api/src/Wire/API/Error/Brig.hs#L47)
* Then a type-level list of `ErrorResponse`s parametrised with some of the errors from Brig (`ErrorResponse 'AllowlistError`).
* And then derive AsUnion with `  deriving (AsUnion RegisterErrorResponses) via GenericAsUnion RegisterErrorResponses RegisterError` which ties everything together. 

Which part of it is redundant outside of wire-server?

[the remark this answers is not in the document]
fisx: I think you can ignore the error stuff, it's kinda neat, but not required and not really finished.  From my understanding the only thing you need is the `AsUnion` instance that explains how you can compute a `Union '[a]` from a `ResponseType a` (following the naming in the definition of `MultiVerb`.

`ResponseType a` can map to a union type, then the `AsUnion` instance is trivial or not necessary, not sure how that looks.

If it maps to an ordinary type, you can map different values of that type to different response types of your union, thus allowing servant to do it's magic.

Not sure that answers your question, or whether I know what I'm talking about...  :)

Do you have any particular commit and corresponding error message that you are struggling with?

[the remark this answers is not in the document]
theophile-scrive: Alright, so I've got this: https://github.com/haskell-servant/servant/pull/1766/files#diff-a4411c940d784b1fef13144b967568c54a5e3986b2c3f5449c5c4297af69a94fR148-R174

However now I have this error:

```
test/Servant/ClientTestUtils.hs:164:13: error: [GHC-83865]
    • Couldn't match type ‘()’ with ‘Person’
      Expected: Maybe Person -> Union (ResponseTypes res)
        Actual: Maybe Person -> Union ('[()] .++ '[()])
    • In the expression: maybeToUnion (toUnion @GetUserMultiVerbErrors)
      In an equation for ‘toUnion’:
          toUnion = maybeToUnion (toUnion @GetUserMultiVerbErrors)
      In the instance declaration for ‘AsUnion res (Maybe Person)’
    |
164 |   toUnion = maybeToUnion (toUnion @GetUserMultiVerbErrors)
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I am not sure where the need for `()` is coming from, since in the Wire codebase it does not appear.

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -145,6 +145,34 @@ instance ToDeepQuery Filter where
...
     ]
 
+-----------------------------
+-- MultiVerb test endpoint --
+-----------------------------
+
+-- Possible errors
+
+type GetUserMultiVerbErrors =
+  '[RespondEmpty 404 "User not found"]
```
fisx, line 155: since this is built into servant, maybe an example with another error, or some differently typed outcome that is not an error, would be better?

like, an end-point that receives a query param with a boolean that decides whether it should return an `Int` or a `Bool`?

fisx: ...  or an integer query param: if it is even, respond with an integer; if it's odd, respond with a boolean; if it's negative, respond with a 400 error?

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -145,6 +147,46 @@ instance ToDeepQuery Filter where
...
+  = NegativeNumber
+  | Even Bool
+  | Odd Int
+  deriving stock (Generic)
+  deriving (AsUnion MultipleChoicesIntResponses)
+    via GenericAsUnion MultipleChoicesIntResponses MultipleChoicesIntResult
+
+instance GSOP.Generic MultipleChoicesIntResult
+
+-- instance MultipleChoicesIntResponses ~ res => AsUnion res MultipleChoicesIntResult where
```
fisx, line 171:
```suggestion
-- instance AsUnion MultipleChoicesIntResponses MultipleChoicesIntResult where
```

as long as you don't mention `res` again in the instance body, the other thing is really just noise.

servant-client/test/Servant/ClientTestUtils.hs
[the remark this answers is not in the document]
fisx: hum, that smells like somewhere the wrong instance of `AsUnion` or `AsConstructor` is picked? have you tried `-ddump-{ds,tc,simpl,splices,...}`?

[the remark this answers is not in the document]
theophile-scrive: Problem has disappeared with your suggestion below. :) I even added an example of possible error in the cookbook and the documentation of `AsUnion`. Please check them out!

theophile-scrive:
@fisx @ysangkok I'd appreciate a review or questions if some things are not clear!

servant/src/Servant/API/MultiVerb.hs
```diff
@@ -0,0 +1,486 @@
+{-# LANGUAGE ApplicativeDo #-}
+{-# LANGUAGE EmptyCase #-}
+
+-- | MultiVerb is a part of the type-level eDSL that allows you to express complex routes
+-- while retaining a high level of precision with good ergonomics.
+
+module Servant.API.MultiVerb
```
fisx, line 7: oh, another thought: should we deprecate `UVerb`?  if not: should we add a comment to the haddocks that `MultiVerb` is strictly more powerful and should be considered as an alternative?

fisx (commented):
sorry for dropping the ball. here are some more comments. i'll continue with the review soon, i promise!  :)

doc/cookbook/generic/Generic.lhs
```diff
@@ -107,7 +107,7 @@ main = do
         ("run-custom-monad":_) -> do
             putStrLn "Starting cookbook-generic with a custom monad at http://localhost:8000"
             run 8000 (appMyMonad AppCustomState)
-        _ -> putStrLn "To run, pass 'run' argument: cabal new-run cookbook-generic run"
+        _ -> putStrLn "To run, pass 'run' argument: cabal run cookbook-generic run"
```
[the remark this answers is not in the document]
theophile-scrive: I'll make separate commits, you're right.

doc/cookbook/multiverb/MultiVerb.lhs
```diff
@@ -0,0 +1,178 @@
...
+  Choices'
+    { choices = choicesHandler
+    }
+
+routesServer :: Routes AsServer
+routesServer =
+  Routes
+    { choicesRoutes = choicesServer
+    , version = versionHandler
+    }
```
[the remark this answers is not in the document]
theophile-scrive: We are moving away from promoting route combinators actually. The enhanced quality of life is worth it, especially for newcomers.

servant/src/Servant/API/MultiVerb.hs
theophile-scrive: We'll deprecate UVerb one day, and yes it should be labelled as an alternative.

doc/cookbook/multiverb/MultiVerb.lhs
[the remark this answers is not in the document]
fisx: Interesting. "We" as in "the servant maintainers"? I'd be a little upset to see the combinators go, but as longs as it's both allowed I don't want to argue :-)

(Wire uses a type `Named` to solve this problem.It also makes it easier to switch between swagger-ui and source code, because we print the names in swagger. And in my opinion the code looks nicer.)

servant/src/Servant/API/MultiVerb.hs
```diff
@@ -0,0 +1,486 @@
...
+-- | This type adds response headers to a 'MultiVerb' response.
+data WithHeaders (headers :: [Type]) (returnType :: Type) (response :: Type)
+
+-- | This is used to convert a response containing headers to a custom type
+-- including the information in the headers.
+class AsHeaders xs a b where
+  fromHeaders :: (NP I xs, a) -> b
+  toHeaders :: b -> (NP I xs, a)
+
+-- single-header empty response
```
fisx, line 137:
```suggestion
-- | single-header empty response
```

servant/src/Servant/API/MultiVerb.hs
```diff
@@ -0,0 +1,486 @@
...
+instance AsHeaders '[a] () a where
+  toHeaders a = (I a :* Nil, ())
+  fromHeaders = unI . hd . fst
+
+-- single-header non-empty response, return value is a tuple of the response and the header
```
fisx, line 142:
```suggestion
-- | single-header non-empty response, return value is a tuple of the response and the header
```

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -187,6 +229,7 @@ type Api =
                                       WithStatus 301 Text]
   :<|> "uverb-get-created" :> UVerb 'GET '[PlainText] '[WithStatus 201 Person]
   :<|> NamedRoutes RecordRoutes
+  :<|> MultipleChoicesInt
```
fisx, line 232:
```suggestion
  :<|> "multiple-choices-int" :> MultipleChoicesInt
```

for consistency with *most* other routes?

fisx (commented):
Sorry for the lag, but I think we're getting there!

Could you go through the unresolved comments one more time and confirm that you're ok with resolving all of them?

If so I'll approve this PR.

After that: @arianvp wanna give it a read, and maybe merge?  @pcapriotti as the author of this code, do you have any feedback?

doc/cookbook/multiverb/MultiVerb.lhs
```diff
@@ -0,0 +1,178 @@
...
+These deriving statements above tie together the responses and the return values, and the order in which they are defined matters. For instance, if `Even` and `Odd` had switched places in the definition of `Result`, this would provoke an error:
+
+```
+• No instance for ‘AsConstructor
+    ((:) @Type Int ('[] @Type)) (Respond 200 "Even number" Bool)’
+        arising from the 'deriving' clause of a data type declaration
+```
+
+> If you would prefer to write an intance of 'AsUnion' by yourself, read more in
+the “Implementing AsUnion manually” section.
```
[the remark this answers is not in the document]
theophile-scrive: I didn't know about `sop-products`. It could be in another PR yes. :)

doc/cookbook/multiverb/MultiVerb.lhs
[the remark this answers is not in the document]
theophile-scrive: > I'd be a little upset to see the combinators go, but as longs as it's both allowed I don't want to argue :-)

The combinators are not going anywhere! :)

servant/src/Servant/API/MultiVerb.hs
```diff
@@ -0,0 +1,486 @@
...
+import Network.HTTP.Types as HTTP
+import Web.HttpApiData (FromHttpApiData, ToHttpApiData, parseHeader, toHeader)
+import qualified Data.CaseInsensitive as CI
+import qualified Data.Sequence as Seq
+import qualified Data.Text as Text
+import qualified Data.Text.Encoding as Text
+
+import Servant.API.TypeLevel.List
+import Servant.API.Stream (SourceIO)
+import Servant.API.UVerb.Union (Union)
```
arianvp, line 57: Do we want to have this cross-dependency on `UVerb`?

It feels like the `Servant.API.Uverb.Union`  module ought to be moved somewhere else . Maybe to `Data.SOP.Utils` or the likes.  As it only contains non-servant specific combinators for working with SOPs

arianvp: Given we're not using any of the functions from that module. We could also just replace `Union` with `NS I` in this module...

arianvp: (This is a bit of a bikeshed; so feel free to ignore)

theophile-scrive: It's a good refactoring to do at a later time, I agree. Let me open a ticket.

theophile-scrive: https://github.com/haskell-servant/servant/issues/1786

doc/cookbook/multiverb/MultiVerb.lhs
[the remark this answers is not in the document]
gdeest: > Wire uses a type Named to solve this problem

Which problem, may I ask ? :-) If the problem you are referring to is "giving a name to routes", then surely there are multiple ways to achieve this. I'd just like to point out that records with `NamedRoutes` have other advantages compared to combinators:

- They make the orders of routes mostly irrelevant when defining a server, etc.
- They make error messages much, much more palatable to newcomers.

Surely, as @theophile-scrive said, there is no need / reason to remove the combinators.

theophile-scrive:
@pcapriotti @fisx I'd love your final feedback on this. :)

servant-client-core/src/Servant/Client/Core/ResponseUnrender.hs
```diff
@@ -0,0 +1,131 @@
+{-# LANGUAGE ApplicativeDo #-}
+module Servant.Client.Core.ResponseUnrender where
```
jkarni, line 2: The name is a bit general. Maybe it could be prefixed by MultiVerb, or the class could be named something like `MultiResponseUnrender`?

servant/src/Servant/API/MultiVerb.hs
```diff
@@ -0,0 +1,486 @@
...
+-- Includes status code, description, and return type. The content type of the
+-- response is determined dynamically using the accept header and the list of
+-- supported content types specified in the containing 'MultiVerb' type.
+data Respond (s :: Nat) (description :: Symbol) (a :: Type)
+
+-- | A type to describe a 'MultiVerb' response with a fixed content type.
+--
+-- Similar to 'Respond', but hardcodes the content type to be used for
+-- generating the response.
+data RespondAs ct (s :: Nat) (description :: Symbol) (a :: Type)
```
jkarni, line 71: It feels a bit sad that there are two locations where a content-type may be specified, and they can be conflicting. It's not obvious to me what:

```
type Responses = '[ RespondAs 'HTML 200 "blah" Int ]
type API = MultiVerb  'GET '[JSON]   Responses     Result
```
Means, and why I should be allowed to say this.

 Is there any reason to have this *besides* `RespondEmpty`? If not, maybe `RespondEmpty` can be it's own datatype?

servant/src/Servant/Types/ResponseList.hs
```diff
@@ -0,0 +1,17 @@
+{-# LANGUAGE DeriveTraversable #-}
+
+-- | This module offers other servant libraries a minimalistic HTTP response type.
+--
+-- It is purely an internal API and SHOULD NOT be used by end-users of Servant.
```
jkarni, line 5: Should the module name be `Servant.Types.Internal.ResponseList` instead? I believe that's what we do elsewhere.

servant/src/Servant/API/TypeLevel/List.hs
```diff
@@ -0,0 +1,9 @@
+module Servant.API.TypeLevel.List 
+    (type (.++)
```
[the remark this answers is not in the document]
jkarni: It is *also* defined [here](https://github.com/haskell-servant/servant/blob/1e6e566eb52a15681d1c8c91cdd5d29360d1680a/servant/src/Servant/API/TypeLevel.hs#L216) and [here](https://github.com/haskell-servant/servant/blob/1e6e566eb52a15681d1c8c91cdd5d29360d1680a/servant-swagger/src/Servant/Swagger/Internal/TypeLevel/API.hs#L34).

jkarni (commented):
Nice feature, and very nice to read PR!

I made some comments, but none of them should be considered blocking!

fisx (approved):
sorry for taking so incredibly long!  :flushed:

servant-client-core/src/Servant/Client/Core/Response.hs
```diff
@@ -51,3 +52,7 @@ instance NFData a => NFData (ResponseF a) where
 
 type Response = ResponseF LBS.ByteString
 type StreamingResponse = ResponseF (SourceIO BS.ByteString)
+
+responseToInternalResponse :: ResponseF a -> InternalResponse a
```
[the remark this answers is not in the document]
theophile-scrive: It's exposed for external consumers at the moment.

servant-client-core/src/Servant/Client/Core/HasClient.hs
```diff
@@ -364,15 +368,13 @@ instance {-# OVERLAPPING #-}
 
         method = reflectMethod $ Proxy @method
         acceptStatus = statuses (Proxy @as)
-    response <- runRequestAcceptStatus (Just acceptStatus) request {requestMethod = method, requestAccept = accept}
+    response@Response{responseBody=body, responseStatusCode=status, responseHeaders=headers}
```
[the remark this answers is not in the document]
theophile-scrive: Yep

servant-client-core/src/Servant/Client/Core/ResponseUnrender.hs
theophile-scrive: Done

servant/src/Servant/API/MultiVerb.hs
theophile-scrive: Good call. @pcapriotti @fisx I'd like your input on this.

servant/src/Servant/Types/ResponseList.hs
theophile-scrive: Done

servant-client/test/Servant/ClientTestUtils.hs
```diff
@@ -145,6 +147,46 @@ instance ToDeepQuery Filter where
...
+  = NegativeNumber
+  | Even Bool
+  | Odd Int
+  deriving stock (Generic)
+  deriving (AsUnion MultipleChoicesIntResponses)
+    via GenericAsUnion MultipleChoicesIntResponses MultipleChoicesIntResult
+
+instance GSOP.Generic MultipleChoicesIntResult
+
+-- instance => AsUnion MultipleChoicesIntResponses MultipleChoicesIntResult where
```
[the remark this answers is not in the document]
theophile-scrive: Yes

servant/src/Servant/API/MultiVerb.hs
pcapriotti: The list of content types passed to `MultiVerb` are used for content negotiation. The one in `Response` is used to render the response. In @jkarni's example, this means that your server will respond to requests that have an `Accept: application/json` header, but then return its response as HTML.

One might think this shouldn't be allowed, but in fact many APIs do this sort of thing, and it's important to have this capability when refactoring existing implementations to use servant.

fisx: thanks @pcapriotti, that makes sense to me!  @theophile-scrive maybe this could go as a comment into the code for future readers?

theophile-scrive: Excellent, thanks a lot.

theophile-scrive:
@fisx would you do me the honour?

servant/src/Servant/Types/Internal/ResponseList.hs
```diff
@@ -0,0 +1,17 @@
+{-# LANGUAGE DeriveTraversable #-}
+
+-- | This module offers other servant libraries a minimalistic HTTP response type.
+--
+-- It is purely an internal API and SHOULD NOT be used by end-users of Servant.
+module Servant.Types.Internal.ResponseList where
```
fisx, line 6: why is it called `ResponseList`?

fisx:
> @fisx would you do me the honour?

merged.  also the honor is all mine :-)  thanks for your infinite patience!

servant/src/Servant/Types/Internal/ResponseList.hs
theophile-scrive: Ah, it used to probably host stuff in relation with IsResponseList but does not anymore. I'll rename it to `Response`.

fisx:
@phadej @jkarni @...

any objections to me cutting a new release tomorrow?  (come to think of it, should i be on the google mailing list so i can ask these things there?  i remember vaguely having been opposed to gmail when that was started, but my opposition appearently didn't make google go away...  :)

tchoutri:
I'd like us to take some time to address some of the tickets that have appeared since the last release before.

tchoutri:
Moreover I'd also be comfortable with advertising the MultiVerb feature first so that people may try it before we cut a release.

PR #1775, opened 2024-08-14
Add instance HasServer (EmptyAPI :> api) context

odr:


tchoutri:
Thank you very much! Please add a [changelog entry](https://github.com/haskell-servant/servant/blob/master/CONTRIBUTING.md#changelog-entries) and it should be good!

tchoutri:
@odr Please do take a look at the link I provided in my message, we use changelog snippets instead of editing the final changelog directly)

odr:
> Thank you very much! Please add a [changelog entry](https://github.com/haskell-servant/servant/blob/master/CONTRIBUTING.md#changelog-entries) and it should be good!



> @odr Please do take a look at the link I provided in my message, we use changelog snippets instead of editing the final changelog directly)

Done!

tchoutri (approved):

PR #1778, opened 2024-08-29
Remove GHC <= 8.10.7 from the support window

tchoutri:
GHC <= 8.10.7 is removed from the support window. The lowest version supported in CI becomes GHC 9.2.1, released on the 29/10/2021.

This grants us the ability to use `-XGHC2021` as default language, as well as `-XOverloadedRecordDot`.

PR #1785, opened 2024-09-23
Exported Servant.Links.addQueryParam

### LINKED ISSUE #1232: Servant.Links.addQueryParam not exported?
I was trying to write a `ToLink` instance for a custom combinator called `AllQueryParams` (that captures all query parameters). Unfortunately, it seems almost impossible without resorting to type-level magic, because neither are the constructors of `Link` exported, nor is `addQueryParams` exported.

Is there an obvious way of doing this, which I might be missing?

Here's my (non-working) code snippet:

```
data AllQueryParams

instance (HasServer api context) => HasServer (AllQueryParams :> api) context where
  type ServerT (AllQueryParams :> api) m = ([(BS.ByteString, Maybe BS.ByteString)] -> ServerT api m)

  hoistServerWithContext _ pc nt s = hoistServerWithContext (Proxy :: Proxy api) pc nt . s

  route _ context subServer = route (Proxy :: Proxy api) context delayed
    where
      delayed = passToServer subServer queryString

instance (HasLink sub) => HasLink (AllQueryParams :> sub) where
  type MkLink (AllQueryParams :> sub) a = ([(BS.ByteString, Maybe BS.ByteString)] -> MkLink sub a)

  toLink toL _ lnk qparams = toLink toL (Proxy :: Proxy sub) $
                             DL.foldl' (\l (k, v) -> addQueryParam (SingleParam (toS k) (toS $ fromMaybe "" v)) l) lnk qparams
```

saurabhnanda:
Closes #1232

### DIFF
--- a/changelog.d/issue-1232
+++ b/changelog.d/issue-1232
@@ -2,8 +2,7 @@ synopsis: Exported addQueryParam
 packages: servant
 prs: #1785
 issues: #1232
-significance: minor
 description: {
   `addQueryParams` is required to define custom `HasLink` instances which actually manipulate the
-  generated query params. This function was not exported earlier and now it is.
+  generated query params. This function was not exported earlier and now it is. 
 }
--- a/changelog.d/issue-1232
+++ b/changelog.d/issue-1232
@@ -1,0 +1,9 @@
+synopsis: Exported addQueryParam
+packages: servant
+prs: #1785
+issues: #1232
+significance: minor
+description: {
+  `addQueryParams` is required to define custom `HasLink` instances which actually manipulate the
+  generated query params. This function was not exported earlier and now it is.
+}
--- a/servant/src/Servant/Links.hs
+++ b/servant/src/Servant/Links.hs
@@ -116,9 +116,6 @@ module Servant.Links (
   , linkQueryParams
   , linkFragment
   , addQueryParam
-  , addSegment
-  , Escaped
-  , escaped
 ) where
 
 import           Data.Kind
--- a/servant/src/Servant/Links.hs
+++ b/servant/src/Servant/Links.hs
@@ -115,6 +115,8 @@ module Servant.Links (
   , linkSegments
   , linkQueryParams
   , linkFragment
+  , addQueryParam
+  , addSegment
 ) where
 
 import           Data.Kind
@@ -187,6 +189,10 @@ import           Web.HttpApiData
 -- | A safe link datatype.
 -- The only way of constructing a 'Link' is using 'safeLink', which means any
 -- 'Link' is guaranteed to be part of the mentioned API.
+--
+-- NOTE: If you are writing a custom 'HasLink' instance, and need to manipulate
+-- the 'Link' (adding query params or fragments, perhaps), please use the the
+-- 'addQueryParam' and 'addSegment' functions.
 data Link = Link
   { _segments    :: [Escaped]
   , _queryParams :: [Param]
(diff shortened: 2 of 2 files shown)


servant/src/Servant/Links.hs
```diff
@@ -187,6 +189,10 @@ import           Web.HttpApiData
 -- | A safe link datatype.
 -- The only way of constructing a 'Link' is using 'safeLink', which means any
 -- 'Link' is guaranteed to be part of the mentioned API.
+--
+-- NOTE: If you are writing a custom 'HasLink' instance, and need to manipulate
+-- the 'Link' (adding query params or fragments, perhaps), please use the the
+-- 'addQueryParam' and 'addSegment' functions.
```
tchoutri, line 195: Good Note, thanks!

tchoutri (approved):

tchoutri:
@saurabhnanda Thanks a lot for this!

Could you please add a new changelog entry in the `changelog.d` file? Just like this one https://github.com/haskell-servant/servant/blob/master/changelog.d/issue-1780

servant/servant.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               servant
-version:            0.20.2
+version:            0.20.3
```
ysangkok, line 3: Probably better to wait with bumping until we do the release. Because it's reasonably likely that we'll merge e.g. two major changes that would both require major bumps.

servant/src/Servant/Links.hs
saurabhnanda: @tchoutri should we also export `addSegment`? Which would force one to also export `Fragment'`? Is there any reason why one would be extending / modifying this behaviour?

saurabhnanda:
> @saurabhnanda Thanks a lot for this!
> 
> Could you please add a new changelog entry in the `changelog.d` file? Just like this one https://github.com/haskell-servant/servant/blob/master/changelog.d/issue-1780

@tchoutri I had created a branch from `v0.20.2` and not `master`. I believe `changelog.d` has been introduced in `master`, along with some GHC-related deprecations? Are those intended? Because I am unable to build `servant` on `Stack LTS-18.28` if I swtich to `master` branch.

```
       
       In the dependencies for servant-0.20.3:
         * base must match >=4.16.4.0 && <4.21, but base-4.14.3.0 is in the Stack configuration (latest matching version is 4.20.0.1).
         * bytestring must match >=0.11 && <0.13, but bytestring-0.10.12.0 is in the Stack configuration (latest matching version is 0.12.1.0).
       The above is/are needed since servant is a build target.
```

tchoutri:
@saurabhnanda changelog-d has been around for quite some time, if I'm not mistaken. 

Regarding the GHC deprecations: yes they are intended, see https://github.com/haskell-servant/servant/pull/1778.

You will need to update the Stack setup to use the LTS of 9.6.6.

saurabhnanda:
@tchoutri added the changelog entry - hope it's fine.

servant/servant.cabal
saurabhnanda: @ysangkok removed the version bump.

changelog.d/issue-1232
```diff
@@ -0,0 +1,9 @@
+synopsis: Exported addQueryParam
+packages: servant
+prs: #1785
+issues: #1232
+significance: minor
```
tchoutri, line 5: @fgaz remind me, is `minor` a valid value for the `significance` field?

servant/src/Servant/Links.hs
tchoutri: Not in this PR, I would like to take the time to think about a different module to export these "unsafe" modifiers for Links.

changelog.d/issue-1232
fgaz: No, we only have `significance: significant` or omitted

ysangkok: @tchoutri btw the docs are here: https://codeberg.org/fgaz/changelog-d

tchoutri: @saurabhnanda Please remove the `significance` line then. :) After that will be good to merge.

tchoutri: @saurabhnanda ping! I believe this would be the last thing before merging. :)

tchoutri (changes requested):

tchoutri:
@saurabhnanda ping! I'd like to get this merged for 0.20.3.0 :)

tchoutri:
@saurabhnanda no worries :)

servant/src/Servant/Links.hs
saurabhnanda: @tchoutri I have removed the exported of the following, for now, but we are already using a fork of `servant` where we ended-up exporting these identifiers:

- escaped
- Escaped
- addSegment

changelog.d/issue-1232
saurabhnanda: @tchoutri have made this change and pushed. Please check.

tchoutri (approved):
Wonderful, thank you for your patience @saurabhnanda!

PR #1789, opened 2024-10-01
Make defaultMakeClientRequest not depend on IO

unclechu:
The `IO` wrap for the return type of this function was never necessary. Inside the function there’s only `return` call that wraps a _pure_ value into `IO`. There is nothing else that would be related to `IO`. The `IO` wrap was introduced in 9df16f9f1a81b0b9ae605820a4348627b102c75d

In order to not break the API again the wrap was preserved but this change makes it to depend on any `Applicative` instead of `IO`. So it still works in `IO`/`MonadIO` context but allows to use something like `Data.Functor.Identity` to work with _pure_ values.

See for more context:
https://github.com/haskell-servant/servant/pull/1595


tchoutri (approved):

tchoutri:
@unclechu Thanks for the patch

PR #1792, opened 2024-10-18
Fix typo in Content-Type docs

knutanderss:
It seems that the Content-Type header for `Plaintext` should be `text/plain;charset=utf8` and not `text/plain;charset-utf8`, as I'm getting `415 Unsupported Media Type` for the former.


tchoutri (approved):

tchoutri:
Thanks a bunch for the correction!

PR #1796, opened 2024-12-10
Use data-default instead of data-default-class

kokobd:
This is a simple version bump to make `servant-auth-server` compatible with `data-default-class 0.2.0.0`.

I noticed that both `data-default` and `data-default-class` are used in this repo. I wonder if there is plan to move to `data-default` for all servant packages, like many other packages on Hackage? For example [`tls`](https://hackage.haskell.org/package/tls-2.1.5/changelog) and [`crypton-connection`](https://hackage.haskell.org/package/crypton-connection-0.4.3/changelog) have recently moved to `data-default`.

I'm happy to update the PR to use `data-default` instead if that's desired.

### DIFF
--- a/servant-quickcheck/servant-quickcheck.cabal
+++ b/servant-quickcheck/servant-quickcheck.cabal
@@ -41,7 +41,6 @@ library
     , bytestring             >=0.11   && <0.13
     , case-insensitive       >=1.2    && <1.3
     , clock                  >=0.7    && <0.9
-    , data-default-class     >=0.0    && <0.3
     , hspec                  >=2.5.6  && <2.12
     , http-client            >=0.7.0  && <0.8
     , http-media             >=0.6    && <0.9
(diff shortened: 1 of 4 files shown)


tchoutri:
@kokobd Hi, thanks for the PR. Yes please a move to `data-default` would be appreciated!

kokobd:
I modified this PR to use `data-default` instead.

tchoutri:
Sorry for being late, thank you again for this PR!

debug-ito:
Hi. Could you guys upload this version of servant-auth-server to hackage?

servant-quickcheck/servant-quickcheck.cabal
```diff
@@ -41,7 +41,6 @@ library
     , bytestring             >=0.11   && <0.13
     , case-insensitive       >=1.2    && <1.3
     , clock                  >=0.7    && <0.9
-    , data-default-class     >=0.0    && <0.2
```
wolfgangwalther, line 44: Looks like this change to `servant-quickcheck` is not reflected on hackage, yet.

This makes the package fail to build in Nixpkgs, because we're at data-default-class 0.2.0.0.

PR #1797, opened 2024-12-16
Install liblzma-dev

ysangkok:
Trying to see whether this fixes the build, which seems like it breaks because of the new version of https://hackage.haskell.org/package/lzma which is using pkg-config.


ysangkok:
Seems that the problem is that it runs freeze before installing system deps.

PR #1804, opened 2025-02-07
Add more multiverb

tchoutri:
This PR adds more instances to ease the transition from UVerb to MultiVerb, and expands on the documentation of `AsHeaders`.

### DIFF
--- a/servant/src/Servant/API/MultiVerb.hs
+++ b/servant/src/Servant/API/MultiVerb.hs
@@ -420,6 +420,10 @@ instance AsConstructor '[] (RespondEmpty code description) where
   toConstructor _ = Nil
   fromConstructor _ = ()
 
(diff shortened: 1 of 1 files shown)


servant/src/Servant/API/MultiVerb.hs
```diff
@@ -146,6 +149,11 @@ instance AsHeaders '[h] a (a, h) where
   toHeaders (t, cc) = (I cc :* Nil, t)
   fromHeaders (I cc :* Nil, t) = (t, cc)
 
+-- | Two headers and an empty response, return value is a tuple of the response and the header
```
fisx, line 152: i don't understand this comment.  return value is the tuple of the two headers, no?  response doesn't need to be returned because it's ()?

PR #1805, opened 2025-02-14
Add new `Servant.API.Range` type

kutyel:
At Scrive we are using this type quite extensively, so I thought it might be a good idea to contribute it back to Servant. Please, let me know which docs or other files am I missing so that this contribution is useful for the servant community and also let me know if you would like to have this as part of the library or not. Thanks in advance!


servant/src/Servant/API/BoundedNat.hs
```diff
@@ -0,0 +1,66 @@
...
+-- instance (KnownNat min, KnownNat max) => ToSchema (BoundedNat min max) where
+--     declareNamedSchema _ =
+--         pure . NamedSchema Nothing $
+--             mempty
+--                 & #type ?~ OpenApiInteger
+--                 & #minimum ?~ fromIntegral (minBound @(BoundedNat min max)) . unBoundedNat
+--                 & #maximum ?~ fromIntegral (maxBound @(BoundedNat min max)) . unBoundedNat
+
+-- instance (KnownNat min, KnownNat max) => ToParamSchema (BoundedNat min max) where
+--     toParamSchema = toSchema
```
tchoutri, line 66: Why is this commented?

tchoutri (changes requested):
Could you please add a a test case to https://github.com/haskell-servant/servant/blob/master/servant-client/test/Servant/ClientTestUtils.hs that makes use of it?

tchoutri:
Thanks a lot @kutyel! Could you please add a changelog entry in https://github.com/haskell-servant/servant/tree/master/changelog.d ?

servant/src/Servant/API/BoundedNat.hs
tchoutri: You'll have to send a patch to https://github.com/biocad/servant-openapi3

servant/src/Servant/API/BoundedNat.hs
```diff
@@ -0,0 +1,69 @@
...
+-- 5
+-- >>> let invalidVal = unsafeBoundedNat 15 :: BoundedNat 1 10
+-- >>> unBoundedNat invalidVal
+-- 15
+-- >>> checkBoundedRange (const $ fail "Out of range") val
+-- MkBoundedNat {unBoundedNat = 5}
+-- >>> checkBoundedRange (const $ fail "Out of range") invalidVal
+-- *** Exception: user error (Out of range)
+
+newtype BoundedNat (min :: Nat) (max :: Nat) = MkBoundedNat {unBoundedNat :: Word}
```
tchoutri, line 33: Could it be renamed to `Range`? Unfortunately `BoundedNat` is more of a description than a name.

servant/src/Servant/API/BoundedNat.hs
```diff
@@ -0,0 +1,69 @@
...
+import           Data.Aeson.Types (modifyFailure)
+import           Data.Bifunctor   (bimap)
+import           Data.Ix
+import           Data.Proxy       (Proxy (Proxy))
+import qualified Data.Text        as T
+import           GHC.Generics     (Generic)
+import           GHC.TypeLits
+import           Servant.API
+
+-- | A newtype wrapper around 'Word' that ensures the value is within a given range.
```
tchoutri, line 16: Any reason to limit to `Word` when the type-level `Natural` can contain arbitrarily-long integers >=0?

servant/src/Servant/API/BoundedNat.hs
```diff
@@ -0,0 +1,69 @@
...
+--
+-- >>> :set -XDataKinds
+-- >>> import Data.Proxy
+-- >>> let val = unsafeBoundedNat 5 :: BoundedNat 1 10
+-- >>> unBoundedNat val
+-- 5
+-- >>> let invalidVal = unsafeBoundedNat 15 :: BoundedNat 1 10
+-- >>> unBoundedNat invalidVal
+-- 15
+-- >>> checkBoundedRange (const $ fail "Out of range") val
```
tchoutri, line 28: I'm quite intrigued by this workflow. Wouldn't it be easier to make a smart creator (parser) that returns a `Maybe (Range n1 n2)` instead of having a validation function called after the fact?

tchoutri (changes requested):

servant/src/Servant/API/BoundedNat.hs
kutyel: I think I can try and refactor to apply this suggestion, thanks for the feedback! 👍🏻

kutyel:
Can I ping @haskell-servant/maintainers for review? 🙏🏻

servant-client/test/Servant/ClientTestUtils.hs
tchoutri, line 1: @kutyel I don't see `usersServer` in the test server (`server = serve api (…`) on line 283.

tchoutri (changes requested):

tchoutri (approved):

kutyel:
@arianvp 🙏🏻 👀

tchoutri:
@kutyel Alright, I'm merging. Please do depend on the commit in an applications to make sure that it works well in real-life scenarios.

PR #1808, opened 2025-03-09
readme for Hackage package https://hackage.haskell.org/package/servant

peterbecich:
Currently https://hackage.haskell.org/package/servant has a short description that points to the tutorial.
The intent of this PR is to add more to this page, so that https://hackage.haskell.org/package/servant serves as an overview for the entire Servant project.

### DIFF
--- a/servant/README.md
+++ b/servant/README.md
@@ -1,0 +1,31 @@
+# servant (API)
+
+Official *servant* packages in [haskell-servant/servant](https://github.com/haskell-servant/servant) repository:
+
+* [**servant (API)**](https://hackage.haskell.org/package/servant)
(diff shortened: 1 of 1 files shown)


tchoutri:
@peterbecich Hi, thanks for tackling this. I don't believe the README should contain a list of packages to such an extent. It should say first and foremost:

1. What servant is
2. Point to the tutorials

Remember that this README will then be bundled in the distribution uploaded to Hackage, and cannot be replaced when it gets out of date. I'd rather put down stuff that is hard to bitrot, and have the more volatile kind of documentation live in the Wiki, or in the docs site. :)

PR #1820, opened 2025-06-09
Deprecate UVerb

fisx:
We now have `MultiVerb`, which is strictly more powerful.

If you find that `MultiVerb` doesn't allow you to do the things you're doing with `UVerb`, please [open an issue](https://github.com/haskell-servant/servant/issues)


ysangkok:
I think this looks good but there are a couple of conflicts.

PR #1825, opened 2025-06-19
Switch to `fourmolu` formatter and reformat all sources

zlondrej:
Code was first formatted using `stylish-haskell` to reduce amount of multi-line import lists as `fourmolu` would put each item on its own line. Then it got one more pass with `fourmolu` with `respectful: false` option.

Some FOURMOLU_DISABLE directives were added in problematic cases (mostly CPP conditional pragmas or foreign imports).

The `fourmolu.yaml` configuration was provided by @tchoutri.


tchoutri (approved):
Need to fix the conflicts, otherwise it looks good.

servant/src/Servant/API/Description.hs
```diff
@@ -1,52 +1,51 @@
...
 -- Example:
 --
 -- >>> :{
---type MyApi = Description
+-- type MyApi = Description
 --  "This comment is visible in multiple Servant interpretations \
 --  \and can be really long if necessary. \
 --  \Haskell multiline String support is not perfect \
 --  \but it's still very readable."
 -- :> Get '[JSON] Book
```
tchoutri, line 45: I suspect the indentation here should be increased, wrt https://github.com/haskell-servant/servant/actions/runs/15779239616/job/44480594133?pr=1825

PR #1840, opened 2025-08-22
Abstract `ByteString` out of `MIME` rendering

dmjio:
This should allow us to abstract over `ByteString` from the `MimeRender` / `MimeUnrender` classes, without breaking any code and preserving existing semantics, while gaining extensibility to other platforms. 

On the client (e.g. browser) dealing with `JSString` / `JSVal` is often more efficient than marshalling by way of `ByteString` (at least for the JS backend). This change should be an easy win, keeping all existing code unaffected (via the type family defaulting rules), but allowing users on different platforms to use a type other than `ByteString`. 

Usage would look like:

```haskell
-- * when using servant-client on native platforms,
-- default to ByteString (existing code should be unaffected)
instance MimeUnrender JSON String where
  mimeUnrender Proxy = eitherDecode

-- * when using servant-client in the browser,
-- you can specify `JSVal`.
instance MimeUnrender OctetStream Blob where
  type MimeUnrenderType Blob = JSVal
  mimeUnrender Proxy = pure . Blob
  ```

### DIFF
--- a/servant/src/Servant/API/ContentTypes.hs
+++ b/servant/src/Servant/API/ContentTypes.hs
@@ -200,7 +200,7 @@ instance
 --
 -- >>> import Network.HTTP.Media hiding (Accept)
 -- >>> import qualified Data.ByteString.Lazy.Char8 as BSC
--- >>> newtype MyContentType = MyContentType String
+-- >>> data MyContentType
 -- >>> newtype MyData = MyData String
 --
 -- >>> :{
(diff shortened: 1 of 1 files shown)


dmjio:
I need to make us of this associated type family in `servant-client-core`, to remove the `ByteString` specific assumptions, then https://github.com/haskell-miso/servant-miso-client can use `servant-client-core` as well (as opposed to rewriting the `HasClient` wheel here).

tchoutri:
@dmjio A changelog entry would be really appreciated for this PR. :)

PR #1844, opened 2025-10-04
Add `OperationId`

blackheaven:
Add `OperationId`, which can direct code and docs generation.

### DIFF
--- a/servant/src/Servant/API/Description.hs
+++ b/servant/src/Servant/API/Description.hs
@@ -48,7 +48,7 @@ data Summary (sym :: Symbol)
 data Description (sym :: Symbol)
   deriving (Typeable)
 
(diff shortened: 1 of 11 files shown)


servant/src/Servant/API/Description.hs
```diff
@@ -47,6 +48,14 @@ data Summary (sym :: Symbol)
 data Description (sym :: Symbol)
   deriving (Typeable)
 
+-- | Add an (uncheck) operation id for (part of) API, useful when generated bindings and OpenAPI docs.
```
divarvel, line 51:
```suggestion
-- | Add an (unchecked) operation id for (part of) API, useful when generating bindings and OpenAPI docs.
```

blackheaven: Thanks :)

tchoutri (approved):

PR #1849, opened 2025-10-26
Infinite steams cookbook recipe

LaurentRDC:
This is a cookbook recipe to serve infinite HTTP streams. There are a lot of subtleties in providing an HTTP infinite stream because data must be sent to the client on a regular basis.

### DIFF
--- a/doc/cookbook/infinite-streams/InfiniteStreams.lhs
+++ b/doc/cookbook/infinite-streams/InfiniteStreams.lhs
@@ -1,0 +1,277 @@
+# Serving infinite streams
+
+Servant provides facilities to work with streams of data. This is handy for cases where the data may take a while to
+fetch, but we can start returning data early. In this cookbook, we will be concerned with serving _infinite_ HTTP
+streams.
+
+HTTP streams have many advantages compared to other streaming standards like websockets: they are simple
+and are well-supported by a broad range of intermediaries, such as proxy servers, content-delivery networks, and
+corporate firewalls.
+
+An _infinite_ HTTP stream differs from a _finite_ stream in two major ways. First, cleaning up resources (such as
+connections) associated with infinite streams is not deterministic. We do not know when the handler will
+terminate -- if at all! Second, we want to ensure that the connection isn't cut off because no data is flowing.
+To keep the connection alive, we will need to send bytes on a regular basis.
+
+This is a Literate Haskell file, so let's get imports out of the way.
+
+```haskell
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumericUnderscores #-}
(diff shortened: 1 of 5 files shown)


LaurentRDC:
https://github.com/kazu-yamamoto/crypton-certificate/issues/21

tchoutri:
@LaurentRDC The package's metadata seems to have been amended on Hackage. :)

doc/cookbook/infinite-streams/InfiniteStreams.lhs
```diff
@@ -0,0 +1,277 @@
...
+    chan <- newChan
+    isDone <- newEmptyMVar
+
+    let -- This is the action that the consumer of the stream
+        -- can run to stop feeding the channel
+        weAreDone = putMVar isDone ()
+
+    -- Writer thread will feed our Chan forever. This is where
+    -- your secret random number generation algorithm would go.
+    -- Unfortunately, we have some technical debt here.
```
tchoutri: > Unfortunately, we have some technical debt here.

What is the intended purpose of this comment?

doc/cookbook/infinite-streams/InfiniteStreams.lhs
```diff
@@ -0,0 +1,277 @@
...
+
+-- from `servant-server`
+import Servant.Server (serveWithContext)
+
+-- from `warp`
+import qualified Network.Wai.Handler.Warp as Warp (run)
+```
+
+We start with our scenario: we are tasked with creating an API which will serve random numbers in real-time. We are
+given a function that creates a producer of characters, and a method for us to stop the producer:
```
tchoutri: > a producer of characters

I thought we were producing an infinite stream of number. There does not seem to be any further mention of characters after that, have I missed something?

tchoutri (changes requested):

doc/cookbook/infinite-streams/InfiniteStreams.lhs
LaurentRDC: It was meant as a joke, as the "random numbers" are not random at all. I'll remove the comment

LaurentRDC:
Thanks for your review!

While addressing your comments, I have also refactored the tricky logic that interleaves integers with `KeepAlive` messages to use `withAsync`. I find this more intuitive, and easier to explain! Let me know what you think.

LaurentRDC:
> What is the reasoning behind the use of throwIO in favour of error? Do you get nicer callstacks or something?

There's no deep reason. I personally think of `error` as "quick-and-dirty`, and 'throwIO` as more proper.

tchoutri (approved):
Perfect, let's go!

LaurentRDC:
Thanks for reviewing @tchoutri !

LaurentRDC:
Very interesting! Thanks for sharing!
So basically, `error` *is* quick and dirty compared to `throwIO`.

PR #1860, opened 2025-12-12
RFC: Delegate Verb/NoContentVerb implementations to MultiVerb.

gdeest:
**NOTE**: This PR is based on #1859 . You may skip the first two commits when reviewing (but the second one is necessary for the tests to pass).

The goal of this PR is to start exploring the standardization of the core of servant around One True Endpoint Combinator (hereinfafter referred to as OTEC) that supports all of our use cases.

Current situation is more or less the following:

- `Verb` / `NoContentVerb` is the historical endpoint combinator in `servant`.
- The `Stream` combinator is distinct from `Verb` and supports advanced streaming usages with custom framing and arbitrary item type.
- `UVerb` improved upon `Verb` by allowing multiple instances, but leaks details of the HTTP protocol into handler types (namely HTTP status codes) and is about to be deprecated (see #1820 ).
- `MultiVerb` is the closest thing we have to OTEC but:

  1) Does not properly support streaming yet. `RespondStreaming` is limited to `SourceIO ByteString` and is not as expressive as the current `Stream` combinator.
  2) Exhibits some hopefully minor semantic differences w.r.t `Verb` that need to be clarified.
  
This PR tries to re-express the `Verb` semantics for `HasServer` and `HasClient` *in terms* of `MultiVerb` instances, in order to explore the feasibility of the OTEC vision, verify that MultiVerb provides everything we need, and work out some backward compatibility solution. It turns out to be easier than I feared, but does require some breaking changes. Notably:

- Verb status codes must now have `KnownStatus` instances, not just `KnownNat`.
- The `Verb` method parameter was (probably) needlessly polykinded ; it must now be restricted to `StdMethod`, as this is what `MultiVerb` does, in order to support delegation.
- Some semantic changes around header handling in clients are. needed for both `MultiVerb` and `Verb`:
    a) `MultiVerb` incorrectly removed duplicate headers when parsing responses (see #1859 ), which made `Verb` test case failed when expressed in terms of MultiVerb. We fix this.
    b) `Verb` client implementation did not really check that headers were provided even when marked as mandatory. We could break user code that relies on some server _not_ honoring the Servant API from which the client was derived.

Hopefully we can do something with the work on this branch. 

Doing the same thing for `Stream` is essentially feasible but requires extending `RespondStreaming` to support the features required by `Stream` (or add a new response type to `MultiVerb`). I have some non-primetime-ready code around that I'll consolidate in a PR later.

The end goal would be:

- To make MultiVerb the OTEC.
- Ultimately, to rename `MultiVerb` as `Verb`, and possibly move current `Verb` to a dedicated module, or rename it to `SimpleVerb`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -51,7 +51,8 @@ packages:
   -- doc/cookbook/open-id-connect
   doc/cookbook/managed-resource
   doc/cookbook/infinite-streams
-  doc/cookbook/openapi3
+  -- Commented out because servant-openapi3 doesn't build with new Verb using StdMethod
+  -- doc/cookbook/openapi3
   doc/cookbook/expose-prometheus
 
 tests: True
--- a/servant-client-core/src/Servant/Client/Core/HasClient.hs
+++ b/servant-client-core/src/Servant/Client/Core/HasClient.hs
@@ -35,7 +35,7 @@ import qualified Data.Text as T
 import qualified Data.Text as Text
 import Data.Text.Encoding (encodeUtf8)
 import Data.Typeable
-import GHC.TypeLits (KnownNat, KnownSymbol, TypeError, symbolVal)
+import GHC.TypeLits (KnownSymbol, TypeError, symbolVal)
 import Network.HTTP.Media (MediaType, matches, parseAccept)
 import qualified Network.HTTP.Media as M
 import qualified Network.HTTP.Media as Media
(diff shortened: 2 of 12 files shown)


ysangkok:
Now that servant-openapi3 is part of this repo, maybe this can be rebased and the cookbook can be fixed instead of commented out.

PR #1875, opened 2026-04-29
Require QuickCheck-2.18

ysangkok:
Since the tests do not need to have wide compatibility like the packages
people use, let's just support one major series (2.18).

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -84,3 +84,6 @@ if(impl(ghc >= 9.6.1))
 -- see https://github.com/commercialhaskell/lts-haskell/issues/409
 -- Change this to exclude 'memory' instead when the ecosystem is ready.
 constraints: ram < 0
+
+-- Allow QuickCheck-2.18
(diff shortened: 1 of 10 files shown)


wolfgangwalther:
This change is a bit annoying for us in Nixpkgs, where we run all tests, but have QuickCheck pinned to the version from whatever stackage snapshot we're using.

Stackage Nightly is currently at QuickCheck 2.16. I need to jailbreak many `servant-...` packages to keep them building.

Would it be possible to at least include the version used in Stackage in the bounds as well?

alexfmpe:
> This change is a bit annoying for us in Nixpkgs, where we run all tests, but have QuickCheck pinned to the version from whatever stackage snapshot we're using.

I think stackage also runs tests so they'll hit the same issue? 
cc @juhp

juhp:
Yes so we disabled the tests for now - I am afraid...

I have to agree - at least supporting stackage nightly would be good if possible

Issue #849, opened 2017-11-06
Superfluous "charset" parameter in JSON content-type


Philonous:
https://github.com/haskell-servant/servant/commit/7793b52d232074e1305cca1227fe8ded7f417686 adds "charset=utf-8" to the content type of json responses, but the parameter isn't defined in the [media-type registration](https://www.iana.org/assignments/media-types/application/json). This causes breakage with hspec-wai (https://github.com/hspec/hspec-wai/issues/46)

Related to: https://github.com/haskell-servant/servant/issues/559

phadej:
I'm sorry, we have ping-ponging this back and forth, and there's always some broken client.
Currently `servant-server` behaves well, as if asked for just `application/json` it will response with JSON.

The [JSON RFC](https://tools.ietf.org/html/rfc7159) says on page 10

>    Note:  No "charset" parameter is defined for this registration.
      Adding one really has no effect on compliant recipients.

So: `hspec-wai` should relax it's mime-type matching, e.g. by using [`mainType`](http://hackage.haskell.org/package/http-media-0.7.1.1/docs/Network-HTTP-Media-MediaType.html#v:mainType) from `http-media`.

Issue #559, opened 2016-07-31
Charset for serving JSON


koterpillar:
When serving JSON with non-ASCII characters in it, `hspec-wai-json` requires the `Content-Type` header to be `application/json; charset=utf-8` (see https://github.com/hspec/hspec-wai/blob/master/hspec-wai-json/src/Test/Hspec/Wai/JSON.hs#L67).

The JSON RFC says that it may be represented using UTF-8, UTF-16 or UTF-32, and the default is UTF-8: https://www.ietf.org/rfc/rfc4627.txt. _However_, the HTTP spec, which governs the `Content-Type` header, says the default is ISO-8859-1, and the charset _should_ be specified: https://www.w3.org/International/articles/http-charset/index

soenkehahn:
@koterpillar: Could you elaborate a bit more, please? It's not entirely clear what this issue is about. How does `servant` behave currently? How would you want it to behave?

koterpillar:
Sorry I was unclear. Currently when serving `JSON`, Servant sets `Content-Type` to `application/json`, and I believe it should be `application/json; charset=utf-8`.

koterpillar:
I am now convinced that Servant's behavior is correct:

- [JSON Interchange format](https://tools.ietf.org/html/rfc7159) says `charset` isn't needed.
- The w3.org link I posted earlier only applies to `text/` content, not to JSON.

PR #1881, opened 2026-05-05
Remove charset from JSON mime type as it is invalid per RFC 7159

### LINKED ISSUE #1715: Get '[JSON] does not imply a header Accept: application/json
I am using servant-client to request JSON information from the hackage API, but it turns out that specifying '[JSON] in the Get request will not add a header `Accept: application/json` to the Request, even though `requestAccept` seems to have the correct application types.

Here's a reproducer:
```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE DeriveAnyClass #-}
module Main where

import GHC.Generics
import Data.Proxy
import Data.Aeson
import Servant.API
import Network.HTTP.Client (newManager, defaultManagerSettings, requestHeaders)
import Servant.Client

data Package
  = Package
    { packageName :: String
    , downloads :: Int
    }
    deriving stock Show
    deriving stock Generic
    deriving anyclass FromJSON

type API = "top" :> Get '[JSON] [Package]

getTop :: ClientM [Package]
getTop = client (Proxy @API)

main :: IO ()
main = do
  manager' <- newManager defaultManagerSettings
  let clientEnv = mkClientEnv manager' (BaseUrl Http "hackage.haskell.org" 80 "packages")
  res <- runClientM getTop (clientEnv{makeClientRequest = \b r -> makeClientRequest clientEnv b r >>=
                                                            \r' -> return r'{requestHeaders = [("Accept", "application/json")]}})
  case res of
    Left err -> putStrLn $ "Error: " ++ show err
    Right packages -> print packages
```
with `build-depends: base, servant, servant-client, aeson, http-client`

## Expected behaviour

I was expecting the `Accept: application/json` to be automatically added to the request, since I'm using a `Get` request and `JSON` as the decoding format.

tchoutri:
See
https://stackoverflow.com/questions/4990095/json-specification-and-usage-of-bom-charset-encoding

Closes #1715 #559 #849

This will enable us to query Hackage, which is overly restrictive and has not changed its behaviour since 2019 https://github.com/haskell/hackage-server/issues/839.

Hackage itself is not wrong, just annoying. RFC 7159 was amended in its favour: https://www.rfc-editor.org/errata/eid5853

cc @ysangkok @alt-romes


ysangkok:
why is it a draft?

tchoutri:
@ysangkok I wanted to ensure CI was passing as expected before submitting for review but I already assigned you so it doesn't really make sense indeed.

ysangkok:
tests still fail:

```
  1) Servant.API.ContentTypes.Servant.API.ContentTypes.handleAcceptH respects the Accept spec ordering
       Falsified (after 7 tests and 1 shrink):
         ZeroToOne 0.3
         ZeroToOne 0.4
         ZeroToOne 0.8
         0
         AcceptHeader "text/plain;charset=utf-8; q=0.8, application/json; q=0.4, application/octet-stream; q=0.3"
         "text/plain;charset=utf-8" /= "text/plain"
```

PR #1816, opened 2025-05-05
Fix Optional ReqBody' (wrap value into Maybe), updated

PPKFS:
I updated #1347.

Tests added, which pass.

PR #1894, opened 2026-07-06
Respect `Optional` in `ReqBody'` in `servant-client`

MagicRB:
This is a continuation of #1816.

Drafted because I haven't tested this properly yet.


MagicRB:
Seems to work fine in my limited testing

PR #1899, opened 2026-07-13
Export internal links

### LINKED ISSUE #1898: Expose Links internals
I'm working on a combinator that needs to perform route-based content negotiation. This is easy enough to write a `HasServer` instance, but the `HasLink` instance needs to *modify* an existing segment of a given `Link`. Rather than expose some bespoke functionality, I think it's best to just expose all of the internals of `Links`.

isovector:
Closes #1898


servant/src/Servant/Internal/Links.hs
```diff
@@ -0,0 +1,692 @@
+{-# LANGUAGE CPP #-}
```
isovector, line 1: This file is copied over; I'm not sure why git doesn't recognize that.

servant/src/Servant/Internal/Links.hs
```diff
@@ -0,0 +1,692 @@
...
+-- >>> safeLink api bad_link
+-- ...
+-- ...Could not ...
+-- ...
+--
+--  This error is essentially saying that the type family couldn't find
+--  bad_link under api after trying the open (but empty) type family
+--  `IsElem'` as a last resort.
+--
+--  @since TODO
```
isovector, line 89: `@since TODO`

tchoutri (approved):

isovector:
@tchoutri is there anything needed on my end to land this?

tchoutri:
@isovector all good. :)

Issue #1100, opened 2018-12-20
Question about QueryParam encoding


NeQuissimus:
Hi, I have just used Servant for the first time.
Everything has been great and easy to use.

However, I have one tiny issue I can't figure out:

I have an API that takes a `QueryParam "q" String`. (This is a call against the GitHub API).
The values usually like similar to this:
`is:open+review-requested:tsteinbach` (Find all PRs I was requested to review)

When I make the actual call, the `:` and `+` have been encoded by Servant (I am sending the request to a `nc -l localhost 8081`).
Is there a way to prevent that? It would appear that the GitHub API does not like the encoded values.

If I use `curl`, the parameters are not encoded and the API returns what I expect (once I point it at the actual GitHub host):
` λ curl -i -H 'Authorization: token XXX' http://localhost:8081/api/v3/search/issues\?q\=is:open+review-requested:tsteinbach`

Thank you very much.

cjnickel:
**Disclaimer** - I'm new to haskell so I could be completely wrong :sweat_smile: 

When using servant for client request to an api the problem looks to me like the [renderQuery](https://hackage.haskell.org/package/http-types-0.12.2/docs/src/Network.HTTP.Types.URI.html#renderQuery) function being executed [here](https://github.com/haskell-servant/servant/blob/97bd6f0a40bf189b479902109b4b89f2889ff58e/servant-client/src/Servant/Client/Internal/HttpClient.hs#L191). `renderQuery` composes `renderQueryBuilder` and results in encoding `+`.
https://github.com/haskell-servant/servant/blob/97bd6f0a40bf189b479902109b4b89f2889ff58e/servant-client/src/Servant/Client/Internal/HttpClient.hs#L191

[...]

PR #1901, opened 2026-07-20
Use explicit query string escaping in clients

### LINKED ISSUE #1779: Inconsistent query parameter handling/encoding between server and client
Seems like the query parameter encoding is all over the place.

`servant-server` is based on `wai` which uses [parseQuery](https://hackage.haskell.org/package/http-types-0.12.4/docs/Network-HTTP-Types.html#v:parseQuery) (https://github.com/yesodweb/wai/blob/master/warp/Network/Wai/Handler/Warp/Request.hs#L108). This will always properly decode both query key and value.

However the `servant-client` introduced a risky approach to client parameter encoding in https://github.com/haskell-servant/servant/pull/1432 (and it seems it wasn't even necessary). Now only query keys are implicitly encoded. `QueryParam'` and `QueryParams` seem to be aware of this (after some bugfixes), and explicitly encode the value using `encodeQueryParam` ([QueryParam'](https://github.com/haskell-servant/servant/blob/4bcd33abd16b7d5f4f1838612d5baa158b2d9fdc/servant-client-core/src/Servant/Client/Core/HasClient.hs#L560) and [QueryParams](https://github.com/haskell-servant/servant/blob/4bcd33abd16b7d5f4f1838612d5baa158b2d9fdc/servant-client-core/src/Servant/Client/Core/HasClient.hs#L609) instances).

However since then, few more query handling types were introduced and they seem to be unaware of the fact. Namely [QueryString](https://github.com/haskell-servant/servant/blob/4bcd33abd16b7d5f4f1838612d5baa158b2d9fdc/servant-client-core/src/Servant/Client/Core/HasClient.hs#L660) (the main issue here is undocumented behavior - both server and client-vise, developers won't know what to expect while using it) and [DeepQuery](https://github.com/haskell-servant/servant/blob/4bcd33abd16b7d5f4f1838612d5baa158b2d9fdc/servant-client-core/src/Servant/Client/Core/HasClient.hs#L675) (no URL-encoding is performed here either).

There have been few issues/PRs related to URL-encoding reported in the past:
 * https://github.com/haskell-servant/servant/issues/1584
 * https://github.com/haskell-servant/servant/issues/1626

Now there's a proposed solution based on https://github.com/fizruk/http-api-data/pull/120 ([toEncodedQueryParam](https://hackage.haskell.org/package/servant-0.20.2/docs/Servant-API.html#v:toEncodedQueryParam)) that is attempting somewhat fixing the symptoms:
 * https://github.com/haskell-servant/servant/pull/1628
 
But the source of the issue is still there. There's nothing saying whether the `requestQueryString :: Query` in the client's `Request' should or should not be URL-encoded.

I think the surefire solution would be to switch to [PartialEscapeQuery](https://hackage.haskell.org/package/http-types-0.12.4/docs/Network-HTTP-Types.html#t:PartialEscapeQuery) for `servant-client`, which forces selection of URL-encoding on the type level and encoding it using `renderQueryPartialEscape`.

And for `servant-server` documenting that `QueryString` returns URL-decoded keys and values would also be a good idea. Not sure whether anything more can be done. Wrapping value in something like `newtype UrlDecodedQueryValue = QD ByteString` (names are just example, not a suggestion) to give the info on the type level seems to be excessive.

kutyel:
## Summary

- represent servant-client-core request queries with PartialEscapeQuery so decoded and already encoded value sections are explicit
- honor toEncodedQueryParam for QueryParam and QueryParams, encode DeepQuery values, and render partial queries in every in-tree client backend
- document server-side decoded query semantics with the DecodedQuery alias
- add regression coverage and a migration guide for custom combinators, middleware, backends, QueryString clients, and server handlers

## Migration

See the [query-string encoding migration guide](doc/migrations/query-string-encoding.md).

## Validation

- affected servant-client-core, servant-client, servant-server, and servant-http-streams test suites
- Fourmolu formatting and idempotence check
- HLint
- Haddock for affected packages

Fixes #1779.

Related: #1626, #1100.


doc/migrations/query-string-encoding.md
tchoutri, line 1: Could you make sure that add this page is part of the ReadTheDocs setup? You'll find instructions on how to generate it locally in ./doc/README.md

kutyel: Done in a68c1b32. The migration guide is now in the root Sphinx toctree. I also built the docs with the pinned requirements and verified that the generated index links to migrations/query-string-encoding.html.

arianvp (approved):

tchoutri (approved):
@kutyel Nice, thank you.

PR #1905, opened 2026-07-30
Fix static routing of explicitly empty segments

isovector:
The route `"hello" :> "" :> "world"` ought to explicitly match `hello//world`, but instead it matches `hello/world`. This PR changes the routing code slightly to support explicitly empty segments when they are present in the routing table, keeping the previous behavior if not.

### DIFF
--- a/changelog.d/pr-1905
+++ b/changelog.d/pr-1905
@@ -1,0 +1,8 @@
+synopsis: Fix routing of explicitly-empty segments
+packages: servant-server
+prs: #1905
+issues:
+description: {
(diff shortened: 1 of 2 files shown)


tchoutri (approved):

tchoutri:
Thanks @isovector. Out of curiosity, is there a usecase or is it just for correctness?

isovector:
I ran into this while trying to make a servant route for a legacy api that makes a distinction between the two cases 😩

tchoutri:
@isovector :(

PR #1907, opened 2026-08-12
Add support for QUERY verb (RFC 10008)

### LINKED ISSUE #1900: support QUERY verb (RFC 10008)
https://datatracker.ietf.org/doc/rfc10008/

sl1depengwyn:
Closes #1900.

## Summary

- Add RFC 10008 `QUERY` support through `Verb`, `NoContentVerb`, `UVerb`, `MultiVerb`, `Stream`, server routing, clients, docs, foreign representations, and QuickCheck.
- Add `Query` status aliases and typed `Accept-Query` response-header support for the RFC 9651 subset required by RFC 10008.
- Require `http-types >= 0.13 && < 0.14`; until 0.13 is released, pin the approved upstream commit for reproducible CI builds.

The `Accept-Query` parser intentionally remains field-specific and private: Servant has no other RFC 9651 consumer or suitable reusable dependency, while a generic implementation would require the complete Structured Fields model and conformance suite.

Swagger 2 and OpenAPI 3 remain unsupported because neither format defines a `QUERY` operation.


servant/src/Servant/API/AcceptQuery.hs
```diff
@@ -0,0 +1,422 @@
...
+
+zero, one, nine, capitalA, capitalZ, lowerA, lowerF, lowerZ :: Word8
+zero = 48
+one = 49
+nine = 57
+capitalA = 65
+capitalZ = 90
+lowerA = 97
+lowerF = 102
+lowerZ = 122
```
sl1depengwyn, line 422: This can be replaced with use of https://hackage.haskell.org/package/word8, but I'm not sure it worth adding a new dependency

clintonmead: It's not a new dependency as `servant` already indirectly depends on `word8`, so making it a direct dependency isn't going to make any difference as `word8` needs to be built to build `servant` anyway. 

Indeed `servant-server` actually *directly* depends on `word8`. 

And `word8` has no dependencies other than `base`. 

I'm not a servant maintainer but I'd go with the dependency over the copy pasta.

clintonmead:
I'd also like to see `QUERY` support in servant. I always find either hacking complex queries into query params or alternative making them `POST` requests when they're really not quite annoying.

Although I probably wouldn't use it until there was OpenAPI support as well.

Problem is that QUERY support needs `OpenAPI 3.2`.

There is a [OpenAPI 3.1 library for Haskell](https://hackage.haskell.org/package/openapi-hs) so probably the best bet would be to get that library to support OpenAPI 3.2 also (perhaps by adding some sort of version type parameter). Then once that is done AND this PR is merged, then the associated [OpenAPI 3.1 library for Servant](https://hackage.haskell.org/package/servant-openapi-hs) could then be upgraded to support OpenAPI 3.2 and the QUERY method in Servant.

So a bit of a long road, but this PR is a good start and I hope makes it into servant to encourage work in these other areas.

Perhaps one temporary hack to consider is having a servant type or option `QUERY_OVER_POST`, which on the server side will accept `POST` and generate `POST` based clients as a backup, i.e. will generate OpenAPI 3 that makes `POST` requests as `QUERY` isn't available there. 

This is a bit of a hack though and ideally we just have proper ecosystem support for `QUERY` and OpenAPI 3.2.

PR #1910, opened 2026-08-15
openapi3 compat with servant-0.20.3

ysangkok:
Because servant-openapi3 needed changes to be compatible with GHC 9.14, but
other servant packages did not, let us adjust the CPP in openapi3 such that it
becomes compilable with the latest released versions on Hackage.

That way, only servant-openapi3 needs a new upload, and other servant packages
can just get new revisions.


ysangkok:
I've just released this to Hackage as servant-openapi3-2.1.0.0.

PR #1911, opened 2026-08-20
Deduplicate `ResponseStatus` and `ResponseBody` type families by moving them to `servant`

clintonmead:
We add a `ResponseStatus` type family to `servant::Servant.API.MultiVerb` and as a result we can remove duplicate definitions in `servant-client` AND `servant-server`.

More importantly, now `servant-openapi3` no longer needs to depend on `servant-server`, reducing its dependency footprint and making it easier to compile to WebAssembly.

Also added a `ResponseDescription` type family, but this isn't used elsewhere, I've just included it for completeness.

The alternative is a [PR to just duplicate the `ResponseStatus` type family into `servant-openapi3`](https://github.com/haskell-servant/servant/pull/1912), but I think this is a better approach. But in the end either PR will achieve what I want which is removing the dependence of `servant-openapi3` on `servant-server`.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -72,6 +72,16 @@ package *
 package HsOpenSSL
   ghc-options: -optc=-Wno-discarded-qualifiers -optc=-Wno-deprecated-declarations -optc=-Wno-incompatible-pointer-types
 
+-- Various examples depend on external packages that are not in the `servant` repository.
+-- At the time of writing, these external packages have upper bounds on `servant-server` 
+-- that are below the version of `servant-server` in this repository.
+-- As a result, without setting `allow-newer` for these packages, `cabal build all` 
+-- will fail, as some examples will fail to build.
+allow-newer: servant-multipart:servant-server
+allow-newer: servant-swagger-ui:servant-server
+allow-newer: servant-swagger-ui-core:servant-server
+allow-newer: servant-pagination:servant-server
+
 -- Print ticks so that doctest type querying is consistent across GHC versions.
 -- This block can be removed when we drop support for GHC 9.4 and below.
 if(impl(ghc >= 9.6.1))
--- a/changelog.d/pr-1911
+++ b/changelog.d/pr-1911
@@ -1,0 +1,14 @@
+synopsis: Deduplicate `ResponseStatus` and `ResponseBody` type families by moving them to `servant`
+packages: servant, servant-server, servant-client-core, servant-openapi3
+prs: #1911
+description: {
+  Previously `ResponseStatus` and `ResponseBody` where defined twice in different modules 
+  with the same name and almost duplicate definitions. Namely associated types of both classes 
+  `ResponseRender` in `servant-server` and `ResponseUnrender` in `servant-client-core`.
+
+  These associated type families have been removed from `servant-server` and 
+  `servant-client-core`, and instead there is now just now a standalone 
+  type family definition of each of these type families in `servant`.
+
+  For completeness, a `ResponseDescription` type family to has been added to `servant` also.
+}
--- a/doc/cookbook/managed-resource/managed-resource.cabal
+++ b/doc/cookbook/managed-resource/managed-resource.cabal
@@ -16,7 +16,7 @@ executable cookbook-managed-resource
                      , aeson >= 2.2
                      , servant >= 0.20 && <0.21
                      , servant-client >= 0.20 && <0.21
-                     , servant-server >= 0.20 && <0.21
+                     , servant-server >= 0.20 && <0.22
                      , warp >= 3.4.9
                      , wai >= 3.2
                      , http-types >= 0.12
--- a/doc/cookbook/using-free-client/using-free-client.cabal
+++ b/doc/cookbook/using-free-client/using-free-client.cabal
@@ -16,9 +16,9 @@ executable cookbook-using-free-client
                      , servant >= 0.20 && <0.21
                      , servant-client >= 0.20 && <0.21
                      , http-client
-                     , servant-client-core >= 0.20 && <0.21
+                     , servant-client-core >= 0.20 && <0.22
                      , base-compat
-                     , servant-server >= 0.20 && <0.21
+                     , servant-server >= 0.20 && <0.22
                      , warp >= 3.4.9
   default-language:    Haskell2010
   ghc-options:         -Wall -pgmL markdown-unlit
--- a/servant-auth/servant-auth-client/servant-auth-client.cabal
+++ b/servant-auth/servant-auth-client/servant-auth-client.cabal
@@ -36,7 +36,7 @@ library
     , containers          >=0.6.5.1  && < 0.9
     , servant-auth        == 0.4.10.0
     , servant             >= 0.20.3   && < 0.21
-    , servant-client-core >= 0.20.3   && < 0.21
+    , servant-client-core >= 0.20.3   && < 0.22
 
   exposed-modules:
       Servant.Auth.Client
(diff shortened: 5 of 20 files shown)


flake.nix
```diff
@@ -60,6 +60,7 @@
                 haskellFormatter
                 haskellLinter
                 nixFormatter
+                pkg-config
```
LaurentRDC: This change seems unrelated?

LaurentRDC (commented):
Overall I think this is a great idea. Thanks for tackling this.

One thing that needs to be done, but isn't in this PR, is everything related to version management and release. Can you add appropriate changelog entries for this?

ysangkok:
Why is there no major version bump here? Surely if ResponseStatus is removed from a class, it will break every consumer of the library that is providing its definition?

flake.nix
clintonmead: I've removed this change from the PR.

But I couldn't get a `cabal build all` working without it.

clintonmead:
@LaurentRDC versions have been bumped. In particular:

1. `servant-client-core` gets major version bump, due to it losing the `ResponseStatus` associated type from the `ResponseUnrender`class
2. `servant-server` gets major version bump, due to it losing the `ResponseStatus` associated type from the `ResponseRender` class
3. `servant-openapi` gets a major version bump also, due to the class `IsSwaggerResponseList` now exposing a `ResponseStatus` from `servant` instead of `servant-server`.
5. `servant` itself only needs a minor version bump as it's just exposing a new name, but again, it's already had a minor version bump in git master above the last release, so no version bump is required.

Also needed to add some `allow-newer` clauses in `cabal.project` to get example packages to build.

`stack build` failed but it wasn't working before the changes also so I haven't touched it.

clintonmead:
Since we're bumping the major version anyway, I've also dedupped `ResponseBody` in the latest commit. 

 I did also have a thought, consider:

```
class IsRequest a where
  type ResponseType a :: Type
  type ResponseStatus a :: Nat
  type ResponseDescription a :: Symbol
  type ResponseBody a :: Type
```

I'm not sure if that's worth it (it's a bit weird having a class with associated types but no actual members) but if people think that will be a better direction to go let me know and I'll give it a go.

clintonmead:
Dedupping `ResponseBody` has broke one of the test builds. I won't be able to look at this til Monday, but in the meantime any other comments appreciated.

clintonmead:
I can't reproduce the build failure.

The error in the [build logs](https://productionresultssa16.blob.core.windows.net/actions-results/41169b8e-4643-4799-9457-53ee714cd040/workflow-job-run-70fdd0bf-b9df-5244-ba8d-23eec1d1978f/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-08-24T00%3A12%3A42Z&sig=M8x9p8L8T3gterTcR25ji0YZkDwHURQxZPNtiZTOOUs%3D&ske=2026-08-24T03%3A29%3A16Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-08-23T23%3A29%3A16Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-08-24T00%3A02%3A37Z&sv=2025-11-05) is as follows:

```
2026-08-21T12:28:43.0237306Z [1 of 1] Compiling Servant.Multipart [Servant.Server.Internal changed]
2026-08-21T12:28:46.3176915Z Configuring test suite 'example' for servant-machines-0.16.1...
2026-08-21T12:28:46.7946920Z Error: [Cabal-6578]
2026-08-21T12:28:46.7956638Z '/usr/local/.ghcup/ghc/9.2.8/bin/ghc-pkg-9.2.8' exited with an error:
2026-08-21T12:28:46.7980217Z ghc-pkg-9.2.8: /home/runner/work/servant/servant/dist-newstyle/packagedb/ghc-9.2.8/servant-multipart-0.12.1-inplace.conf: getModificationTime:getFileStatus: does not exist (No such file or directory)
```

I don't really understand why this is failing to build. There is no direct dependency on `servant-multipart` in `servant-machines`. Nor can I find anything even related to multipart in `servant-machines`.

clintonmead:
I've rebased the branch on `master` and repushed, but the workflow isn't running, so I'm not sure it made a difference to the build issue.

**Edit**: Workflow has run now and the build succeeded, don't really understand why the rebase made a difference though.

LaurentRDC (commented):
I still think we can improve the "packaging" of this change. Most importantly, there are no changelog entries anywhere

clintonmead:
I've added a `changelog.d` entry as per [`CONTRIBUTING.md`](https://github.com/clintonmead/servant/blob/master/CONTRIBUTING.md).

I note that I have made package version changes in this PR, on the advice of @LaurentRDC and @ysangkok, but `CONTRIBUTING.md` says:

> Important: please do not modify the versions of the servant packages you are sending patches for.

If these package version changes break the servant release workflow this these version changes can be removed from the PR, but if such package version changes are now expected for some or all PRs, perhaps the `CONTRIBUTING.md` should be updated to reflect that.

LaurentRDC (approved):
Thanks @clintonmead !

ysangkok:
I don't think we should let the major versions drift apart. It seems like previously, for major releases, most packages have been released together. I understand that in theory, we could release everything independently, I just don't know if it is a net-gain to even attempt. It becomes easier to think about if the possibly combination matrix of libraries is smaller. So I would prefer if all packages that have previously moved in lockstep continued to do so. That means that all packages that were on v0.20 would move to v0.21, and they would all require building with v0.21 of their dependencies. And all other packages with dependencies with a lower bound on v0.20 would move to a lower bound on v0.21. It's just too easy to allow a version combination that doesn't actually build. Remember that CI wouldn't actually verify that the older version actually works with the newer version of the other library.

clintonmead:
@ysangkok I'm unsure of how you want me to proceed here. I've done the version bumps as requested, even though [CONTRIBUTING.md](https://github.com/clintonmead/servant/blob/master/CONTRIBUTING.md) says I shouldn't do such version bumps at all. 

Would it be better if I just removed all the version and dependency changes as suggested in `CONTRIBUTING.md` and as part of your next release you can just perform your usual process of bumping versions and updating dependencies?

It seems to me that there's only 4 packages here that have their versions in sync, namely `servant`, `servant-server`, `servant-client-core` and `servant-client`.

Under this change, `servant` needs a minor version bump and `servant-client` is actually completely unchanged, the change in `servant-client-core` does not impact `servant-client`. 

But despite this I could bump `servant` and `servant-client` to `0.21.0.0`, like `servant-server` and `servant-client-core`, even though nothing has changed in `servant-client` and `servant` has had no removals or adjustments to existing exports. 

Another alternative if one wants to keep the major versions in sync but doesn't want to do unnecessary major version bumps when not required by the [PVP](https://pvp.haskell.org/) would be to proceed with the PR as is, but the next time `servant`/`servant-client` does require a major version bump is to perhaps skip a major version. So lets say this PR is merged then a few months down the track there both `servant-server` and `servant-client` require major version bumps, `servant-server` could move from `0.21.x` -> `0.22.0` and `servant-client` could skip a major version from `0.20.x` -> `0.22.0`. There's nothing in the PVP that says version numbers can't have gaps.

This will keep the major version numbers close (but not exactly the same) whilst not breaking downstream libraries with major version bounds when there's no reason to do so.

ysangkok:
Oh I didn't realize bumping is not supposed to be part of PRs.

I think you have good ideas around release management but I would hate for it to block this PR.

So let's merge this if @tchoutri thinks it looks good.

tchoutri (approved):
Very nice work Clinton, this is a much-appreciated contribution.

PR #1912, opened 2026-08-20
Eliminate `servant-openapi3` dependency on `servant-server`

clintonmead:
This is done by duplicating the type family `ResponseStatus` inside `servant-openapi3` directly.

This is a simpler approach than [the PR I've opened to move ResponseStatus to `servant` proper](https://github.com/haskell-servant/servant/pull/1911), but I think that PR it better because unlike this PR it reduces the net amount of code.

But either being merged will fix my issue, namely removing the dependence of `servant-openapi3` on `servant-server`, a dependence which was only added last release and I don't think is required, and makes it harder to compile `servant-openapi3` to WASM.

### DIFF
--- a/servant-openapi3/servant-openapi3.cabal
+++ b/servant-openapi3/servant-openapi3.cabal
@@ -74,7 +74,6 @@ library
                      , http-media                >=0.7.1.3  && <0.9
                      , lens                      >=5.3.6    && <5.4
                      , servant                   >=0.20.3   && <0.21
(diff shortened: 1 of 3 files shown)


LaurentRDC:
I also prefer the other approach, although it complicates release management

ysangkok:
I think we can close this since we decided on #1911.

clintonmead:
Agreed [#1991](https://github.com/haskell-servant/servant/pull/1911) obsoletes this. Closing.

PR #1918, opened 2026-09-08
No headers bodytype

lrworth:
This is a renewal of [this PR against biocad/servant-openapi3](https://github.com/biocad/servant-openapi3/pull/56).

Stop `BodyTypes` incorrectly adding `NoContent` to the list of body types if it appears within a `Headers` (on endpoints that set response headers but don't return a response body).

### DIFF
--- a/servant-openapi3/CHANGELOG.md
+++ b/servant-openapi3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased
 ----------
 
+* The `HasOpenApi` instance that recurses through `UVerb` responses no
+  longer demands `ToSchema NoContent`.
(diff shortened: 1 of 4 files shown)


JackKelly-Bellroy:
Thanks for doing this.

ysangkok:
The compilation failure seems related to this change, but I haven't looked very much into it. Would you say it is a problem with this PR, or a problem with the CI suite?

lrworth:
> The compilation failure seems related to this change, but I haven't looked very much into it. Would you say it is a problem with this PR, or a problem with the CI suite?

As rare as it is these days, PEBKAC.

LaurentRDC (approved):
This is great, thanks!

